PR #3, opened 2019-03-23
Use Python entry point

SylvainCorlay:


setup.py
```diff
@@ -35,7 +35,8 @@ def __str__(self):
         ['include/py_interface.cpp', 'include/parsing.cpp', 'include/solver.cpp', 'include/thirdparty/simdjson/simdjson.cpp'],
         include_dirs=[
             get_pybind_include(),
-            get_pybind_include(user=True)
+            get_pybind_include(user=True),
+            os.path.join(libsolv_prefix, 'include')
```
SylvainCorlay: It appears that this was missing when libsolv is not in the default include directory for the compiler.

PR #25, opened 2019-05-24
Add check to allow empty arguments to mamba

eteq:
Currently if one simply does ``mamba``, the result is an ``IndexError: tuple index out of range``.  Regular conda instead yields the equivalent of ``conda --help``.  I suspect this is just a bug, so this PR does what I think is the natural fix to ensure the standard conda behavior if no argument is given.


wolfv:
Awesome, thanks!

PR #46, opened 2019-07-03
fixes non-complete json extraction

wolfv:


wolfv:
@mdehollander this should fix the installation for qiime

sry 'bout that!

PR #51, opened 2019-07-10
fix update by adding changed packages to transaction

wolfv:


wolfv:
this might also fix #39 – do you have an easy way to test that @isuruf ?

PR #87, opened 2020-03-25
Prevent updating python if not specified

davidbrochart:


wolfv:
Looks great! Thanks a lot :)

PR #96, opened 2020-03-31
Substitute url_w_credentials for url_w_subdir

dmaljovec:
What?
=====

Uses `url_w_credentials` in place of `url_w_subdir` when looking up
channels in the priority map.

Why?
====

We use a private channel repository in conda and we are getting an
error here.
Stepping through the source code in a debugger it looks like the dict
key values of `_channel_priority_map` match the values of
`url_w_credentials` for cases where we are logged in (and do not need
 more than the channel name) and when we are logged out of anaconda
 (and thus require a url + token as our channel).

### DIFF
--- a/mamba/mamba_env.py
+++ b/mamba/mamba_env.py
@@ -43,7 +43,7 @@ def mamba_install(prefix, specs, args, env, *_, **kwargs):
 
     for x in index:
         # add priority here
-        priority = len(_channel_priority_map) - _channel_priority_map[x.url_w_subdir][1]
+        priority = len(_channel_priority_map) - _channel_priority_map[x.url_w_credentials][1]
         subpriority = 0 if x.channel.platform == 'noarch' else 1
         if os.path.exists(x.cache_path_solv):
             cache_file = x.cache_path_solv


dmaljovec:
Hi all, great work with this project. We are really excited about trying out mamba to help speed up some of our CI builds, but given we are using some projects that are hosted on a private anaconda channel, we are struggling with using mamba as a drop-in replacement for conda in our automated build system.

We are seeing great speed up (I think our rudimentary analysis is something like over 4x as fast) for a fairly bare bones conda environment setup on a local build where I can use `anaconda login` to circumvent the issue solved by this PR.

Thanks for putting this project together, and let us know if you need any more information for this.

dmaljovec:
CC: @teresa-recursion

dmaljovec:
Sorry, I have not added any tests to go along with this PR, but I can verify that using this fork of mamba successfully builds our environments with privately hosted dependencies in our CI system.

The timing is a bit apples to oranges right now, as I am building my fork from source in a docker image  which also includes pulling in extra dependencies for the build process (`g++`) rather than publishing it and installing it from our private channel. That said, there does seem to be an improvement in the timing already.

wolfv:
Hi @dmaljovec thanks a lot for this PR. Indeed we don't do a lot of testing with private channels so this is very much appreciated!

Thanks also for the kind words!

dmaljovec:
It appears there is not a continuous deployment for this repo which is fine, but I was curious to know what your release schedule looks like so we can have an expectation of when we can start using this from `conda-forge`? There is not a huge rush on our part, it is more just about expectation setting. 

Alternatively, we could also deploy a version to our private channel with this change if the releases are really spread out. I am just debating whether that effort is worth it at this point.

wolfv:
Hi @dmaljovec we had some nice fixes this week so a release will come very soon (latest friday).
Does that sound good to you?

dmaljovec:
That’s perfect. Again no rush. It is better that it be correct than done quickly.

wolfv:
0.1.2 is released on conda forge now. Will take a little while longer to be available

PR #98, opened 2020-04-01
add experimental C++

wolfv:
This contains a branch where I attempt to port everything necessary for `FastSubdirData.py` to C++ so we can get rid of some of the Python.

This will need further love, but it shows how we can use librepo to download metadata for us and how we can structure the C++ to resemble conda. 

We should expand on this, but for the moment we might take a direction where we still use the conda "frontend" to extract the data and correct all the paths, and then copy the necessary data into pybind11-exposed classes.

E.g. we would have a `mamba.Channel` class that is backed by C++ and contains fields like `name`, `url`, `repodata_fn` (which is tedious to get from the raw arguments right now).
This class + the `Subdir` class would then take care of downloading the metadata to the cache.

### DIFF
--- a/experiments/CMakeLists.txt
+++ b/experiments/CMakeLists.txt
@@ -1,0 +1,30 @@
+cmake_minimum_required(VERSION 3.2)
+
+project(testxxx)
+
+set(CMAKE_CXX_FLAGS -std=c++17)
+
+set(CMAKE_BUILD_TYPE "Debug")
+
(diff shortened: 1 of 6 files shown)


wolfv:
note that one needs to additionally have `skyr-url` installed (which is relying on another package called `expected`).

wolfv:
This has been superseded by everything added last month :)

PR #105, opened 2020-04-06
parallel downloads with librepo

wolfv:


wolfv:
Note:

we _will_ unvendor the indicators library once there is a conda package for it!

wolfv:
Remaining issues with this code:

- Python needs to be cleaned up, and FastSubdirData needs to be removed for good.
- We might have to use cURL directly because conda set's some headers, and also adds some header data to the metadata: 

It is adding these:

```
    headers = {}
    if etag:
        headers["If-None-Match"] = etag
    if mod_stamp:
        headers["If-Modified-Since"] = mod_stamp

    headers['Accept-Encoding'] = 'gzip, deflate, compress, identity'
    headers['Content-Type'] = 'application/json'
```

and saving these with the metadata:

```
    add_http_value_to_dict(resp, 'Etag', saved_fields, '_etag')
    add_http_value_to_dict(resp, 'Last-Modified', saved_fields, '_mod')
    add_http_value_to_dict(resp, 'Cache-Control', saved_fields, '_cache_control')
```

We could also ignore this and save the mamba metadata in a different place, disregarding these headers and just using a timeout set through the setting (which is how linux is usually doing it I guess).

wolfv:
@JohanMabille @davidbrochart 

I have written down some C++ stuff to tackle :)
Most of this concerns the subdirdata.hpp file.

librepo:

for metadata download we can't set or retrieve custom headers, but this would be needed for full compatibility of the downloaded metadata with conda. 
the alternative is be to use cURL (libCURL) directly through C++.

currently there is a setting in librepo to manage the maximum number of parallel downloads. 

we should figure out if we can limit the maximum number of downloads with taskflow as well. If yes, that would be great.

Maybe we should have one job with N connection sub-threads where each thread has a connection for downloading data.


Other C++ stuff:

figure out how we can use <filesystem> with C++...

better wrapper around libarchive (https://github.com/wolfv/mamba/blob/79f98d4b48c6effaff04a8c2ab2d2ea083e3ecd0/include/subdirdata.hpp#L33-L75)

Make sure that local URL's work

Make sure that cache directory exists from within C++. E.g. this line fails if the `~/miniconda3/pkgs/cache/` folder does not exist. Some functions for filesystem path creation etc. are in the experimental branch here https://github.com/wolfv/mamba/blob/c7fdd67572ff568b107f43fe7b9e6c58d897fff3/experiments/path.hpp#L116-L143

	https://github.com/wolfv/mamba/blob/79f98d4b48c6effaff04a8c2ab2d2ea083e3ecd0/include/subdirdata.hpp#L258

Conan-Kudo:
> librepo:
> for metadata download we can't set or retrieve custom headers, but this would be needed for full compatibility of the downloaded metadata with conda.

I think that adding the ability to do this in the librepo API would be a welcome addition. There are many cases where this makes sense, even for rpm ecosystem repodata fetching.

PR #110, opened 2020-04-07
Repoquery

wolfv:
First iteration for adding `repoquery` as a command.

These kind of queries already work:

```
(base) ~/Programs/mamba:master$ mamba repoquery --installed --whatrequires pyzmq
...

Executing the query pyzmq

jupyter_client, 6.0.0, py_0
notebook, 6.0.3, py37_0
done.
```

And:

`mamba repoquery xtensor` will show all available xtensor versions, `mamba repoquery xtensor>=0.15` all version >= 0.15 etc.

We will need 2 different output modes for this content:

- JSON (using nlohmann_json)
- Table-like output

cc @marimeireles

### DIFF
--- a/include/py_interface.cpp
+++ b/include/py_interface.cpp
@@ -5,6 +5,7 @@
 #include "pool.hpp"
 #include "transaction.hpp"
 #include "repo.hpp"
+#include "query.hpp"
 
 #include <pybind11/pybind11.h>
 #include <pybind11/stl.h>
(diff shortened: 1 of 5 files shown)


xhochy:
Not sure whether you are aware but I like the functionality that is currently provided by https://github.com/rvalieris/conda-tree and also its API (if you need a bit of inspiration). `repoquery --installed --whatrequires` is probably doing the same as it does in `dnf` but it seems a bit unintuitive for me.

marimeireles:
Hey @xhochy thanks for that :)
You think that output would be better than a table @wolfv?

wolfv:
thanks @xhochy for dropping by.

I think those are not excluding each other. Sometimes you want a table, sometimes you want a tree and sometimes we might want json :)

The natural representation for the two queries presented above would be tables or a flat json.

wolfv:
ok merging this now, but we'll be iterating :)

PR #112, opened 2020-04-07
add prefix data reading in C++

wolfv:
This will remove our need for writing out the temporary JSON to a file, and shifts reading of the Prefix Data (i.e. installed packages to C++).


wolfv:
Missing: 

need to pass in virtual packages from Python.

PR #114, opened 2020-04-08
Support channel::package

davidbrochart:


davidbrochart:
I cannot add tests with local channels, even conda doesn't allow `channel_a::pkg` if `channel_a` is local, it tries to get it from `https://conda.anaconda.org/channel_a`.

dhirschfeld:
Sounds like the default channel alias interfering:
https://conda.io/projects/conda/en/latest/user-guide/configuration/use-condarc.html#set-ch-alias

Does using a `file://` url work?

davidbrochart:
> Does using a `file://` url work?

no, it doesn't either.

dhirschfeld:
Yeah, I wasn't sure that the `channel::pkg` syntax worked with full urls - seems like it probably *should* though.

davidbrochart:
I filed an issue: https://github.com/conda/conda/issues/9828

wolfv:
I skipped merging this PR because I'd like to not parse the JSON from Python. I'll try to come up with some refactoring to do this through libsolv :)

wolfv:
I have just merged the exact same idea, implemented in C++ #277 :tada;

PR #135, opened 2020-04-20
C++ 14 compat for win

davidbrochart:


wolfv:
C:\Users\wolfv\Miniconda3\Lib\site-packages\mamba\mamba_api.cp38-win_amd64.pyd

that is the location

wolfv:
I was also thinking maybe we need to switch to windows-2016? I think everything on conda-forge should be compatible with VS 2017 but I don't know about VS 2019 :(

davidbrochart:
Trigger CI.

PR #141, opened 2020-04-20
Refactored validate.hpp

JohanMabille:


include/util.hpp
```diff
@@ -1,85 +1,40 @@
...
-    fs::path& path()
-    {
-        return m_path;
-    }
+    bool ends_with(const std::string_view& str, const std::string_view& suffix);
+    bool starts_with(const std::string_view& str, const std::string_view& prefix);
+    void to_human_readable_filesize(std::ostream& o, double bytes, std::size_t precision = 0);
 
-    operator fs::path()
+    class TemporaryDirectory
```
wolfv, line 24: btw I'd also like to have an TemporaryFile like this

PR #154, opened 2020-04-21
JSON output

davidbrochart:


davidbrochart:
When `--json` is set, conda still prints some non-JSON stuff (Package Plan, etc.). I guess this can be removed when we move more code to C++.
Also, because we still let conda do the LINK, the JSON output doesn't include it at the moment. Again, this can be fixed in the future.

.github/workflows/main.yml
```diff
@@ -31,7 +31,7 @@ jobs:
         run: |
           conda config --add channels conda-forge
           conda config --set channel_priority strict
-          conda create -q -y -n mamba-tests python=$PYTHON_VERSION pip pybind11 libsolv libarchive libcurl nlohmann_json conda cxx-compiler
+          conda create -q -y -n mamba-tests python=$PYTHON_VERSION pip pybind11 pybind11_json libsolv libarchive libcurl nlohmann_json conda cxx-compiler
```
wolfv, line 34: pybind11 json isn't really necessary anymore, right?

davidbrochart: Indeed :-)

src/transaction.cpp
```diff
@@ -214,12 +214,18 @@ namespace mamba
...
         m_transaction = solver_create_transaction(solver);
         init();
+        Console::instance().json_down("actions");
     }
 
     MTransaction::~MTransaction()
     {
         LOG(INFO) << "Freeing transaction.";
         transaction_free(m_transaction);
+        Console::instance().json_write("PREFIX", Context::instance().target_prefix);
```
wolfv, line 224: Can we simplify this to use the fancy nlohmann::json syntax (

```
{
{"PREFIX", target_prefix},
{"dry_run", ...}
}
```

(This is the example in the readme:

```
// instead, you could also write (which looks very similar to the JSON above)
json j2 = {
  {"pi", 3.141},
  {"happy", true},
  {"name", "Niels"},
  {"nothing", nullptr},
  {"answer", {
    {"everything", 42}
  }},
  {"list", {1, 0, 2}},
  {"object", {
    {"currency", "USD"},
    {"value", 42.99}
  }}
};
```

wolfv: also on win, PREFIX is lowercase? should it be lowercase or uppercase?

wolfv: ah, wait, that isn't true, just saw the lowercase prefix below, sorry :)

src/transaction.cpp
```diff
@@ -275,10 +281,11 @@ namespace mamba
         queue_free(&pkgs);
     }
 
-    auto MTransaction::to_conda() -> to_conda_type 
+    auto MTransaction::to_conda() -> to_conda_type
     {
-        to_install_type to_install_structured; 
+        to_install_type to_install_structured;
         to_remove_type to_remove_structured;
+        Console::instance().json_down("FETCH");
```
wolfv, line 288: I don't think this is the correct place. `to_conda` just turns the transaction to a JSON format that conda & mamba use to interface with each other.

Maybe we should have a dedicated function to print the transaction as JSON?

.github/workflows/main.yml
```diff
@@ -31,7 +31,7 @@ jobs:
         run: |
           conda config --add channels conda-forge
           conda config --set channel_priority strict
-          conda create -q -y -n mamba-tests python=$PYTHON_VERSION pip pybind11 libsolv libarchive libcurl nlohmann_json conda cxx-compiler
+          conda create -q -y -n mamba-tests python=$PYTHON_VERSION pip pybind11_json libsolv libarchive libcurl nlohmann_json conda cxx-compiler
```
wolfv, line 34: I think we shoudl remove `pybind11_json` because it's not used, but keep `pybind11`, right?

davidbrochart: Right, it's not used anymore.

src/transaction.cpp
davidbrochart: Regarding the more concise syntax, it's possible.

include/output.hpp
```diff
@@ -247,6 +248,16 @@ namespace mamba
...
 
+        nlohmann::json json_log;
+        void json_write(const std::string& key, bool value);
+        void json_write(const std::string& key, const std::string& value);
+        void json_write(const nlohmann::json& j);
+        void json_append(const std::string& value);
+        void json_append(const nlohmann::json& j);
+        void json_down(const std::string& key);
+        void json_up();
+        void json_print();
```
JohanMabille, line 259: I wonder if this API should not be in a dedicated class (JSON builder / formatter or whatever) and call `Console::print()` when the json is finished. This violates the principle of single responsibility.

src/output.cpp
```diff
@@ -345,5 +345,67 @@ namespace mamba
...
+    {
+        if (Context::instance().json)
+            json_hier.erase(json_hier.rfind('/'));
+    }
+
+    void Console::json_print()
+    {
+        if (Context::instance().json)
+            std::cout << std::setw(4) << Console::instance().json_log.unflatten() << std::endl;
+    }
```
JohanMabille, line 409: This should acquire the mutex, which strengthens my feeling that this (and all the json related methods) should be in a dedicated class.

include/output.hpp
davidbrochart: Agreed!

src/output.cpp
davidbrochart: :+1:

PR #176, opened 2020-04-24
use dynamic libsolv version

wolfv:


src/common_write.c
```diff
@@ -24,6 +24,17 @@
 
 #define MAMBA_SOLV_VERSION MAMBA_TOOL_VERSION "_" LIBSOLV_VERSION_STRING
 
+static char*
+mamba_tool_version()
+{
+  static char MTV[30];
+  MTV[0] = '\0';
```
wolfv, line 31: @JohanMabille do you have an idea how to make this nicer?

JohanMabille:
```cpp
static char* mamba_tool_version()
{
    std::string mtv = std::string(MAMBA_TOOL_VERSION) + "_" + std::string(solv_version);
    return mtv.c_str();
}
```

But you need to define out of common_write.c since it has external C linkage.

PR #115, opened 2020-04-09
initial version of conda history parsing

wolfv:

PR #181, opened 2020-04-27
Added url and patch_spec from #115

JohanMabille:
Files taken from #115. The biggest change is the refactoring of pystring. The methods were directly added to `utils.hpp`, we can also move them to a dedicated file (like `string_utils.hpp`).

The `split` and `rsplit` overloads that accept string separators have not been implemented yet. I will add them in a dedicated PR.

### DIFF
--- a/include/match_spec.hpp
+++ b/include/match_spec.hpp
@@ -1,0 +1,34 @@
+#ifndef MAMBA_MATCH_SPEC
+#define MAMBA_MATCH_SPEC
+
+#include <regex>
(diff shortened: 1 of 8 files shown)


wolfv:
do you think we should merge this already? match spec is ~50% complete with the parsing -- but it's not used anywhere right now so I guess it would be fine.

JohanMabille:
Yep this would allow any of us to iterate on it when he/she has time.

PR #184, opened 2020-04-27
Added overloads of split and rsplit

JohanMabille:


wolfv:
Closing this for now as this is the functionality that we now have in the source (but it's the only one, the one-char is out for the moment as it didn't behave as expected :)

PR #185, opened 2020-04-28
initial implementation of conda history parsing

wolfv:
@marimeireles I had to rebase the PR because we pushed some of the files from the other PR to master already.

What's missing in this PR are functions to modify the history.

The history records all actions that conda / mamba did on a specific prefix such as installing or removing a package.

One needs to distinguish packages that are installed on **user** request (those are the "specs"), and just installed packages.

What we need now is to have a function where we can pass in new specs (packages that were asked to be installed, removed or neutered), and where we can give a list of newly installed and newly deleted packages and then write them out to the History file in the same format as conda does:

```
==> 2020-04-24 15:06:47 <==
# cmd: /home/wolfv/miniconda3/bin/mamba install pytorch -c defaults
# conda version: 4.8.3
+conda-forge/linux-64::ninja-1.10.0-hc9558a2_0
+defaults/linux-64::_pytorch_select-0.2-gpu_0
+defaults/linux-64::blas-1.0-mkl
+defaults/linux-64::cudatoolkit-10.1.243-h6bb024c_0
+defaults/linux-64::cudnn-7.6.5-cuda10.1_0
+defaults/linux-64::intel-openmp-2020.0-166
+defaults/linux-64::mkl-2020.0-166
+defaults/linux-64::mkl-service-2.3.0-py37he904b0f_0
+defaults/linux-64::pytorch-1.4.0-cuda101py37h02f0884_0
# update specs: ['pytorch']
```

The interface can probably look something like:

```
History
{
...

add_entry(const std::vector<std::string>& update_specs, const std::vector<std::string>& remove_specs, ..., )
{
}

}
```

### DIFF
--- a/include/history.hpp
+++ b/include/history.hpp
@@ -1,0 +1,62 @@
+#ifndef MAMBA_HISTORY
+#define MAMBA_HISTORY
+
+#include <vector>
+#include <regex>
+#include <set>
+
+#include "thirdparty/filesystem.hpp"
+
+#include "output.hpp"
+// #include "prefix_data.hpp"
+#include "match_spec.hpp"
+
+namespace fs = ghc::filesystem;
+
+namespace mamba
+{
+
+class History
+{
+public:
+
+    // History(const std::shared_ptr<PrefixData>& prefix)
+    //     : m_prefix_data(prefix)
+    // {
+    // }
+
+    History(const std::string& prefix);
+
+    struct ParseResult
+    {
+        std::string head_line;
+        std::set<std::string> diff;
+        std::vector<std::string> comments;
+    };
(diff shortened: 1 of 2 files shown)


wolfv:
Ah the cmake of the experiment has to add two files (if you are using explicit C++ files now):

```
	${CMAKE_SOURCE_DIR}/../src/history.cpp
	${CMAKE_SOURCE_DIR}/../src/match_spec.cpp
	${CMAKE_SOURCE_DIR}/../src/url.cpp
```

wolfv:
@marimeireles I just merged the branch so you can also work from master on the missing bits :)

PR #188, opened 2020-04-29
Add method that allows to add history to file

marimeireles:
+ fix regex bugs in match_spec
I checked the other occurrence of `regex_match`, `subdirdata.cpp` and it seems fine.


include/history.hpp
```diff
@@ -36,12 +36,12 @@ class History
 
     struct UserRequest
     {
+        std::string date;
         std::string cmd;
         std::string conda_version;
-        std::string date;
```
marimeireles, line 41: I'm changing the order here to be the same as the one in the file.

src/history.cpp
```diff
@@ -11,17 +12,17 @@ namespace mamba
     std::vector<History::ParseResult> History::parse()
     {
         std::vector<ParseResult> res;
-        fs::path history_file_path = fs::path(m_prefix_data) / "conda-meta" / "history";
-        LOG_INFO << "parsing history: " << history_file_path;
+        m_history_file_path = fs::path(m_prefix_data) / "conda-meta" / "history";
+        LOG_INFO << "parsing history: " << m_history_file_path;
 
-        if (!fs::exists(history_file_path))
+        if (!fs::exists(m_history_file_path))
```
marimeireles, line 18: Decided to create a member in the class because I'd have to do the same line:
`         fs::path history_file_path = fs::path(m_prefix_data) / "conda-meta" / "history";` again in the method I added.

src/history.cpp
```diff
@@ -71,10 +72,10 @@ namespace mamba
             std::cmatch ematch;
             std::vector<std::string> pkg_specs;
             const char* text_iter = elems.c_str();
-            while (std::regex_search(text_iter, (const char*) elems.c_str() + elems.size(), ematch, elems_pat))
```
marimeireles, line 74: This is not really necessary, cpp can figure it out having just `const char*` as an input.

src/history.cpp
```diff
@@ -170,4 +171,60 @@ namespace mamba
...
+                }
+                if (request.update.size() > 0)
+                {
+                    out << "# update specs: [";
+                    std::string update_string;
+                    for (auto u : request.update)
+                    {
+                        update_string += "'" + u + "',";
+                    }
+                        update_string.back() = ']';
```
wolfv, line 206: I see a few too many tabs or spaces here :)

src/history.cpp
```diff
@@ -170,4 +171,60 @@ namespace mamba
...
+                        out << update_string << std::endl;
+                }
+                if (request.remove.size() > 0)
+                {
+                    out << "# remove specs: [";
+                    for (auto r : request.remove)
```
wolfv, line 212: Maybe move this repeated code to write a "Python list" to a lambda function that converts a vector of strings to this pythonic std::string?

src/history.cpp
```diff
@@ -170,4 +171,60 @@ namespace mamba
...
+                    {
+                        out << "'" << r << "',";
+                    }
+                        out << "]" << std::endl;
+                }
+                if (request.neutered.size() > 0)
```
marimeireles, line 218: Neutered and removed needs an update string logic, otherwise it will format things wrongly.
Forgot about this.

src/history.cpp
marimeireles: Can you explain a bit more?
What would this lambda function do? It'd modify the `u`s from the `for` and format them the right way?

wolfv: I was thinking you can make a lambda funtion taht takes two inputs: 

1. "remove", "update", "neuter"
2. std::vector<string>

and then the lambda spits out a std::string like `# XXXXXX specs: [ '....', '...', ...]`

so something like

```
auto to_string_list = [](const std::string& action, const std::vector<std::string>& specs) -> std::string {
    make_my_string
}
```

what do you think? would that work?

marimeireles: It makes sense.
Much more fancy
gonna implement that
thanks Wolf

wolfv:
awesome!

PR #191, opened 2020-04-29
add cpp tests

wolfv:
cc @marimeireles


wolfv:
@JohanMabille you will also like this :)

PR #195, opened 2020-05-04
Add history test

marimeireles:


marimeireles:
There is something weird happening locally to me, the CMake for the tests don't work at all because it was missing dependencies and a few other things.
I basically copied a bunch of stuff from the experiment CMake into it. Does it work in your machine @wolfv?
Maybe my environment is weird :(

wolfv:
great! yeah the current CMake is working on my machine and on CI... if you have a build folder, maybe remove it and restart? That's what I do usually.

I think you need to add the test to the test/CMakeLists here: https://github.com/QuantStack/mamba/blob/1f9bc5f98f5f2ef36c74aaadabcb629ca1cb6bcb/test/CMakeLists.txt#L5-L8

Then it'll also run on github (on the ubuntu machines :)

marimeireles:
> yeah the current CMake is working on my machine and on CI

Yeah, it makes sense :p

And thanks, forgot about that.

wolfv:
Hey, I am not convinced this isn't just testing two empty strings :)
I am not sure that if the file doesn't exist, it will fail with ifstream.

You might have to install the text files with CMake so that they appear at the correct location for the tests.

marimeireles:
You mean...
`install(FILES history_data DESTINATION . )`
Like this?

wolfv:
yep!

test/test_history.cpp
```diff
@@ -0,0 +1,31 @@
...
+
+        std::ifstream original_history_file("original_history_data");
+        std::stringstream original_history_buffer;
+        original_history_buffer << original_history_file.rdbuf();
+
+        std::ifstream aux_history_file("aux_history_data");
+        std::stringstream aux_history_buffer;
+        aux_history_buffer << aux_history_file.rdbuf();
+
+        ASSERT_EQ(original_history_buffer.str(), aux_history_buffer.str());
```
wolfv, line 23: if those two buffers are empty, it will still compare true, right?
Can we add some checks to make sure we're actually comparing something here?

marimeireles: I can do that, but isn't it redundant? `original_history_buffer` will never be empty, because there will always be content inside  the file `original_history_data`.

wolfv: have you tested what happens when you delete the file and it doesn't exist? it's possible that there is no loud error...

wolfv: have you tested what happens when you delete the file and it doesn't exist? it's possible that there is no loud error...

wolfv: Also the way I see it right now is that your aux_history_file is actually empty, so I am pretty sure you are comparing two empty strings here otherwise the test shouldn't pass :)

And you aren't using the `user_requests` that you read from the other history file as far as i can see.

marimeireles: No I didn't, but the file is not going to be deleted by any "natural" cause, if that makes sense. Only if the user does something they shouldn't.
But I see your point, I'm going to test it and see what happens and depending on what happens I'll throw an error. Thanks for the feedback 🐺

marimeireles: Ohh, yeah, you're totally right.

wolfv:
hey I just merged the other PR so you might have to rebase :/ sorry about that!

marimeireles:
```
-- Performing Test HAS_FLTO
  Cannot find source file:

-- Performing Test HAS_FLTO - Success
    test_history.cpp

```

Wut, why?

wolfv:
you put it in a subfolder :)

marimeireles:
Oh yeah, how does windows and macos find it tho

wolfv:
The CMake based tests are only run on Linux right now.

marimeireles:
I can't hardcode the path like I'm doing because GHA uses a different kind of environment to run tests?

```
unknown file: Failure
C++ exception with description "Couldn't open file: ../history_test/conda-meta/history" thrown in the test body.
```

wolfv:
are you able to run the tests locally (also from a clean slate?)

wolfv:
You need to make sure that the test files are correctly installed as well :) you can check how the tests are run on github in the `.github/...` folder

marimeireles:
Ok, I'm gonna check it in my mac, but it's working fine on my linux.
And thanks :)

wolfv:
usually we have a `build` folder, and removing it should bring everyting back to pristine state.

e.g. 

```
mkdir build; cd build;
cmake .. -DCMAKE_INSTALL_PREFIX=$CONDA_PREFIX -DENABLE_TESTS=ON
make test
```

and then to reset

```
cd ..; rm -rf build;
```

marimeireles:
Yeah, I've done that on my linux and it works. But I have a different CMake there, that's why I'm gonna run on mac. I currently can't reproduce this. :(

```
          mkdir cmake_build
          cd cmake_build
          cmake .. -DCMAKE_INSTALL_PREFIX=$CONDA_PREFIX -DENABLE_TESTS=ON
          make test -j2
```

GHA is doing the same thing that I am, I just couldn't find where it runs cpp tests.

wolfv:
make test is a magic command that executes the C++ tests :) (it's in the cmake).

marimeireles:
It also works locally for me, so, should I rebase to merge? @wolfv

wolfv:
noice! yeah rebasing and squashing would be good!

PR #199, opened 2020-05-04
Stringstuff

wolfv:


wolfv:
@JohanMabille now you'll have some tests to work against :)

PR #204, opened 2020-05-07
Fix cpp-tabulate version requirement

davidbrochart:


wolfv:
hey sorry for bothering you but wouldn't it be beter to put the `"` like `"cpp-tabulate>=1.12"`? I think it makes it clearer

PR #206, opened 2020-05-07
Moved PackageRecord into PackageInfo

davidbrochart:
Removed PackageRecord and merged it with PackageInfo, which tried to keep the attributes and the JSON object in sync.


wolfv:
gr8!

PR #236, opened 2020-05-13
Channel

JohanMabille:
Partial implementation of channel. I think it's better to start integrating at this point and iterate rather than reviewing a huge PR when everything is done.

### DIFF
--- a/include/channel.hpp
+++ b/include/channel.hpp
@@ -79,7 +79,6 @@ namespace mamba
 
         Channel m_channel_alias;
         channel_map m_custom_channels;
-
     };
 
 }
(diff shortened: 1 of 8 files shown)


wolfv:
I read through it – looks very good! Happy to merge, but I merged your later PRs and now we need a rebase :)

I think we can expose the channel thing to python in the py_interface and try to integrate it under the `--mamba-experimental` flag?

JohanMabille:
I need to implement the url/urls methods (they are used in `util.py` and `mamba.py`) and plug `ChannelContext` into `Context` and then we can expose them to the Python.

PR #244, opened 2020-05-14
Produce correct JSON output on --dry-run --json

xhochy:


xhochy:
Probably should add a test for this in `tests/`

test/test_all.py
```diff
@@ -48,3 +53,29 @@ def test_track_features():
...
+    output = subprocess.check_output(mamba_cmd, shell=True).decode()
+
+    if use_json:
+        # assert that the output is parseable parseable json
+        json.loads(output)
+    elif not experimental:
+        # Assert the non-JSON output is in the terminal.
+        assert "Total download" in output
+
+    if not experimental:
```
xhochy, line 79: The `not experimental` cases here should be `xfail`s but then new breakage wouldn't be detected in `experimental` either.

wolfv:
thanks!

PR #248, opened 2020-05-14
Add version info to the banner

dhirschfeld:
If we're going to have an enormous banner we may as well use it to display some useful information by default! This is just a suggestion - totally support any other formatting or feel free to close if you don't think it's worthwhile.


wolfv:
Awesome, totally makes sense.

Also mamba --version should show something different (currently it shows the conda version).

PR #257, opened 2020-05-15
Channel

JohanMabille:
@wolfv It seems that we have to implement some features from MultiChannel for channel prioritization. However I think we can really simplify it (a simple additional multiname member in the Channel class, that remains empty if the Channel is not a multi channel).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -38,6 +38,7 @@ include_directories(
 
 set(MAMBA_SOURCES
     ${CMAKE_SOURCE_DIR}/src/activation.cpp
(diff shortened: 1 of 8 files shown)


wolfv:
weird, tests aren't running ...

JohanMabille:
yep, let me push an empty commit.

JohanMabille:
This is overgreen!

JohanMabille:
This is ready for a review. There are still some issues when using the C++ implementation of Channel from the C++, but I think this one can be merged  nevertheless.

PR #268, opened 2020-05-18
respect dry run in transaction execute

wolfv:


davidbrochart:
I think this is due to JSON output not being called, I can fix this if you let me push to this PR.

PR #278, opened 2020-05-19
platforms and urls with default argument

JohanMabille:


src/context.cpp
```diff
@@ -7,6 +7,46 @@
 
 namespace mamba
 {
+    namespace
+    {
+// armv6l and armv7l
+#if defined(__arm__) || defined(__thumb__)
```
wolfv, line 13: nice! I wonder if we should not do this at compile time ... 
one might want to be able to override this dynamically.
Or at least we shoudl respect hte CONDA_SUBDIR env param that e.g. Uwe used for debugging here: https://github.com/QuantStack/mamba/issues/180

wolfv: anyways we can fix this later, maybe in Context have a subdir var.

JohanMabille: we can initialize a data member in the context that can be overriden and have the `platform` method returning this data member instead of the constant (that was my first implementation by the way).

PR #280, opened 2020-05-20
Repoquery json

wolfv:


mamba/utils.py
```diff
@@ -67,7 +67,7 @@ def init_api_context(use_mamba_experimental=False):
         context.quiet = True
         if use_mamba_experimental:
             context.json = False
-            context.dry_run = False
+            # context.dry_run = False
```
wolfv, line 70: @xhochy do you remember why we needed to set this to false here?

xhochy: @davidbrochart probably knows better but I think this had to do that the non-dry-run things  are all done in the C++ code.

xhochy: Still, an issue that is "commented out" in a unit test, is that `mamba create --dry-run` still creates the environment directory which it shouldn't: https://github.com/QuantStack/mamba/blob/bccd8ae94282712163c32493738d7507abea0ad4/test/test_all.py#L79-81

wolfv: yes, that is because of this here. when keeping it dry_run in the context it works fine.

wolfv: ok i tested a bunch of commands and it seems to be an improvement :)

PR #284, opened 2020-05-20
Implemented MultiChannel logic, calculate_channel_urls and get_channe…

JohanMabille:
…l_urls

I started reimplementing all the MultiChannel logic with composite pattern and so on until I realize we just don't need them.

### DIFF
--- a/src/channel.cpp
+++ b/src/channel.cpp
@@ -518,9 +518,83 @@ namespace mamba
         return Channel::from_name(value);
     }
 
+    /************************************
+     * utility functions implementation *
+     ************************************/
+
     Channel& make_channel(const std::string& value)
     {
         return Channel::make_cached_channel(value);
+    }
+
+    void append_channel_urls(const std::string name,
+                             const std::vector<std::string>& platforms,
+                             bool with_credential,
+                             std::vector<std::string>& result,
+                             std::set<std::string>& control)
+    {
+        bool ret = !control.insert(name).second;
+        if (ret)
+            return;
+
+        std::vector<std::string> urls = make_channel(name).urls(platforms, with_credential);
+        std::copy(urls.begin(), urls.end(), std::back_inserter(result));
+    }
+
+    std::vector<std::string> get_channel_urls(const std::vector<std::string>& channel_names,
+                                              const std::vector<std::string>& platforms,
+                                              bool with_credential)
+    {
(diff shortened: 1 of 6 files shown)


JohanMabille:
What's going on here? The build is systematically cancelled.

wolfv:
I haven't cancelled anything. Either github is has issues or they want us to pay for CI time?

test/test_channel.cpp
```diff
@@ -72,5 +79,27 @@ namespace mamba
         EXPECT_EQ(urls2[0], "https://conda.anaconda.org/conda-forge/win-64");
         EXPECT_EQ(urls2[1], "https://conda.anaconda.org/conda-forge/noarch");
     }
+
+    TEST(Channel, calculate_channel_urls)
+    {
+        std::vector<std::string> urls = { "conda-forge", "defaults" };
+        std::vector<std::string> res = calculate_channel_urls(urls, true);
+        EXPECT_EQ(res.size(), 6);
+        EXPECT_EQ(res[0], "https://conda.anaconda.org/conda-forge/linux-64");
```
wolfv, line 88: yeah! :tada:

src/channel.cpp
```diff
@@ -518,11 +518,85 @@ namespace mamba
...
+            names.reserve(channel_names.size() + 1 + ctx_channels.size());
+            if (use_local)
+            {
+                names.push_back(LOCAL_CHANNELS_NAME);
+            }
+            std::copy(channel_names.begin(), channel_names.end(), std::back_inserter(names));
+            if (prepend)
+            {
+                std::copy(ctx_channels.begin(), ctx_channels.end(), std::back_inserter(names));
+            }
```
wolfv, line 591: what happens if prepend is false?

wolfv: ah so the opposite of prepend is not append, it's just to ignore the ctx_channels?

JohanMabille: Nothing.

JohanMabille: Btw I kept the name from conda, but I think we should rename it into something more explicit, like "append_context_channel".

Also in our case, the channels_names and the ctx_channels are exactly the same (the context always contains those from the command line + configuration), and I wonder if we'll have a use case where we pass different channel names than those from the command line.

PR #292, opened 2020-05-22
WIP - Implemented get_index

JohanMabille:


JohanMabille:
The issue comes from the instantiation of `PackageRecord`. It ends up in `conda.Channel` constructor with an argument of type `mamba_api.Channel`, making this [condition] false while it should be true.

PR #296, opened 2020-05-25
Fixed URLHandler for file scheme

JohanMabille:


wolfv:
should a path be handled by the URL handler?
We could also do some preprocessing to add a `file://` scheme?

JohanMabille:
`file` is a valid scheme for an URL / URI, so I think we should handle it like any other URL. The issue is well explained [here](https://curl.haxx.se/docs/faq.html#file_URLs_containing_drive_le), except that libcurl sees an empty localhost instead of a drive letter; but the result is the same: part of the path is removed (in our case, the leading '/').

TLDR: the library strictly implements the URL specification while most of the time we don't want to specify a host for files.

src/channel.cpp
```diff
@@ -356,7 +356,8 @@ namespace mamba
                                                      const std::string& path)
     {
         std::string spath = std::string(rstrip(path, "/"));
-        std::string url = URLHandler().set_host(host).set_port(port).set_path(spath).url();
+        std::cout << "spath = " << spath << std::endl;
```
wolfv, line 359: happy to merge but this looks like a debug cout?

JohanMabille: ahem, let' sremove it ;)

PR #306, opened 2020-05-29
Initial implementation of QueryResult

JohanMabille:
This is an early preview of the new QueryResult class so we can review / comment.

### DIFF
--- a/include/output.hpp
+++ b/include/output.hpp
@@ -164,7 +164,7 @@ namespace mamba
             void add_row(const std::vector<FormattedString>& r);
             void add_rows(const std::string& header, const std::vector<std::vector<FormattedString>>& rs);
 
-            void print();
+            std::ostream& print(std::ostream& out);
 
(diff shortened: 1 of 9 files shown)


wolfv:
Looks good to me.
I think one question / problem we still have is how to collect the data:

Do we store everything as a tree? (search could be stored with everything having a nullptr as parent?). Or do we construct the tree on the fly from the given values?

JohanMabille:
My idea was to store everything as a list (meaning this is where the ownership is), and have list / tree view on it. The tree view is always built for depends and whoneeds, and on the fly for search.

Also we store a tree, not a graph, so not sure to get what you mean about nullptr as parent.

JohanMabille:
@wolfv I think this is ready to be merged. I've tested mamba repoquery (all commands) on xtensor, it works like a charm.

wolfv:
awesome

PR #331, opened 2020-06-13
correcting test for SOLVER_UPDATE

wolfv:


wolfv:
@xhochy this is the fix for the `--no-deps` & `--only-deps` issue you mentioned on the QuantStack chat. really sorry!

PR #347, opened 2020-06-24
Gracefully handle non-existent env

alanhdu:
Fix https://github.com/TheSnakePit/mamba/issues/346


wolfv:
Thanks! By the way, conda has the same deficiency, so if you were interested you could also contribute this fix to conda :)

PR #355, opened 2020-06-24
Directed graph implementation

JohanMabille:
@wolfv I still have a little issue when printing the tree (the prefix is not correct, I'll fix it when I have time in the week end), but this should give you an idea of the new query result. The "XXX already visited" has been removed from the table and the json output. It is still present in the tree output.

EDIT: issue fixed, this is ready for a review. I haven't added the possibility to filter out the "xxx already visited" in the tree/graph view, I can do it in a dedicated PR once this one has been merged (so it will be easier to retrieve diff if we experience some bug later).


wolfv:
This is what it looks like right now for me ... does it look like that for you as well?
On the right side is conda-tree, that we can use as a reference.

Screenshot from 2020-06-29 19-28-59

JohanMabille:
Mmmmn no. I have a segfault Oo. Can you print the output of `mamba repoquery depends xtensor` ?

JohanMabille:
That should work now, I fixed my noobiness.

wolfv:
nice, looks much better! 

I get a segfault with packages that aren't installed, e.g. `mamba repoquery depends coincbc -t`  (this wokrs fine with `mamba repoquery depends coincbc -t --all`

Maybe we can handle that more gracefully?
Everything else looks great now!

JohanMabille:
Fixed.

PR #356, opened 2020-06-25
initial work for cmd.exe activation

wolfv:
This is some initial work to make micromamba work on Windows. 

It successfully writes to the windows registry to add mciromamba activation to the autostart of the command prompt, and comes with a couple of batch scripts that are executed.

However, currently for some weird reason, this seems to break when getting the instance of a ChannelContext.
I need to figure out how to properly debug Windows.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,6 +6,7 @@ project (mamba)
 
 if (MSVC)
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /DNOMINMAX")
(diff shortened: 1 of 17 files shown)


dhirschfeld:
Looking forward to testing this out!

In case it's relevant/useful, there's an interesting PR in `conda` about JSON format activation:
https://github.com/conda/conda/pull/10045

wolfv:
hmm, interesting, that would be easy to achieve with our framework here as well :)

By the way, powershell is also done, and seems to work fine!

PR #361, opened 2020-06-26
Update README.md

tanimislam:
I have decided to make, as clear as I could, how to install packages using `mamba` instead of `conda`


JohanMabille:
Documentation only, no need to wait for CI

Issue #360, opened 2020-06-26
mamba env update removes packages from existing environment


minrk:
`mamba env update` should add specs from an `environment.yml` to an existing environment. However, it appears to *replace* the environment instead, ignoring what's already there and removing any packages not in the given `environment.yml`

Full repro steps

```yaml
# environment.yml
dependencies:
- xz
```

```
mamba create -n test zlib
mamba env update -n test -f environment.yml
mamba list -n test
```

This pattern is how repo2docker installs user packages, so I ran into this when testing if mamba is ready for another go over there, but this issue is a blocker.

---

Complete output, including demonstration that conda adds packages, while mamba removes them:

<details>

```
(base) minrk[14:51]/tmp $ mamba create -n test zlib
Looking for: ['zlib']

Transaction

  Prefix: /Users/minrk/conda/envs/test

  Updating specs:

   - zlib


  Package  Version  Build          Channel                  Size
──────────────────────────────────────────────────────────────────
  Install:
──────────────────────────────────────────────────────────────────

  zlib      1.2.11  h01d97ff_1006  conda-forge/osx-64     Cached

  Summary:

  Install: 1 packages

  Total download: 0  B

──────────────────────────────────────────────────────────────────

Preparing transaction: done
Verifying transaction: done
Executing transaction: done
#
# To activate this environment, use
#
#     $ conda activate test
#
# To deactivate an active environment, use
#
#     $ conda deactivate

(base) minrk[14:52]/tmp $ conda list -n test
# packages in environment at /Users/minrk/conda/envs/test:
#
# Name                    Version                   Build  Channel
zlib                      1.2.11            h01d97ff_1006    conda-forge

(base) minrk[14:52]/tmp $ cat environment.yml
dependencies:
  - xz
(base) minrk[14:52]/tmp $ conda env update -f environment.yml
Collecting package metadata (repodata.json): done
Solving environment: done
#
# To activate this environment, use
#
#     $ conda activate /Users/minrk/conda
#
# To deactivate an active environment, use
#
#     $ conda deactivate

(base) minrk[14:52]/tmp $ conda list
^C
CondaError: KeyboardInterrupt

(base) minrk[14:53]/tmp $ conda env update -n test -f environment.yml
Collecting package metadata (repodata.json): done
Solving environment: done
Preparing transaction: done
Verifying transaction: done
Executing transaction: done
#
# To activate this environment, use
#
#     $ conda activate test
#
# To deactivate an active environment, use
#
[...]

PR #371, opened 2020-06-30
support env update

wolfv:
this should help #360 

However, conda tries to _freeze_ the packages that are already installed, which we're not attempting here (we could do this with `SOLVER_LOCK` on each of the already installed specs).
Also, while the creation of new envs with pip packages works, the pip interop is currently disabled in mamba. So only the last environment should use pip, I guess (or the ones before shouldn't expect pip interop to work properly).

I will release a new version of mamba tonight, maybe you can give it a shot. Let me know if the mentioned points are blockers for you.


wolfv:
@minrk

PR #377, opened 2020-07-01
link libstdc++ and libgcc statically as well

wolfv:
Looks like this works really well (even worked on centos 6).


wolfv:
Thanks @jjhelmus :)

PR #393, opened 2020-07-10
Add copyright headers

SylvainCorlay:


CMakeLists.txt
```diff
@@ -1,3 +1,9 @@
+# Copyright (c) 2019, QuantStack and Mamba Contributors
```
JohanMabille, line 1: Should LICENSE.md be updated to have the same copyright date?

SylvainCorlay: Fixed. Good catch.

Issue #389, opened 2020-07-09
mamba run throws away exit code


analog-cbarber:
This is similar to the corresponding bug in conda (https://github.com/conda/conda/issues/8385).

The conda cli no longer returns an int but instead a Response class with stdout, stderr and rc fields.
The exit code is the rc field.

```python
>>> from conda.cli.main import generate_parser
>>> from conda.cli.main_run import execute as run
>>> parser = generate_parser()
>>> result = run(parser.parse_args(['run','false'], parser)
ERROR conda.cli.main_run:execute(32): Subprocess for 'conda run ['false']' command failed.  (See above for error)
>>> result
Response(stdout='', stderr='', rc=1)
>>> result.fc
1
```

mamba doesn't check for this return type so it ends up throwing away the return type:

https://github.com/TheSnakePit/mamba/blob/66fb193293f13bd8028ed02e86f1439ff66a45b8/mamba/mamba.py#L769-L771

that code should be extended to check for Response object and return result code.

wolfv:
do you want to open a PR? Would be wonderful

analog-cbarber:
Sure, when I have the time. The fix itself should be easy.

wolfv:
fixed with #395

PR #395, opened 2020-07-10
Fix issue #389: mamba run throws away exit code

analog-cbarber:
Did not see an easy way to add a test for this, but tested it manually.


wolfv:
Thanks!

PR #410, opened 2020-07-17
Rollback transaction when interrupted

### LINKED ISSUE #129: Implement CTRL+C and rollback consistently
CTRL+C during download doesn't work right now ... we need to figure out how to handle this correctly.

JohanMabille:
Fixes #129


wolfv:
Looks great!

PR #421, opened 2020-07-22
Implemented interruptible threads

JohanMabille:


wolfv:
@JohanMabille do you think we could use https://github.com/taskflow/taskflow and the run until true with the condition?
I was thinking about using taskflow in the future anyways, for adding multi-threaded linking and MD5 sum verification of environments...

JohanMabille:
That might be a solution although I don't know how it will interact with the curl multi download feature. Threads for extracting the packages after download are launched from callbacks, this is equivalent to changing the graph while we execute it.

JohanMabille:
@wolfv this is ready for a review. The idea is the following:

- When running a single thread, or many threads that do not acquire any ressources that should be freed upon interruption, we rely on a simple interruption handler as before.
- If many threads are spawned, which will acquire ressources that must be freed upon interruption, we switch to the second mode, where a cleaner thread is spawn and wait for a SIGINT signal. If he receives one, it sets the sig_interrupted_flag, and wait ofr other threads to finish before invoking the cleanup callback. To get the desired behavior, the user should not instantiated `std::thread` directly, but rather `mamba::thread`, which can be interrupted. A new function, `interrupt_point` checks if the  `sig_interrupted` flag has been set. This allows to control where the threads can be interrupted.
- the interruption guard allows to easily switch between the two modes.

The only drawback for now is that features such as `std::async` cannot be used in interruptible mode. However, I don't think this is a big deal, and I can implement them later if this is really required.

wolfv:
Looks great!

Do you think we should add the interrupt checks in the archive unpacking? https://github.com/TheSnakePit/mamba/blob/master/src/package_handling.cpp

This is running inside the thread after download is finished (but mutexed in such a way that only one thread unpacks an archive at a time).

wolfv:
there is one `while(true)` and one `for(;;)` loop

JohanMabille:
Ha indeed, that would be usefull for large packages.

src/package_handling.cpp
```diff
@@ -86,6 +88,7 @@ namespace mamba
         
         for (;;)
         {
+            interruption_point();
```
wolfv, line 91: can we delete the partially extracted data?

wolfv:
I think after interrupting extraction, we should try to delete the extracted files.
The main reason is that we currently don't check if the extracted files match up with our expectation in the micromamba installation process.

JohanMabille:
Totally agree!

JohanMabille:
I actually need to fix additional problems before we merge it:

- interrupting while mamba waits for a confirmation create an inifinite loop. Interrupting again stops the loop but not mamba, only a kill can stop it.
- interrupting while extracting an archive randomly results in a memory corruption / double free.

wolfv:
ok I'll hold off with merging until your GO

JohanMabille:
@wolfv this should be fine now.

wolfv:
nice!

PR #430, opened 2020-07-24
Added unit test for thread_utils

JohanMabille:


include/thread_utils.hpp
```diff
@@ -17,6 +17,10 @@
 
 namespace mamba
 {
+
+#ifdef MAMBA_TEST_SUITE
```
wolfv, line 21: leftover?

JohanMabille: oups

PR #431, opened 2020-07-24
fix noarch linking on Win

wolfv:
For noarhc entrypoints we need to hard-link or copy `conda.exe` and rename it.
Clearly this is hacky (but that's how conda does it).
We should probably ship the contents of conda.exe as binary buffer inside mamba / micromamba on Windows to create this file even without conda being installed.

### DIFF
--- a/src/link.cpp
+++ b/src/link.cpp
@@ -112,7 +112,7 @@ namespace mamba
         win_script += "-script.py";
         std::ofstream out_file(m_context->target_prefix / win_script);
         #else
-        std::ofstream out_file(target);        
+        std::ofstream out_file(target);
         #endif
 
         fs::path python_path;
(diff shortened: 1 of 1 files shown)


wolfv:
@dhirschfeld do you know (by any chance) how these magical `.exe` entrypoints work on Windows?
As mentioned above, python entrypoints are apparently `.exe` files with the same name that somehow then start python and run the script (with the same name in the same folder).

wolfv:
Also while this helps to make the installation of noarch packages work, there is another issue that comes up when unlinking packages. The one where a file that is in use cannot be deleted.

The cool thing is that this is all coming up with the current `boa` work, so that's fruitful at least :) 
Also this doesn't impact regular "mamba"-operation since it's using Conda's link implementation.

dhirschfeld:
> do you know (by any chance) how these magical .exe entrypoints work on Windows?

IIRC before exe's the entrypoints were just `bat` files that called `..\python.exe name-script.py %*`. Now they're actually executable binaries but I think they just do exactly the same - i.e. the `name-script.py` files still sit alongside the `name.exe` binaries so I think the exes still just call `..\python.exe name-script.py %*`. I guess they're just a minimal c program that calls `spawn` (or something like that)

dhirschfeld:
> The one where a file that is in use cannot be deleted.

This is a constant source of pain on Windows 😬. I think that the native `MoveFileEx` api is supposed to cope with that (see [`os.replace`](https://github.com/python/cpython/blob/master/Modules/posixmodule.c#L4376)).

Looks like this was implemented in `conda` in https://github.com/conda/conda/pull/8024

PR #432, opened 2020-07-24
Does not update python when updating enrivonment

JohanMabille:
This should allow binder to test/use mamba.

cc @SylvainCorlay @wolfv

### DIFF
--- a/mamba/mamba_env.py
+++ b/mamba/mamba_env.py
@@ -75,8 +75,15 @@ def mamba_install(prefix, specs, args, env, *_, **kwargs):
         repo.set_priority(priority, subpriority)
         repos.append(repo)
 
+    
     solver = api.Solver(pool, solver_options)
     solver.add_jobs(specs, api.SOLVER_INSTALL)
+
+    python_version = [irec.version for irec in installed_pkg_recs if irec.name == 'python']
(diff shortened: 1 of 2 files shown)


mamba/mamba_env.py
```diff
@@ -75,8 +75,15 @@ def mamba_install(prefix, specs, args, env, *_, **kwargs):
         repo.set_priority(priority, subpriority)
         repos.append(repo)
 
+    
     solver = api.Solver(pool, solver_options)
     solver.add_jobs(specs, api.SOLVER_INSTALL)
+
+    python_version = [irec.version for irec in installed_pkg_recs if irec.name == 'python']
+    if python_version:
+        python_constraint = MatchSpec('python==' + version).conda_build_form()
```
wolfv, line 84: yeah this code wasn't tested, right?
version looks undefined here ;)

I have opened a PR with the implementation I had on disk and forgot to commit, sorry about that!

#433

wolfv:
This one is superseded by #433

mamba/mamba_env.py
JohanMabille: Oo I tested it locally and it worked. I guess I forgot to push some commits :/

PR #438, opened 2020-08-02
Add pre-commit and black

marcelotrevisani:
For now I am proposing to add ``pre-commit`` for the python interface.
I am also adding a new github workflow to be able to run all the linters
Removed unused imports
Fixed some small errors as well
using ``black`` to format the code as well

To use ``pre-commit`` in your environment just install it and in the root folder run ``pre-commit install``. After that every-time you try to commit something it will run all the checks.


mamba/_version.py
```diff
@@ -1,3 +1,2 @@
-# -*- coding: utf-8 -*-
 version_info = (0, 4, 4)
-__version__ = '.'.join(map(str, version_info))
+__version__ = ".".join(map(str, version_info))
```
marcelotrevisani, line 2: it would be nice to change this part and add ``setuptools_scm``
I can do it in another PR if you agree to that

wolfv: Can we keep the version_info tuple though? We parse it in CMake to put the same version number into the mamba_lib / micromamba binary here: https://github.com/TheSnakePit/mamba/blob/5e801e3ecebf98b1c3655938eee051850dcfb7b0/CMakeLists.txt#L51-L60

wolfv:
Looks great, I love this. 
We just need to fix the version parsing in CMake apparently: https://github.com/TheSnakePit/mamba/pull/438/checks?check_run_id=936863596#step:6:38

Let me know if you want to do it or if I should fix that on Monday.

mamba/_version.py
marcelotrevisani: Sure. I will change it, no problem
For a new PR I was thinking to add ``setuptools_scm`` and get the version at runtime calling the python function from the C++ code. As ``pybind11`` is already a dependency that should be simple to do.
What do you think?

marcelotrevisani:
> Looks great, I love this.
> We just need to fix the version parsing in CMake apparently: https://github.com/TheSnakePit/mamba/pull/438/checks?check_run_id=936863596#step:6:38
> 
> Let me know if you want to do it or if I should fix that on Monday.

I will do it later, no problem! :)

wolfv:
do you want to squash the commits a little? then I'll merge :+1: 
Thanks a lot for the contribution

marcelotrevisani:
> do you want to squash the commits a little? then I'll merge 
> Thanks a lot for the contribution

Done! :)

PR #449, opened 2020-08-13
Makes mamba easy to find to other projects dependable on it

marimeireles:
* I tried to give name to sessions (like, this is the build part, this is the linking part) to make it more organized;
* Created a new dir `include/mamba` to add the libs in to follow other QS's projects pattern;
* Added a version 1.0.0 to the binary, but I'm not sure if this is right. Was just copying the pattern on `xeus`

Thanks to @SylvainCorlay and @JohanMabille, maybe you can review it too?


SylvainCorlay:
Well done! Happy to see mamba's cmake cleaned up. I will be reviewing this asap.

PR #459, opened 2020-08-19
Update header

SylvainCorlay:
New header


SylvainCorlay:
It is a bit more neutral, and the black & white theme matches conda-forge.

PR #470, opened 2020-08-25
Add list to micromamba

marimeireles:
The printing part is not my favorite at all, but if we're using this `multimap` thing that I found very handy I think it's the way to go.
Let me know if you have different ideas to do it with maps or vectors. I thought on different implementations, but all of them looked equally non ideal in my head! :/

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -70,7 +70,8 @@ struct formatted_pkg
     std::string name, version, build, channel;
 };
 
-bool compareAlphabetically(const formatted_pkg &a, const formatted_pkg &b)
+bool
+compareAlphabetically(const formatted_pkg& a, const formatted_pkg& b)
 {
     return a.name < b.name;
 }
@@ -364,7 +365,7 @@ list_packages()
 
     std::vector<formatted_pkg> packages;
 
-    //order list of packages from prefix_data by alphabetical order
+    // order list of packages from prefix_data by alphabetical order
     for (auto package : prefix_data.m_package_records)
     {
         formatted_pkgs.name = package.second.name;
@@ -384,7 +385,7 @@ list_packages()
 
     std::sort(packages.begin(), packages.end(), compareAlphabetically);
 
-    //format and print table
+    // format and print table
     printers::Table t({ "Name", "Version", "Build", "Channel" });
     t.set_alignment({ printers::alignment::left,
                       printers::alignment::left,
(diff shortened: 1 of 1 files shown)


src/main.cpp
```diff
@@ -338,6 +339,63 @@ install_specs(const std::vector<std::string>& specs, bool create_env = false)
...
+    auto& ctx = Context::instance();
+    PrefixData prefix_data(ctx.target_prefix);
+    prefix_data.load();
+
+    std::multimap<std::string, std::string, std::less<std::string>> ordered_packages;
+
+    std::cout << "List of packages in environment: " << ctx.target_prefix << std::endl;
+
+    //order list of packages from prefix_data by alphabetical order
+    for (auto package : prefix_data.m_package_records)
```
wolfv, line 354: Nice work! 

I wonder if it wouldn't be better to either use a std::tuple or std::array as a container for the 4 elements, or as another very nice option to use a struct.

you can make a new struct definition and I think it can be local to the function as well, e.g.


struct formatted_pkg
{
	std::string name, version, channel;
}

and then have the map be std::map<std::string, formatted_pkg> etc.

src/main.cpp
```diff
@@ -381,8 +439,20 @@ init_create_parser(CLI::App* subcom)
     });
 }
 
-std::string
-version()
+void
+init_list_parser(CLI::App* subcom)
+{
+    subcom->add_option("specs", create_options.specs, "Specs to install into the new environment");
+    init_network_parser(subcom);
```
wolfv, line 446: do we need this? I thought we're printing the local packages here so I guess no need for network options?

src/main.cpp
```diff
@@ -4,12 +4,15 @@
 //
 // The full license is in the file LICENSE, distributed with this software.
 
+#include <algorithm>
```
marimeireles, line 7: Should prbb remove this? We're keeping clean header wise right.

src/main.cpp
```diff
@@ -62,6 +65,17 @@ static struct
...
+{
+    std::string name, version, build, channel;
+};
+
+bool
+compareAlphabetically(const formatted_pkg& a, const formatted_pkg& b)
+{
+    return a.name < b.name;
+}
+
```
marimeireles, line 78: Ist das ok?
Had to write the struct here to use in this function! :/

wolfv: fine for now :)

PR #473, opened 2020-08-26
Added ghcFilesystem dependency

JohanMabille:


wolfv:
I think what we currently have seems to work reasonably well with the conda deps.

PR #475, opened 2020-08-27
Add missing libs

marimeireles:
Oops! Seems like we forgot to add some mamba libs.

@JohanMabille


marimeireles:
Hum! I don't get why this test is failing: https://github.com/TheSnakePit/mamba/runs/1037744655
Is it because we were waiting for segfault and because we didn't have the lib imported correctly it was segfaulting maybe?

wolfv:
test failure seems unrelated :)

PR #479, opened 2020-08-29
Use ~/.netrc if present

skumagai:
When private packages are hosted in a private package repository, the repository can be behind access cotrol methods. This PR adds the simplest form with .netrc file. By the way, conda also supports .netrc too.


wolfv:
Thanks for your contribution!

PR #483, opened 2020-09-03
fix prefix reading and add remove function to micromamba

wolfv:


marimeireles:
We're having this on Windows:

```
  C:\Miniconda\envs\mamba-tests\Library\include\solv/repo.h(150): warning C4244: 'argument': conversion from '__int64' to 'Id', possible loss of data
    src/package_info.cpp(182): error C2440: 'static_cast': cannot convert from 'nlohmann::basic_json<std::map,std::vector,std::string,bool,int64_t,uint64_t,double,std::allocator,nlohmann::adl_serializer,std::vector<uint8_t,std::allocator<_Ty>>>' to 'std::vector<std::string,std::allocator<_Ty>>'
            with
```

Doing something like this:
```
 for (auto d : j["depends"])
{
         depends.push_back(d);
}
```
Works for me, but looking around the web I wondered if this https://github.com/nlohmann/json/issues/861 is not the actual problem, this issue is closed, but I found another one that's still open and assess a similar problem.

Did you fix the problem that was grabbing incomplete data?

PR #506, opened 2020-09-17
try micromamba installation

wolfv:


dhirschfeld:
you could try the shortpath name:
```
C:/PROGRA~1/7-Zip/7z.exe
```

dhirschfeld:
```powershell
> (New-Object -ComObject Scripting.FileSystemObject).GetFolder("C:\Program Files\7-Zip").ShortPath
C:\PROGRA~1\7-Zip
```

wolfv:
thanks @dhirschfeld I just saw that `tar.exe` is also available. Hopefully that works :)

dhirschfeld:
tar and curl are both available on modern windows

wolfv:
hmm seems to stall though ..

wolfv:
thanks dave, your trick actually worked with 7zip! :)

PR #515, opened 2020-09-24
Update the mamba banner to point to mamba-org

jtpio:
Now that it shows up on the https://mybinder.org front page :) 

image


wolfv:
thanks!

PR #518, opened 2020-09-24
add creation of conda and tar.bz2 packages

wolfv:
This adds some code to create .tar.bz2 and .conda packages from mamba. This will be useful in boa. Maybe this code should actually live in `boa`... 

Also we'll soon add transmutation to boa, and here we can control the compression for zstd. In conda-forge-package-handling the compression is hard coded to a very high value which makes transmutation very expensive.


wolfv:
Note we should also add a build of zstd with multithreading enabled.

PR #541, opened 2020-10-20
Use cpp-terminal

martinRenou:


martinRenou:
For some reason `ctrl + c` does not work, and I am not really happy with the UX

wolfv:
it's possible that CTRL+C is captured by our interrupt handling. If that's the case the global `is_sig_interrupted` variable should be true.
We could also see if we can detach that signal handler during `prompt`.

martinRenou:
@wolfv do you see cpp-terminal being used in other places in Mamba?

I actually find it a bit more low level than I thought. And it feels a bit overkill in comparison to our simple `getline` solution.

martinRenou:
I realize I am mostly copying what was done in the `prompt` example in cpp-terminal. So instead of doing this I think it would be nicer to use the `prompt` example directly, which will fit our needs. 

See https://github.com/certik/terminal/pull/49 for a PR adding prompt to the public API in cpp-terminal

martinRenou:
@wolfv this is now using cpp-terminal master. It's a lot cleaner than what I was starting to do yesterday :)

martinRenou:
I wonder if we should not have the terminal instance as a property of the `Console` class, so it gets reused at other places if needed, and a new one is not created every time there is a prompt call.

wolfv:
yes, I think there are other places where we could use Terminal. For example, we have our own cursor - movement code in the output.hpp file, that we could use from terminal-hpp...

I don't know if you're interested in tackling that? We can do it later, too...

PR #609, opened 2020-11-22
fix explicit noarch compilation

wolfv:
@mariusvniekerk this should additionally fix the issues with noarch installation in explicit envs.


wolfv:
Tested, and works as expected.

PR #629, opened 2020-12-11
ENH build and release micromamba

beckermr:
This PR builds and releases micromamba when the repo is tagged on master.

It will upload 2 artifacts per platform. The first will be the actual executable. The second is a .tar.bz2 archive that looks like the conda package. This can be used in place of the download server that redirects to anaconda.org.

To do:

- [x] add win-64 build
- [x] ~add osx-arm64 build~
- [x] add job that depends on the rest that turns the pre-release into a release


beckermr:
cc @wolfv

beckermr:
OK @wolfv, this one is ready for review. I have skipped the osx-arm64 build for now since it is difficult. I am also basically ripping the windows build from the feedstock. No hard feelings if you don't want this extra cruft here, but this is one way to do the static builds.

.github/workflows/release-micromamba.yml
```diff
@@ -0,0 +1,202 @@
...
+            -DCMAKE_INSTALL_PREFIX=$CONDA_PREFIX \
+            -DCMAKE_BUILD_TYPE="Release" \
+            -DBUILD_EXE=ON \
+            -DBUILD_BINDINGS=OFF \
+            -DBUILD_STATIC=ON \
+            -DBUILD_SHARED=OFF \
+            -DSTATIC_DEPENDENCIES=ON \
+            -DUSE_VENDORED_CLI11=ON
+
+          make -j2
```
beckermr, line 70:
```suggestion
          make -j2
          ./micromamba -h
```

README.md
```diff
@@ -168,11 +168,21 @@ cmake .. \
...
+
+When tags are pushed to the repo, a CI job will make a pre-release on GitHub
+with the built `micromamba` artifacts. You then need to use the GitHub UI to
+remove the pre-release status and release the artifacts.
+
+Note that the pre-release is essential since otherwise the GitHub links to the
+latest release will point to partial releases, breaking any workflows that
+depend on those links.
+
+## Support us
```
beckermr, line 181:
```suggestion
## Support Us
```

.github/workflows/release-micromamba.yml
```diff
@@ -0,0 +1,194 @@
...
+          conda list
+
+      - name: build micromamba
+        shell: bash -l {0}
+        run: |
+          conda activate mamba-tests
+          mkdir -p ~/.conda
+          touch ~/.conda/environments.txt
+
+          curl -Ls https://github.com/CLIUtils/CLI11/releases/download/v1.9.1/CLI11.hpp -o include/mamba/CLI.hpp
```
wolfv, line 52: Can we use this `CLI11.hpp` file for win and unix? https://github.com/conda-forge/micromamba-feedstock/blob/master/recipe/CLI11.hpp

This is a later master version of CLI11, and the (important) difference is that one can put multiple `-c` arguments anywhere on the command line. The latest release can only handle `-c` arguments properly when they come at the end of the CLI arguments.

I asked the CLI team for a release but they are quite busy. cc @henryiii

wolfv:
awesome stuff @beckermr! 

Would you mind squashing the history a bit?
Happy to merge after we figure out the CLI11 stuff.

beckermr:
I'm looking at this and wondering if maybe we should try to build the feedstock itself by cloning, adjusting the recipe, and building. I'll take a look later.

wolfv:
Hi @beckermr I have some micromamba source builds here now: https://github.com/mamba-org/boa-forge/releases

This includes a build (cached) of all dependencies of micromamba. One nice upside of that is that we can disable features at compile time (e.g. we don't need all features of cURL activated for the micromamba build, thus shrinking the binary size a little bit).

This runs in github actions and we also have an autotick bot for new versions :)

https://github.com/mamba-org/boa-forge/actions/runs/509548489

Let me know what you think!

beckermr:
As long as it produces a stable url and will be maintained I am happy to use that over this PR. Do what you think is best!

wolfv:
hey @beckermr I finally got the release pipeline thing up & running over at "boa-forge", and we get automatically uploaded assets from "latest everything":

https://github.com/mamba-org/boa-forge/releases

It's not really a nightly release as we're not building master but the latest tagged version. Will have to see if we change this up somehow. 

It could be interesting to test these binaries. 

We can get them through API, e.g.

`curl -H "Accept: application/vnd.github.v3+json" https://api.github.com/repos/mamba-org/boa-forge/releases`

wolfv:
I think we have a pretty elaborate Github Action with many after-release tests over at boa-forge now. I am considering t rename it to `micromamba-forge` :)

I've also recently added cross-compiled linux-aarch64 and osx-arm64 binaries.

Going to close this.

PR #661, opened 2021-01-19
Fix virtual package handling in member_env

cbalioglu:
The command `mamba env create -f <path/to/environment.yml>` fails in case one of the packages in the dependency closure has a dependency to a virtual package such as `__glibc` or `__cuda`.

The `if` condition that has been removed in this PR was taking virtual packages into account only when called with the `mamba env update` verb. With this PR the condition has been removed as virtual packages are considered to be always available regardless whether an environment is being created or updated.


wolfv:
thanks a lot!

cbalioglu:
You are welcome! Thanks for merging it!

wolfv:
It's released with mamba 0.7.10

PR #672, opened 2021-02-02
raise sigint instead of sigterm

wolfv:


test/test_thread_utils.cpp
```diff
@@ -6,35 +6,85 @@
 
 namespace mamba
 {
-    TEST(thread_utils, interrupt)
+    namespace
     {
-#if defined(__unix__) || (defined(__APPLE__) && defined(__MACH__))
+        std::mutex res_mutex;
         int res = 0;
```
JohanMabille, line 12: This should be a local variable to the test_interruption_guard function

test/test_thread_utils.cpp
```diff
@@ -6,35 +6,85 @@
...
+        res = 0;
+        test_interruption_guard(true);
+        EXPECT_EQ(res, -95);
+    }
+
+    TEST(thread_utils, no_interrupt)
+    {
+        res = 0;
+        test_interruption_guard(false);
+        wait_for_all_threads();
```
JohanMabille, line 64: This is a bit annoying, this should be handled automatically in the interruption mechanism. Otherwise, the main thread needs to wait for the parallel task to complete before it resumes.

Maybe the `wait_for_all_threads` should always be called in the `interruption_guard` destructor?

wolfv: it is handled when we do have an interruption (but not in the `false` case).

test/test_thread_utils.cpp
wolfv: fixde

include/mamba/thread_utils.hpp
```diff
@@ -51,15 +51,11 @@ namespace mamba
...
     void wait_for_all_threads();
-    void notify_cleanup();
-
-    // Should be called by the main thread to ensure
-    // it does not exit before the cleaning thread has
-    // terminated.
-    void wait_for_cleanup();
 
     void register_cleaning_thread_id(std::thread::native_handle_type);
     std::thread::native_handle_type get_cleaning_thread_id();
```
JohanMabille, line 56: I guess these two methods can be removed too

wolfv: right.

src/thread_utils.cpp
```diff
@@ -19,13 +18,58 @@ namespace mamba
...
+    }
+
+    int default_signal_handler(sigset_t sigset)
+    {
+        int signum = 0;
+        // wait until a signal is delivered:
+        sigwait(&sigset, &signum);
+        sig_interrupted.store(true);
+        // notify all waiting workers to check their predicate:
+        sig_interrupted_cv.notify_all();
```
JohanMabille, line 50: I think this can be removed, nobody never waits on this conditional variable

PR #673, opened 2021-02-02
Fixed interruption mechanics

JohanMabille:
This assumes we don't check `is_sig_interrupted` outside of the scope of an `interruption_guard` object.


JohanMabille:
superseded with #672

Issue #686, opened 2021-02-08
mamba fails to use cached repos it the tar file has been cleaned


s22chan:
starting with https://github.com/mamba-org/mamba/commit/d361e1e2d57b33a360314a5926b391235ac7e868

it appears that all cached packages must be extracted, but if an  environment has called `conda clean -a`, then the validation code doesn't check if the packages are already extracted and throws an EXTRACT_ERROR with the missing tar.bz2 file.

I'm not sure how you want to handle detection of extracted packages (exists? timestamp?) so it wouldn't be an obvious fix for me to implement.

wolfv:
I think the state of this is much better now.

PR #693, opened 2021-02-10
add a simple reposerver which supports the authentication methods

wolfv:
@s22chan I am planning to use this simple server implementation in the tests, to make sure that we will not have any further regressions.


s22chan (approved):
Thanks @wolfv, that'd be a great addition. I'm not sure if you saw my comment though but I falsely attributed my build failures to the new basic auth change that got integrated. 

It turned out the transaction error was due to the extraction portion at https://github.com/mamba-org/mamba/issues/686. Still not sure why I didn't see any logs related to that at with `-vvv` but maybe it was noise.

Still thanks for this, it'll eliminate one thing to check if any regressions happen.

PR #702, opened 2021-02-22
read config for rc files

adriendelsalle:
Description
--

Add capability to read configuration from various places : 
- config files (`.condarc` files) : located at `prefix`, user home level, sys level or specified by env variable
- CLI parameters
- env variables

### DIFF
--- a/include/mamba/config.hpp
+++ b/include/mamba/config.hpp
@@ -1,0 +1,73 @@
+// Copyright (c) 2019, QuantStack and Mamba Contributors
+//
+// Distributed under the terms of the BSD 3-Clause License.
+//
+// The full license is in the file LICENSE, distributed with this software.
+
+#ifndef MAMBA_CONFIG_HPP
+#define MAMBA_CONFIG_HPP
+
+#include "mamba/context.hpp"
+#include "environment.hpp"
+#include "mamba/mamba_fs.hpp"
+#include "mamba/output.hpp"
+#include "mamba/util.hpp"
+
+#include <yaml-cpp/yaml.h>
+
+
+namespace mamba
+{
+    class Configurable
+    {
+    public:
+        Configurable();
+
+        const YAML::Node& get_config();
+        std::vector<fs::path> get_sources();
+        std::vector<fs::path> get_valid_sources();
+
+        std::string dump(bool show_sources = false);
+
+    protected:
+        Configurable(std::string unique_source);
+
+        static Context& ctx()
+        {
+            return Context::instance();
+        };
+
+        static bool has_config_extension(const std::string& file);
+        static bool looks_config_file(const fs::path& path);
+
+        static YAML::Node load_rc_file(const fs::path& file);
+
+        void load_config();
+        void load_config_files();
+        void update_sources();
+
+        void build_prepend_seq(const std::vector<std::shared_ptr<YAML::Node>>& configs,
+                               const std::string& key,
+                               const std::map<std::shared_ptr<YAML::Node>, std::string>& node2src,
+                               YAML::Node& result,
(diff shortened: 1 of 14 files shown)


include/mamba/config.hpp
```diff
@@ -0,0 +1,73 @@
...
+            for (auto& channel : new_channels)
+            {   // Remove duplicates
+                auto c = std::find(ctx.channels.begin(), ctx.channels.end(), channel);
+                if (c != ctx.channels.end())
+                {
+                    ctx.channels.erase(c);
+                }
+            }
+
+            ctx.channels.insert(ctx.channels.begin(), new_channels.begin(), new_channels.end());
```
wolfv, line 36: I think it would be better to have this function somewhere else, because we need to merge channels from the command line as well.

Also FYI there is the `--override-channels` option on the `conda` command line that we'll have to handle as well.

Looks good though!

wolfv: Also I just realized that there is an additional complication because the same channel can appear in "different" forms:

- conda-forge
- https://conda.anaconda.org/conda-forge

is the same channel. So that function should probably normalize the channel names to URLs and then make sure that they are unique and sorted in the right way.

Anyways, we can keep that also for later.

wolfv: FYI we have the code for this normalization in channel.hpp

adriendelsalle: Thanks a lot for your comments @wolfv !
I will improve this first shot to handle those cases

wolfv:
it's best to install `pre-commit` (e.g. from conda-forge) and use `pre-commit run --all` and/or install the hooks `precommit install-hooks` or somehing like that.

wolfv:
I quickly had a look and I think we should only do the conf file reading in micromamba, not `mamba-mamba` since there it's done by conda. 
So we can remove the windows error by not including the files that use `yaml-cpp`

adriendelsalle:
This should be ready for review @wolfv

src/channel.cpp
```diff
@@ -641,8 +641,10 @@ namespace mamba
 
     Channel ChannelContext::build_channel_alias()
     {
+        auto& ctx = Context::instance();
+        std::string alias = ctx.channel_alias != "" ? ctx.channel_alias : DEFAULT_CHANNEL_ALIAS;
```
wolfv, line 645: there is also `ctx.channel_alias.empty()` to test if a string has any value.

src/config.cpp
```diff
@@ -0,0 +1,329 @@
...
+    }
+
+    bool Configurable::has_config_extension(const std::string& file)
+    {
+#ifdef _WIN32
+        std::string sep = "\\";
+#else
+        std::string sep = "/";
+#endif
+        return rsplit(file, sep, 1).back() == ".condarc"
```
wolfv, line 72: can we not use `fs::path(...).filename()` for this?

src/config.cpp
```diff
@@ -0,0 +1,329 @@
...
+               || rsplit(file, sep, 1).back() == ".mambarc" || ends_with(file, ".yml")
+               || ends_with(file, ".yaml");
+    }
+
+    bool Configurable::looks_config_file(const fs::path& path)
```
wolfv, line 77: function name coudl be more descriptive (e.g. `is_config_file`)

src/config.cpp
wolfv: there is aslo `.extension()` or someething like that to get the file extension.

wolfv:
awesome! looks really great, just these two minor comments and it's good to go!

adriendelsalle:
Thanks! I did the suggested modifications!

Issue #631, opened 2020-12-14
mamba env update installing into wrong path


buhrmann:
Hi, when using `mamba env update -n myenv --file requirements.yml`, and in an active `myenv` environment, the command creates a new nested environment. E.g. if my current environment is at `miniforge3/env/myenv`, then the command installs the desired packages in `miniforge3/env/myenv/env/myenv`,  which is not what conda does, and probably not what's expected. When no environment is activated (i.e. base), the update works correctly.

wolfv:
Do you have `mamba` installed in `myenv`? 

mamba needs to be _only_ in the base environment currently.

buhrmann:
Yes, that could be it. Sorry, didn't know!

wolfv:
yeah, it's a limitation of the way mamba and conda work together.

buhrmann:
Ok, thanks. I've got another problem with `mamba env update`, not sure whether to open a new issue. Basically, it seems to ignore the pinned file in `myenv/conda-meta/pinned`. If I do a normal `mamba install ... some_pkg`, in contrast, it is respected. I also tried with `mamba install --file requirements.yml`, but that command fails with`CondaValueError: could not parse 'name: myeny ...' and it seems it doesn't work with the usual conda yml structure (starting with name: prefix)?

wolfv:
indeed, there might be an issue since `mamba env` and `mamba create` are implemented differently (it's a historic artifact from conda ... )

PRs welcome!

buhrmann:
Sorry, just to confirm... there is currently no way to make `mamba env` respect the pinned file, nor for `mamba install` to consume a conda environment file, right? Just asking because the latter seems to be possible with micromamba, e.g.

wolfv:
yes, micromamba is a completley different implementation for the CLI where I removed the weird differences between "conda env" and "conda create".

For conda & mamba, it's true that `conda env` handles `yaml` files, and `conda create` text or explicit env files.

Disrespecting the pinnings is something that can be fixed in `mamba_env.py` (https://github.com/mamba-org/mamba/blob/master/mamba/mamba_env.py) by copying what we're doing in `mamba.py` to handle pinned specs.

buhrmann:
Ok, thanks. I was trying to write some install scripts that would use micromamba in production docker images, and mamba in local installs. But seems that for now that isn't possible since one only accepts conda environment yaml files, and the other only plain text files.

Haydnspass:
[...]

PR #709, opened 2021-02-24
[DOC]: Make explicit installation into base env (closes #708)

### LINKED ISSUE #708: Update README to explain base env installation requirement
I would like to submit a simple PR to make this point (https://github.com/mamba-org/mamba/issues/631#issuecomment-744407332) explicit in the docs.

It's not immediately apparent that mamba *must* be installed into the `base` Conda env. I was also caught by the same `mamba env update -n foo -f bar.yml` gotcha as #631, because experience has taught me the hard way that I should keep my base env as minimal as possible.

ralexx:
It's not immediately apparent from the project README that mamba must be installed into the base Conda env. I was also caught by the same `mamba env update -n foo -f bar.yml` gotcha as https://github.com/mamba-org/mamba/issues/631#issuecomment-744407332.

Since the project documentation site is not linked from the project README, this PR updates `README.md` to reflect the installation instructions in `docs/getting_started.md`, and adds a note to the doc page that foreshadows the unexpected behavior of #631.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,10 +36,10 @@ The ecosystem also consists of `quetz`, an open source conda package server and 
 
 ### Installation
 
-It's advised to install mamba from conda-forge. If you already have conda:
+It's advised to install mamba from conda-forge. If you already have conda, install mamba into the base environment:
 
 ```
-conda install mamba -c conda-forge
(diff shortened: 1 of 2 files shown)


wolfv:
Thanks!

PR #710, opened 2021-02-25
Pin packages

adriendelsalle:
Description
--

Add capability to pin packages:
- from rc files
- from `<prefix>/conda-meta/pinned` file

Also auto-pin python version if already installed in prefix and specs does not include python at least pinned with minor version.

Add `--no-pin` flag to `create` and `install` CLI sub-commands to disable pinned specs from files (rc or prefix located).

### DIFF
--- a/src/micromamba/main.cpp
+++ b/src/micromamba/main.cpp
@@ -454,7 +454,7 @@ int RETRY_SUBDIR_FETCH = 1 << 0;
 int RETRY_SOLVE_ERROR = 1 << 1;
 
 void
-install_specs(std::vector<std::string>& specs, bool create_env = false, int is_retry = 0)
+install_specs(const std::vector<std::string>& specs, bool create_env = false, int is_retry = 0)
 {
     auto& ctx = Context::instance();
 
--- a/src/micromamba/main.cpp
+++ b/src/micromamba/main.cpp
@@ -545,13 +545,6 @@ install_specs(std::vector<std::string>& specs, bool create_env = false, int is_r
     PrefixData prefix_data(ctx.target_prefix);
     prefix_data.load();
 
-    if (!create_options.no_pin)
-    {
-        pin_file_specs(prefix_data.path() / "conda-meta/pinned", specs);
-        pin_config_specs(ctx.pinned_packages, specs);
-    }
-    pin_python_spec(prefix_data, specs);
-
     auto repo = MRepo(pool, prefix_data);
     repos.push_back(repo);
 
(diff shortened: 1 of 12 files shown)


src/pinning.cpp
```diff
@@ -0,0 +1,71 @@
...
+namespace mamba
+{
+    void pin_python_spec(const PrefixData& prefix_data, std::vector<std::string>& specs)
+    {
+        bool py_found = false;
+        std::string py_spec = "";
+
+        for (auto spec : specs)
+        {
+            if (spec.find("python") != std::string::npos)
```
wolfv, line 21: I am not sure this works. Can we not use the MatchSpec facilities for proper parsing?

wolfv:
Note:

this is how it's done from the Python side: https://github.com/mamba-org/mamba/blob/53e31e4c4973e5b715ae00f6416907be2cae9fd3/mamba/mamba.py#L537-L538

src/pinning.cpp
adriendelsalle: fixed!

adriendelsalle:
> Note:
> 
> this is how it's done from the Python side:
> 
> https://github.com/mamba-org/mamba/blob/53e31e4c4973e5b715ae00f6416907be2cae9fd3/mamba/mamba.py#L537-L538

Thanks @wolfv ! I missed that point and now add pins to the solver.

src/micromamba/main.cpp
```diff
@@ -599,6 +602,18 @@ install_specs(const std::vector<std::string>& specs, bool create_env = false, in
     MSolver solver(pool, { { SOLVER_FLAG_ALLOW_DOWNGRADE, 1 } });
     solver.add_jobs(create_options.specs, SOLVER_INSTALL);
 
+    if (!create_options.no_pin)
+    {
+        solver.add_pins(file_pins(prefix_data.path() / "conda-meta/pinned"));
```
wolfv, line 607: this is not cross-platform. 

Use `...path() / "conda-meta" / "pinned"`

src/pinning.cpp
```diff
@@ -0,0 +1,57 @@
...
+
+namespace mamba
+{
+    std::string python_pin(const PrefixData& prefix_data, const std::vector<std::string>& specs)
+    {
+        std::string pin = "";
+        std::string py_version;
+        MatchSpec ms;
+
+        try
```
wolfv, line 20: hmm, I am more a fan of (from stackoverflow):

```
MyMapType::const_iterator iter = mymap.find(key);

if(iter != mymap.end())
{
    // iter is item pair in the map. The value will be accessible as `iter->second`.
}
```

wolfv:
this code looks great. Two nitpicks, then it's ready to be merged!

src/micromamba/main.cpp
adriendelsalle: nice catch, thanks!

src/pinning.cpp
adriendelsalle: Thanks again for the suggestion!

wolfv:
cool. Maybe we should make the minor-pin behavior configurable in the future. 

It could be interesting / useful for other programming languages well (e.g. Ruby 2.X / 3.X, Lua, ...).

Issue #697, opened 2021-02-15
libcurl certificate revocation checks break conda compatibility in restricted network scenarios


mparry:
I've been evaluating mamba as a drop-in conda replacement for our internal usage and so far it looks great. However, there's one problem that I've run into where on many of our servers, that have no internet access, repository connections fail due to failed certificate revocation checks. (We have internal conda repo proxies/mirrors configured, so the servers can connect to those, but not anything outside, including revocation servers.)

Would it be possible to either a) disable these checks entirely, which would arguably offer maximum compatibility with conda, or b) at least provide an option to disable them?

mparry:
Note that this is very straightforward to do:

```cpp
curl_easy_setopt(handle, CURLOPT_SSL_OPTIONS, CURLSSLOPT_NO_REVOKE);
```

I'd be happy to open a PR for option a, if that's acceptable.

wolfv:
Hi @mparry indeed, I haven't run across this one yet. 

Would be great if you can open a PR. Maybe we add this as an option for `--ssl-verify`? 

e.g. `--ssl-verify no-revoke`?

mparry:
I guess it depends what you're aiming for with mamba exactly. If you want it to work as closely as possible to conda out of the box then, since conda / requests don't do these checks at all, I think it should just be turned off by default (maybe with an option to re-enable it).

tl-hbk:
This is something I'd like to see added as well.

@wolfv I was wondering if you had any thoughts on @mparry's last comment about conda not doing these checks or if you'd prefer to keep these checks as an option if there should be a different way other than `ssl-verify` since I'm under the impression that that would prevent passing a custom cacert bundle.

wolfv:
thanks @tl-hbk for fixing this. @mparry once the next release of mamba/micromamba is out, you can use the env var `MAMBA_SSL_NO_REVOKE=1` to disable revocation checking.

Issue #628, opened 2020-12-11
`ssl_verify` `.condarc` option doesn't work if pointing to a custom crt


dhirschfeld:
In our linux containers we point the `.condarc` [`ssl_verify`](https://docs.conda.io/projects/conda/en/latest/user-guide/configuration/use-condarc.html#ssl-verification-ssl-verify) option to our custom crt file:
```
ssl_verify: /etc/ssl/certs/ca-certificates.crt
```

`conda` picks this up and we can connect to our internal repo server but it appears `mamba` ignores this setting and we get SSL errors when trying to connect.

```
RuntimeError: Download error (60) SSL peer certificate or SSH remote key was not OK [https://conda.smc.stanwell.com/stanwell/uat/noarch/repodata.json]
```
<br>

If however, we set the `REQUESTS_CA_BUNDLE` environment variable, then `mamba` works fine:
```
export REQUESTS_CA_BUNDLE=/etc/ssl/certs/ca-certificates.crt
```

It would be good if `mamba` would also use the `ssl_verify` `.condarc` option when pointing to a custom crt.

dhirschfeld:
Looks related to https://github.com/mamba-org/mamba/issues/151#issuecomment-617535208

wolfv:
thanks for opening this issue. I will have to check why we're not using the option from condarc.

tl-hbk:
This change here seems to have been what caused the `.condarc` option to stop working

https://github.com/mamba-org/mamba/commit/5c2951cc181d383e131fc345d439fc2878fe94db#diff-83165d5f8a167348297e4c513a32c2e9b78cfcd82f7a134a770683d710073989L95-L98

dhirschfeld:
It looks like the code assumes that `ssl_verify` is either a boolean or an empty string and doesn't admit the possibility to specify a path to `ca-certificates.crt`?

![image](https://user-images.githubusercontent.com/881019/109069600-2a38ad80-773d-11eb-8673-0a0dae31c151.png)

https://github.com/mamba-org/mamba/blob/9997ff826585b7f7ee613714d98e0941986c42f2/mamba/utils.py#L158-L166

wolfv:
this should be fixed by #712. I'll release today or on Monday.

PR #712, opened 2021-02-25
Fetch SSL changes

tl-hbk:
This PR resolves the following issues
* [Use custom cacert path in .condarc](https://github.com/mamba-org/mamba/issues/628)
* [Add option to skip libcurl revocation checks](https://github.com/mamba-org/mamba/issues/697)

### DIFF
--- a/src/fetch.cpp
+++ b/src/fetch.cpp
@@ -89,6 +89,13 @@ namespace mamba
             }
         }
 
+        std::string ssl_no_revoke_env
+            = std::getenv("MAMBA_SSL_NO_REVOKE") ? std::getenv("MAMBA_SSL_NO_REVOKE") : "0";
+        if (Context::instance().ssl_no_revoke || ssl_no_revoke_env != "0")
+        {
+            curl_easy_setopt(m_handle, CURLOPT_SSL_OPTIONS, CURLSSLOPT_NO_REVOKE);
+        }
+
         std::string& ssl_verify = Context::instance().ssl_verify;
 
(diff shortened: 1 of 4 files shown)


src/fetch.cpp
```diff
@@ -89,6 +89,11 @@ namespace mamba
             }
         }
 
+        if (Context::instance().ssl_no_revoke || std::getenv("MAMBA_SSL_NO_REVOKE") != nullptr)
```
wolfv, line 92: should we check the value of the env variable?

src/micromamba/main.cpp
```diff
@@ -159,6 +160,9 @@ init_network_parser(CLI::App* subcom)
 {
     subcom->add_option(
         "--ssl_verify", network_options.ssl_verify, "Enable or disable SSL verification");
+    subcom->add_option("--ssl_no_revoke",
```
wolfv, line 163: I would prefer using `--ssl-no-revoke`, and could you add the default at teh end of the description? (i.e. `(default: false)`

wolfv:
This is a great PR, thanks! Once you've made the changes and squashed the commits I'll merge immediately .

src/fetch.cpp
tl-hbk: Sure thing.  Any preference on what values to support for it though (true, 1)?

wolfv: I am voting for `0` == false, everything else true?

tl-hbk: Sounds good to me.

PR #714, opened 2021-02-25
Override or disable rc file(s)

adriendelsalle:
Description
---

Add `--no-rc` flag and `--rc-file` option:
- `--no-rc` disables *all* configuration files
- `--rc-file` overrides *all others* configuration files (unique rc file)

Discussion
---
We should consider making `Configurable` class a singleton to avoid reading files multiple times.

### DIFF
--- a/include/mamba/config.hpp
+++ b/include/mamba/config.hpp
@@ -21,16 +21,20 @@ namespace mamba
     class Configurable
     {
     public:
-        Configurable();
-        Configurable(std::string unique_source);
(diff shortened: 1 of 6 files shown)


adriendelsalle:
I used a singleton pattern to avoid reading files multiple times at construction.
Now this operation can be done explicitly by calling `Configurable::instance().load()`

At this time, internals of the class are still `protected` and not `private`, for testing reasons. Let me know if you have ideas @wolfv !

PR #727, opened 2021-03-01
fix multi-prefix replacement in binary files

wolfv:
@tdegeus this should fix the issue you were seeing in pkgconfig. I am going to add tests now.


tdegeus:
Awesome!

PR #736, opened 2021-03-03
Add linking options

adriendelsalle:
Description
---

Details:
- add `always_copy` link feature
- add `always_softlink` link feature
- add `allow_softlinks` link fallback
- add flags in umamba CLI and set options properly

### DIFF
--- a/include/mamba/transaction_context.hpp
+++ b/include/mamba/transaction_context.hpp
@@ -9,6 +9,7 @@
 
 #include <string>
 
+#include "context.hpp"
 #include "mamba_fs.hpp"
 
 namespace mamba
@@ -25,7 +26,7 @@ namespace mamba
     {
     public:
         TransactionContext() = default;
-        TransactionContext(const fs::path& prefix, const std::string& py_version);
+        TransactionContext(Context& ctx, const fs::path& prefix, const std::string& py_version);
 
         bool has_python;
         fs::path target_prefix;
@@ -33,6 +34,9 @@ namespace mamba
         fs::path python_path;
         std::string python_version;
         std::string short_python_version;
+        bool allow_softlinks = false;
+        bool always_copy = false;
+        bool always_softlink = false;
     };
 }  // namespace mamba
 
--- a/include/mamba/transaction_context.hpp
+++ b/include/mamba/transaction_context.hpp
@@ -26,7 +26,7 @@ namespace mamba
     {
     public:
         TransactionContext() = default;
-        TransactionContext(Context& ctx, const fs::path& prefix, const std::string& py_version);
+        TransactionContext(const fs::path& prefix, const std::string& py_version);
 
         bool has_python;
         fs::path target_prefix;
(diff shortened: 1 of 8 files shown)


src/link.cpp
```diff
@@ -898,10 +898,44 @@ namespace mamba
...
+                fs::create_hard_link(src, dst, ec);
+
+                if (ec)
+                {
+                    softlink = m_context->allow_softlinks;
+                    copy = !softlink;
+                }
+                else
+                {
+                    LOG_INFO << "hard linked " << src << " --> " << dst;
```
wolfv, line 918: return here? otherwise it can _also_ softlink?

wolfv: ok i think i get it now, looks fine!

src/micromamba/main.cpp
```diff
@@ -915,6 +927,12 @@ init_install_global_parser(CLI::App* subcom)
         ->allow_extra_args(false);
     subcom->add_flag(
         "--no-pin", create_options.no_pin, "Ignore pinned packages (from config or prefix file)");
+    subcom->add_flag(
+        "--allow_softlinks", create_options.allow_softlinks, "Allow softlinks when hardlinks fail");
```
wolfv, line 931: I prefer `--allow-softlinks` vs underscore

include/mamba/transaction_context.hpp
```diff
@@ -25,14 +26,17 @@ namespace mamba
...
-        TransactionContext(const fs::path& prefix, const std::string& py_version);
+        TransactionContext(Context& ctx, const fs::path& prefix, const std::string& py_version);
 
         bool has_python;
         fs::path target_prefix;
         fs::path site_packages_path;
         fs::path python_path;
         std::string python_version;
         std::string short_python_version;
+        bool allow_softlinks = false;
```
wolfv, line 37: is this used?

wolfv (approved):
looks awesome!

src/link.cpp
adriendelsalle: cool thanks for the review!

src/micromamba/main.cpp
adriendelsalle: it's a typo sorry, will be fixed!

include/mamba/transaction_context.hpp
adriendelsalle: Yes it is:

When a hardlink fails, we have this fallback :
```
softlink = m_context->allow_softlinks;
copy = !softlink;
```

src/link.cpp
```diff
@@ -898,10 +898,44 @@ namespace mamba
             return std::make_tuple(validate::sha256sum(dst), rel_dst);
         }
 
-        if (path_data.path_type == PathType::HARDLINK)
+        if (path_data.no_link || (path_data.path_type == PathType::HARDLINK))
```
wolfv, line 901: maybe change the order of comparison here because path_data.path_type == HARDLINK is much more likely (speed boost!)

adriendelsalle: done ! huge boost ! :)

PR #742, opened 2021-03-04
Improve search of cuda virtual package

adriendelsalle:
Description
---

`nvidia-smi` binary is currently used to get the `cuda` driver version.
On Windows, this utility may be not located in the `path`, then:
- check if `nvcuda.dll` is found in the path
- if true, search in all `C:\Windows\System32\DriverStore\FileRepository` subdirs if one starting with `nv` contains `nvidia-smi.exe` binary file

Also lower log message level to `debug` if `cuda` is not found, since it's will be a pretty common case.

### DIFF
--- a/src/virtual_packages.cpp
+++ b/src/virtual_packages.cpp
@@ -92,6 +92,49 @@ namespace mamba
 
             if (ec)
             {
+                out = "";
+
+                // Windows fallback
+                bool may_exist = false;
+                std::string path = env::get("PATH");
+                std::vector<std::string> paths = split(path, env::pathsep());
+
+                for (auto& p : paths)
+                {
+                    if (fs::exists(concat(p, env::pathsep(), "nvcuda.dll")))
+                    {
+                        may_exist = true;
+                        break;
+                    }
+                }
+
+                std::string base = "C:\\Windows\\System32\\DriverStore\\FileRepository";
+                if (may_exist)
+                {
+                    for (auto& p : fs::directory_iterator(base))
+                    {
+                        if (starts_with(p.path().string(), "nv")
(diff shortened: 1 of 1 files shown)


src/virtual_packages.cpp
```diff
@@ -92,7 +92,50 @@ namespace mamba
...
+                    {
+                        may_exist = true;
+                        break;
+                    }
+                }
+
+                std::string base = "C:\\Windows\\System32\\DriverStore\\FileRepository";
+                if (may_exist)
+                {
+                    for (fs::path p : fs::directory_iterator(base))
```
wolfv, line 113: I think `auto&` is better

src/virtual_packages.cpp
```diff
@@ -92,7 +92,50 @@ namespace mamba
...
+                    {
+                        if (p.string().startswith("nv")
```
wolfv, line 115: then maybe have to use `p.path().string()...`

src/virtual_packages.cpp
```diff
@@ -92,7 +92,50 @@ namespace mamba
...
+                            && fs::exists(concat(p, env::pathsep(), "nvidia-smi.exe")))
```
wolfv, line 116: maybe not use concat, but rather `p / "nvidia-smi.exe"`

src/virtual_packages.cpp
```diff
@@ -92,7 +92,50 @@ namespace mamba
...
+                        {
+                            std::string err;
+                            std::string f = fs::exists(concat(p, env::pathsep(), "nvidia-smi.exe"));
```
wolfv, line 119: ? looks liek a bug (assigning bool to string)

adriendelsalle: it is !

Issue #666, opened 2021-01-26
Add support for use_only_tar_bz2 conda flag


rs2:
See:
* https://stackoverflow.com/a/58717064
* https://github.com/mamba-org/mamba/blob/d43f49927262b709d8da7dea13d1509583c0ae2a/mamba/utils.py#L142-L173

wolfv:
Thanks for the bug report. Indeed one will need to modify the behavior in libsolv (add a flag) to skip those two if's: https://github.com/openSUSE/libsolv/blob/4bc791c0d235eb14bfe4c5da607206bfdfa6983d/ext/repo_conda.c#L270-L273

PR's are warmly welcome!

tl-hbk:
@wolfv does this PR solve this issue?  https://github.com/openSUSE/libsolv/pull/434  I assume the only follow up is modifying mamba to call `repo_add_conda` with the flag if `use_only_tar_bz2` is set

wolfv:
awesome, yes, this looks great!

tl-hbk:
@wolfv when do you think you'll be able to release this?

wolfv:
End of this week hopefully!

PR #744, opened 2021-03-05
Add support for use_only_tar_bz2

tl-hbk:
Change for https://github.com/mamba-org/mamba/issues/666.  Will need a new release of libsolv but I'd like for the windows regex fix to be merged before requesting that.

EDIT: Will need to rerun tests after new libsolv is available with the new flag.  Tested this locally already with a locally built copy of libsolv and works as expected.

### DIFF
--- a/include/mamba/context.hpp
+++ b/include/mamba/context.hpp
@@ -100,6 +100,8 @@ namespace mamba
 
         std::vector<std::string> pinned_packages = {};
 
+        bool use_only_tar_bz2 = false;
+
         static Context& instance();
 
         Context(const Context&) = delete;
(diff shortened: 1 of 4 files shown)


wolfv:
@tl-hbk thanks a lot for these changes! Sooner or later it would be cool to add this feature to micromamba, too. Maybe @adriendelsalle can comment on what changes to the `rc` file reading and CLI flags are necessary.

adriendelsalle:
Sure, will do soon-ish!

wolfv:
We've added the two patches from libsolv master to the libsolv feedstock (that means the build with _1 should have the new functionality).

tl-hbk:
Awesome thanks for doing that

Issue #733, opened 2021-03-03
[windows] UNC path not supported by mamba, supported by conda


adament:
In conda on windows it is possible to refer to a channel on a windows share using a UNC path using
```yaml
channels:
- file:////hostname/share
- file://\\hostname\share
```
Curl supports using the `file:////hostname/share` URL to refer to a UNC path. However it seems that mamba replaces the quadruple slash with a double slash before passing it on to curl as `file://hostname/share`. Would it be possible to keep the quadruple slash and pass it to libcurl? We are using conda in a Windows environment for internal package repositories and UNC paths and fileshares provides a simple way to apply access control to the repositories.

I have tested with:
```
mamba 0.7.14
conda 4.9.2
```

wolfv:
A PR would be very welcome!

adament:
I think I have identified the issue to [line 463](https://github.com/mamba-org/mamba/blob/243bf801e99fc2215f88fd87c66574aeb85a52e9/src/channel.cpp#L463) in the `fix_win_path` function in `channel.cpp`. However given the [test](https://github.com/mamba-org/mamba/blob/243bf801e99fc2215f88fd87c66574aeb85a52e9/test/test_channel.cpp#L28) it seems the like the incompatibility with conda was intentional.

I will try simply removing that line and changing the test and see if that fixes my issue, if so I will make a pull request.

adament:
Turns out this was not correct. It seems that the current implementation of fix_win_path just matches win_path_backout in conda.common.path so when calling conda.core.index.calculate_channel_urls the result has stripped the initial // from the UNC url as well. I do not know where conda makes that back into a UNC path again.

adament:
My current guess is that conda fixes the problem created by stripping the UNC prefix by reintroducing it again in url_to_path in conda.common.path.

wolfv:
if there is absolutely no cleaner way, then yeah, I guess we have to do it.

btw. just to make sure: can you check that this URL format works with curl? (`curl file:////.... -o test.json`) so that we know we won't hit problems in curl later on.

adament:
This should be closed with the merge of #749

wolfv:
Thanks @adament !

PR #749, opened 2021-03-09
[RFC] Add support for UNC file:// urls

adament:
Conda and mamba convert `file:////hostname/share` UNC urls to `file://hostname/share` when reading them from e.g. condarc, in conda they are then converted back to UNC urls in the function `url_to_path` which is used by the `LocalFSAdapter`. Implement a similar hack for mamba as discussed in #733.

Note: I have not been able to get the test suite building on my machine, so I have not run the test suite against this change.

### DIFF
--- a/src/fetch.cpp
+++ b/src/fetch.cpp
@@ -6,6 +6,7 @@
 
 #include <string_view>
 #include <thread>
+#include <regex>
 
 #include "mamba/fetch.hpp"
 #include "mamba/context.hpp"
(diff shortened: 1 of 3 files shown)


wolfv:
Can you either add some tests or post some expected values from this function here?
I think for proper testing it might be easier to factor this function out into a standalone function. 

Is this only relevant on Windows?

Is this going to convert `file:///blabla` to `file:////blabla`? Did you check that regular file URL's still work fine?

Besides, this is looking really good! I just don't have a good understanding of UNC urls yet, and in what condition you need to transform them back.

adament:
Good idea splitting it into a separate function, and yes I will add some tests. Do you have any preferences for where to place the function?

As I read the curl manpage it is only relevant for Windows:
```
FILE

Read or write local files. curl does not support accessing file:// URL remotely, but when running on Microsoft Windows using the native UNC approach will work.
```

It should leave `file:///absolute/path` alone and only modify `file://host/share` but some tests would be good to support this. Furthermore it leaves `file://localhost/absolute/path` alone for localhost, 127.0.0.1, and ::1 as this is what conda does, but I have not tested that curl actually supports localhsot file urls.

wolfv:
I think this function can go in a similar place as `fix_win_path`. 
Tests can go into the C++ tests we have in the test subdir.

Thanks for tackling this!

wolfv:
@adament do you think you'll find the time to finish this?

adament:
Yes I hope to get back to it at the end of this week. But if somebody else has time and interest in picking it up they are very welcome.

adament:
@wolfv I have tried adding tests but I have not managed to build the tests on my machine and I cannot quite tell if the CI jobs run the C++ tests?

wolfv:
Yes, the CI should run them! Thanks. I'll check it out tomorrow or Monday

wolfv:
Thanks @adament !

wolfv:
You can find the result of the CI run of these tests here: https://github.com/mamba-org/mamba/pull/749/checks?check_run_id=2149502434#step:6:313

Looks good!

PR #753, opened 2021-03-09
support for larger build numbers

pbauwens-kbc:
Conda packages with a .dev and timestamp suffix caused an error when the version number parts were converted to an integer (using stoi). Updated to use stoull.

### DIFF
--- a/src/link.cpp
+++ b/src/link.cpp
@@ -981,8 +981,8 @@ namespace mamba
 
         auto py_ver_split = split(m_context->python_version, ".");
 
-        if (std::stoi(std::string(py_ver_split[0])) >= 3
-            && std::stoi(std::string(py_ver_split[1])) > 5)
+        if (std::stoull(std::string(py_ver_split[0])) >= 3
+            && std::stoull(std::string(py_ver_split[1])) > 5)
         {
             // activate parallel pyc compilation
             command.push_back("-j0");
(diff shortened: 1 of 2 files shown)


wolfv:
Thanks, this is great!

wolfv:
@pbauwens-kbc btw I just checked the netrc issue you were mentioning. We do use `netrc-optional` here: https://github.com/mamba-org/mamba/blob/e80c1420e668e26fac0144a94660e6a0bc635b3b/src/fetch.cpp#L108

It's the same as `curl https://blabla --netrc-optional`. 

Is your netrc file at a different location?

pbauwens-kbc:
Public

yes, thx. I noticed it in the code. It turned out to be an issue on my Mac.

I have nog clue though what was going wrong, even regular curl did not work. When I ran with the -vvv option it showed error messages that the machine name could not be found  in the netrc file while it was clearly there.

I called a colleague this morning to help me investigate but all of a sudden it worked. The only thing that changed was an overnight update of OSX and the virus scanner we are forced to use.

So all is good 🙂

I am going to look into using boa too, as a replacement for conda-build. Can I expect similar issues there (I mean with the stoi)?

Peter


________________________________

wolfv:
for the moment, yes, until a new release of mamba is published on conda-forge.

boa is using mamba through the Python interface.

wolfv:
great wrt to the netrc issue, btw!

PR #754, opened 2021-03-09
Fix activate command example in "How to use mamba" docs

jfcorbett:
In docs section ["how to use"](https://mamba.readthedocs.io/en/latest/how_to_use.html), an example is given with the erroneous command:

    mamba activate myjlabenv

Now fixed to:

    conda activate myjlabenv

since the one exception (when using mamba as a drop-in replacement for conda) is that [you should still use conda for activation and deactivation](https://mamba.readthedocs.io/en/latest/getting_started.html).


wolfv:
Thanks!

PR #764, opened 2021-03-10
Improve config

adriendelsalle:
Description
---
This is a big PR, so let me explain a bit the design intent:
- Use a `Configuration` singleton in charge of storing all the config options
- Use a `Configurable` class to hold all the information relation to a unique config option
  - value, group, description, long description
  - set corresponding cli and/or env variables and/or `Context` variable
    - the `Configurable` is in charge to "push" config value in the `Context` (if linked to a `Context` variable)
  - rc configurable, configured (rc, cli or env var)
- Store the `Configurable`s in `Configuration` using type erased `ConfigurableInterface` interface
 
Details:
- rename enums following [Google C++ style guide](https://google.github.io/styleguide/cppguide.html#Enumerator_Names) to fix a collision with macro on Windows
- keep track of creation order for a nicer display
- add `micromamba config describe` sub command
- improve `micromamba config list` sub command
- add capability to store CLI values with a different type than `Configurable`'s one
  - useful to disambiguate `undefined` and `false` states of a CLI boolean "flag" value
- add post-merge `Configurable` hook to allow complex behaviors
  - "merge" = computation of the final config value given default value/rc files/CLI/env var


src/configuration.cpp
```diff
@@ -0,0 +1,702 @@
...
+    {
+        auto& ctx = Context::instance();
+
+        // Channels
+        insert(Configurable("channels", &ctx.channels)
+                   .group("Channels")
+                   .set_env_var_name("CONDA_CHANNELS")
+                   .description("Define the list of channels")
+                   .long_description(unindent(R"(
+                    The list of channels where the packages will be search for.
```
wolfv, line 106: search -> searched

src/configuration.cpp
```diff
@@ -0,0 +1,702 @@
...
+                    The list of channel names and/or urls used for the 'defaults'
+                    multichannel.)")));
+
+        insert(Configurable("override_channels_enabled", &ctx.override_channels_enabled)
+                   .group("Channels")
+                   .set_env_var_name()
+                   .description("Permit use of the --overide-channels command-line flag"));
+
+        // Network
+        insert(Configurable("local_repodata_ttl", &ctx.local_repodata_ttl)
```
wolfv, line 128: did you rename this? I think I had a flag for this before, but unsure what name I used.

src/configuration.cpp
```diff
@@ -0,0 +1,702 @@
...
+                   .group("Network")
+                   .set_env_var_name()
+                   .description("Verify SSL certificates for HTTPS requests")
+                   .long_description(unindent(R"(
+                        'ssl_verify' can be either an empty string (regular SSL verification),
+                        the string "<false>" to indicate no SSL verification, or a path to
+                        a directory with cert files, or a cert file..)"))
+                   .set_post_build_hook(detail::ssl_verify_hook));
+
+        insert(Configurable("cacert_path", std::string(""))
```
wolfv, line 149: is the option then translated to `--cacert-path` (underscores to dashes?)

src/configuration.cpp
```diff
@@ -0,0 +1,702 @@
...
+                        WARNING: overrides 'ssl_verify' if provided and 'ssl_verify'
+                        also contains a path to SSL certificates.)")));
+
+        insert(Configurable("ssl_no_revoke", &ctx.ssl_no_revoke)
+                   .group("Network")
+                   .set_env_var_name()
+                   .description("SSL certificate revocation checks")
+                   .long_description(unindent(R"(
+                        This option tells curl to disable certificate revocation checks.
+                        It's only working for 'Schannel' back-end.
```
wolfv, line 165: Schannel -> Windows.

WARNING: this option loosens the SSL security, and by using this
                         flag you ask for exactly that.)
-> WARNING: this option loosens the SSL security.

src/configuration.cpp
```diff
@@ -0,0 +1,702 @@
...
+                        WARNING: this option loosens the SSL security, and by using this
+                        flag you ask for exactly that.)")));
+        // Solver
+        insert(Configurable("channel_priority", &ctx.channel_priority)
+                   .group("Solver")
+                   .set_env_var_name()
+                   .description("Define the channel priority ('strict', 'flexible' or 'disabled')")
+                   .long_description(unindent(R"(
+                    Accepts values of 'strict', 'flexible', and 'disabled'. The default
```
wolfv, line 174: actually, we don't really implement anything for "flexible", only strict and disabled.

src/configuration.cpp
```diff
@@ -0,0 +1,702 @@
...
+                    dependencies, rather than raising an unsatisfiable error. With channel
+                    priority disabled, package version takes precedence, and the
+                    configured priority of channels is used only to break ties. In
+                    previous versions of conda, this parameter was configured as either
+                    True or False. True is now an alias to 'flexible'.)")));
+
+        insert(Configurable("pinned_packages", &ctx.pinned_packages)
+                   .group("Solver")
+                   .set_env_var_name()
+                   .description("A list of package specs to pin for every environment resolution"));
```
wolfv, line 188: does this want a path to a file? or a list as cmd line options?

src/configuration.cpp
```diff
@@ -0,0 +1,702 @@
...
+        if (on_mac || on_linux)
+        {
+            m_sources = { "/etc/conda/.condarc",       "/etc/conda/condarc",
+                          "/etc/conda/condarc.d/",     "/etc/conda/.mambarc",
+                          "/var/lib/conda/.condarc",   "/var/lib/conda/condarc",
+                          "/var/lib/conda/condarc.d/", "/var/lib/conda/.mambarc" };
+        }
+        else
+        {
+            m_sources = { "C:/ProgramData/conda/.condarc",
```
wolfv, line 295: Does this work? Usually double backslash is more "correct" on windows.

wolfv (changes requested):
OK, after reading through all the changes, I got that the config flags and the config system are still seperate.

It looks good to me, but I need to be sure that it works as advertised :) Going to build it locally and test a bit more.

wolfv:
conda config uses --show-sources with an additional `s`. I think it would be good to do the same here.

wolfv:
Don't use default base prefix if no prefix is passed for create (It's not possible to overwrite 'base' env. Install in 'base' instead?: [y/N] n)

wolfv:
zsh is gone from the shell init?

wolfv:
If env var is empty, then don't check if there is a condarc file .

wolfv:
giving --rc-file that does not exist should throw an error.

src/configuration.cpp
adriendelsalle: the corresponding CLI flag is `--repodata-ttl`

src/configuration.cpp
adriendelsalle: I don't get it, yes this configurables is used to store value passed using `--cacert-path` CLI option

src/configuration.cpp
adriendelsalle: done

src/configuration.cpp
adriendelsalle: done

src/configuration.cpp
adriendelsalle: You're right, I modified the description.
I also remove `strict_channel_priority` from the context since what's matter is `channel_priority`.
If we don't want to implement later a `flexible` option, we can also keep the former and remove `channel_priority` ?

src/configuration.cpp
adriendelsalle: This has the same type as the context value, which is `std::vector<std::string>>` that represent the specs.

src/configuration.cpp
adriendelsalle: I will change that and maybe add some tests for Windows

wolfv:
when executing only the hook, the `micromamba info` shows the (fallback) root prefix, even though the $CONDA_PREFIX variable is empty

wolfv:
On first run, micromamba warns about using default root prefix. However, the message of whcih fallback prefix is used is now `INFO` and thus nobody knows what the default root prefix is.

src/configuration.cpp
adriendelsalle: fixed, not tested yet

adriendelsalle:
> conda config uses --show-sources with an additional `s`. I think it would be good to do the same here.

I changes the CLI flags of the `config list` sub command (`--sources`, `--descriptions`, etc.). it could be refactored later

> Don't use default base prefix if no prefix is passed for create (It's not possible to overwrite 'base' env. Install in 'base' instead?: [y/N] n)
> 
> (Conda message: CondaValueError: either -n NAME or -p PREFIX option required,
> try "conda create -h" for more details)

done, `ERROR   No target prefix specified` raised

> zsh is gone from the shell init?

fixed

> If env var is empty, then don't check if there is a condarc file .

done, `micromamba config sources -vv` doesn't show `DEBUG   Configuration not found at ''` anymore when `CONDARC` or `MAMBARC` env vars are not set or empty

> giving --rc-file that does not exist should throw an error.

done, and `--rc-file` and `--no-rc` are now exclusive. tests added

> micromamba info

should be fixed now

> On first run, micromamba warns about using default root prefix. However, the message of whcih fallback prefix is used is now `INFO` and thus nobody knows what the default root prefix is.

also fixed

adriendelsalle:
@wolfv I updated this PR according your review and requested changes!
Let me know if it looks like ok now.

include/mamba/configuration.hpp
```diff
@@ -0,0 +1,1117 @@
...
+            {
+                rhs = mamba::VerificationLevel::kWarn;
+            }
+            else if (str == "disabled")
+            {
+                rhs = mamba::VerificationLevel::kDisabled;
+            }
+            else
+            {
+                throw std::runtime_error("'VerificationLevel' wrong value");
```
wolfv, line 69: I think this should list the accepted values.

include/mamba/configuration.hpp
```diff
@@ -0,0 +1,1117 @@
...
+            {
+                if (m_value == 1)
+                {
+                    return true;
+                }
+                else
+                {
+                    return false;
+                }
+                throw std::runtime_error("Should not happen!");
```
wolfv, line 200: This runtime error can never be hit, so might as well remove it. Or remove the else branch and replace it with return false.

src/micromamba/clean.cpp
```diff
@@ -19,24 +19,32 @@ init_clean_parser(CLI::App* subcom)
...
-}
+    auto& config = Configuration::instance();
 
-void
-load_clean_options(Context& ctx)
-{
-    load_general_options(ctx);
-    load_prefix_options(ctx);
-    load_rc_options(ctx);
+    auto& clean_all = config.insert(
```
wolfv, line 24: just for me to understand: this is technically slightly over-engineered since these options are only ever set through CLI, right? We could also just keep the old way here, right?

src/micromamba/common_options.cpp
```diff
@@ -0,0 +1,489 @@
...
+
+void
+channels_hook(std::vector<std::string>& channels)
+{
+    auto& config = Configuration::instance();
+
+    // Workaround to silent the hook warning
+    // TODO: resolve configurable deps/workflow to compute only once
+    auto verbosity = MessageLogger::global_log_severity();
+    MessageLogger::global_log_severity() = mamba::LogSeverity::error;
```
wolfv, line 256: these enums have not been adjusted to the new kError format?

include/mamba/configuration.hpp
adriendelsalle: :+1:

include/mamba/configuration.hpp
adriendelsalle: nice catch, it was intended to be `else if(m_value == -1)`

src/micromamba/clean.cpp
adriendelsalle: Indeed.
I tried not to mix different approaches to have a single way to define and use configuration.
It also gives you capability to turn on rc configuration or more useful here, config by environment variable in one line : `.set_env_var_name()`

adriendelsalle: it will also make much more sense when moving the high-level API of `mamba`: `install`, `update`, `clean`, etc. from `micromamba` to `mamba`/`libmamba` itelf. Because you will maybe want to be able to configure these using rc files (`clean_all: true` in rc file makes `clean` command always cleaning packages/tarballs/index)

src/micromamba/common_options.cpp
adriendelsalle: done

wolfv:
can we add a `--no-env` flag and disregard all environment variables easily?

src/micromamba/constructor.cpp
```diff
@@ -16,13 +17,23 @@ using namespace mamba;  // NOLINT(build/namespaces)
...
-                     "Extract given tarball into prefix");
+                     extract_conda_pkgs.set_cli_config(0),
+                     extract_conda_pkgs.description());
+
+    auto& extract_tarball = config.insert(Configurable("constructor_extract_tarball", false)
+                                              .group("cli")
+                                              .rc_configurable(false)
+                                              .description("Extract given tarball into prefix"));
+    subcom->add_flag(
+        "--extract-tarballs", extract_tarball.set_cli_config(0), extract_tarball.description());
```
wolfv, line 36: this flag has _changed_??

It's used in the miniforge constructor project and this will definitely break things.

src/micromamba/create.cpp
```diff
@@ -18,28 +18,16 @@ set_create_command(CLI::App* subcom)
...
-                    install_specs(create_options.specs, false);
-                }
-                else
-                {
-                    exit(1);
-                }
-            }
-            catch_existing_target_prefix(ctx);
-            install_specs(create_options.specs, true);
+            check_target_prefix(false, false, true, false);
```
wolfv, line 29: hmmm ... this API doesn't speak to me. 

we could fake named arguments using `/*someparam=*/false, ...
I need to check the signature of check_target_prefix

src/micromamba/install.cpp
```diff
@@ -28,88 +30,75 @@ init_install_parser(CLI::App* subcom)
...
-        load_install_options(ctx);
+        load_configuration();
 
-        if (!create_options.specs.empty() && !ctx.target_prefix.empty())
+        auto& specs = configuration.at("specs").value<std::vector<std::string>>();
+
+        if (!specs.empty())
         {
-            install_specs(create_options.specs, false);
+            check_target_prefix(true, true, true, true);
```
wolfv, line 100: here it would be interesting to use boolean flags:

`check_target_prefix(MM_CREATE | MM_INSTALL)...`

src/micromamba/constructor.cpp
adriendelsalle: fixed, sorry for that

src/micromamba/install.cpp
wolfv: or multiple named functions that call to a `check_target_prefix_impl`

like `check_target_prefix_install`

src/micromamba/create.cpp
adriendelsalle: I'm not really happy with that either. I will try to improve it

src/micromamba/install.cpp
```diff
@@ -183,7 +171,11 @@ install_specs(const std::vector<std::string>& specs, bool create_env, int solver
         sdir->load();
         multi_dl.add(sdir->target());
         subdirs.push_back(sdir);
-        if (ctx.strict_channel_priority)
+        if (ctx.channel_priority == ChannelPriority::kDisabled)
```
wolfv, line 174: i think it would be better to consider flexible and disabled the same way, and strict differently

wolfv: which is hte behavior from before

src/micromamba/install.cpp
```diff
@@ -285,8 +273,8 @@ install_specs(const std::vector<std::string>& specs, bool create_env, int solver
     bool success = solver.solve();
     if (!success)
     {
-        std::cout << "\n" << solver.problems_to_str() << std::endl;
```
wolfv, line 288: can we revert these changes here?

This is not a log message, it's an output

src/micromamba/install.cpp
```diff
@@ -308,7 +296,6 @@ install_specs(const std::vector<std::string>& specs, bool create_env, int solver
         repo_ptrs.push_back(&r);
     }
 
-    std::cout << std::endl;
```
wolfv, line 311: ?

src/micromamba/install.cpp
adriendelsalle: `flexible` looks like a mix between `strict` for specs and `disabled` for deps.. I don't know

src/micromamba/install.cpp
wolfv: Although in the future we might want to adjust this so that we get the proper json output

wolfv: But LOG_ERROR has implications also with regard to `--json` and `--quiet`

src/micromamba/install.cpp
```diff
@@ -325,78 +312,83 @@ install_specs(const std::vector<std::string>& specs, bool create_env, int solver
...
-                {
-                    create_options.channels.push_back(c);
-                }
+                channels.add_rc_value(yaml_channels, file);
             }
             else
             {
-                std::cout << termcolor::yellow << "WARNING: No channels specified in " << file
-                          << termcolor::reset << std::endl;
+                LOG_DEBUG << "No 'channels' specified in file: " << file;
```
wolfv, line 366: Do we have a global warning if no channels are specified?

adriendelsalle: I don't think so. I will add one

src/micromamba/install.cpp
adriendelsalle: done

src/micromamba/install.cpp
adriendelsalle: also done, I understand distinction made between log and output.

src/micromamba/clean.cpp
```diff
@@ -45,17 +53,31 @@ set_clean_command(CLI::App* subcom)
...
 
-        load_clean_options(ctx);
+        if (!(clean_all || clean_index || clean_pkgs || clean_tarballs))
+        {
+            std::cout << "Nothing to do." << std::endl;
+            return;
+        }
 
-        Console::print("Cleaning up ... ");
+        LOG_INFO << "Collect information...";
```
wolfv, line 72: I think it's fine to print some output by default with this command so that the user knows something is going on.

src/micromamba/clean.cpp
```diff
@@ -146,20 +168,23 @@ set_clean_command(CLI::App* subcom)
...
+            if (!ctx.dry_run)
             {
-                Console::print("No cached tarballs found");
-            }
-            else if (!ctx.dry_run && Console::prompt("\n\nRemove tarballs", 'y'))
-            {
-                for (auto& tbr : to_be_removed)
+                if (to_be_removed.size() == 0)
+                {
+                    LOG_INFO << "No cached tarballs found";
```
wolfv, line 178: Same here, giving some output by default is nice

PR #788, opened 2021-03-23
C api

adriendelsalle:
Description
---

Create a C-API to ease re-use of `libmamba`.


include/mamba/c_api/mambac.h
```diff
@@ -0,0 +1,46 @@
...
+
+    void mamba_install(char* prefix, char** specs, char** channels);
+
+    void mamba_update(char* prefix, char** specs, char** channels);
+
+    void mamba_list(char* prefix, char* regex);
+
+    void mamba_config_list();
+
+    void set_opt(int option, int value);
```
wolfv, line 25: since C has no namespaces we should use `mamba_set_opt`

wolfv:
maybe we can put the high-level functions into a `src/highlevel/...` folder?

adriendelsalle:
Thanks for the early review! That is still WIP.

> maybe we can put the high-level functions into a `src/highlevel/...` folder?

Indeed, that's an idea! We could also take the leap to split public and private headers and keep as public headers only the high-level API. It would be much easier then to know where the API may or may not evolve :).

wolfv:
I think we want more than the high-level headers as the public API.

PR #808, opened 2021-03-29
Add --channel-alias option to umamba CLI

adriendelsalle:
Description
---

Add --channel-alias option to umamba CLI: in `install`, `create` and `update` sub commands
Add python tests


adriendelsalle:
Closes https://github.com/mamba-org/mamba/pull/808

PR #810, opened 2021-03-30
Fix spec files + no_rc

adriendelsalle:
Description
---

Spec files are currently considered as rc files.
This causes issue when using the `--no-rc` flag.
- read spec files as CLI configuration
- add tests

### DIFF
--- a/src/micromamba/install.cpp
+++ b/src/micromamba/install.cpp
@@ -353,10 +353,10 @@ parse_file_options()
 
             if (config["channels"])
             {
-                YAML::Node yaml_channels;
+                std::vector<std::string> yaml_channels;
                 try
                 {
-                    yaml_channels = config["channels"];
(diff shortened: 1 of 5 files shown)


src/micromamba/install.cpp
```diff
@@ -373,15 +382,40 @@ parse_file_options()
 
             if (config["name"])
             {
-                env_name.add_rc_value(config["name"], file);
+                if (env_name.cli_configured())
+                {
+                    throw std::runtime_error(
```
wolfv, line 387: I think we need to allow that i think. can you check what conda does there?
IMO CLI should override yaml file here.

adriendelsalle: This will be fixed by https://github.com/mamba-org/mamba/pull/816
Thanks for the review @wolfv

PR #812, opened 2021-03-30
Support spec files in update operation

adriendelsalle:
Description
---

Add support of spec files for `update` operation/sub command
Add tests

### DIFF
--- a/src/micromamba/update.cpp
+++ b/src/micromamba/update.cpp
@@ -37,6 +37,7 @@ set_update_command(CLI::App* subcom)
     init_update_parser(subcom);
 
     subcom->callback([&]() {
+        parse_file_options();
         load_configuration(MAMBA_ALLOW_ROOT_PREFIX | MAMBA_ALLOW_FALLBACK_PREFIX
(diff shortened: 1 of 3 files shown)


src/micromamba/update.cpp
```diff
@@ -37,6 +37,7 @@ set_update_command(CLI::App* subcom)
     init_update_parser(subcom);
 
     subcom->callback([&]() {
+        parse_file_options();
```
wolfv, line 40: what happens with the `name` variable of the yaml here?

adriendelsalle: In that case if `-n` is passed to the cli, it will override the one of the yaml file.
But if `-n` is not passed, then it uses the name of the yaml file instead of getting the fallback value from `CONDA_PREFIX` env var.
Sounds good or would it be better to **never** use the env name from yaml file ?

wolfv: usually, `install` and `update` operate on the "active environment". This would somewhat break this assumption (since it would update "another" environment).
That's why I am not sure about what behavior we want.

PR #818, opened 2021-03-31
add simple context debugging

wolfv:
This should give us speedier CLI tests

### DIFF
--- a/src/micromamba/common_options.cpp
+++ b/src/micromamba/common_options.cpp
@@ -104,6 +104,15 @@ init_general_options(CLI::App* subcom)
                                       .description("Only display what would have been done"));
     subcom->add_flag("--dry-run", dry_run.set_cli_config(false), dry_run.description())
         ->group(cli_group);
+
+#ifdef ENABLE_CONTEXT_DEBUGGING
+    static bool debug_context = false;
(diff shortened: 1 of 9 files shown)


src/micromamba/common_options.cpp
```diff
@@ -180,7 +180,7 @@ init_network_parser(CLI::App* subcom)
 
     auto& ssl_no_revoke = config.at("ssl_no_revoke").get_wrapped<bool>();
     subcom
-        ->add_option(
+        ->add_flag(
```
adriendelsalle, line 183: looks like not related to this PR

src/micromamba/install.cpp
```diff
@@ -93,6 +93,13 @@ set_install_command(CLI::App* subcom)
         parse_file_options();
         load_configuration();
 
+
+        if (true)
```
adriendelsalle, line 97: I guess you want to print the context if a CLI flag is passed

src/micromamba/install.cpp
```diff
@@ -93,6 +93,13 @@ set_install_command(CLI::App* subcom)
...
+        {
+            Context::instance().debug_print();
+            exit(0);
```
adriendelsalle, line 100: Would be better to return instead of using `exit` to avoid exiting the R or Python interactive shell when using directly libmamba shared lib

PR #821, opened 2021-03-31
mamba core API

adriendelsalle:
Description
---

Create a low level `core` API from existing `libmamba` files to prepare a higher level API from `micromamba` ones.


adriendelsalle:
@wolfv could you please review that, especially the top level `CMakeLists` file ?
Thanks!

include/mamba/core/version.hpp
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) 2019, QuantStack and Mamba Contributors
```
wolfv, line 1: this file is generated from version.hpp.in and shouldn't be commited

adriendelsalle: Nice catch, sorry for that. It's fixed

JohanMabille: I think this pattern should be reversed at some point (read version.hpp form CMake instead of generating the code with CMake). It is usually considered as a bad practice to generate code from your build system since it makes it difficult for someone to build the sources with a different build system.

PR #822, opened 2021-03-31
Improve solver display of problems

adriendelsalle:
Description
---

Have a nicer display of solver problems if some are encountered:
- remove extra newline if only 1 problem encountered
- display problems as a list

before:
```
Encountered problems while solving.
Problem: nothing provides requested xtensora

ERROR   Could not solve for environment specs
```
after:
```
- nothing provides requested xtensora
ERROR   Could not solve for environment specs
```

### DIFF
--- a/src/core/solver.cpp
+++ b/src/core/solver.cpp
@@ -328,10 +328,10 @@ namespace mamba
         for (int i = 1; i <= count; i++)
         {
             queue_push(&problem_queue, i);
-            problems << "Problem: " << solver_problem2str(m_solver, i) << "\n";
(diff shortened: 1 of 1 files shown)


wolfv:
I actually like to have some newlines in the output so that we get some visual "breaks". 
Maybe we can also have a bit of an indentation in front of the `-`?

```
  - nothing provides requested xtensora
  - conflict between xtensor-io and github
 
ERROR   Could not solve for environment specs
```

adriendelsalle:
Sounds good, will modify this and rebase on master

PR #824, opened 2021-04-01
Add a high-level API

adriendelsalle:
Description
---

Add high level api. This is a first implementation that will need further improvements.

Main modifications are:
- Move high-level API from `micromamba` codebase
- Move configuration from core to api
- Clean micromamba by removing useless headers

Side modifications are:
- allow `create` operation without specifications, resulting to an empty environment containing a `conda-meta/history` file
- add a `remove_all` argument to `remove` operation and add the corresponding `-a,--all` flag in `micromamba`
- add more control on `target_prefix` checks using 2 news options `MAMBA_EXPECT_EXISTING_PREFIX` and `MAMBA_ALLOW_NOT_ENV_PREFIX`
- add a `explicit_install` configurable (not rc configurable) to trigger explicit installation of specs
  - use it when parsing spec files with explicit specs
  - works with both `install` and `create` operations
  - add tests in the Python based test suite
- declare `UMAMBA_ONLY` flag when building the C++ test suite to avoid link error
- remove `version.hpp` generated file and update gitignore
- make the `banner` configurable from the context
- add an `api` level of configuration for `Configurable`
  - `api`>`cli`>`env var`>`rc`
- add capability to set or clear all `Configurable` values (`api`, `cli`, etc.)


include/mamba/api/construct.hpp
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) 2019, QuantStack and Mamba Contributors
+//
+// Distributed under the terms of the BSD 3-Clause License.
+//
+// The full license is in the file LICENSE, distributed with this software.
+
+#ifndef MAMBA_API_CONSTRUCT_HPP
```
wolfv, line 7: this part is really only used in conda constructor context and there we use micromamba. I think that could be moved back to micromamba?

adriendelsalle: sure we can move it back to `micromamba` only then

src/api/create.cpp
```diff
@@ -0,0 +1,32 @@
...
+#include "mamba/api/install.hpp"
+
+
+namespace mamba
+{
+    void create()
+    {
+        auto& config = Configuration::instance();
+
+        config.load(!MAMBA_ALLOW_ROOT_PREFIX & !MAMBA_ALLOW_FALLBACK_PREFIX
```
wolfv, line 17: what does the `!MAMBA_ALLOW_ROOT_PREFIX & ... ` syntax achieve? Does that work? For inverting bits it's usually `~....` if I am not mistaken.

wolfv: !... should cast to bool, and then will likely return `false` so that should be `0 & 0 & 0` here? Or am I missing something?

src/api/install.cpp
```diff
@@ -0,0 +1,560 @@
...
+                    | MAMBA_ALLOW_EXISTING_PREFIX);
+
+        auto& install_specs = config.at("specs").value<std::vector<std::string>>();
+        auto& use_explicit = config.at("explicit_install").value<bool>();
+
+        if (!install_specs.empty())
+        {
+            if (use_explicit)
+            {
+                detail::install_explicit_specs(install_specs);
```
wolfv, line 40: personally, I'd prefer these two functions to be (at least) part of the API, if not the only API.

src/api/create.cpp
adriendelsalle: oO I should check.
I could have used `0` instead of this, but it looked more informative to understand what options are set to false instead

src/api/list.cpp
```diff
@@ -0,0 +1,129 @@
...
+namespace mamba
+{
+    void list(const std::string& regex)
+    {
+        auto& config = Configuration::instance();
+
+        config.at("show_banner").get_wrapped<bool>().set_value(false);
+        config.load(MAMBA_ALLOW_ROOT_PREFIX | MAMBA_ALLOW_FALLBACK_PREFIX
+                    | MAMBA_ALLOW_EXISTING_PREFIX | MAMBA_ALLOW_MISSING_PREFIX
+                    | !MAMBA_ALLOW_NOT_ENV_PREFIX | MAMBA_EXPECT_EXISTING_PREFIX);
```
wolfv, line 24: again !MAMBA ... which is difficult to interpret

adriendelsalle: will work on that

src/api/install.cpp
adriendelsalle: I guess when adding a low-level API we should move it there because it only relies on `Context` configuration, which is not the case of high-level API that need the `Configuration` to be loaded to work correctly.

I can move it outside this namespace detail.

src/api/update.cpp
```diff
@@ -0,0 +1,50 @@
...
+
+        if (update_all)
+        {
+            PrefixData prefix_data(ctx.target_prefix);
+            prefix_data.load();
+
+            for (const auto& package : prefix_data.m_package_records)
+            {
+                auto name = package.second.name;
+                if (name != "python")
```
wolfv, line 34: this is probably not new, but I just wanted to note that we shouldn't just remove python here.

It shoudl work to update python minor versions with the python-pinning behavior you implemented, right?

adriendelsalle: > I just wanted to note that we shouldn't just remove python here.

I don't get it

This is done to not update `python` when using `-a,--all`/`update_all`

src/api/create.cpp
adriendelsalle: I added some tests to check it works as expected. Negation of int just gives the information that the flag is not set.
```
EXPECT_EQ(!MAMBA_EXPECT_EXISTING_PREFIX, 0);
EXPECT_EQ((!MAMBA_ALLOW_ROOT_PREFIX) | MAMBA_ALLOW_EXISTING_PREFIX, MAMBA_ALLOW_EXISTING_PREFIX);
```

adriendelsalle: I can remove that and use 0 instead, but I think it's much less informative

src/api/install.cpp
wolfv: no it's fine, keep it in that namespace. we can move that to core later.

src/api/update.cpp
wolfv: I think we do want to update python, but keep the minor version. 

So python 3.7.4 should update to 3.7.8, but not to 3.9.1

Is that a wrong assumption?

src/api/create.cpp
adriendelsalle: I added a `MAMBA_NOT_...` for each possible check, that is a bit verbose but easier to read!

PR #826, opened 2021-04-02
Add C API

adriendelsalle:
Description
---

Add a C API for high-level operations.

Other: remove use of `exit` in `remove` operation when aborted, return instead


wolfv:
nice!

Issue #814, opened 2021-03-31
[micromamba] allow creation of empty environment


wolfv:
When creating an empty environment, we should drop an empty "history" file into `$ENV/conda-meta/history` so that the directory is recognizable as a conda / mamba environment.

adriendelsalle:
It's implemented as part of https://github.com/mamba-org/mamba/pull/824

What could/should be improved is to return just after having creating the folders and the `history` file, preventing the download of channels repodata files and the solver to run with empty specs.

adriendelsalle:
We can close this now @wolfv ?

PR #827, opened 2021-04-02
Improve creation of empty env

adriendelsalle:
Description
---

Creating en empty environment is a nice feature. It has been enabled by https://github.com/mamba-org/mamba/pull/824.

This PR improves the initial implementation by:
- only create an empty `conda-meta/history` file without running the solver
- printing to the console a `Empty environment created at prefix: $prefix`
- adding a `success: true` entry to the JSON logger
- adding tests to the python based test suite

Solves https://github.com/mamba-org/mamba/issues/814


include/mamba/api/install.hpp
```diff
@@ -7,36 +7,10 @@
...
-
-#include <string>
-#include <vector>
-
 
 namespace mamba
 {
     void install();
-
-    void install_specs(const std::vector<std::string>& specs,
```
wolfv, line 25: hmm, I find moving these to `create.hpp` a bit counterinutive ... 

two other options: 

- merging create and install?
- adding a third file with these implementations?

adriendelsalle: Yeah maybe! I did that to avoid circular import. And it looked quite natural that create is a superset of install.
Let me find out something :)

PR #836, opened 2021-04-06
Create behavior with root_prefix

adriendelsalle:
Description
---

This PR is related to the behavior of `create` operation at `root_prefix`.

The current behavior is:
- not allow `target_prefix` to be `root_prefix` in `create` operation
  - a work around is to put some relative `/./` in either the `target_prefix` or the `root_prefix` to cheat. This has to be fixed
  - using `fs::weakly_canonical` only transform into canonical path the existing part of the prefix. The previously mentioned hack still works when appending `/./` to a non-preexisting `root_prefix`
- allow `create` elsewhere
  - even with no specs provided (to get an empty env), will be soonish supported. See #827.

`install` operation is not permitted at a non-existing `target_prefix`.
It's then really difficult to know how to `create`/`install` in `base` env/`root_prefix`.

Proposition
---

I would propose to:
- allow `target_prefix` to be `root_prefix` in `create` op
  - i.e. remove `MAMBA_NOT_ALLOW_ROOT_PREFIX` check (only use for `create`) 
- always create a `root_prefix` structure when the directory doesn't exist, during `root_prefix_hook` (e.g. before `target_prefix` checks)
  - empty `envs` and `pkgs` directories
  - empty `conda-meta/history` file

It will result in:
- `create` in `base` env/`root_prefix` will always throw an `Overwriting root prefix is not permitted` error
  - using `fs::weakly_canonical` will now work because the `root_prefix` is always created *before* `target_prefix` checks 
- `install` in `base` env/`root_prefix` will always be possible, even for a non pre-exisiting `root_prefix`

### DIFF
--- a/src/api/configuration.cpp
+++ b/src/api/configuration.cpp
@@ -169,6 +169,17 @@ namespace mamba
                                 (then restart or source the contents of the shell init script))");
             }
 
+            if (!fs::exists(prefix))
+            {
+                fs::create_directories(prefix / "conda-meta");
+                fs::create_directories(prefix / "envs");
+                fs::create_directories(prefix / "pkgs");
+                std::fstream history;
+                history.open(prefix / "conda-meta" / "history", std::fstream::out);
+                history.close();
+            }
+            prefix = fs::weakly_canonical(prefix);
+
             auto& ctx = Context::instance();
             ctx.envs_dirs = { prefix / "envs" };
             ctx.pkgs_dirs = { prefix / "pkgs" };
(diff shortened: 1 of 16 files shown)


src/api/configuration.cpp
```diff
@@ -169,6 +169,17 @@ namespace mamba
                                 (then restart or source the contents of the shell init script))");
             }
 
+            if (!fs::exists(prefix))
+            {
+                fs::create_directories(prefix / "conda-meta");
+                fs::create_directories(prefix / "envs");
```
wolfv, line 175: might be better to use path::touch? https://github.com/mamba-org/mamba/blob/79d3342a04bc57ef21b99c0b926b8a46b24dbafa/include/mamba/core/fsutil.hpp#L48

adriendelsalle: thanks for the pointer! It should be fixed now

PR #838, opened 2021-04-07
Add config set, get, append and prepend for micromamba

### LINKED ISSUE #797: [micromamba] implement config --set / --append ... commands
It would be cool to implement the equivalent to `conda config --set / ...` commands for micromamba:

- [x] `--set`
- [x] `--prepend`
- [x] `--append`
- [x] `--get`
- [x] `--remove`
- [x] `--remove-key`

These functions should read the current config from the `~/.mambarc` file, do the necessary modifications and write out the file.

marimeireles:
Closes #797

### DIFF
--- a/src/micromamba/config.cpp
+++ b/src/micromamba/config.cpp
@@ -4,6 +4,8 @@
 //
 // The full license is in the file LICENSE, distributed with this software.
 
+#include <fstream>
+
 #include "common_options.hpp"
 
 #include "mamba/api/config.hpp"
@@ -92,18 +94,12 @@ set_config_describe_command(CLI::App* subcom)
 void
 set_config_set_command(CLI::App* subcom)
 {
-}
-
-
-void
-set_config_get_command(CLI::App* subcom)
-{
     auto& config = Configuration::instance();
 
-    auto& key = config.insert(
-    Configurable("key", std::string("")).group("Output, Prompt and Flow Control").description("A shell type"));
+    auto& set_key = config.insert(
+    Configurable("set_key", std::vector<std::string>({})).group("Output, Prompt and Flow Control").description("Set configuration value on rc file"));
 
-    subcom->add_option("key", key.set_cli_config(""), key.description());
+    subcom->add_option("set_key", set_key.set_cli_config({}), set_key.description());
 
     subcom->callback([&]() {
         config.at("use_target_prefix_fallback").set_value(true);
@@ -113,9 +109,53 @@ set_config_get_command(CLI::App* subcom)
                        | MAMBA_ALLOW_NOT_ENV_PREFIX | MAMBA_NOT_EXPECT_EXISTING_PREFIX);
         config.load();
 
-        YAML::Emitter out;
+        auto srcs = config.sources();
+        auto valid_srcs = config.valid_sources();
+        std::ofstream rc_file;
 
-        std::cout << config.dump(MAMBA_SHOW_CONFIG_VALUES, { key.value() });
+        for (auto s : srcs)
+        {
+            auto found_s = std::find(valid_srcs.begin(), valid_srcs.end(), s);
+            if (found_s != valid_srcs.end())
+            {
+                rc_file.open(env::expand_user(s).string(), std::ios::app);
+                for (std::size_t i = 0; i < set_key.value().size() - 1 ; i++)
+                {
+                    rc_file << set_key.value()[i] << ": "<< set_key.value()[i + 1] << std::endl;
+                }
+                rc_file.close();
+            }
+            else
+            {
+                std::cout << env::expand_user(s).string() + " (invalid)" << std::endl;
+            }
+        }
+
+        config.operation_teardown();
+    });
+
+}
+
+
+void
+set_config_get_command(CLI::App* subcom)
+{
+    auto& config = Configuration::instance();
+
+    auto& get_key = config.insert(
+    Configurable("get_key", std::string("")).group("Output, Prompt and Flow Control").description("Display configuration value from rc file"));
+
+    subcom->add_option("get_key", get_key.set_cli_config(""), get_key.description());
+
+    subcom->callback([&]() {
+        config.at("use_target_prefix_fallback").set_value(true);
+        config.at("show_banner").set_value(false);
+        config.at("target_prefix_checks")
+            .set_value(MAMBA_ALLOW_EXISTING_PREFIX | MAMBA_ALLOW_MISSING_PREFIX
+                       | MAMBA_ALLOW_NOT_ENV_PREFIX | MAMBA_NOT_EXPECT_EXISTING_PREFIX);
+        config.load();
+
+        std::cout << config.dump(MAMBA_SHOW_CONFIG_VALUES, { get_key.value() }) << std::endl;
 
         config.operation_teardown();
     });
--- a/src/micromamba/config.cpp
+++ b/src/micromamba/config.cpp
@@ -6,6 +6,7 @@
 
 #include <fstream>
 #include <cstdlib>
+#include <typeinfo>
 
 #include "common_options.hpp"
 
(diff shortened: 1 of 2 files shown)


src/micromamba/config.cpp
```diff
@@ -184,6 +184,44 @@ set_config_describe_command(CLI::App* subcom)
...
+    {
+        auto& positional_arg = key.compute().value();
+        auto& configs = group_it.second;
+
+        for (auto& c : configs)
+        {
+            if (c->name() == positional_arg)
+            {
+                std::cout << "🍉🦀\n";
+                detail::print_configurable(out, *c, false);
```
marimeireles, line 216: I'm calling this function here because I need to get the value of a key `c` on this configurations dictionary. Is there a better way to do it?
This throws a segfault. Should I debug it deeper or do you have an idea of why it isn't working?

src/micromamba/config.cpp
```diff
@@ -184,6 +184,44 @@ set_config_describe_command(CLI::App* subcom)
...
+    auto& key = config.insert(Configurable("key", std::string())
+                                .group("cli")
+                                .rc_configurable(false)
+                                .description("Writes a new value for a key"));
+    subcom->add_option(
+        "key", key.set_cli_config({}), key.description());
+    YAML::Emitter out;
+
+    auto& specs = config.at("config_specs").get_wrapped<std::vector<std::string>>();
+    subcom->add_option("specs", specs.set_cli_config({}), specs.description());
```
marimeireles, line 201: Also, if i understood correctly this might be too much to be here, right? Should I create an init config?

adriendelsalle: no necessarily, depends if this init will be shared with another sub command or not

src/micromamba/config.cpp
adriendelsalle: I guess you get a dangling reference to the yaml emitter `out` that is define outside the callback.
You can also easily get the `ConfigurableInterface` by key in the `Configuration` using `at` as if it was a `std::map`:
```
subcom->callback([&]() {
    auto& c = config.at(key.compute().value());
    YAML::Emitter out;
    detail::print_configurable(out, c, false);
    std::cout << out.c_str() << std::endl;
});
```

src/micromamba/config.cpp
```diff
@@ -184,6 +184,44 @@ set_config_describe_command(CLI::App* subcom)
     });
 }
 
+void
+set_config_get_command(CLI::App* subcom)
+{
+    auto& config = Configuration::instance();
+
+    auto& key = config.insert(Configurable("key", std::string())
```
adriendelsalle, line 192: to disambiguate you can prepend the `Configurable` and the associated CLI option with `config_get_` for example

marimeireles: osome
I see
Thank you! :cherry_blossom:

adriendelsalle:
I also forgot to tell you that you need to load the configuration at the beginning of the callback to collect all the sources (including rc files) and compute resulting values.
Then you don't have to call `compute` anymore on a configurable because all the configurables are computed during this previous call.

marimeireles:
>load the configuration

How do you do that?

Also, a different context question:
In the `set` mode I need to set the variables somewhere. So,
1. How do I choose where to set these variables? Should we have a default for now?
2. Where is the data for the other places persisted? There are two other places to set variables, right? I only found how to do it for `rc` files (ofc!).

Thanks @adriendelsalle :)

adriendelsalle:
hey @marimeireles 
Just call `config.load(target_checks)` with the correct `target_checks` corresponding to your sub command.
Here you can use the same as `config list`, which is:
```
config.load(MAMBA_ALLOW_ROOT_PREFIX | MAMBA_ALLOW_FALLBACK_PREFIX
                    | MAMBA_ALLOW_EXISTING_PREFIX);
```
You may also want to disable the banner before loading:
```
config.at("show_banner").get_wrapped<bool>().set_value(false);
```

For the `config set` sub command, you have to check what `conda`is doing but I guess it's setting configuration value in a rc file.
The default rc file should be in our case `~/.mambarc`. We can provide extra location using CLI flags `--user` (=default location), `--root_prefix`, `--target-prefix`, etc.

marimeireles:
Are we adding tests for these or..?

adriendelsalle:
> Are we adding tests for these or..?

sure we are ! :)
For `config get`:
- this idea is probably to write a temporary rc file
- read that only source
- parse the result to be what was written previously
For `config set`:
- create a new test root prefix
- create a new empty env in that root prefix
- do the same checks and clean-up
- most annoying part is to save a pre-existing config file in home directory to not mess your config in dev mode :)

I can help a bit on that if you need

marimeireles:
Hey @adriendelsalle, I did something simpler following the example on the other tests. I'll push it here, so you can have a look!
Thanks for the reply :)

marimeireles:
Sorry @adriendelsalle, I hadn't submit my review that's why you didn't know what I was talking about!
If you can look at it and comment it'd be great :)
Thank youuu and sorry :see_no_evil:

src/micromamba/config.cpp
```diff
@@ -88,6 +137,324 @@ set_config_describe_command(CLI::App* subcom)
...
+{
+    auto& config = Configuration::instance();
+
+    auto& file_path = config.insert(Configurable("config_set_file_path", fs::path())
+                                     .group("cli")
+                                     .description("File path to set configuration"),
+                                     true);
+    subcom->add_option("--file", file_path.set_cli_config(fs::path()), file_path.description());
+
+    path::touch(file_path.value().string(), true);
```
marimeireles, line 152: `file_path` is empty! :/

marimeireles:
There's one thing @adriendelsalle pointed out that could be improved... that is the `is_key_valid` function. I don't really get what you said Adrien, I could just use `find`? I don't remember anymore why I wrote this function like this, just that it was one of the first things I wrote and I got inspired by your code on `list -a`.

I think that's the only thing left that I had to ask... All of the other problems I raised are now fixed.
Let me know if you need more tests! I kinda ran out of ideas to write them, maybe smth come up to you peeps.

marimeireles:
I also believe I squashed all of the useless commits, but lemme know if you want something dryer.
Sorry to upload it so late! See you Monday, lovely weekend! :heart:

src/micromamba/config.cpp
```diff
@@ -88,6 +137,336 @@ set_config_describe_command(CLI::App* subcom)
...
+void
+set_config_file_command(CLI::App* subcom)
+{
+    auto& config = Configuration::instance();
+
+    auto& file_path = config.insert(Configurable("config_set_file_path", fs::path())
+                                        .group("cli")
+                                        .description("File path to set configuration"),
+                                    true);
+    subcom->add_option("--file", file_path.set_cli_config(fs::path()), file_path.description());
```
wolfv, line 150: can we also handle `--system`, `--env` and `--home`? 

And the default value shoudl be `--home` -> `env::home_directory() / ".condarc"`

marimeireles: Jawohl, I'm on it :)

marimeireles:
I still need to write tests for the `--env` and `--system` flags.
I didn't implement a `--home` one because, as conda, mamba sets the `home` env as the default.

marimeireles:
Actually, I have some questions:
1. for the `--system` flag, the user normally won't have permission to create files under the sys path, I'm not the one who should handle these permissions, right? The user should figure out.
2. should I write tests for those?

src/micromamba/config.cpp
```diff
@@ -88,6 +185,422 @@ set_config_describe_command(CLI::App* subcom)
...
+                       | MAMBA_ALLOW_NOT_ENV_PREFIX | MAMBA_NOT_EXPECT_EXISTING_PREFIX);
+        config.load();
+
+        fs::path rc_source = env::expand_user(env::home_directory() / ".condarc");
+        std::ofstream rc_file;
+
+        std::string prepend_key = prepend_map.value().front();
+        std::string prepend_value = prepend_map.value().back();
+
+        if (file_path.configured())
```
wolfv, line 245: there is a bit of code-duplication going on now. Do you think we can refactor that into a single callback? You can move the code that gets the config object in the callback, too.

You could have a function that is called:

```
enum CONFIG_MOD_ACTION { SET, APPEND, PREPEND };

void modify_config(CONFIG_MOD_ACTION action)
{
    auto& config = Config::instance();
    config.load();
    auto& filepath = config.at("config_set_file_path").get_wrapped<fs::path>();
     ...
}

// and then in the callbacks:

subcom->callback([&]() {
    modify_config(CONFIG_MOD_ACTION::SET);
}
...
```

marimeireles: Sorry, I just saw the comment now.
Yeah, there's a bunch of code duplication in these methods.

PR #843, opened 2021-04-08
Update dependencies

### LINKED ISSUE #841: Dependency issue with development installation
Hi!

I've been trying to build the last version of Mamba on a fresh Ubuntu 20.04 VM, following the instructions in the readme. When running `pip install -e .`, I run into the following error:
```
    src/core/repo.cpp: In member function 'bool mamba::MRepo::read_file(const string&)':
    src/core/repo.cpp:247:60: error: 'CONDA_ADD_USE_ONLY_TAR_BZ2' was not declared in this scope
      247 |         int flags = Context::instance().use_only_tar_bz2 ? CONDA_ADD_USE_ONLY_TAR_BZ2 : 0;
          |                                                            ^~~~~~~~~~~~~~~~~~~~~~~~~~
    error: command 'gcc' failed with exit status 1
```

Somehow it seems related to 748060c. When running `mamba install ... libsolv ... -c conda-forge` from the readme, version 0.7.16 is installed on my machine but version >=0.7.18 is required since that commit.

Aratz:
Description
---------------
- Update dependency to `libsolv`
- Improve `README.md`

Closes #841

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,9 +138,9 @@ For more instructions (including OS X) check out https://gist.github.com/wolfv/f
 
 ### Development installation
 
-Make sure to have the following requirements in your conda environment:
+Make sure you install mamba's requirements in your conda environment:
 
-`mamba install cmake compilers pybind11 libsolv libarchive libcurl nlohmann_json pip cpp-filesystem yaml-cpp reproc-cpp -c conda-forge`
+`mamba env update --name <conda_env> --file environment-dev.yml`
 
 If you build mamba in a different environment than base, you must also install conda in
 that environment:
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -7,7 +7,7 @@ dependencies:
   - ninja
   - nlohmann_json
   - pybind11
-  - libsolv
+  - libsolv >=0.7.18
   - libarchive
   - libcurl
   - cxx-compiler


README.md
```diff
@@ -138,9 +138,9 @@ For more instructions (including OS X) check out https://gist.github.com/wolfv/f
 
 ### Development installation
 
-Make sure to have the following requirements in your conda environment:
+Make sure you install mamba's requirements in your conda environment:
 
-`mamba install cmake compilers pybind11 libsolv libarchive libcurl nlohmann_json pip cpp-filesystem yaml-cpp reproc-cpp -c conda-forge`
+`mamba env update --name <conda_env> --file environment-dev.yml`
```
adriendelsalle, line 143: I guess we could propose to use `install` command to be consistent with text `install mamba's requirements` ?
But also propose to `update` an existing env :)

adriendelsalle (commented):
Thanks for that @Aratz !
A minor update would be great :)

README.md
Aratz: Sure! `conda install` doesn't like yaml files though, so I had to place the dependencies in a new text file. Thanks for the prompt answer :)

adriendelsalle (approved):
Thanks!

README.md
```diff
@@ -138,9 +138,9 @@ For more instructions (including OS X) check out https://gist.github.com/wolfv/f
 
 ### Development installation
 
-Make sure to have the following requirements in your conda environment:
+Make sure you install mamba's requirements in your conda environment:
 
-`mamba install cmake compilers pybind11 libsolv libarchive libcurl nlohmann_json pip cpp-filesystem yaml-cpp reproc-cpp -c conda-forge`
+`mamba install --file requirements.txt -c conda-forge`
```
maresb, line 143: It would be more conda-centric to use the command `mamba env update --name mamba-dev --file environment-dev.yml`. Then we can get rid of `requirements.txt`.

maresb: I think `conda` should be added to `environment-dev.yml` as per the comment in `README.md` about building mamba in a different environment than base.

I have success with the following commands:
```bash
docker run --rm -it condaforge/mambaforge
git clone https://github.com/Aratz/mamba.git
cd mamba
git checkout fix/841

echo "  - conda" >> environment-dev.yml
rm requirements.txt

mamba env update --name mamba-dev --file environment-dev.yml

conda activate mamba-dev
pip install -e .
```

Aratz: Ah, this is what I initially proposed in 6f0baae. What's the main difference between `conda install` and `conda env update` ? They seem quite interchangeable to me so I'd be fine either way.

maresb: @Aratz, that's a very good question. I wish I knew the answer.

I pulled that particular command from a script I wrote a while back. At the time I think I had a good reason for choosing `conda env update`. It may have been that I was working around some conda bug which has since been fixed. Unfortunately some aspects of conda remind me of [git](https://xkcd.com/1597/) ;)

maresb: I also find it very bizarre that mamba doesn't install `0.7.18` in the first place.

Perhaps there's some conflict with having all packages in their latest versions, and it pseudorandomly decides that `libsolv` loses?

adriendelsalle: > I think `conda` should be added to `environment-dev.yml` as per the comment in `README.md` about building mamba in a different environment than base.
> 
> I have success with the following commands:
> 
> ```shell
> docker run --rm -it condaforge/mambaforge
> git clone https://github.com/Aratz/mamba.git
> cd mamba
> git checkout fix/841
> 
> echo "  - conda" >> environment-dev.yml
> rm requirements.txt
> 
> mamba env update --name mamba-dev --file environment-dev.yml
> 
> conda activate mamba-dev
> pip install -e .
> ```

you're right, `conda` is missing!

> Ah, this is what I initially proposed in [6f0baae](https://github.com/mamba-org/mamba/commit/6f0baaea48bb2c4384dba93ed760db70327dd9d5). What's the main difference between `conda install` and `conda env update` ? They seem quite interchangeable to me so I'd be fine either way.

We have to distinguish `conda` and `mamba` here.
You point an issue about `mamba` implementation that only handles install-like command using a spec file with `env` sub-command:
https://github.com/mamba-org/mamba/blob/38b374c83952385fc3d7f9fc57e2a264b924bbd2/mamba/mamba_env.py#L107

We should mentioned that in the original issue @Aratz opened.
Maybe it could also help to run use `micromamba` while we are fixing this:
```
docker run --rm -it mambaorg/micromamba
apt update && apt install git -y
git clone https://github.com/mamba-org/mamba.git
cd mamba

echo "  - conda" >> environment-dev.yml
micromamba create -f environment-dev.yml
micromamba activate mamba-dev
pip install -e .
```

adriendelsalle:
Hi @Aratz !
I don't understand your last changes about `gmock`. There is no need of that package for building C++ test suite.
Can you provide what issue you faced without install it ?

Aratz:
@adriendelsalle when I tried to compile two days ago, I got an error message mentioning  `gmock`. After installing it, mamba compiled normally so I figured it was another package that had been left out.

However, today when trying to reproduce the issue, everything compiled normally :expressionless:

I'll reset this commit until I can find how to reproduce this issue and understand what the problem is :monocle_face:

adriendelsalle:
I was able to reproduce building in a micromamba docker container.
GTest and GMock are part of the same project and the CMake install includes both target. Since we don't really like to install unnecessary deps, I opened a PR to enable building GTest from source instead https://github.com/mamba-org/mamba/pull/853

I also realized that the CI tests wer broken due to that and incorrectly returned successful. So thanks for the pointer!!

Aratz:
Happy to help :)

adriendelsalle:
EDIT: we just fixed the `gmock` error without downloading sources by importing only GTest using `add_library` on unix OS.

wolfv:
@Aratz thanks for your contributions! I am not sure if we should have another file (requirements.txt) when we already have the dev-env.yml file.

@maresb I think you hit that annoying thing in conda (and by proxy mamba) that `conda env` and `conda install ...` are two completely separate commands and only `conda env ...` can read `yaml` files. That's probably why you used conda env update instead of `conda install --file devenv.yml`.

Since mamba should usually be in base can we change the command to install the devenv into base? maybe by changing the install directory? 

Does

```
mamba env update --name  base --file environment-dev.yml
```

work?

My main concern is that we should only have one list of dependencies, ideally a list that we also use in the CI. Otherwise we'll get out of sync again.

Aratz:
I agree it's better with only one file, `mamba env update --name  base --file environment-dev.yml` works fine, I'll reset my branch to 6f0baae then :slightly_smiling_face:

PR #845, opened 2021-04-09
Ultra dry tests

adriendelsalle:
Description
---

Create ultra dry tests to fasten CI.
Uses `Configuration` status after being loaded instead of `dry_run`/`--dry-run` that still requires fetching the repodata.json and running the solver.

### DIFF
--- a/include/mamba/api/configuration.hpp
+++ b/include/mamba/api/configuration.hpp
@@ -18,6 +18,33 @@
 #include <functional>
 
 
+#ifdef ENABLE_CONTEXT_DEBUG_PRINT
+#define CONTEXT_DEBUGGING                                                                          \
+    if (Configuration::instance().at("print_context_only").value<bool>())                          \
+    {                                                                                              \
+        Context::instance().debug_print();                                                         \
+        exit(0);                                                                                   \
+    }
+#define CONFIG_DEBUGGING                                                                           \
+    if (Configuration::instance().at("print_config_only").value<bool>())                           \
+    {                                                                                              \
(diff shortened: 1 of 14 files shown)


adriendelsalle:
@wolfv This PR allows to save about 10min on the CI but also masks some issues met when executing real operations.
For example, the recent issue about Windows long paths not supported in `cmd.exe` causing errors when installing noarch packages is not caught here.

Any recommendations on where to run normal/dry/ultra_dry tests ?

wolfv:
I think everything about CLI parsing can be run in ultra-dry mode.

include/mamba/api/configuration.hpp
```diff
@@ -18,6 +18,33 @@
...
 
 
+#ifdef ENABLE_CONTEXT_DEBUG_PRINT
+#define CONTEXT_DEBUGGING                                                                          \
+    if (Configuration::instance().at("print_context_only").value<bool>())                          \
+    {                                                                                              \
+        Context::instance().debug_print();                                                         \
+        exit(0);                                                                                   \
+    }
+#define CONFIG_DEBUGGING                                                                           \
```
wolfv, line 28: this is defined also a couple of lines below

wolfv: sorry. misread!

adriendelsalle: :)

adriendelsalle: :)

wolfv:
looks great!

PR #846, opened 2021-04-10
Change pybind11 to a build dependency

maresb:
This closes https://github.com/conda-forge/mamba-feedstock/issues/109

Adding pybind11 to `install_requires` makes it a runtime requirement, so I created a `pyproject.toml` and added it instead as a build requirement.


wolfv:
Thanks!

PR #849, opened 2021-04-13
Fix implicit conversion of fs::path on windows

adriendelsalle:
Description
---

`cpp-filesystem` uses native storage since 1.5.0 (see https://github.com/gulrak/filesystem#v150).
We need at least to define the `GHC_WIN_DISABLE_WSTRING_STORAGE_TYPE` to fix compiling issues with newest versions installed using `conda-forge` (and available from today).

I also needed to fix some C++ tests failing on Windows.


wolfv:
still some failing tests?

wolfv:
looks good to me. just the lint :)

adriendelsalle:
Will fix that, I also have to check that everything is fine with cmd.exe and bash on windows. will ping you :)

Issue #852, opened 2021-04-13
Bug of handling package cache in secondary locations


wenjuno:
### The problem

When the package cache exists in a secondary location, mamba will fail to create environment if it needs the package cache. This can be reproduced by these steps:

1. Create a new environment with some packages (e.g. python).
2. Set the "pkgs_dirs" or CONDA_PKGS_DIRS environment variable to a different directory.
3. Create a new environment with the same spec.

Mamba will fail to extract the package cache file and it thinks the archive file is in the first cache location. The error message from the exception is a bit misleading probably because it was thrown after the actual error so it doesn't have much clue there: "RuntimeError: Found incorrect download. Aborting". To see more helpful error message, you need to add "-v" to see the logging from C++ code and (sometimes) it contains the paths of the tarballs it tries to extract: "ERROR   Error when extracting package: /home/wenjun/tmp/conda_pkgs_2/ld_impl_linux-64-2.33.1-h53a641e_7.conda: Could not open archive for reading."

I think this is the root cause of #779. 

### My solution

I think the problem is in this function
https://github.com/mamba-org/mamba/blob/38b374c83952385fc3d7f9fc57e2a264b924bbd2/src/core/transaction.cpp#L223-L241

The `cache_path` passed in is the first writable cache directory. The tarball path and the extracted cache path (`dest_dir`) are synthesized by using `cache_path` as prefix. The `query` function will return true regardless of the package cache location, so when the package cache is in a secondary location and it cannot see the `dest_dir`, it will try to extract the tarball but the synthesized tarball path points to the first cache location and it's wrong (inconsistent).

I think there are four cases here we need to handle (sorry for the badly formatted pivot table, I'm not sure how to format it):

| same_cache_dir \ dest_exists | Yes |    No    |
|------------------------------|-----|----------|
|  Yes                         | Use | Extract  |
|   No                         | Use | Download |

*same_cache_dir*: The first writable cache directory (`cache_path`) is the same as the directory of first available cache (tarball and/or extracted), so we can safely extract the tarball to the same directory if we need to.
*dest_exists*: The extracted package cache exists in the _directory of first available cache_.

Note here that if the extracted package cache exists we don't really care about whether the tarball exists and where the tarball is because conda will use the extracted cache. One thing that puzzles me is that if the extracted cache is in the first location and tarball is in a secondary location conda still tries to unzip the tarball in-place, which looks unnecessary to me but I don't think it makes difference here.

Any comment is appreciated. Thanks.

[...]

PR #856, opened 2021-04-15
Handle package cache in secondary locations.

wenjuno:
This is the fix for #852. I changed my mind about the solution and I think it's a better and easier one:

1. If there is extracted cache, use it, otherwise next.
2. If there is tarball in the first writable cache path, extract it, otherwise next.
3. Run the full download pipeline.

I have two questions that I think might be related to the decision I'm not aware of but I'm very curious about them:

1. In `py_interface.cpp` the `execute` member function of `MTransaction` is wrapped in a lambda function. Why not use the member function's address?
2. The member `PackageCacheData::m_valid_cache` is `std::map`. Why not use `std::unorder_map` instead? I basically follow the style and use `std::map` for the cache in `MultiPackageCache`.

Thanks.

### DIFF
--- a/include/mamba/core/package_cache.hpp
+++ b/include/mamba/core/package_cache.hpp
@@ -40,6 +40,7 @@ namespace mamba
         void set_writable(Writable writable);
         Writable is_writable();
         fs::path get_pkgs_dir() const;
+        void clear_query_cache(const PackageInfo& s);
 
         bool query(const PackageInfo& s);
 
@@ -59,11 +60,15 @@ namespace mamba
         MultiPackageCache(const std::vector<fs::path>& pkgs_dirs);
         PackageCacheData& first_writable();
 
-        bool query(const PackageInfo& s);
+        fs::path query(const PackageInfo& s);
+        fs::path first_cache_path(const PackageInfo& s, bool return_empty = true);
         std::vector<PackageCacheData*> writable_caches();
+
+        void clear_query_cache(const PackageInfo& s);
 
     private:
         std::vector<PackageCacheData> m_caches;
+        std::map<std::string, std::string> m_path_cache;
     };
 }  // namespace mamba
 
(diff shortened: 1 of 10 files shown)


wenjuno:
Previous failure was because I was using cpp-filesystem 1.4 and it has a bug with `exists` function such that it returns true for dangling symlink. The bug was fixed in 1.5.4: https://github.com/gulrak/filesystem/issues/107. In the last commit I add a check with `is_symlink` so if the file is a symlink it will consider it exist no matter whether it's a dangling symlink or not.

wolfv:
Hi @wenjuno thanks so much for this PR! We're currently preparing a release, and we'll review this right after. I hope we can make another release early next week to get this in!

wolfv:
btw. we also just updated to the latest cpp-filesystem (and faced some issues). Hopefully those issues are resolved for now (mostly around long path support on Windows). I am happy to hear that the update also fixed something for you!

wenjuno:
Hi @wolfv, thanks. I see you changed the master branch, do you want me to rebase my branch?

To be clear, the cpp-filesystem 1.5 didn't fix something for me, maybe my comment above is a bit confusing. What I mean is I was using 1.4 for local testing and it has the bug so the code worked for me but CI has started using 1.5 around the time so it failed the CI test the first time. I went back and found out the code that relies on the 1.4 behavior (the `exists` function) and fixed that in my latest commit. The change of behavior also reflects in the master branch CI. You can see the warning message: "WARNING Missing file from package cache: "/home/runner/mambaroot/pkgs/_openmp_mutex-4.5-1_gnu/lib/libgomp.so.1"". After this warning mamba will remove the package cache of _openmp_mutex. Later on mamba will extract the package cache from tarball but the logic in master doesn't check the package cache again so it doesn't remove the package cache again so it works when linking the package files.

Thanks.

wenjuno:
There is a fork in the history so I rebased my three commits.

wolfv:
thanks for fixing this symlink extract bug and improving the situation with multiple package caches! Excellent PR

wenjuno:
Thanks, I'm glad I can help. I will close the associated issue.

PR #859, opened 2021-04-16
Fix computation of target prefix

adriendelsalle:
Description
--

Fix the computation of target prefix using `fs::weakly_canonical` currently misplaced.
Also catch the trailing fs separators due to possible trailing not existing directories such as `/some/./not/existing/prefix/./`.

### DIFF
--- a/src/api/configuration.cpp
+++ b/src/api/configuration.cpp
@@ -125,7 +125,6 @@ namespace mamba
                                     the current working directory, use './some_prefix')")
                                                 .c_str());
                 }
-                prefix = fs::absolute(prefix);
             }
             else
             {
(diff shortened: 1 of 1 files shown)


src/api/configuration.cpp
```diff
@@ -125,14 +125,21 @@ namespace mamba
...
             }
             else
             {
                 bool use_fallback = config.at("use_target_prefix_fallback").value<bool>();
                 if (use_fallback)
                     prefix = std::getenv("CONDA_PREFIX") ? std::getenv("CONDA_PREFIX") : "";
             }
+
+#ifdef _WIN32
+            std::string sep = "\\";
```
wolfv, line 137: do we not have a function for this? ... 

or is there a function in fs::path?

wolfv: there is preferred_seperator in https://en.cppreference.com/w/cpp/filesystem/path

adriendelsalle: I searched for, but no.
maybe we can create one if you prefer

wolfv: there is preferred_seperator in https://en.cppreference.com/w/cpp/filesystem/path

PR #860, opened 2021-04-16
Improve base env creation

adriendelsalle:
Description
--

We currently automatically create empty base env when creating an env (even when itself empty) but `micromamba create base` is prohibited by `target_prefix_checks`.

Features:
- allow use of "base" `env_name` or `target_prefix`==`root_prefix` if specs is empty
  - this check is moved from `target_prefix_checks` to `create` operation
  - the prompt for possibly overwriting the prefix is transfered downstream to the operation
- always try to create base env, not only when `root_prefix` does not exist
  - fixes cases where an existing directory wants to be used as root prefix

Refactor tests accordingly and also remove "non-dry" tests already covered by config "ultra-dry" tests to speed-up


src/api/configuration.cpp
```diff
@@ -172,12 +179,10 @@ namespace mamba
                                 (then restart or source the contents of the shell init script))");
             }
 
-            if (!fs::exists(prefix))
-            {
-                path::touch(prefix / "conda-meta" / "history", true);
-                fs::create_directories(prefix / "envs");
-                fs::create_directories(prefix / "pkgs");
-            }
+            path::touch(prefix / "conda-meta" / "history", true);
```
wolfv, line 182: what happens if you want to initialise a root prefix in a location that already contains some files?

wolfv: ok sorry, forget what I said. this check is probably somewhere else :)

adriendelsalle: here is just a touch and the create dir ops just fail without exception if the directories exist.
do you think it's ok ?

wolfv: yes!

adriendelsalle:
https://github.com/mamba-org/mamba/issues/814

PR #872, opened 2021-04-20
Fix channel_priority computation

adriendelsalle:
Description
--

Fix `--strict-channel-priority` and `--no-channel-priority` that are currently broken.
There is currently a multiple computation conflict and error thrown with `channel_priority` computation.

Add tests


tdegeus:
Thanks! It would be great to release a patch if this is merged!

wolfv:
we're on it! :)

PR #879, opened 2021-04-21
C-API teardown on error

adriendelsalle:
Description
--

Call configuration operation teardown when catching an error in C-API.
This prevent memory effect on next operation.


wolfv:
we should maybe move to a RAII pattern for this at some point.

PR #883, opened 2021-04-21
Make platform rc configurable

adriendelsalle:
Description
--

The `platform` is used to get/fetch the packages from channels that correspond to your system.
It's already possible to modify it using `CONDA_SUBDIR` env variable but it could be pretty convenient to have RC configurable. 

For example, ones can then place an rc file in a given env that will act has it was on another platform to install corresponding packages. Pretty convenient when using emulators or interpreted code.

Add tests

### DIFF
--- a/include/mamba/core/context.hpp
+++ b/include/mamba/core/context.hpp
@@ -17,6 +17,60 @@
 
 namespace mamba
 {
+    namespace
+    {
+// Linux
+#if defined(__linux__)
+#if __x86_64__
+        static const char MAMBA_PLATFORM[] = "linux-64";
+#elif defined(i386)
+        static const char MAMBA_PLATFORM[] = "linux-32";
+// armv6l and armv7l
+#elif defined(__arm__) || defined(__thumb__)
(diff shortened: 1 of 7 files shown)


include/mamba/core/context.hpp
```diff
@@ -108,8 +162,9 @@ namespace mamba
 
         void set_verbosity(int lvl);
 
-        static std::string platform();
-        static std::vector<std::string> platforms();
+        std::string host_platform = MAMBA_PLATFORM;
+        std::string platform = MAMBA_PLATFORM;
```
wolfv, line 166: you said: no api change!

wolfv:
@beckermr this is the PR that implements a "rc-settable" platform. So one can add `platform: osx-64` to an RC file (and rc-files can be per-environment). 

One interesting use-case is for M1-macbooks as that allows people to run older Python through Rosetta 2 (I tried running Python 3.6 yesterday, that worked really well!).

Do you think we should file an issue on conda?

PR #891, opened 2021-04-23
Make 'env update' update packages if they're already installed

### LINKED ISSUE #854: 'update' with an environment file inconsistent with conda
Running an update using an environment file does not work as expected to update installed packages, using both micromamba and mamba

With a basic environment file:
```yaml
# env-1.yml
channels:
- conda-forge
dependencies:
- python>=3.8
- numpy=1.18.1
- scipy
```
running `micromamba create -p $PWD/env -f env-1.yml` gives an envionment with python 3.8.8 and numpy 1.18.1, as expected.

Now I'd like to update scipy to its latest version, so I run `micromamba update -p $PWD/env -f env-1.yml`. This now installs python 3.9.2 and numpy 1.20.2, breaking the numpy constraint in the environment file.

If on the other hand I use mamba, and run `mamba env create --prefix $PWD/env -f env-1.yml` to get python 3.8.8 and numpy 1.18.1 again, then do an update with `mamba env update --prefix $PWD/env -f env-1.yml` nothing happens as expected - the  constraints are already satisfied. Now if I change the environment file itself to remove the pin on numpy:
```yaml
# env-2.yml
channels:
- conda-forge
dependencies:
- python>=3.8
- numpy
- scipy
```
and run `mamba env update --prefix $PWD/env -f env-2.yml` to get the latest numpy (presumably 1.20.2), mamba reports `All requested packages already installed`, and doesn't update numpy.

Using conda directly with `conda env update --prefix $PWD/env -f env-2.yml` updates numpy to 1.20.2 as expected.

**Expected Behaviour**:

After creating an environment with `env-1.yml`:
* Running an update with `env-1.yml` keeps numpy pinned to 1.18.1
* Running an update with `env-2.yml` removes the numpy pin, installing some future version e.g. 1.20.2

ScottWales:
This is an attempt to fix #854. The issue appears to be that `api.SOLVER_INSTALL` won't do anything if a package is already installed, so `env update` doesn't update packages.

I've split the install jobs, so that packages newly added still use `api.SOLVER_INSTALL`, but if a package already exists it will instead be updated with `api.SOLVER_UPDATE`. This matches the behavior of `conda env update`.

### DIFF
--- a/mamba/mamba_env.py
+++ b/mamba/mamba_env.py
@@ -66,8 +66,8 @@ def mamba_install(prefix, specs, args, env, *_, **kwargs):
     # Also pin the Python version if it's installed
     # If python was not specified, check if it is installed.
     # If yes, add the installed python to the specs to prevent updating it.
+    installed_names = [i_rec.name for i_rec in installed_pkg_recs]
     if "python" not in [s.name for s in match_specs]:
-        installed_names = [i_rec.name for i_rec in installed_pkg_recs]
         if "python" in installed_names:
             i = installed_names.index("python")
             version = installed_pkg_recs[i].version
--- a/mamba/mamba_env.py
+++ b/mamba/mamba_env.py
@@ -105,7 +105,11 @@ def mamba_install(prefix, specs, args, env, *_, **kwargs):
     if pinned_specs_info:
         print(f"\n  Pinned packages:\n\n{pinned_specs_info}\n")
 
-    solver.add_jobs(specs, api.SOLVER_INSTALL)
+    install_specs = [s for s in specs if MatchSpec(s).name not in installed_names]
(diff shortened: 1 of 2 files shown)


wolfv:
@ScottWales thanks, this is great! 

I think we need to figure one other thing out: how we can update _with_ a spec.

E.g. if I have numpy 1.14 installed and I want to upgrade (but to a maximum of 1.16, when 1.20 is available) I should be able to do `mamba update "numpy<=1.16"`. I am going to have a look at how to do that.

ScottWales:
Thanks @wolfv, being able to control the version would be good as well

wolfv:
Thanks! I've opened a PR to constrain the version when updating: #892

PR #893, opened 2021-04-23
Deactivate CA certs search offline

adriendelsalle:
Description
--

Do not search for ca certs in offline mode
Move curl ssl init in ssl verify hook, that way it is not called in offline mode


adriendelsalle:
https://github.com/conda-forge/miniforge/pull/108
https://github.com/mamba-org/mamba/issues/755

PR #895, opened 2021-04-23
Micromamba only implementations

adriendelsalle:
Description
--

Add a UMAMBA_STATIC definition to only call init curl ssl in that case
Not throw an error on pkg cache only on mamba, should be thrown using libmamba through rhumba for example.
Remove a guard on `add_virtual_packages`, only called in `install_specs` (not called in mamba)

### DIFF
--- a/src/micromamba/CMakeLists.txt
+++ b/src/micromamba/CMakeLists.txt
@@ -31,6 +31,8 @@ target_link_libraries(micromamba PRIVATE mamba-static)
 message("Using static dependencies for micromamba: ${STATIC_DEPENDENCIES}")
 
 if (STATIC_DEPENDENCIES AND UNIX)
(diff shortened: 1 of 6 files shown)


src/micromamba/CMakeLists.txt
```diff
@@ -31,6 +31,8 @@ target_link_libraries(micromamba PRIVATE mamba-static)
 message("Using static dependencies for micromamba: ${STATIC_DEPENDENCIES}")
 
 if (STATIC_DEPENDENCIES AND UNIX)
+    add_definitions(-DUMAMBA_STATIC)
```
wolfv, line 34: this is the wrong place to add it, isn't it? we're not building the library here? Am I missing something?

adriendelsalle: it's totally the wrong place, sorry for that

PR #904, opened 2021-04-26
ignore failing pyc compilation for noarch packages

wolfv:


src/core/link.cpp
```diff
@@ -999,18 +999,28 @@ namespace mamba
...
+        auto [_, ec] = reproc::run(
+            wrapped_command, options, reproc::sink::string(out), reproc::sink::string(err));
 
-        if (ec)
+        if (ec || !err.empty())
         {
-            throw std::runtime_error(ec.message());
+            LOG_INFO << "noarch pyc compilation failed (cross-compiling?). " << ec.message();
+            LOG_INFO << err;
+            return {};
```
adriendelsalle, line 1023: I would prefer to warn on error that are authorized to fail, it's not a regular info log level

wolfv: it would clutter the logs for cross-compilation a lot. Also this is actually mostly harmless.

PR #907, opened 2021-04-26
PR template

adriendelsalle:
Description
--

Add a PR template to clarify what basic info would be great to be filled when opening a PR :)


wolfv:
I hate PR templates :/

adriendelsalle:
I guess we can find the right balance on this.
I agree that  tons of unnecessary check boxes are not helpful at all. But I also believe that raw PR with only a title to help understanding what's done here is not enough at all.

PR #909, opened 2021-04-27
Remove target prefix checks for activation

adriendelsalle:
Description
---

Shell activate was using `target_prefix` checks even if it uses its own `shell_prefix` Configurable value.
Deactivate `target_prefix` checks for this operation to avoid wrong errors thrown.

improve and clean tests (cc @ashwinvis)

### DIFF
--- a/test/micromamba/test_shell.py
+++ b/test/micromamba/test_shell.py
@@ -161,7 +161,7 @@ class TestShell:
     @pytest.mark.parametrize("prefix_is_root", [False, True])
     @pytest.mark.parametrize("prefix_exists", [False, True])
     @pytest.mark.parametrize(
-        "prefix_type", ["shrinked_prefix", "expanded_prefix" "name"]
+        "prefix_type", ["shrinked_prefix", "expanded_prefix", "name"]
     )
     def test_activate(self, shell_type, prefix_is_root, prefix_exists, prefix_type):
         skip_if_shell_incompat(shell_type)
--- a/test/micromamba/test_shell.py
+++ b/test/micromamba/test_shell.py
@@ -160,18 +160,18 @@ class TestShell:
     @pytest.mark.parametrize("shell_type", ["bash", "posix", "powershell", "cmd.exe"])
     @pytest.mark.parametrize("prefix_is_root", [False, True])
     @pytest.mark.parametrize("prefix_exists", [False, True])
-    @pytest.mark.parametrize("expanded_home", [False, True])
-    @pytest.mark.parametrize("prefix_type", ["prefix", "name"])
-    def test_activate(
-        self, shell_type, prefix_is_root, prefix_exists, prefix_type, expanded_home
-    ):
+    @pytest.mark.parametrize(
+        "prefix_type", ["shrinked_prefix", "expanded_prefix" "name"]
+    )
+    def test_activate(self, shell_type, prefix_is_root, prefix_exists, prefix_type):
         skip_if_shell_incompat(shell_type)
 
         if prefix_exists:
             # Create the environment for this test, so that it exists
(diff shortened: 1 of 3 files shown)


test/micromamba/test_shell.py
```diff
@@ -152,20 +158,22 @@ def custom_shell(shell):
             assert res["actions"]["print"][0] == ""
 
     @pytest.mark.parametrize("shell_type", ["bash", "posix", "powershell", "cmd.exe"])
-    @pytest.mark.parametrize("root", [False, True])
-    @pytest.mark.parametrize("env_exists", [False, True])
+    @pytest.mark.parametrize("prefix_is_root", [False, True])
+    @pytest.mark.parametrize("prefix_exists", [False, True])
     @pytest.mark.parametrize("expanded_home", [False, True])
     @pytest.mark.parametrize("prefix_type", ["prefix", "name"])
```
ashwinvis, line 164: expanded_home case can be combined with prefix_type
```suggestion
    @pytest.mark.parametrize("prefix_type", ["prefix", "prefix_expanded_home", "name"])
```
and then made into a separate condition in the `if prefix_type ...` block

test/micromamba/test_shell.py
```diff
@@ -152,20 +158,22 @@ def custom_shell(shell):
...
+        self, shell_type, prefix_is_root, prefix_exists, prefix_type, expanded_home
+    ):
         skip_if_shell_incompat(shell_type)
 
-        if env_exists:
+        if prefix_exists:
             # Create the environment for this test, so that it exists
             create("-n", TestShell.env_name, "-q", "--offline", no_dry_run=True)
         else:
             shutil.rmtree(TestShell.root_prefix)
```
ashwinvis, line 174: This may not cover `prefix_is_root = False`. 
```suggestion
            if prefix_is_root:
                shutil.rmtree(TestShell.root_prefix)
            else:
                shutil.rmtree(TestShell.prefix)
```

test/micromamba/test_shell.py
adriendelsalle: thanks for the suggestion, yes we can have a lighter test matrix here avoiding unnecessary tests

test/micromamba/test_shell.py
adriendelsalle: I don't get this one, the target prefix of the test is located at root_prefix/envs/env_name == target_prefix.
So removing the root_prefix that is always created at test setup as an empty dir should do the job right ?

adriendelsalle: But yeah it's more *accurate* to consider that only the target_prefix doesn't exist. And when it's not the root prefix, we keep the root prefix.
The else condition is not necessary since at test setup we always get an empty root prefix
Will do that.

adriendelsalle:
Thanks for your review @ashwinvis , fresh eyes are always appreciated

test/micromamba/test_shell.py
```diff
@@ -160,18 +160,18 @@ def custom_shell(shell):
     @pytest.mark.parametrize("shell_type", ["bash", "posix", "powershell", "cmd.exe"])
     @pytest.mark.parametrize("prefix_is_root", [False, True])
     @pytest.mark.parametrize("prefix_exists", [False, True])
-    @pytest.mark.parametrize("expanded_home", [False, True])
-    @pytest.mark.parametrize("prefix_type", ["prefix", "name"])
-    def test_activate(
-        self, shell_type, prefix_is_root, prefix_exists, prefix_type, expanded_home
-    ):
+    @pytest.mark.parametrize(
+        "prefix_type", ["shrinked_prefix", "expanded_prefix" "name"]
```
ashwinvis, line 164: Missed a comma before `"name"`.

adriendelsalle: good catch thx

PR #910, opened 2021-04-27
cURL lazy init

adriendelsalle:
Description
--

Lazy loading of cURL allow to search for system CA certificates on linux only at first download init.
It prevent from errors in offline mode but also in operation that don't require download.

Skip cURL SSL init when `ssl_verify` is set to false (`<false>`).

### DIFF
--- a/src/api/configuration.cpp
+++ b/src/api/configuration.cpp
@@ -46,42 +46,9 @@ namespace mamba
             {
                 if (value.empty() || (value == "true") || (value == "1") || (value == "<true>"))
                 {
-                    if (on_linux)
-                    {
-                        std::array<std::string, 6> cert_locations{
-                            "/etc/ssl/certs/ca-certificates.crt",  // Debian/Ubuntu/Gentoo etc.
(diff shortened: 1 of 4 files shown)


src/api/configuration.cpp
```diff
@@ -46,42 +46,9 @@ namespace mamba
...
-                        {
-                            LOG_ERROR << "ssl_verify is enabled but no ca certificates found";
-                            throw std::runtime_error("No CA certificates found.");
-                        }
-                    }
-                    else
-                    {
-                        value = "<system>";
-                    }
+                    value = "<system>";
```
wolfv, line 49: will this value contain `<system>` for rhumba now? in that case, that would be the wrong value (for non-statically compiled rhumba).

adriendelsalle: I just moved the find operation to init_curl_ssl, and also added back the pre-processor guard UMAMBA_STATIC that prevent setting SSL backends into that same function

wolfv:
unfortunately failing tests now :(

PR #914, opened 2021-04-28
Fix stack activate

wolfv:


wolfv:
This simplifies the activation logic a bit by changing how we use CLI11. 

We can now pass `micromamba shell activate --stack $prefix` without having to specifiy `-p`, which puts more of the parsing into the C++ side and makes the shell functions more straightforward.

PR #916, opened 2021-04-30
Add initial crypto sign and verify

adriendelsalle:
Description
--

Initial implem of cryptographic functions
Add message signature and verification using libsodium `ed25519` (resp. `sign` and `verify`)
Add libsodium dependency
Add tests

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -72,9 +72,11 @@ add_definitions(-DGHC_WIN_DISABLE_AUTO_PREFIXES)
 
 # Dependencies
 # ============
+set(CMAKE_MODULE_PATH "${CMAKE_MODULE_PATH}" "${CMAKE_CURRENT_SOURCE_DIR}/cmake/")
 
 find_package(Threads REQUIRED)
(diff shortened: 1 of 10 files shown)


wolfv:
awesome! I think we need to add libsodium to the requirements now.

CMakeLists.txt
```diff
@@ -72,9 +72,11 @@ add_definitions(-DGHC_WIN_DISABLE_AUTO_PREFIXES)
 
 # Dependencies
 # ============
+set(CMAKE_MODULE_PATH "${CMAKE_MODULE_PATH}" "${CMAKE_CURRENT_SOURCE_DIR}/cmake/")
 
 find_package(Threads REQUIRED)
 find_package(nlohmann_json REQUIRED)
+find_package(Sodium REQUIRED)
```
wolfv, line 79: maybe we should move this behind a config flag until the feature is ready to be rolled out? So that we can keep the builds more lightweight until we're ready. Doesn't have to happen in this PR, but maybe before the next release.

adriendelsalle: ok!

PR #927, opened 2021-05-10
Conda root prefix

adriendelsalle:
Description
---

Improve interoperability with `conda` by adding a free-function to set `MAMBA_ROOT_PREFIX` using `conda config --show root_prefix --json` output.
Expose it through the C-API.

### DIFF
--- a/src/api/configuration.cpp
+++ b/src/api/configuration.cpp
@@ -12,6 +12,10 @@
 #include "mamba/core/fetch.hpp"
 #include "mamba/core/fsutil.hpp"
 
+#include <reproc++/run.hpp>
+
+#include <nlohmann/json.hpp>
+
 #include <algorithm>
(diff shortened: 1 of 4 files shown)


src/api/configuration.cpp
```diff
@@ -384,6 +388,33 @@ namespace mamba
...
+            auto j = nlohmann::json::parse(out);
+            return j.at("root_prefix").get<std::string>();
+        }
+    }
+
+    void use_conda_root_prefix(bool force)
+    {
+        if (!Configuration::instance().at("root_prefix").configured() || force)
+        {
+            env::set("MAMBA_ROOT_PREFIX", get_conda_root_prefix());
```
wolfv, line 414: did you properly test this? This is not re-configuring the config? Is it properly propagated to the context?

PR #928, opened 2021-05-10
Progress bar

JohanMabille:


wolfv:
nice! do you want to run pre-commit?

`mamba install pre-commit -c conda-forge && pre-commit run --all` should fix the lint.

JohanMabille:
Done ;)

wolfv:
I just tried. Looks nice!

One thing I believe we should fix before merging is that the progress bar is still printed during the transaction execution:

Screenshot 2021-05-12 at 16 22 41

And then I think it would be nice if we could show an aggregated Mb/s next to the running stop watch on the right, and toggle through the currently downloading items. But we can keep those two things for another PR.

JohanMabille:
ok let's fix the issue. I think the other changes can be pushed to a new PR.

PR #933, opened 2021-05-11
Use openSSL for package signatures

adriendelsalle:
Description
---

Use `openSSL` cryptographic functions for package signing and verification
Remove `libsodium` dep
Add `bin2hex` and `hex2bin` utils from `libsodium`
Update tests


dhirschfeld:
I must be out of the loop - how do you sign packages? Is this new functionality?

wolfv:
Hi @dhirschfeld yeah, we're following [`conda-content-trust`](https://github.com/conda/conda-content-trust).

The main point from my POV is to sign `repodata.json` which will be useful in mirroring situations. I've also implemented a small server to do that in https://github.com/mamba-org/mamba/pull/931

Note that conda-content-trust is _based_ on the TUF spec, but it deviates quite substantially. But still, the TUF spec is a good place to get familiar with the ideas. https://theupdateframework.io/security/

wolfv:
btw. conda-content-trust on Anaconda side is being rolled out _only to some commercial customers_ -- so we won't see it in conda-forge for a while.

dhirschfeld:
Thanks for the info @wolfv - good to know...

PR #944, opened 2021-05-14
Artefact verification spec update

adriendelsalle:
Description
---

implement update capability of spec version on root role for artefact verification
implement spec version 0.6 root role (consistent with `conda-content-trust`)
add tests for spec v1 and v0.6

### DIFF
--- a/src/core/validate.cpp
+++ b/src/core/validate.cpp
@@ -17,6 +17,7 @@
 #include <iostream>
 #include <string>
 #include <set>
+#include <regex>
 
 
 namespace validate
@@ -49,6 +50,11 @@ namespace validate
 
     role_file_error::role_file_error() noexcept
         : trust_error("Invalid role file. Aborting.")
+    {
+    }
+
+    spec_version_error::spec_version_error() noexcept
+        : trust_error("Unsupported specification version. Aborting.")
     {
     }
 
@@ -304,6 +310,11 @@ namespace validate
         return verify_gpg_hashed_msg(data, bin_pk, bin_signature);
     }
 
+    bool operator<(const RoleSignature& rs1, const RoleSignature& rs2)
+    {
+        return rs1.keyid < rs2.keyid;
+    };
+
     /**
      * Binary to hexadecimal converter.
      * from https://github.com/jedisct1/libsodium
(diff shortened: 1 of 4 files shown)


adriendelsalle:
There are probably still missing parts to make the top-level roles complete update process work, but it should be almost done for the root roles now! @wolfv

src/core/validate.cpp
```diff
@@ -694,87 +1129,44 @@ namespace validate
...
-                                mandatory_roles.end(),
-                                current_roles.begin(),
-                                current_roles.end(),
-                                std::inserter(diff, diff.end()));
-            LOG_ERROR << "Missing roles while loading 'root' metdata: " << mamba::join(", ", diff);
-            throw role_metadata_error();
-        }
+    void RepoTrust::update_all_roles()
+    {
+        // p_root->update();
```
wolfv, line 1170: ?

wolfv:
can you reword the WIP commit?

Looks good to me. 

I would like to have a clear picture why we need the V1 support.
IMO we should really concentrate on v06 for now and then figure out how and if we are ever going to support V1 or if we need to come up with some other scheme altogether. For example, I think right now it's not clear if we are ever going to use / need the key-schema that V1 is using (since we could just as well continue to only use ed22519 keys as keyid like conda-content-trust does). So we're adding code here where we can't really be sure if we're going to use it in the future, and then I'll have to maintain it indefinitely :)

If we want to go with V1 we should write a MEP (Mamba Enhancement Proposal) and outline how V1 is better, what problems it solves and how we're going to use it.

wolfv:
Also (in general, since this is not part of this PR) for the bin2hex and hex2bin functions -- I think with C++ we can come up with a more concise solution.

E.g. we have this to go from bin to hex in mamba already:

https://github.com/mamba-org/mamba/blob/e21185c1d3adeb099362620271874e81ea594e40/include/mamba/core/util.hpp#L258-L268

And here is a short hex2bin: https://stackoverflow.com/a/18906270

But we can fix that another time

src/core/validate.cpp
wolfv: maybe this function should be removed if it's not implemented yet? :)

adriendelsalle: sure it will be done

adriendelsalle:
> can you reword the WIP commit?

sure 

> Looks good to me.

:+1: thanks

> I would like to have a clear picture why we need the V1 support.

v0.6 is not an official release of [The Update Framework specification](https://github.com/theupdateframework/specification) but more a custom [conda-content-trust](https://github.com/conda/conda-content-trust) initial implementation that may continue moving.

> If we want to go with V1 we should write a MEP (Mamba Enhancement Proposal) and outline how V1 is better, what problems it solves and how we're going to use it.

Security features have to be specified in details and audited. Any adaptation should be fully documented and reviewed by experts about consequences/risks. I would then recommend to write a MEP about adaptations/changes vs v1.

> IMO we should really concentrate on v06 for now and then figure out how and if we are ever going to support V1 or if we need to come up with some other scheme altogether. 

I decided here to keep the initial v1.0.17 implementation of the `root role` because with minimal effort it was possible to figure out a design that allows to update spec version (a change in the metadata schema) assuming basic interface such as `root role` will still exists and lists others top-level roles public keys. It is fully described in [TAP14](https://github.com/theupdateframework/taps/blob/master/tap14.md).

Going or not to this specific version is not the point of this PR, but we now have a fully tested and more likely to be future-proof design.
It also provides a ready-to-use code base to convert v0.6 to v1 format that makes us 100% compliant with official and audited specification of TUF.

> For example, I think right now it's not clear if we are ever going to use / need the key-schema that V1 is using (since we could just as well continue to only use ed22519 keys as keyid like conda-content-trust does). So we're adding code here where we can't really be sure if we're going to use it in the future, and then I'll have to maintain it indefinitely :)

Totally agree, so we should write about how we propose to simplify the metadata for implementation of the spec than plans to only rely on a single key scheme/type (ED25519).

On this example, GPG keys require to add extra information (GPG header) to the message (role metadata) to verify the signature. v0.6 put that information in the `signatures` map, using an additional `other_headers` key/val:
- key header is part of the key definition, not signature
- this header is then not part of the signed section itself, and I can't status about potential security risk

> and then I'll have to maintain it indefinitely :)

I'm also maintainer of this project, so I would say *we* !

PR #949, opened 2021-05-17
only set CURLOPT_CAINFO to nullptr if system and static

wolfv:


wolfv:
@adriendelsalle i think that fixes everything

PR #953, opened 2021-05-19
refactor documentation

adriendelsalle:
Description
---

add doc for configuration from user perspective: how to set it, etc.
add doc for general concepts
update info about micromamba
update installation section to not only refer to mamba
move user guide to appropriate section


wolfv:
looks really really nice

PR #963, opened 2021-05-25
Implement key_mgr role of content trust

adriendelsalle:
Description
---

Factorize implementation across roles
Implement `key_mgr` role of `conda-content-trust` (~v0.6.0 specification)
Implement expiration timestamp
Add tests

### DIFF
--- a/src/core/util.cpp
+++ b/src/core/util.cpp
@@ -756,4 +756,61 @@ namespace mamba
         fs::remove(m_path);
     }
 
+    std::string timestamp(const std::time_t& utc_time)
+    {
+        char buf[sizeof("2011-10-08T07:07:09Z")];
+        strftime(buf, sizeof(buf), "%FT%TZ", gmtime(&utc_time));
+        return buf;
+    }
+
+    std::time_t utc_time_now()
+    {
+        std::time_t now;
+        std::time(&now);
+        gmtime(&now);
+        return now;
+    }
+
+    std::string utc_timestamp_now()
+    {
+        return timestamp(utc_time_now());
+    }
+
+    std::time_t parse_utc_timestamp(const std::string& timestamp, int& error_code) noexcept
+    {
+        error_code = 0;
+        struct std::tm tt = { 0 };
+
+        if (sscanf(timestamp.data(),
+                   "%04d-%02d-%02dT%02d:%02d:%02dZ",
+                   &tt.tm_year,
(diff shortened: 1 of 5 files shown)


src/core/util.cpp
```diff
@@ -756,4 +756,43 @@ namespace mamba
...
+        char buf[sizeof("2011-10-08T07:07:09Z")];
+        strftime(buf, sizeof(buf), "%FT%TZ", gmtime(&utc_time));
+        return buf;
+    }
+
+    std::time_t utc_time_now()
+    {
+        std::time_t now;
+        std::time(&now);
+        gmtime(&now);
```
wolfv, line 770: is this doing the same thing twice?

adriendelsalle: it's not, `std::time` get the time since Epoch using the current calendar while `gmtime` converts it to UTC.

src/core/util.cpp
```diff
@@ -756,4 +756,43 @@ namespace mamba
...
+        if (sscanf(timestamp.data(),
+                   "%04d-%02d-%02dT%02d:%02d:%02dZ",
+                   &tt.tm_year,
+                   &tt.tm_mon,
+                   &tt.tm_mday,
+                   &tt.tm_hour,
+                   &tt.tm_min,
+                   &tt.tm_sec)
+            != 6)
+            return -1;
```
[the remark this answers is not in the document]
adriendelsalle: I will write some noexcept variants to be able to report metadata error in top-level roles files of content trust

src/core/validate.cpp
```diff
@@ -953,14 +1096,15 @@ namespace validate
...
-            auto allkeys = keys();
+            auto allkeys = all_keys();
             std::map<std::string, Key> v1_keys = allkeys.at("root").keys;
             v1_keys.insert(allkeys.at("key_mgr").keys.cbegin(), allkeys.at("key_mgr").keys.cend());
 
             v1_equivalent_root["keys"] = v1_keys;
             v1_equivalent_root["_type"] = "root";
             v1_equivalent_root["version"] = version();
             v1_equivalent_root["spec_version"] = "1.0.17";
+            v1_equivalent_root["expires"] = "some";
```
wolfv, line 1107: some?

adriendelsalle: good catch, it was a test. I will fix it

PR #970, opened 2021-05-28
Implement RepoChecker interface for a content verification

adriendelsalle:
Description
---

implement `RepoChecker` workflow
improve compatible or upgrade checks of spec
improve tests using `gmock`, add tests on `RepoChecker` for v0.6.0 spec version using `PkgMgrRole`
update `CMakeLists` to use `gmock` on C++ tests
update CI steps to install `gmock`

### DIFF
--- a/src/core/util.cpp
+++ b/src/core/util.cpp
@@ -211,6 +211,26 @@ namespace mamba
         return str.size() >= prefix.size() && 0 == str.compare(0, prefix.size(), prefix);
     }
 
+    bool any_starts_with(const std::vector<std::string_view>& str, const std::string_view& prefix)
+    {
+        for (auto& s : str)
+        {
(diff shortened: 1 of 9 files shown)


src/core/util.cpp
```diff
@@ -211,6 +211,26 @@ namespace mamba
...
+    {
+        for (auto& s : str)
+        {
+            if (starts_with(s, prefix))
+                return true;
+        }
+        return false;
+    }
+
+    bool starts_with_any(const std::string_view& str, const std::vector<std::string_view>& prefix)
```
wolfv, line 224: how about using std::any_of (https://en.cppreference.com/w/cpp/algorithm/all_any_none_of)

adriendelsalle: Thanks for the comment I'll try to refactor that later

PR #972, opened 2021-05-31
Use cURL to fetch content trust roles metadata

adriendelsalle:
Description
---

Fetch content trust roles metadata using `cURL`
Check for freeze attack on `root` or `pkg_mgr` metadata
Add capability to silent logs on `DownloadTarget` (also move a log from constructor to `perform`)
Add `fetch_error` and `freeze_error`
add C++ tests

### DIFF
--- a/include/mamba/core/fetch.hpp
+++ b/include/mamba/core/fetch.hpp
@@ -76,11 +76,6 @@ namespace mamba
             return m_active_logs;
         }
 
-        inline curl_off_t get_downloaded_size() const
-        {
-            return downloaded_size;
-        }
-
         void set_result(CURLcode r);
         bool finalize();
 
--- a/include/mamba/core/fetch.hpp
+++ b/include/mamba/core/fetch.hpp
@@ -45,6 +45,8 @@ namespace mamba
         const std::string& name() const;
 
         void init_curl_target(const std::string& url);
+
+        bool resource_exists();
         bool perform();
         CURL* handle();
 
@@ -64,16 +66,6 @@ namespace mamba
         inline bool ignore_failure() const
         {
             return m_ignore_failure;
-        }
-
-        inline void set_active_logs(bool yes)
-        {
-            m_active_logs = yes;
-        }
-
-        inline bool active_logs() const
-        {
-            return m_active_logs;
         }
 
         void set_result(CURLcode r);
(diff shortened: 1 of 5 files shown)


include/mamba/core/fetch.hpp
```diff
@@ -66,6 +66,21 @@ namespace mamba
             return m_ignore_failure;
         }
 
+        inline void set_active_logs(bool yes)
```
wolfv, line 69: do we need this? Context::verbosity is the same, no?

include/mamba/core/fetch.hpp
```diff
@@ -66,6 +66,21 @@ namespace mamba
...
+        }
+
+        inline bool active_logs() const
+        {
+            return m_active_logs;
+        }
+
+        inline curl_off_t get_downloaded_size() const
+        {
+            return downloaded_size;
```
wolfv, line 81: do we need this? downloaded size is public.

include/mamba/core/fetch.hpp
adriendelsalle: it's needed to silent specifically all the logs emitted in a `DownloadTarget`, since the target can be downloaded later and the possible callback can also emit some logs.

include/mamba/core/fetch.hpp
adriendelsalle: right, I could use directly `downloaded_size`

adriendelsalle:
I removed the unnecessary `get_downloaded_size` method! See my comment for the other one.

include/mamba/core/fetch.hpp
wolfv: I still don't understand ... the logs that I see all use LOG_INFO. what's the issue with that? This looks like introducing another log level for one particular use-case? Is that really required? 

I am happy if you change the logs in the callbacks to be less verbose if that's the goal.

adriendelsalle: Some logs are emitted in `perform` (previously in constructor, moved in this PR) and `set_result`.
There is no point of logging failed downloads when in most cases they are allowed to fail, it's totally misleading.

We could remove this later when having capability to check existence of the target on the server using header request instead.

adriendelsalle:
@wolfv I reused head request implementation of @afranchuk in https://github.com/mamba-org/mamba/pull/934 to get rid of the special logging behavior in `DownloadTarget`.

@afranchuk I mentioned you as co-author of this commit.

PR #975, opened 2021-06-01
add menu install shortcut and more long paths support for Windows

wolfv:


src/core/menuinst.cpp
```diff
@@ -0,0 +1,455 @@
...
+                             const std::string& arguments,
+                             const fs::path& work_dir,
+                             const fs::path& icon_path,
+                             int icon_index)
+        {
+            IShellLink* pShellLink = nullptr;
+            IPersistFile* pPersistFile = nullptr;
+
+            HRESULT hres;
+            LOG_WARNING << "Creating shortcut with "
```
adriendelsalle, line 68: why a warning and not info log level ?
a warning should be reserved to something that went wrong but not preventing to continue the operation (using a fallback for example) or to something that represent a potential risk of failure (using experimental mode, etc.)

src/core/shell_init.cpp
```diff
@@ -198,7 +199,7 @@ namespace mamba
         }
         catch (const std::exception& e)
         {
-            std::cerr << e.what() << '\n';
+            LOG_INFO << "No AutoRun key detected.";
```
adriendelsalle, line 202: LOG_INFO -> LOG_DEBUG

src/core/shell_init.cpp
```diff
@@ -243,6 +244,47 @@ namespace mamba
             }
         }
     }
+
+    void enable_long_paths_support()
```
adriendelsalle, line 248: really nice to have it, but same comment as for the header file this should be split in another PR

adriendelsalle (commented):
Hi @wolfv thanks for this one!

Few comments here, would it be possible to:
- split long paths support in another PR?
- change some messages log levels?
- maybe move operations workflows doc to an advanced usage section (those can be pretty useful for both advanced users and developers)?

Thanks!

wolfv:
checklist:

- [ ] figure out if admin / regular user install path for link
- [ ] write some tests

src/core/menuinst.cpp
wolfv: just for debug

PR #978, opened 2021-06-02
Add experimental artifact verification

adriendelsalle:
Description
---

add `RepoChecker` to Channel
add lazy loading of the `RepoChecker`
add `verify_artifacts` to `Context` and `Configuration`
- add artifact `--verify-artifacts` flag to micromamba CLI, requires `--experimental` to be also set to be effective
- `verify_artifacts` is rc and env var configurable using respectively `verify_artifacts` key and `MAMBA_VERIFY_ARTIFACTS`

get index_file from a `MRepo`
refactor `Channel` to avoid circular dependency in base_url
add cache_name_from_url free function
add index_error to disambiguate with package_error
add `RepoIndexChecker` and `RepoChecker` `verify_package` package to allow verification of a specific package instead of using `verify_index` all the time
implement `verify_package` on v0.6.0 spec `PkgMgr`
define reference and cache path for top-level content trust roles
use cached root metadata if any
add docstrings

### DIFF
--- a/include/mamba/core/context.hpp
+++ b/include/mamba/core/context.hpp
@@ -134,7 +134,7 @@ namespace mamba
 
         VerificationLevel safety_checks = VerificationLevel::kWarn;
         bool extra_safety_checks = false;
-        bool artifact_verif = false;
+        bool verify_artifacts = false;
 
         // debug helpers
         bool keep_temp_files = std::getenv("MAMBA_KEEP_TEMP") ? 1 : 0;
--- a/include/mamba/core/context.hpp
+++ b/include/mamba/core/context.hpp
@@ -134,6 +134,7 @@ namespace mamba
 
         VerificationLevel safety_checks = VerificationLevel::kWarn;
         bool extra_safety_checks = false;
+        bool artifact_verif = false;
 
         // debug helpers
         bool keep_temp_files = std::getenv("MAMBA_KEEP_TEMP") ? 1 : 0;
--- a/src/core/subdirdata.cpp
+++ b/src/core/subdirdata.cpp
@@ -4,12 +4,12 @@
 //
 // The full license is in the file LICENSE, distributed with this software.
 
-#include "openssl/md5.h"
-
 #include "mamba/core/mamba_fs.hpp"
 #include "mamba/core/output.hpp"
 #include "mamba/core/package_cache.hpp"
 #include "mamba/core/subdirdata.hpp"
+#include "mamba/core/url.hpp"
+
 
 namespace decompress
 {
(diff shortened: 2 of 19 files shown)


src/core/subdirdata.cpp
```diff
@@ -428,7 +423,7 @@ namespace mamba
 
     MRepo MSubdirData::create_repo(MPool& pool)
     {
-        RepoMetadata meta{ m_url,
+        RepoMetadata meta{ m_index_url,
```
wolfv, line 426: maybe it would be better to use `repodata_json_url` or `repodata_url`?

adriendelsalle: if we prefer to use a conda semantic repodata = package index, yes.
I can refactor that

wolfv: otherwise it's just more confusing in the end since people might wonder what the index is :)

wolfv:
It's British English vs American English, but I think `artifact` is more commonly used in conda / conda-forge than artefact.

E.g. https://github.com/conda-forge/artifact-validation

Would you mind switching to that?

adriendelsalle:
> Would you mind switching to that?

It's done

include/mamba/core/context.hpp
```diff
@@ -134,6 +134,7 @@ namespace mamba
 
         VerificationLevel safety_checks = VerificationLevel::kWarn;
         bool extra_safety_checks = false;
+        bool artifact_verif = false;
```
wolfv, line 137: how about `verify_artifacts`?
It's almost the same length and proper words. Would also be good for yaml option or flag

`verify_artifacts: true`

`--verify-artifacts`

src/micromamba/common_options.cpp
```diff
@@ -327,6 +327,6 @@ init_install_options(CLI::App* subcom)
                     { "enabled", "warn", "disabled" },
                     safety_checks.description());
 
-    auto& av = config.at("artifact_verif").get_wrapped<bool>();
-    subcom->add_flag("--artifact-verif", av.set_cli_config(0), av.description());
+    auto& av = config.at("verify_artifacts").get_wrapped<bool>();
+    subcom->add_flag("--verif-artifacts", av.set_cli_config(0), av.description());
```
wolfv, line 331: --verify-artifacts (missing `y`)

include/mamba/core/context.hpp
adriendelsalle: done, this is RC configurable using `verify_artifacts` and also environment variable configurable using `MAMBA_VERIFY_ARTIFACTS`

src/micromamba/common_options.cpp
adriendelsalle: arghhh, thanks! :)

Issue #991, opened 2021-06-08
Micromamba create fails with long prefix (Windows 10)


avaissi:
Micromamba create command fails with long prefix path on Windows 10 Pro.
LongPathsEnabled registry is set to 1, ie. long path support should be set true for my system.

I've attached the CMD output as a log file.

Note: Create command fails right after package download if prefix is set long enough.
`ERROR   The filename or extension is too long.
        : C:\Users\avaissi\Documents\tmp_rcc\my_very_long_path_this_is\my_very_long_path_this_is\my_very_long_path_this_is\my_very_long_path_this_is\my_very_long_path_this_is\my_very_long_path_this_is\my_very_long_path_this_is\my_very_long_path_this_is\envs\lab-env\conda-meta`

Windows specs:
Edition	Windows 10 Pro
Version	20H2
Installed on	‎11/‎06/‎2020
OS build	19042.985
Experience	Windows Feature Experience Pack 120.2212.2020.0

System specs:
Processor	Intel(R) Core(TM) i9-8950HK CPU @ 2.90GHz   2.90 GHz
Installed RAM	32,0 GB (31,7 GB usable)
System type	64-bit operating system, x64-based processor

[mamba_log_longpath_issue.log](https://github.com/mamba-org/mamba/files/6616025/mamba_log_longpath_issue.log)

adriendelsalle:
Hi @avaissi , thank you for opening this issue!

It's probably related to https://github.com/mamba-org/mamba/issues/857.
I'll try to reproduce.

adriendelsalle:
I'm able to reproduce.

At least, we need to:
- [ ] remove the `GHC_WIN_DISABLE_AUTO_PREFIXES` definition added here https://github.com/mamba-org/mamba/blob/c1516d4cb67df8bbfcba8514e1f8309d2a980108/CMakeLists.txt#L72
- [ ] fix #857 issue when running `cmd.exe` subprocesses with long paths
- [ ] fix upstream `GHC::filesystem` implementation of `create_directories`
  - see issue https://github.com/gulrak/filesystem/issues/125
  - see PR https://github.com/gulrak/filesystem/pull/126

wolfv:
@adriendelsalle can we close this one?

PR #994, opened 2021-06-09
Provide long paths support on Windows

### LINKED ISSUE #857: Long paths support on Windows
Description
---

We currently disabled long path support on Windows in `cpp-filesystem` using `GHC_WIN_DISABLE_AUTO_PREFIXES` flag because `cmd.exe` doesn't support call to executable specified using the `\\?\` prefix.
This was causing an error when executing the compile script for noarch packages at install time.

See https://github.com/gulrak/filesystem/releases/tag/v1.5.0

adriendelsalle:
Description
---

Provide long paths support on Windows

- [x] remove GHC_WIN_DISABLE_AUTO_PREFIXES definition
- [x] fix #857 issue when running cmd.exe subprocesses with long paths
  - add documentation on pitfalls
- [x] fix upstream GHC::filesystem implementation of create_directories
  - see issue gulrak/filesystem#125
  - see PR gulrak/filesystem#126

Related issue https://github.com/mamba-org/mamba/issues/991

### DIFF
--- a/src/core/shell_init.cpp
+++ b/src/core/shell_init.cpp
@@ -267,6 +267,7 @@ namespace mamba
                 size = GetModuleFileNameW(NULL, (wchar_t*) buffer.c_str(), (DWORD) buffer.size());
             } while (new_size == size);
         }
+        buffer.resize(buffer.find(L'\0'));
         return fs::absolute(buffer);
 #elif defined(__APPLE__)
         uint32_t size = PATH_MAX;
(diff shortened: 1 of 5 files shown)


src/core/shell_init.cpp
```diff
@@ -454,6 +455,9 @@ namespace mamba
     std::string get_hook_contents(const std::string& shell)
     {
         fs::path exe = get_self_exe_path();
+
+        std::cerr << exe.string() << std::endl;
```
wolfv, line 459: this should probably be removed, right?

wolfv:
great stuff!

src/core/shell_init.cpp
adriendelsalle: totally! sorry for this one!

adriendelsalle:
@wolfv could you please review that ? We should make a patch release for that pretty soon.

PR #1007, opened 2021-06-16
Fix constructor invalid `repodata_record`

### LINKED ISSUE #993: Found corrupted repodata_record file warning
I'm trying to use `boa` / `mamba` to debug requirement conflicts over in https://github.com/conda-forge/qt-feedstock/pull/194 
and seeing some strange errors, like 

```
WARNING Found corrupted repodata_record file "/Users/runner/miniforge3/pkgs/libcxx-11.1.0-habf9029_0/info/repodata_record.json"
```
~~and
package clangxx_osx-64-11.0.0-hc3ec130_0 requires clang_osx-64 11.0.0 h05bbb7f_0, but none of the providers can be installed~~

Unfortunately I don't have a Mac and so can't reproduce or investigate locally, but let me know if there is any other useful information I can provide.

adriendelsalle:
Description
---

Fix constructor invalid `repodata_record`: missing `size` and `sha256` for validation
Improve logging

Closes #993

### DIFF
--- a/src/core/package_handling.cpp
+++ b/src/core/package_handling.cpp
@@ -446,7 +446,8 @@ namespace mamba
                 {
                     if (is_warn || is_fail)
                     {
-                        LOG_WARNING << "Missing file from package cache: " << full_path;
+                        LOG_WARNING << "Invalid package cache, file '" << full_path.string()
+                                    << "' is missing";
                         return false;
                     }
                 }
@@ -458,8 +459,8 @@ namespace mamba
                     if (p.path_type != PathType::SOFTLINK
                         && !validate::file_size(full_path, p.size_in_bytes))
                     {
-                        LOG_WARNING << "Size incorrect, file modified in package cache "
-                                    << full_path;
+                        LOG_WARNING << "Invalid package cache, file '" << full_path.string()
+                                    << "' has incorrect size";
                         is_invalid = true;
                         if (is_fail)
                         {
@@ -469,8 +470,8 @@ namespace mamba
                     if (full_validation && !is_invalid && p.path_type != PathType::SOFTLINK
                         && !validate::sha256(full_path, p.sha256))
                     {
-                        LOG_WARNING << "SHA256 checksum incorrect, file modified in package cache "
-                                    << full_path;
+                        LOG_WARNING << "Invalid package cache, file '" << full_path.string()
+                                    << "' has incorrect SHA-256 checksum";
                         if (is_fail)
                         {
                             return false;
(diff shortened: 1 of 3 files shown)


src/micromamba/constructor.cpp
```diff
@@ -82,7 +82,9 @@ construct(const fs::path& prefix, bool extract_conda_pkgs, bool extract_tarball)
         {
             if (is_package_file(entry.path().filename().string()))
             {
+                LOG_TRACE << "Extracting " << entry.path().filename() << std::endl;
```
wolfv, line 85: we shoudl keep cout here. it's good to see the progress.

src/core/package_handling.cpp
```diff
@@ -481,10 +480,10 @@ namespace mamba
         }
         catch (...)
         {
-            LOG_WARNING << "Could not read paths.json from " << pkg_folder << std::endl;
```
wolfv, line 484: not sure that trace is the correct replacement for WARNING

src/core/package_handling.cpp
```diff
@@ -446,7 +446,7 @@ namespace mamba
                 {
                     if (is_warn || is_fail)
                     {
-                        LOG_WARNING << "Missing file from package cache: " << full_path;
+                        LOG_TRACE << "Missing file from package cache: " << full_path;
```
wolfv, line 449: Here I'd keep the warning as it could be a hacker :)

src/micromamba/constructor.cpp
adriendelsalle: I didn't removed it, just added the same info in the logs

src/core/package_handling.cpp
adriendelsalle: this method is called here https://github.com/mamba-org/mamba/blob/42475c057869356f6652e9ef9ba7f94f520567fc/src/core/package_cache.cpp#L226 and the warning will be emitted.
This is not necessary to emit 2 warnings for the same invalid cache, so I put the details in `trace`

src/micromamba/constructor.cpp
wolfv: But the miniforge / mambaforge users won't see anything or am I mistaken? Since the TRACE log is not going to be displayed.

src/core/package_handling.cpp
adriendelsalle: same as the other log, 2 warnings are not necessary for the same invalid cache, it's just flooding the logs

src/core/package_handling.cpp
wolfv: this stuff almost _never_ happens and when it happens it would help us to locate a bug more easily since the users would see the (more precise) warnings right away. So it could help us with future bugs

src/micromamba/constructor.cpp
adriendelsalle: have a look at line 86, it's still printed in `cout`

src/core/package_handling.cpp
wolfv: Again this should almost never happen, and I quite like to get this information when I (myself) changed the files in the package cache, just so that I can figure out bugs that I made myself quicker. In the second warning we're not printing which file has been changed.

src/micromamba/constructor.cpp
wolfv: then i don't know why we need the trace log :)

wolfv:
AwEsOmE! :)

Issue #1002, opened 2021-06-12
Improve error message when Python version conflicts with package dependencies?


krinsman:
MWE: (assumes base conda is running Python 3.9 or Python 3.8, so created environment has that too)

```
conda create --name test
conda activate test
conda config --env --add channels bioconda
conda config --env --add channels conda-forge
conda install --yes mamba
mamba install pasta
```

The resulting error message (since the package as of this writing has evidently not been updated to use higher Python versions)

```
Encountered problems while solving:
  - nothing provides pymongo 3.3.0 py27_0 needed by pasta-1.7.8-0
```

This makes it sound like the problem is with conda-forge not providing the package pymongo. But when running

```
conda install pasta
```

one gets the real source of the error

```
UnsatisfiableError: The following specifications were found
to be incompatible with the existing python installation in your environment:

Specifications:

  - pasta -> python[version='>=2.7,<2.8.0a0|>=3.7,<3.8.0a0|>=3.6,<3.7.0a0|>=3.5,<3.6.0a0|>=2.7,<3']

Your python: python=3.9

If python is on the left-most side of the chain, that's the version you've asked for.
When python appears to the right, that indicates that the thing on the left is somehow
not available for the python version you are constrained to. Note that conda will not
change your python version to a different minor version unless you explicitly specify
that.
```

Personally I think the error message from conda is more informative here than that from mamba. But if you disagree then this is a non-issue. I just thought it might help to be brought to your attention if you weren't already aware.

adriendelsalle:
Hey @krinsman thank you for opening this issue!

I agree this is not informative enough.
The logs contain a `Pinning Python to '...'` message that could be helpful to also display at console.
I could be also pretty useful to be able to disable Python automatic pinning using configuration (`--no-py-pin`/`--py-pin`)

adriendelsalle:
I just edited my answer because the solution proposed will maybe not work on `mamba` but only on `micromamba`.

In the PR opened, I propose at least to add python pin to the list of pinned packages in `mamba` (and do the exact same in `micromamba`):
```

Looking for: ['pasta']

bioconda/linux-64        Using cache
bioconda/noarch          Using cache
conda-forge/linux-64     Using cache
conda-forge/noarch       Using cache

Pinned packages:
  - python 3.9.2

Encountered problems while solving:
  - nothing provides pymongo 3.3.0 py27_0 needed by pasta-1.7.8-0
```
We still need to improve `libsolv` problem itself

[...]

PR #1010, opened 2021-06-17
Improve python auto pinning

adriendelsalle:
Description
---

Add console message with pinned packages, including python
- modify `mamba` to include python pin
- add it to `micromamba` to have the exact same display

Add configurable `no_py_pin` to disable python pinning
Add micromamba CLI flags for pinning or not python, `--no-py-pin` and `--py-pin`
Add test

Display of pinned packages is:
```
Pinned packages:
  - python 3.9.*
```

Related to https://github.com/mamba-org/mamba/issues/1002

### DIFF
--- a/include/mamba/core/solver.hpp
+++ b/include/mamba/core/solver.hpp
@@ -56,6 +56,7 @@ namespace mamba
 
         const std::vector<MatchSpec>& install_specs() const;
         const std::vector<MatchSpec>& remove_specs() const;
+        const std::vector<MatchSpec>& pinned_specs() const;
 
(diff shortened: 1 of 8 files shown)


wolfv:
I think ideally we'd print all pinned packages as part of the transaction

adriendelsalle:
> I think ideally we'd print all pinned packages as part of the transaction

For now the transaction is created from a successful `Solver::solve`, that would mean that the `Solver` could be part of the transaction itself (which makes sense to me) but it's deeper work and I would prefer to open a dedicated issue/PR for that

PR #1018, opened 2021-06-17
add pretty print to repoquery search

madhur-tandon:


marimeireles:
Hey @madhur-tandon this is how we lint everything `pre-commit run --all-files --verbose` :)

madhur-tandon:
@marimeireles The issue is that the linter also modifies files that I didn't touch. So it should ideally change only 4 files of the PR if at all. But it affects many other ones too. Not sure if I would want to include so many unrelated changed files (just because of lint) in this PR.

PR #1023, opened 2021-06-18
Refactor `micromamba` config prepend and append sequence

### LINKED ISSUE #1019: Config prepend/append should raise on scalar types
Description
---

The current implementation of `micromamba config prepend` or `micromamba config append` should raise when passing a key of a scalar type.

`./micromamba config prepend json true` gives the following YAML file:
```
json:
  - true
```
which is incorrect and will raise errors during later use

cc @marimeireles

### LINKED ISSUE #1020: Config prepend/append values handling
Description
---

This is both a bug and feature request about  `micromamba config append` or  `micromamba config prepend` values handling:
- bug
  - passing no arguments makes `micromamba config append` crash (segfault)
  - passing a single argument to `micromamba config append` is considered as both a key and value, which is incorrect
- feature request
  - being able to handle multiple values

Such as:
```
micromamba config append
```
returns help or an explicit error.
```
micromamba config append channels
```
return an error stating at least one value should be specified.
```
micromamba config append channels val1,val2
```
just appends `val1` and `val2` in that order to `channels` (same comma separated syntax used for environment variables)
```
micromamba config append channels val1,val2 default_channels foo,bar
```
to set multiple key/values pairs

### LINKED ISSUE #1021: Config prepend/append keys should be rc configurable
Description
---

`prepend` or `append` works with any `Configurable` key to an rc file, even if this `Configurable` is not rc configurable.
We should also filter this property.

adriendelsalle:
Description
---

refactor `micromamba` config prepend and append sequence
add capability to define multiple key/val pairs
catch empty args or no values passed cases
check if keys match rc configurables configs
define config append or prepend spec as required
update and add tests

cc @marimeireles 

Closes #1019
Closes #1020 
Closes #1021

### DIFF
--- a/include/mamba/api/configuration.hpp
+++ b/include/mamba/api/configuration.hpp
@@ -592,6 +592,8 @@ namespace mamba
         self_type& compute(const int options = 0,
                            const ConfigurationLevel& level = ConfigurationLevel::kFile);
 
+        bool is_valid_serialization(const std::string& value) const;
+
         void reset_compute_counter();
 
         void lock();
@@ -814,6 +816,20 @@ namespace mamba
     }
 
     template <class T>
+    bool Configurable<T>::is_valid_serialization(const std::string& value) const
+    {
+        try
+        {
+            detail::Source<T>::deserialize(value);
+            return true;
+        }
+        catch (...)
+        {
+            return false;
+        }
+    }
+
+    template <class T>
     void Configurable<T>::lock()
     {
         m_lock = true;
@@ -1010,6 +1026,8 @@ namespace mamba
 
             virtual bool has_single_op_lifetime() const = 0;
 
+            virtual bool is_valid_serialization(const std::string& value) const = 0;
+
             virtual bool locked() = 0;
 
             virtual void reset_compute_counter() = 0;
@@ -1157,6 +1175,11 @@ namespace mamba
             bool locked()
             {
                 return p_wrapped->locked();
+            }
+
+            bool is_valid_serialization(const std::string& value) const
+            {
+                return p_wrapped->is_valid_serialization(value);
             }
 
             void reset_compute_counter()
@@ -1441,6 +1464,11 @@ namespace mamba
         bool locked()
         {
             return p_impl->locked();
+        }
+
+        bool is_valid_serialization(const std::string& value) const
+        {
+            return p_impl->is_valid_serialization(value);
         }
 
         self_type& set_rc_yaml_value(const YAML::Node& value, const std::string& source)
--- a/include/mamba/api/configuration.hpp
+++ b/include/mamba/api/configuration.hpp
@@ -385,6 +385,8 @@ namespace mamba
                               std::vector<std::string>& source);
 
             static T deserialize(const std::string& value);
+
+            static bool is_sequence();
         };
 
         template <class T>
@@ -401,6 +403,8 @@ namespace mamba
                               std::vector<std::string>& source);
 
             static std::vector<T> deserialize(const std::string& value);
+
+            static bool is_sequence();
         };
 
         template <class T>
@@ -439,6 +443,12 @@ namespace mamba
         }
 
         template <class T>
+        bool Source<T>::is_sequence()
+        {
+            return false;
+        }
+
+        template <class T>
         void Source<std::vector<T>>::merge(const std::map<std::string, std::vector<T>>& values,
                                            const std::vector<std::string>& sources,
                                            std::vector<T>& value,
@@ -467,6 +477,12 @@ namespace mamba
         std::vector<T> Source<std::vector<T>>::deserialize(const std::string& value)
         {
             return YAML::Load("[" + value + "]").as<std::vector<T>>();
+        }
+
+        template <class T>
+        bool Source<std::vector<T>>::is_sequence()
+        {
+            return true;
         }
     }
 
@@ -593,6 +609,8 @@ namespace mamba
                            const ConfigurationLevel& level = ConfigurationLevel::kFile);
 
         bool is_valid_serialization(const std::string& value) const;
+
+        bool is_sequence() const;
 
         void reset_compute_counter();
 
@@ -830,6 +848,12 @@ namespace mamba
     }
 
     template <class T>
+    bool Configurable<T>::is_sequence() const
+    {
+        return detail::Source<T>::is_sequence();
+    }
+
+    template <class T>
     void Configurable<T>::lock()
     {
         m_lock = true;
@@ -1028,6 +1052,8 @@ namespace mamba
 
             virtual bool is_valid_serialization(const std::string& value) const = 0;
 
+            virtual bool is_sequence() const = 0;
+
             virtual bool locked() = 0;
 
             virtual void reset_compute_counter() = 0;
@@ -1180,6 +1206,11 @@ namespace mamba
             bool is_valid_serialization(const std::string& value) const
             {
                 return p_wrapped->is_valid_serialization(value);
+            }
+
+            bool is_sequence() const
+            {
+                return p_wrapped->is_sequence();
             }
 
             void reset_compute_counter()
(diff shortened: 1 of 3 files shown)


marimeireles:
Seems good :)
But I think I'll check it on Monday!

src/micromamba/config.cpp
```diff
@@ -194,194 +178,150 @@ set_config_path_command(CLI::App* subcom)
...
+    config.operation_teardown();
+}
 
-        // if the rc file is being modified, it's necessary to rewrite it
-        rc_file.open(rc_source.string(), std::ofstream::in | std::ofstream::trunc);
-        rc_file << rc_YAML << std::endl;
+void
+set_config_prepend_command(CLI::App* subcom)
+{
+    set_config_sequence_command(subcom);
```
marimeireles, line 315: Maybe here we add the specific description.

test/micromamba/test_config.py
```diff
@@ -500,25 +500,55 @@ def test_file_append_single_input(self):
         ).splitlines() == ["channels:", "  - flowers"]
 
     def test_file_append_multiple_inputs(self):
+        with open(TestConfigModifiers.home_rc_path, "w") as f:
+            f.write("channels:\n  - foo")
+
+        config(
+            "append",
+            "channels",
+            "condesc,mambesc",
```
marimeireles, line 509: Is there a difference between adding args as an independent string and as you did?
Is this just to remove the `splitlines()` command?

src/micromamba/config.cpp
```diff
@@ -194,194 +178,150 @@ set_config_path_command(CLI::App* subcom)
...
+void
+set_sequence_to_yaml(YAML::Node& node,
+                     const std::string& key,
+                     const std::string& value,
+                     const SequenceAddType& opt)
+{
+    if (!is_valid_rc_sequence(key, value))
+    {
+        if (!is_valid_rc_key(key))
+            LOG_ERROR << "Invalid key '" << key << "' or not rc configurable";
```
marimeireles, line 230: I thought we were throwing errors on `std::cout`, when to use LOG_ERROR and std?

test/micromamba/test_config.py
```diff
@@ -529,25 +559,55 @@ def test_file_prepend_single_input(self):
...
+
+    def test_file_prepend_invalid_input(self):
+        with pytest.raises(subprocess.CalledProcessError):
+            config("prepend", "--file", TestConfigModifiers.home_rc_path)
+
+        with pytest.raises(subprocess.CalledProcessError):
+            config("prepend", "@#A321", "--file", TestConfigModifiers.home_rc_path)
 
     def test_file_append_and_prepend_inputs(self):
         config(
```
marimeireles, line 613: I think we should add tests for the following cases:

1. Non vector key prepending/appending:
`./micromamba config prepend json true` and `./micromamba config append json true`

2. Vector key appending + non-vector key appending:
`./micromamba config prepend channels wawah json true`

Awesome work! :)
Thank you for this @adriendelsalle. The code is shining!

marimeireles (approved):

src/micromamba/config.cpp
```diff
@@ -194,194 +178,150 @@ set_config_path_command(CLI::App* subcom)
                                           .group("cli")
                                           .description("Set configuration on system's rc file"),
                                       true);
-    subcom->add_flag("--system", system_path.set_cli_config(0), system_path.description());
+    auto* system_flag
```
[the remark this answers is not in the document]
adriendelsalle: flags are commutative so it's already done by setting `--env` excluding `--system`

include/mamba/api/configuration.hpp
```diff
@@ -1443,6 +1466,11 @@ namespace mamba
             return p_impl->locked();
         }
 
+        bool is_valid_serialization(const std::string& value) const
```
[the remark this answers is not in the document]
adriendelsalle: I'm not 100% sure of your question, is it related to type erasure ?
If so:
- we don't know what's the type of `Configurables` and need an interface to manipulate them
  -  we could also use dynamic polymorphism to pick the right implementation at runtime with a huge `if... else...` conditional approach
  - it's much more expressive to declare/add/register/insert a new `Configurable` in the `Configuration` singleton and use init value with type deduction
  - then just manipulate `Configurable` without its type using the type erased interface
- I guess some of forwarded calls could be improved at some point

src/micromamba/config.cpp
```diff
@@ -194,194 +178,150 @@ set_config_path_command(CLI::App* subcom)
...
-
-        fs::path rc_source = env::expand_user(env::home_directory() / ".condarc");
-        std::ofstream rc_file;
-
-        std::string prepend_key = prepend_map.value().front();
-        std::string prepend_value = prepend_map.value().back();
+    auto& specs = config.insert(Configurable("config_set_sequence_spec",
+                                             std::vector<std::pair<std::string, std::string>>({}))
+                                    .group("Output, Prompt and Flow Control")
+                                    .description("Add value to the beginning of a configuration"),
```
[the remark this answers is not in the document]
adriendelsalle: it's overridden in `append` but yes we could definitely keep it generic or empty then also override `prepend`

src/micromamba/config.cpp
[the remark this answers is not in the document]
marimeireles: Hum, I see! :) cool

src/micromamba/config.cpp
adriendelsalle: logs are currently displayed through `std::cerr`.
we use `Console::print` for information that needs to be printed to user on `std::cout` to be able to silent it when using `--json`, `--quiet` or other similar flags

src/micromamba/config.cpp
adriendelsalle: yep :+1:, see my comment above

test/micromamba/test_config.py
adriendelsalle: sorry I don't get it, could you please point me what modification you're focusing on ?

test/micromamba/test_config.py
adriendelsalle: totally agree with the missing tests, will do that ;)
thanks for those kind words and for taking time to review @marimeireles :+1:

src/micromamba/config.cpp
marimeireles: Yeah, I saw it was overwritten, was just thinking about code clarity :)

src/micromamba/config.cpp
marimeireles: Ah, okay, so `std::cout` is for info only and errors should be handled by `std::cerr`?

include/mamba/api/configuration.hpp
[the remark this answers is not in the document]
marimeireles: So the reason why we have to jump through so many hoops here is because of the type erasure?

I don't know if my question is related to type erasure. It's a broad question of why did you choose this way of structuring the code. I don't understand why we have a `Wrapper` class a `BaseWrapper` a diff. impl of the `is_valid_serialization` method for each one of them and an impl. for the `Configurable` class.
I was wondering if you're following some design model or if you just came up with this due the circumstances you had.

test/micromamba/test_config.py
marimeireles: Just wondering because before it was

```
config("append", "channels", "a", "b")
```
but now you changed to:

```
config("append", "channels", "a,b")
```
and I'm asking why.

adriendelsalle: The previous behavior was to pick the last value only, and drop the intermediate one (`a` in that case).
I changed the behavior to:
- be able to handle multiple key/val pairs (as vector of string pairs)
- use the exact same `deserialization` used for environment variable. So there is no surprise you can use the same syntax when you set `CONDA_CHANNELS` to `a,b`

src/micromamba/config.cpp
adriendelsalle: you should never use `std::cout` directly but prefer `Console::print` to be able to silent it.
every logs messages you go to the appropriate level using macros (`LOG_INGO`, LOG_DEBUG`, etc.). For the moment yes those log messages go to the stderr there could be more sinks (file, etc.)

PR #1028, opened 2021-06-22
fix wrong update spec in history

madhur-tandon:


wolfv:
I think there are a couple more interesting cases, can we check that we handle them correctly?

`xtensor==0.12.*` -> `xtensor=0.12`

`xtensor==0.12*` -> `xtensor=0.12`

madhur-tandon:
Just checked, works for both.

PR #1033, opened 2021-06-23
Parse and use explicit platform specifications.

### LINKED ISSUE #794: Custom platform support

afranchuk:
Rather than trying to figure out whether a channel string/url contains a platform, we instead make platform specification explicit with `CHANNEL[PLATFORM1,PLATFORM2,...]`.

There is a refactor of the channel class here to hide implementation details/functions which are only relevant for testing.

This is a rework related to discussion at #934, and will close #794.

Note that this introduces breaking CLI changes as well as breaking Python API changes (the `Channel` class is improved), so these should be noted in the changelog.

### DIFF
--- a/src/api/install.cpp
+++ b/src/api/install.cpp
@@ -352,32 +352,34 @@ namespace mamba
             load_tokens();
         }
 
-        for (auto& url : channel_urls)
+        for (auto& chan_url : channel_urls)
         {
-            auto& channel = make_channel(url);
-            std::string repodata_full_url = concat(channel.url(true), "/repodata.json");
-
-            auto sdir
-                = std::make_shared<MSubdirData>(concat(channel.name(), "/", channel.platform()),
-                                                repodata_full_url,
-                                                cache_dir / cache_fn_url(repodata_full_url));
-
-            sdir->load();
-            multi_dl.add(sdir->target());
-            subdirs.push_back(sdir);
-            if (ctx.channel_priority == ChannelPriority::kDisabled)
+            auto& channel = make_channel(chan_url);
+            for (auto& [platform, url] : channel.platform_urls(true))
             {
-                priorities.push_back(std::make_pair(0, max_prio--));
-            }
-            else  // Consider 'flexible' and 'strict' the same way
-            {
-                if (channel.name() != prev_channel_name)
+                std::string repodata_full_url = concat(url, "/repodata.json");
+
+                auto sdir
+                    = std::make_shared<MSubdirData>(concat(channel.name(), "/", platform),
+                                                    repodata_full_url,
+                                                    cache_dir / cache_fn_url(repodata_full_url));
+
+                sdir->load();
+                multi_dl.add(sdir->target());
+                subdirs.push_back(sdir);
+                if (ctx.channel_priority == ChannelPriority::kDisabled)
                 {
-                    max_prio--;
-                    prev_channel_name = channel.name();
+                    priorities.push_back(std::make_pair(0, max_prio--));
                 }
-                priorities.push_back(
-                    std::make_pair(max_prio, channel.platform() == "noarch" ? 0 : 1));
+                else  // Consider 'flexible' and 'strict' the same way
+                {
+                    if (channel.name() != prev_channel_name)
+                    {
+                        max_prio--;
+                        prev_channel_name = channel.name();
+                    }
+                    priorities.push_back(std::make_pair(max_prio, platform == "noarch" ? 0 : 1));
+                }
             }
         }
         if (!ctx.offline)
--- a/src/api/install.cpp
+++ b/src/api/install.cpp
@@ -4,6 +4,8 @@
 //
 // The full license is in the file LICENSE, distributed with this software.
 
+#include <set>
+
 #include "mamba/api/configuration.hpp"
 #include "mamba/api/install.hpp"
 
@@ -333,7 +335,10 @@ namespace mamba
             LOG_WARNING << "No 'channels' specified";
         }
 
-        std::vector<std::string> channel_urls = calculate_channel_urls(ctx.channels);
+        std::vector<std::string> channel_urls = ctx.channels;
+        // Always append context channels
+        auto& ctx_channels = Context::instance().channels;
+        std::copy(ctx_channels.begin(), ctx_channels.end(), std::back_inserter(channel_urls));
 
         std::vector<std::shared_ptr<MSubdirData>> subdirs;
         MultiDownloadTarget multi_dl;
@@ -352,8 +357,13 @@ namespace mamba
             load_tokens();
         }
 
+        // Keep track of urls used to ignore duplicates
+        std::set<std::string> visited_urls;
         for (auto& chan_url : channel_urls)
         {
+            if (visited_urls.count(chan_url) > 0)
+                continue;
+            visited_urls.insert(chan_url);
             auto& channel = make_channel(chan_url);
             for (auto& [platform, url] : channel.platform_urls(true))
             {
@@ -362,7 +372,8 @@ namespace mamba
                 auto sdir
                     = std::make_shared<MSubdirData>(concat(channel.name(), "/", platform),
                                                     repodata_full_url,
-                                                    cache_dir / cache_fn_url(repodata_full_url));
+                                                    cache_dir / cache_fn_url(repodata_full_url),
+                                                    platform == "noarch");
 
                 sdir->load();
                 multi_dl.add(sdir->target());
--- a/src/api/install.cpp
+++ b/src/api/install.cpp
@@ -357,20 +357,14 @@ namespace mamba
             load_tokens();
         }
 
-        // Keep track of urls used to ignore duplicates
-        std::set<std::string> visited_urls;
-        for (auto& chan_url : channel_urls)
+        for (auto channel : get_channels(channel_urls))
         {
-            if (visited_urls.count(chan_url) > 0)
-                continue;
-            visited_urls.insert(chan_url);
-            auto& channel = make_channel(chan_url);
-            for (auto& [platform, url] : channel.platform_urls(true))
+            for (auto& [platform, url] : channel->platform_urls(true))
             {
                 std::string repodata_full_url = concat(url, "/repodata.json");
 
                 auto sdir
-                    = std::make_shared<MSubdirData>(concat(channel.name(), "/", platform),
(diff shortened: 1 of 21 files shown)


afranchuk:
The clang-format linting failed, but it's all on files that I never touched? I automatically clang-format any files I change.

src/mamba/py_interface.cpp
```diff
@@ -212,19 +215,33 @@ PYBIND11_MODULE(mamba_api, m)
...
+                     if (!first)
+                         s += ",";
+                     s += platform;
+                     first = false;
+                 }
+                 s += "]";
+                 return s;
+             });
 
     m.def("get_channel_urls", &get_channel_urls);
```
afranchuk, line 244: It's not clear whether `get_channel_urls` is important to keep around or not. I didn't yet remove it because it is the only function in the C++ code which uses the `get_custom_multichannels` function, which seemed odd to me.

afranchuk:
The other failing tests seem to be using an installed version of mamba rather than one built from source? That's why the python tests are failing (they work if I run pytest myself).

test/micromamba/test_linking.py
```diff
@@ -100,9 +100,9 @@ def test_cross_device(self, allow_softlinks, always_copy):
 
         create_args = ["xtensor", "-p", "/tmp/testenv", "--json"]
         if allow_softlinks:
-            create_args.append("--allow-softlinks", no_dry_run=True)
+            create_args.append("--allow-softlinks")
         if always_copy:
-            create_args.append("--always-copy", no_dry_run=True)
```
wolfv, line 105: I think we should keep the `no_dry_run=True` here because we check if we're linking the packages correctly by looking at inode etc.

afranchuk: Uh, that fails when I run the test, saying you can't have keyword args in `list.append`? `no_dry_run=True` is there just below in the call to `create`.

afranchuk:
Any comment on the failing tests? Do I need to rebase on master again?

wolfv:
@afranchuk I am currently pushing a lot to get this released: https://github.com/openSUSE/libsolv/pull/457

Will have to take a deep dive into this PR after.

I think some issues should be easily fixable, though (e.g.     AttributeError: 'Channel' object has no attribute 'platform_urls' ... 

I think there are legitimate failures.

afranchuk:
I was able to reproduce the pytest errors (I didn't realize I had accidentally excluded them before), and I think I fixed all of them (by changing how some parts of the python code work).

afranchuk:
I can run clang-format on everything, but it's still a bit odd that the linting for that is failing on files I never changed.

afranchuk:
Looks like there's still one test failing, let me see if I can track down why.

afranchuk:
The `test_env_update` test seems to pass an invalid argument (`-f`) to `mamba`. This would work with micromamba, however.

wolfv:
The CI mentions:

```
platform_url(): incompatible function arguments. The following argument types are supported:\\n    1. (self: mamba.mamba_api.Channel, platform: str, with_credentials: bool = True) -> str\\n\\nInvoked with: [linux-64,noarch], 'linux-64'; kwargs: with_credential=False\
```

afranchuk:
Bah, simple typo.

afranchuk:
Weird though, I ran the tests myself and it didn't have that error.

wolfv:
alright, looks like one remaining issue with windows for mamba: stdout = b'{\r\n  "command": "C:\\\\Miniconda\\\\Scripts\\\\mamba create -p C:\\\\Users\\\\runneradmin\\\\AppData\\\\Local\\\\T...ict[split_anaconda_token(remove_auth(c))[0]]\\nKeyError: \'file://d/a/mamba/mamba/test/channel_b/win-64\'\\n"\r\n}\r\n'


Maybe it's the trailing slash? or the `//` vs`\\`...?

afranchuk:
Yeah, unfortunately at the moment I don't have a windows machine to fix the windows errors more rapidly, but I'll try a few things.

afranchuk:
I added some context to the exception, it looks like it's being queried with `file://d/...` but the actual paths are `file://D:/...`. I guess I'm missing some sanitization step?

afranchuk:
It's not immediately obvious what is altering the file path to `d` rather than `D:`; based on existing tests I think `D:` is the desired/correct path component.

afranchuk:
I think the issue is the `remove_auth` conda function being called, which calls `parse_urls` which has this: https://github.com/conda/conda/blob/33a142c16530fcdada6c377486f1c1a385738a96/conda/_vendor/urllib3/util/url.py#L186. I assume the difference is that now we use `api.Channel` to parse urls rather than conda's `Channel`, which results in a mismatch.

afranchuk:
Basically their implementation is wrong for windows `file://` uris.

afranchuk:
Okay, looks like all that's wrong is the linting (though I can't fix that myself apparently since my `clang-format` disagrees with a lot). If you're okay with the workaround I did for that windows issue it should be in a good state.

wolfv:
Great work! Fixing the linting is easy for me (just run pre-commit run —all)

I’ll need to have a deep dive into this refactor. I saw that you’re using std::optional which is one thing I’m not sure about (since we need to support pretty old macOS versions) but we could quite easily refactor that to pointers or similar

wolfv:
I just ran the linter, let's see if it passes now :)

afranchuk:
Awesome! Yeah I have clang-format 12, beware of differences in the future!

src/api/install.cpp
```diff
@@ -4,6 +4,8 @@
 //
 // The full license is in the file LICENSE, distributed with this software.
 
+#include <set>
```
afranchuk, line 7: No longer needed.

- [ ] Remove

wolfv:
I am testing this PR right now, gonna write down observations as I go:

When using micromamba, and I have `channels: [conda-forge]` configured in my condarc file, I get this download:

```
(base) ➜  cbuild git:(explicit-platforms) ✗ micromamba create ninja -n tlinx -c "conda-forge[linux-64, noarch]"

                                           __
      /_/

Found conda-prefix at '/Users/wolfvollprecht/micromamba/envs/tlinx'. Overwrite?: [y/N] y
conda-forge/osx-arm64    [====================] (00m:00s) Done
conda-forge/noarch       [====================] (00m:02s) Done
conda-forge/noarch       [====================] (00m:02s) Done
conda-forge/linux-64     [====================] (00m:06s) Done
```

IMO the command line here should overwrite the other configured conda-forge config, shouldn't it?

wolfv:
- there is currently no implicit noarch (when explicitly selecting channels). 

This means this fails now like this:

```
(base) ➜  cbuild git:(explicit-platforms) ✗ micromamba create ninja -n tlinx -c "conda-forge33[linux-64, osx-arm64]" --override-channels

                                           __
      /_/

conda-forge33/osx-arm64  [====================] (00m:00s) 404 Failed
conda-forge33/linux-64   [====================] (00m:00s) 404 Failed
  - nothing provides requested ninja

ERROR   Could not solve for environment specs
```

I think it would be better in this case to fail with a donwload error of some sort (which we did otherwise with the existence of the noarch channel). If we explicitly request some platforms, we coudl require that all of them resolve to something on the server..?

wolfv:
Let's fix the remaining issues on master :)

afranchuk:
Sounds good! As to your comments:

> IMO the command line here should overwrite the other configured conda-forge config, shouldn't it?

I'm not so sure. You could certainly make a decision here about how you think it should behave, but I think the behavior you saw is okay/expected, same as if you had specified `-c conda-forge[linux-64,noarch] -c conda-forge`. I understand why you might want to have the behavior that you expected, but that's kind of what `--override-channels` is for. You could certainly make that other case work the way you expect, but I think that's more complicated and not quite as straightforward to understand.

> I think it would be better in this case to fail with a donwload error of some sort (which we did otherwise with the existence of the noarch channel). If we explicitly request some platforms, we coudl require that all of them resolve to something on the server..?

This seems like an error priority issue. I'd argue that _any_ 404 received for a configured channel should take precedence as a final user-visible error over an error that a package could not be resolved. Are there any use cases where you'd want a 404 to be ignored?

wolfv:
@afranchuk yes, if you request a channel that doesn't implement the default platform the error should be ignored.

e.g. you can use `defaults` from osx-arm64 but you'll only get the `noarch` platform. When mixing channels + platforms it's likely that it might sometimes fail

wolfv:
@afranchuk it seems that the load_channels implementation also changed slightly, correct? Would you mind detailing what I'll have to adjust in `boa` to make it work? 

https://github.com/mamba-org/boa/blob/5482bd4a8eea22c93d7f2d87692e93f0264f9876/boa/core/solver.py#L110-L112

Currently it doesn't seem to load the `noarch` platform.

Maybe we could actually revert this behavior in a way that keeps boa working?

afranchuk:
@wolfv `load_channels` should be functionally identical, however the index it returns is different. The index has the same keys, however values (which used to be `conda.Channel`) are now a dict with the following entries:
* `channel`: the `mamba.Channel`
* `platform`: the platform name (string)
* `url`: the channel url (string)

wolfv:
I think one problem is that when giving a platform explicitly now noarch isn't implicitly loaded. That's breaking boa right now but I'll come up with a fix tomorrow :)

afranchuk:
Oh yeah you're right, `load_channels` is the same but the use of `mamba.Channel` rather than `conda.Channel` changes that resolution behavior.

PR #1047, opened 2021-06-30
blacklist lower prio packages

wolfv:


wolfv:
Merging this for now, until we have a way to do this in libsolv directly

adriendelsalle:
I'm not sure adding this in `MSolver::add_jobs` is the right place, it may be called multiple times.
Is it also possible to add tests, especially when it modifies the solver behavior?

wolfv:
It's added in Solver::solve.

For tests, I think this specific feature is covered in libsolv.

adriendelsalle:
> It's added in Solver::solve.

ah yes sorry the code was collapsed and I got confused

> For tests, I think this specific feature is covered in libsolv.

the unit test about blacklisting is tested in libsolv but the usage we have in `mamba` is not

PR #1057, opened 2021-07-05
The RE recognizing tokens is too narrow

asterbini:
When using mamba with tokens sometimes the token part is not recognixed because it contains the '_' char.
This patch enlarge the RE to all chars different from '/'

### DIFF
--- a/src/core/url.cpp
+++ b/src/core/url.cpp
@@ -24,7 +24,7 @@ namespace mamba
 
     void split_anaconda_token(const std::string& url, std::string& cleaned_url, std::string& token)
     {
-        std::regex token_re("/t/([a-zA-Z0-9-]*)");
+        std::regex token_re("/t/([^/]*)");
         auto token_begin = std::sregex_iterator(url.begin(), url.end(), token_re);
         if (token_begin != std::sregex_iterator())
         {


wolfv:
can I ask where you get this token from? Is that from anaconda.org or from a quetz instance?

asterbini:
> can I ask where you get this token from? Is that from anaconda.org or from a quetz instance?

It's from Anaconda, I get it when I am logged on in anaconda-navigator

wolfv:
does your channel / username start with something like `A_`? is one of the first two letters an underscore?
it's super weird, the tokens of anaconda.org are pretty predictable (xx-abcdef123-1234-1234-1234-123412312312) --> they follow a schema of a UUID4 and there absolutely cannot be any characters outside of [1-9A-Za-z]

wolfv:
To note: the `xx-` starting two characters are teh first two chars of the channel the token is for.

asterbini:
> To note: the `xx-` starting two characters are teh first two chars of the channel the token is for.

I think they are the first two letters of my username on anaconda.org (a_sterbini).

wolfv:
that makes sense, thanks for the clarification.

asterbini:
I think also this line in util.hpp should be changed
'''
const static std::regex token_re("/t/([a-zA-Z0-9-]*)");
'''

wolfv:
I'll have a look at this today

PR #1059, opened 2021-07-06
finalize extraction progress bar

wolfv:
cc @adriendelsalle @JohanMabille I think this could fix the progress bar issues.


wolfv:
that doesn't seem to work at all :)

PR #1095, opened 2021-07-21
Improvement: Log actual MD5/Size/SHA256 when we fail to validate

johnhany97:
## Before this PR
Currently when mamba logs the validation error, we only see information about what we expect, with no info on what was expected. Example:
```
ERROR File not valid: file size doesn't match expectation "/tmp/caches/CONDA_PKGS_DIRS/future-0.18.2-py36h5fab9bb_3.tar.bz2"

Expected: 733415
```
or
```
ERROR File not valid: MD5 sum doesn't match expectation "/tmp/caches/CONDA_PKGS_DIRS/pip-21.1.3-py36h06a4308_0.tar.bz2"

Expected: 5ffa50ebfad2aab590348cff27322d32
```
## After this PR
I propose we additionally log the actual value we got, not just the expected value. This should help facilitate debugging issues and help understand what was going on.

### DIFF
--- a/src/core/transaction.cpp
+++ b/src/core/transaction.cpp
@@ -100,7 +100,7 @@ namespace mamba
         {
             LOG_ERROR << "File not valid: file size doesn't match expectation " << m_tarball_path
                       << "\nExpected: " << m_expected_size
-                      << "\nActual:: " << size_t(m_target->downloaded_size) << "\n";
+                      << "\nActual: " << size_t(m_target->downloaded_size) << "\n";
             m_progress_proxy.mark_as_completed("File size validation error.");
             m_validation_result = SIZE_ERROR;
         }
--- a/src/core/transaction.cpp
+++ b/src/core/transaction.cpp
@@ -103,26 +103,32 @@ namespace mamba
                       << "\nActual: " << size_t(m_target->downloaded_size) << "\n";
             m_progress_proxy.mark_as_completed("File size validation error.");
             m_validation_result = SIZE_ERROR;
+            return;
         }
         interruption_point();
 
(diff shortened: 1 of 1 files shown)


johnhany97:
Unsure why `umamba_tests` are failing checks. The error seems to be a `443: Connection refused` which makes me suspect this is not relevant to the changes introduced in the PR.

adriendelsalle (commented):
Thanks for this PR @johnhany97 ! 

Maybe you could just store the checksums instead of recomputing them for logging?

> Unsure why umamba_tests are failing checks.

We had some issues with the CI caused by runner and dependency updates, rebasing should fix it!

wolfv:
@adriendelsalle can you have another look? I refactored the code slightly.

src/core/transaction.cpp
```diff
@@ -99,27 +99,36 @@ namespace mamba
...
+                m_validation_result = SHA256_ERROR;
+                m_progress_proxy.mark_as_completed("SHA256 sum validation error.");
+                LOG_ERROR << "File not valid: SHA256 sum doesn't match expectation "
+                          << m_tarball_path << "\nExpected: " << m_sha256
+                          << "\nActual: " << sha256sum << "\n";
+            }
+            return;
         }
-        else
+        if (!m_md5.empty())
```
adriendelsalle, line 123: this is changing the behavior, previously the `MD5` was not checked if the `SHA256` was provided

adriendelsalle (approved):
LGTM @wolfv 

Just confirm that you want to change the behavior to always check `MD5` if provided

src/core/transaction.cpp
wolfv: there is a return above

Issue #1099, opened 2021-07-23
Python pinning is too precise


mparry:
Currently mamba will pin Python at its exact semver, i.e. including the patch component. This seems incorrect, e.g. when running `mamba update --all` in an environment. I realised recently that, despite doing so semi-regularly, several of my envs had Python 3.9.1 instead of 3.9.5, which isn't something I would generally want.

Wouldn't it be more correct and/or the expected behaviour to only pin `major.minor` for it?

(I realise this functionality is based upon conda's, but I didn't think conda pinned the patch version. That said, quite a while since I used conda's solver, for obvious reasons!)

wolfv:
Can you send a PR to fix this? https://github.com/mamba-org/mamba/blob/ceb85460fd22f6cb2e7446622b655dff7036718a/mamba/mamba.py#L492-L496

mparry:
Sure. The associated PR seems to do the trick.

adriendelsalle:
https://github.com/mamba-org/mamba/pull/1101#discussion_r680092430

There is something to investigate about the `python` pinning because even if the `SOLVER_LOCK` are correctly placed on solvables not matching the spec in `micromamba`, `python` is not part of the update specs when running `micromamba update --all`

adriendelsalle:
Just checked quickly but it was a deliberate choice.
See https://github.com/mamba-org/mamba/blob/ceb85460fd22f6cb2e7446622b655dff7036718a/src/api/update.cpp#L38

Running
`micromamba create -n pins python=3.7.0` installs `python=3.7.0`
`micromamba update -n pins -a` proposes `python=3.7.10` (pinning is `python=3.7.*`)
`micromamba update -n pins -a python` proposes `python=3.9.6`

mparry:
This issue concerns mamba though, not micromamba. (I hadn't actually tested the latter - we don't use it much currently.)

adriendelsalle:
@mparry yep thanks for the precision, I mention it because the solving step should be the exact same for both, which is not the case.

mparry:
Addressed by #1101

PR #1101, opened 2021-07-26
Only pin major.minor version of Python; allow patches in update --all

mparry:
Addresses #1099

### DIFF
--- a/mamba/mamba.py
+++ b/mamba/mamba.py
@@ -435,8 +435,9 @@ def install(args, parser, command="install"):
     # and that they are name-only specs
     if isupdate and context.update_modifier == UpdateModifier.UPDATE_ALL:
         for i in installed_names:
-            if i != "python":
-                specs.append(MatchSpec(i))
+            if i.startswith("__"):
+                continue
+            specs.append(MatchSpec(i))
 
         prefix_data = PrefixData(prefix)
         for s in args_packages:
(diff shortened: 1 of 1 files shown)


mparry (commented):
Note that I didn't make any changes to the similar code over in `mamba/mamba_env.py` because I figured the current behaviour was probably correct in that case, if it relates to env creation from specs

mamba/mamba.py
```diff
@@ -555,7 +554,6 @@ def install(args, parser, command="install"):
 
         pinned_specs_info = ""
         if python_constraint:
-            solver.add_pin(python_constraint)
             pinned_specs_info += f"  - {python_constraint}\n"
```
mparry, line 557: Of course, this isn't actually implemented as a pin now, so I don't know if you would prefer this output to change somehow. It did seem useful to retain something, but I'm happy to tweak. As things stand, it will look like this:

```
Pinned packages:
  - python 3.9.*
```

wolfv: what's wrong about keeping it a `pin`? As far as I can see it's currently added to the regular "install specs"?

mparry: That was the first thing I tried, naturally, but it didn't seem to work: regardless of how I phrased the looser 'pin spec', it retained the current patch version. For example, I had `python-3.9.4` installed and with a pin of either `python==3.9` or `python==3.9.*` it still kept `3.9.4`. With this regular 'install spec' of `python==3.9.*` instead, it offers `3.9.5`, as it should.

Please let me know if there's some better way. I didn't really want to start digging into the solver side - that seems likely to require beyond the amount of time I can spend on this.

mparry: (From a very quick at the solver, it seems to me that if `SOLVER_LOCK` is applied to a package, that package will then simply go unchanged, and the rest of its spec is a bit irrelevant.)

adriendelsalle: The `SOLVER_LOCK` is only placed on solvables/packages that are not matching the spec.
So it should work as expected to create a relaxed pin (without the patch).

mparry: As I said above, that's what I initially tried, e.g.

```python
python_constraint = MatchSpec(f"python=={major_minor_version}").conda_build_form()
```

It doesn't work. This current solution does, though I'm entirely prepared to accept it may not be the ideal approach.

wolfv:
The problem with the previous solution was the following: when update all, we _removed_ python from the specs that should be updated, therefore no update of python was attempted.

I modified things a little bit to keep `python` in the specs to update + add a pin for the current minor version. This way we should get the desired behavior for python updates.

Let me know what you think :)

wolfv:
Thanks @mparry for the initial work on this!

mparry:
Looks like you got to the real root of the problem, and the changes work for me locally. Thanks!

PR #1103, opened 2021-07-27
Feature: Support compatibility constraints

JerryMaoQC:
### Summary
Provide a new feature allowing users to specify additional packages they want the solved environment to be compatible with, but ***without actually downloading or installing*** those packages.

### Use case
In many of our workflows, the software dependencies in our production environments are a strict subset of our development environments. For example, one would not need `pytest` or `gtest` in production. These differing sets of dependencies can result in differently solved environments (see example), which is suboptimal for ensuring software behaves identically during development and production.

One possible solution is to install additional dependencies in production to enforce consistency with the development environment. However, it does not really make sense to waste production resources in this way.

This feature allows you to create environments ***as if*** those additional dependencies were also present. The resulting environment is compatible with those additional dependencies, but does not contain them because they are not actually required. All installed packages would then have the same consistent version.

### Example
Here we describe a toy and somewhat contrived example. Imagine if, the production environment requires only `numpy=1.21.1`, and the development environment ***additionally*** requires `pytest=6.0.1`. Then, let's try:
```bash
mamba create -yq -n example-prod -c conda-forge numpy=1.21.1
mamba create -yq -n example-dev  -c conda-forge numpy=1.21.1 pytest=6.0.1
conda list -n example-prod "^python$"
conda list -n example-dev  "^python$"
```
The resulting environments contain Python 3.9.6 in prod, but only Python 3.8.10 in dev. This is inconsistent, which isn't very good. Although this is a very contrived example (eg. the developer can simply update to the latest pytest), with more complex dependencies a simple solution does not always exist (eg. perhaps the latest version still only supports Python 3.8).

Given the new feature, we can instead create the production environment with the following command to maintain compatibility without installing pytest.
```bash
mamba create -yq -n example-prod -c conda-forge numpy=1.21.1 --compatible pytest=6.0.1
```
*(We're also open to renaming this command-line argument to something else instead of `-x,--compatible`.)*

### Implementation
We add a component to the transaction's filter to ensure that only relevant specs are actioned upon. After the environment is solved, we perform a multi-source BFS through the dependency graph, using as starting nodes the set of specs the user has explicitly requested mamba to install. Any specs not visited by this BFS are filtered out of the transaction.

### Tasks checklist
- [x] Implement feature and link to micromamba
- [ ] Discuss and reach consensus about implementation and usage of the feature
- [ ] Add tests (after above discussion)
- [ ] Add bindings to expose functionality to mamba

FYI: @wolfv @fhoehle


wolfv:
Hi @JerryMaoQC thanks for tackling this! Looks like excellent work :)

I am on vacation for 2 weeks, but will be happy to have a look upon return.

wolfv:
Hi Jerry, I will have a look at this PR today and give some feedback. Sorry for the long delay!

JerryMaoQC:
Thanks Wolf! I've also rebased the PR to bring in your update for the CI dependencies.

wolfv:
Hmm, I am currently wondering if we can find a way to achieve this by leveraging libsolv. Instead of manipulating the transaction, we could also first solve, then create a "fake" installed repo, and remove packages from that fake installation through libsolv. I believe that would make it possible to remove a bit of code from this, right?

I also agree that it seems to make most sense for a "create" operation, more than for an installation op.

I need to think a bit more about this ...

JerryMaoQC:
I also agree that it seems possible to get this functionality out of libsolv by invoking libsolv twice. First to get the plan to install all of the specs, and second to remove the constraint packages, using a frozen/locked environment.

I also think it could be logical to do this using libsolv, instead of manipulating the transaction to emulate what libsolv would do. Do you think it would be a good idea to change the implementation to do it this way? My understanding is this requires us to take the output of the first libsolv call, process it, and then send it as input to the second libsolv call. I'm not fully familiar with how to use the libsolv API though, so please let me know if you think this is the best approach.

wolfv:
Hi @JerryMaoQC again, sorry that this process is a bit slow. 

I see two possibilities for a clean solution:

First create a solve, then add pins for all packages from the initial solution (possible using solver.add_pin(...). Then one could run another job and should have all dependencies completely pinned down.

The other approach would be to generalize this functionality _even more_ with one specially interesting benefit that would allow us to test mamba's resolutions in a much nicer way: to execute "fake" transactions that influence the special "installed" repo.

I've written down my thoughts here: https://gist.github.com/wolfv/e6f99294cc92c4811e396eb8e8ea3a88

wolfv:
Note, what I am describing in the gist -- for your use-case we'd need to transform an annotated yaml into the structure proposed there. But that should be straightforward once we have the execution engine for these multistage YAMLs :)

Let me know what you think!

JerryMaoQC:
Hi @wolfv, thanks for the ideas and sorry about the slow reply. I agree that the first solution (2 jobs with pinning) is a very clean implementation of the feature, and I also agree that the multi-stage pipeline is an even more general solution.

Regarding the multi-stage pipeline, are there any other use-cases you foresee, other than the compatibility constraints and testing? I feel like some additional applications might better justify adding a new syntax to the yaml; otherwise the first solution seems sufficient.

wolfv:
Another use case could be a package REPL (similar to what Julia has) where one could go like:

```
> add numpy
+ numpy 1.2
> add tensorflow
... adding tensorflow
+ tensorflow      4.3
+ tensorflow-base 4.5
^ numpy 1.2 -> 3.2
> remove matplotlib
- matplotlib 2.1
> execute
... installing blabla ...
```

So that the transaction is only generated + executed when `execute` is called (or similar).

wolfv:
Another use case could be a package REPL (similar to what Julia has) where one could go like:

```
> add numpy
+ numpy 1.2
> add tensorflow
... adding tensorflow
+ tensorflow      4.3
+ tensorflow-base 4.5
^ numpy 1.2 -> 3.2
> remove matplotlib
- matplotlib 2.1
> execute
... installing blabla ...
```

So that the transaction is only generated + executed when `execute` is called (or similar).

JerryMaoQC:
Thanks for the example! If you don't mind, may I ask why a user might wish to use a REPL in this way, without running code in between, such as to add numpy and tensorflow in separate steps? I'm sorry about all the additional questions, I really appreciate your help with my understanding!

wolfv:
@JerryMaoQC this would be like an "interactive" dry-run.
Another place where this could be very useful is in `boa`: we have an interactive mode where you can run a recipe, and if it fails, one can change the `build`, `host` and `run` environment. For example the user could say `add package B` to the host environment, and then we could execute that step later when the user issues the `build` command.

jonashaag:
@AntoinePrv thanks for the pointer, it works pretty well, I think we might be able to use this.

UX wise it would be nice to be able to pass a `pinned` file when using `micromamba create`, maybe we will add that.

PR #1108, opened 2021-08-03
Fix `mamba` not writable cache error

### LINKED ISSUE #1098: mamba install fails due to RuntimeError: Permission denied: <global conda>/pkgs/cache
After updating mamba in the global base environment (only accessible via elevated priviliges)
```
conda activate <some local environment>
mamba install <some package>
# the same is true if I try
mamba create -n <some new env> <some packages>
```
fails with
```
RuntimeError: Permission denied: '/opt/conda/pkgs/cache/9ee0c365'
```
The global `.condarc` I have
```
pkg_dirs:
  - /opt/conda/pkgs
  - $HOME/.conda/pkgs
envs_dirs:
  - /opt/conda/envs
  - $HOME/.conda/envs
```
while my USER `.condarc` has
```
pkg_dirs:
  - $HOME/.conda/pkgs
envs_dirs:
  - $HOME/.conda/envs
```
In the printed report coming after the error I noticed that
```
package cache : <global conda>/pkgs
                         $USER/.conda/pkgs
envs directories : $USER/.conda/envs
                          <global conda>/envs
```
the order of package cache is reversed while the envs directries look fine.
I also tried installing via
```
CONDA_PKGS_DIRS=$HOME/.conda/pkgs mamba install <some package>
```
which fails with the same issue but in the report after the rror
```
package cache : $USER/.conda/pkgs
envs directories : $USER/.conda/envs
                          <global conda>/envs
```
which suggests that
> `mamba` seems to use only the global package cache and ignores settings in the `rcfiles` or via environment variables

I have the following versions
```
mamba --version
# mamba 0.15.2
# conda 4.10.3
```
tested on centos7.

adriendelsalle:
Description
---

Fix `mamba` `runtime_error` when `root_prefix` cache in not writable
Use the first writable cache dir in `Channel` constructor to re-use `Context.pkgs_dirs` filled using `conda` api

The bug has been introduced in https://github.com/mamba-org/mamba/pull/1033, by using `libmamba` `Channel` instead of `conda`'s one.

Closes #1098

### DIFF
--- a/src/api/configuration.cpp
+++ b/src/api/configuration.cpp
@@ -219,7 +219,15 @@ namespace mamba
 
             auto& ctx = Context::instance();
             ctx.envs_dirs = { prefix / "envs" };
-            ctx.pkgs_dirs = { prefix / "pkgs" };
+
+            if (std::getenv("CONDA_PKGS_DIRS") != nullptr)
+            {
+                ctx.pkgs_dirs = { fs::path(std::getenv("CONDA_PKGS_DIRS")) };
+            }
+            else
+            {
+                ctx.pkgs_dirs = { prefix / "pkgs" };
+            }
         }
         /*
                 void log_level_hook(LogLevel& lvl)
--- a/src/api/install.cpp
+++ b/src/api/install.cpp
@@ -290,15 +290,7 @@ namespace mamba
         auto& no_py_pin = config.at("no_py_pin").value<bool>();
         auto& retry_clean_cache = config.at("retry_clean_cache").value<bool>();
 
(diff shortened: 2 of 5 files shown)


adriendelsalle:
@wolfv any idea why the tests you introduced recently are failing on Windows?

Will try to debug that if you don't have time to have a look.

adriendelsalle:
The exact same tests are failing on `0.15.2` workflow re-ran: https://github.com/mamba-org/mamba/runs/3233778749?check_suite_focus=true

Issue #996, opened 2021-06-11
Document log levels


adriendelsalle:
Description
---

Add a section about log levels in developer docs:
- what are available the log levels
- how they should be used

Maybe also add a section in advanced usage to point to levels description

adriendelsalle:
I like this description

```
TRACE level: this is a code smell if used in production. This should be used during development to track bugs, but never committed to your VCS.
DEBUG level: log at this level about anything that happens in the program. This is mostly used during debugging, and I’d advocate trimming down the number of debug statement before entering the production stage, so that only the most meaningful entries are left, and can be activated during troubleshooting.
INFO level: log at this level all actions that are user-driven, or system specific (ie regularly scheduled operations…)
NOTICE level: this will certainly be the level at which the program will run when in production. Log at this level all the notable events that are not considered an error.
WARN level: log at this level all events that could potentially become an error. For instance if one database call took more than a predefined time, or if an in-memory cache is near capacity. This will allow proper automated alerting, and during troubleshooting will allow to better understand how the system was behaving before the failure.
ERROR level: log every error condition at this level. That can be API calls that return errors or internal error conditions.
FATAL level: too bad, it’s doomsday. Use this very scarcely, this shouldn’t happen a lot in a real program. Usually logging at this level signifies the end of the program. For instance, if a network daemon can’t bind a network socket, log at this level and exit is the only sensible thing to do.
```

PR #1109, opened 2021-08-03
Handle multiple caches in umamba

adriendelsalle:
Description
---

Handle multiple caches in `micromamba`, to be able to use first writable one as done in `conda` or `mamba` (relying on `conda` implementation)


wolfv:
Was this ready besides the (now fixed) CI failures?

adriendelsalle:
I'm not sure, I'll rebase and write some tests

src/mamba/py_interface.cpp
```diff
@@ -75,7 +75,7 @@ PYBIND11_MODULE(mamba_api, m)
         .def("clear", &MRepo::clear);
 
     py::class_<MTransaction>(m, "Transaction")
-        .def(py::init<MSolver&, MultiPackageCache&, const std::string&>())
```
wolfv, line 78: this will be a problem for `boa`.

include/mamba/core/context.hpp
```diff
@@ -109,7 +109,15 @@ namespace mamba
 
         // TODO check writable and add other potential dirs
         std::vector<fs::path> envs_dirs = { root_prefix / "envs" };
-        std::vector<fs::path> pkgs_dirs = { root_prefix / "pkgs" };
+        std::vector<fs::path> pkgs_dirs
+            = { root_prefix / "pkgs",
+                fs::path("~") / ".mamba" / "pkgs"
+#ifdef _WIN32
+                ,
+                fs::path(std::getenv("APPDATA") ? std::getenv("APPDATA") : "") / ".mamba" / "pkgs"
```
wolfv, line 117: hm, i think there is a better way: https://github.com/mamba-org/mamba/blob/da87c2675370f2557369a982ac40aee809995e41/src/core/menuinst.cpp#L252-L253

src/api/install.cpp
```diff
@@ -334,11 +330,11 @@ namespace mamba
...
         MultiDownloadTarget multi_dl;
-        std::unique_ptr<LockFile> subdir_download_lock;
+
+        std::vector<std::unique_ptr<LockFile>> subdir_download_locks;
         if (!ctx.offline)
-        {
-            subdir_download_lock = std::make_unique<LockFile>(cache_dir / "mamba.lock");
-        }
+            for (const auto& c : ctx.pkgs_dirs)
+                subdir_download_locks.push_back(std::make_unique<LockFile>(c / "mamba.lock"));
```
wolfv, line 337: do we really need to lock multiple? isn't it enough to lock the `first writable`?

src/mamba/py_interface.cpp
```diff
@@ -46,6 +46,7 @@ PYBIND11_MODULE(mamba_api, m)
 
     py::class_<fs::path>(m, "Path")
         .def(py::init<std::string>())
+        .def("string", [](fs::path& self) -> std::string { return self.string(); })
```
wolfv, line 49: can we use `__str__`?

src/api/install.cpp
adriendelsalle: I removed that part to a clever lock, only when working with that cache (because we even don't know if the first writable one will be needed or not, or if the packages will be picked from one or multiple caches)

include/mamba/core/context.hpp
adriendelsalle: great thx

src/mamba/py_interface.cpp
adriendelsalle: yes I bet downstream projects will be impacted. the current signature is unrelevant regarding multiple caches

src/mamba/py_interface.cpp
adriendelsalle: sure

include/mamba/core/context.hpp
adriendelsalle: do you know the name corresponding to appdata? this one is the userprofile

wolfv: I think we need to first factor out the code from `menuinst.cpp` and move it to some utils file.
Then we could use `https://docs.microsoft.com/en-us/windows/win32/shell/knownfolderid` on of the folderid's here (e.g. `FOLDERID_LocalAppData` or one of the others, not sure which one is the appropriate one).

src/core/subdirdata.cpp
```diff
@@ -197,16 +235,16 @@ namespace mamba
     {
         if (m_target->result != 0 || m_target->http_status >= 400)
         {
-            LOG_INFO << "Unable to retrieve repodata (response: " << m_target->http_status
-                     << ") for " << m_repodata_url;
+            LOG_DEBUG << "Unable to retrieve repodata (response: " << m_target->http_status
```
wolfv, line 238: why is this log debug? It seems INFO would be appropriate?

src/core/subdirdata.cpp
```diff
@@ -106,65 +108,101 @@ namespace mamba
...
+
+                    auto cache_age_seconds
+                        = std::chrono::duration_cast<std::chrono::seconds>(cache_age).count();
+                    if ((max_age > cache_age_seconds || Context::instance().offline))
+                    {
+                        // valid json cache found
+                        if (!m_loaded)
+                        {
+                            LOG_DEBUG << "Using JSON cache";
+                            LOG_TRACE << "Cache age: " << cache_age_seconds << "/" << max_age
```
wolfv, line 157: this should be log info.

src/core/subdirdata.cpp
```diff
@@ -249,22 +329,25 @@ namespace mamba
...
         m_mod_etag["_etag"] = m_target->etag;
         m_mod_etag["_mod"] = m_target->mod;
         m_mod_etag["_cache_control"] = m_target->cache_control;
 
-        LOG_INFO << "Opening: " << m_json_fn;
-        std::ofstream final_file(m_json_fn);
-
-        create_cache_dir();
+        LOG_DEBUG << "Opening '" << json_file.string() << "'";
+        path::touch(json_file, true);
```
wolfv, line 345: What's the point of the `touch` here?

src/core/util.cpp
```diff
@@ -1021,6 +1021,21 @@ namespace mamba
...
+        {
+            auto ptr = std::make_unique<LockFile>(path);
+            return ptr;
+        }
+        catch (...)
+        {
+            LOG_WARNING << "LockFile creation for path '" << path.string()
+                        << "' failed, continuing without it";
+        }
+        return NULL;
```
wolfv, line 1036: `NULL` is `C`, `nullptr` is `C++`.

wolfv:
Looks great, these will be awesome changes! Can't wait to release micromamba with all this :)

include/mamba/core/util.hpp
```diff
@@ -172,18 +174,17 @@ namespace mamba
...
         int m_pid;
 #endif
-
-        bool set_lock(bool blocking) const;
-
         int read_pid() const;
         bool write_pid(int pid) const;
+
+        bool set_fd_lock(bool blocking) const;
+        bool lock_non_blocking();
```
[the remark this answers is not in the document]
adriendelsalle: yes it's on purpose, I just realized that we need a factory to create `LockFile` with error handling and `try_lock` is a good and explicit name for that.
So I renamed some private methods with more explicit ones

src/api/install.cpp
```diff
@@ -58,9 +58,9 @@ namespace mamba
         options.redirect.parent = true;
         options.working_directory = cwd.c_str();
 
-        std::cout << "\n"
-                  << termcolor::cyan << "Installing " << pkg_mgr
-                  << " packages: " << join(", ", deps) << termcolor::reset << std::endl;
+        Console::stream() << std::endl
```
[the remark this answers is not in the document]
adriendelsalle: no problem, the point was more to use `Console::stream()` to avoid extra new lines when running quiet mode

src/api/install.cpp
```diff
@@ -307,15 +308,10 @@ namespace mamba
...
-        fs::path cache_dir = pkgs_dirs / "cache";
-        try
-        {
-            fs::create_directories(cache_dir);
-        }
-        catch (...)
-        {
-            throw std::runtime_error("Could not create `pkgs/cache/` dirs");
-        }
+        // fs::path cache_dir = pkgs_dir / "cache";
```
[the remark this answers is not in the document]
adriendelsalle: :+1:

src/core/subdirdata.cpp
adriendelsalle: a log info should be limited to the actions the user asked for, not internal machinery
I posted [here](https://github.com/mamba-org/mamba/issues/996#issuecomment-937637727) what I would like to use as a log policy

src/core/subdirdata.cpp
adriendelsalle: humm yeah maybe it's a warning here

src/core/subdirdata.cpp
wolfv: OK.

src/core/subdirdata.cpp
wolfv: Imaging being an sysadmin trying to debug why mamba won't load cached files though ... those are the users of our logs (and we are, when people are filing bugs). Making these logs to TRACE gives us little chance to figure out whats going on.

src/core/package_cache.cpp
```diff
@@ -46,214 +43,218 @@ namespace mamba
...
-                if (extract_dir_valid)
+                if (valid)
                 {
-                    extract_dir_valid = validate(extract_dir);
+                    valid = validate(extracted_dir);
                 }
             }
-            if (!extract_dir_valid)
+            /*
+            if (!valid)
```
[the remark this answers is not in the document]
adriendelsalle: sorry I forgot to move that part of the code
- a method `has_valid_extracted_dir` should not try to remove the directory but only check it
- we also need to deal with read-only caches
I'll handle that case

PR #1126, opened 2021-08-12
Add Python bindings for content trust core API

adriendelsalle:
Description
---

Add Python bindings for content trust core API
Add overloads and Python bindings to g:
- generate an `ed25519` key pair
- sign data

Add ``PkgMgrRole` and `Key.json_str` bindings
Make `V06RoleBaseExtension` a public base of roles
Add v06 roles constructors using json string
Add tests
Remove `nl` namespace alias to avoid undetected errors in Python bindings

### DIFF
--- a/src/mamba/py_interface.cpp
+++ b/src/mamba/py_interface.cpp
@@ -8,6 +8,8 @@
 #include <pybind11/pybind11.h>
 #include <pybind11/stl.h>
 
+#include "nlohmann/json.hpp"
+
 #include "mamba/core/channel.hpp"
 #include "mamba/core/context.hpp"
 #include "mamba/core/package_handling.hpp"
@@ -20,7 +22,10 @@
 #include "mamba/core/transaction.hpp"
 #include "mamba/core/url.hpp"
 #include "mamba/core/util.hpp"
+#include "mamba/core/validate.hpp"
 #include "mamba/core/virtual_packages.hpp"
+
+#include <stdexcept>
 
 namespace py = pybind11;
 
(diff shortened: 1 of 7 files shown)


src/mamba/py_interface.cpp
```diff
@@ -246,6 +254,82 @@ PYBIND11_MODULE(mamba_api, m)
...
+            return signature;
+        },
+        py::arg("data"),
+        py::arg("secret_key"));
+
+    py::class_<validate::Key>(m, "Key")
+        .def_readwrite("keytype", &validate::Key::keytype)
+        .def_readwrite("scheme", &validate::Key::scheme)
+        .def_readwrite("keyval", &validate::Key::keyval)
+        .def("to_json",
```
wolfv, line 274: Do we really need `pybind11_json`? Because I would expect these functions to return a `std::string` that contains the JSON.

adriendelsalle: I see, in fact I was also using it in `RootImpl`, `KeyMgrRole` and `PkgMgrRole` constructors.
It was pretty handy in `Quetz` but I can try to do another way to remove this extra dep

wolfv: why not just:

```
std::stringstream out;
out << j.dump(4);
return out.str();
```

wolfv: maybe I am not understanding correctly what pybind json does? I understand it creates an object out of the string?

But do we need that given that we already access variables on the `Key` object through `.def_readwrite...`?

adriendelsalle: for `Key` it's definitely because I was lazy ;)
for the main content roles, I just created factories to have Python constructor taking json strings instead of either `fs::path` or `json` types. The factory allows to disambiguate the signature between `std::string` and `fs::path`. Maybe there is also another way to do it.

adriendelsalle: I didn't needed the factory pattern, just added a new constructor

wolfv:
LGTM

PR #1131, opened 2021-08-17
downgrade libcurl 7.75.0

syslaila:
When bumping the version of mamba 0.9.0 to a version >=0.9.2 we would constantly get a 401 Unauthorised response; "Multi-download failed." when retrieving the repodata for some channels. Note: this only happens when we have multiple channels when solving an environment. It seems like the issue comes from the version of libcurl, mamba 0.9.0 depends on libcurl 7.75.0, whilst mamba 0.9.2 depends on 7.76.1. (there have been quite a few changes between these 2 versions: https://curl.se/changes.html)
I was able to reproduce the issue by creating an environment with mamba 0.9.0 and bumping the libcurl version to 7.76.1. 
```
load_channels(self.pool, self.channels, [], prepend=False)
File \"/tmp/solver/miniconda3/lib/python3.6/site-packages/mamba/utils.py\", line 99, in load_channels
    use_cache=use_cache,\n  File \"/tmp/solver/miniconda3/lib/python3.6/site-packages/mamba/utils.py\", line 74, in get_index
        is_downloaded = dlist.download(True)\nRuntimeError {exception_message}", 
        {"exception_message": "Multi-download failed."},  
```
the channels= list of internal conda channels of the format: [https://:<token>@<channel_url>/repodata.json]

From what I've seen we are getting a permission denied because the token (for some channels) is not passed to the request made from mamba through libcurl. We use the same token for all channel.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -45,7 +45,7 @@ jobs:
         run: |
           conda config --add channels conda-forge
           conda config --set channel_priority strict
-          conda install -n base python=$PYTHON_VERSION pip pybind11 libsolv libsodium libarchive "libcurl=7.76.1=*_0" nlohmann_json cpp-filesystem conda cxx-compiler ccache cmake gtest gmock reproc-cpp yaml-cpp
+          conda install -n base python=$PYTHON_VERSION pip pybind11 libsolv libsodium libarchive "libcurl=7.75.0=*_0" nlohmann_json cpp-filesystem conda cxx-compiler ccache cmake gtest gmock reproc-cpp yaml-cpp
         env:
           PYTHON_VERSION: ${{ matrix.python-version }}
       - name: Install dependencies
@@ -147,7 +147,7 @@ jobs:
           export MAMBA_ROOT_PREFIX=~/mambaroot
           export MAMBA_EXE=$(pwd)/micromamba
           . $MAMBA_ROOT_PREFIX/etc/profile.d/mamba.sh
-          micromamba create -y -p ~/build_env pybind11 libsolv libsodium libarchive "libcurl=7.76.1=*_0" nlohmann_json cxx-compiler ccache cmake gtest gmock cpp-filesystem reproc-cpp yaml-cpp cli11 -c conda-forge
+          micromamba create -y -p ~/build_env pybind11 libsolv libsodium libarchive "libcurl=7.75.0=*_0" nlohmann_json cxx-compiler ccache cmake gtest gmock cpp-filesystem reproc-cpp yaml-cpp cli11 -c conda-forge
         env:
           PYTHON_VERSION: ${{ matrix.python-version }}
       - name: build tests
@@ -208,7 +208,7 @@ jobs:
           export MAMBA_ROOT_PREFIX=~/mambaroot
           export MAMBA_EXE=$(pwd)/micromamba
           . $MAMBA_ROOT_PREFIX/etc/profile.d/mamba.sh
-          micromamba create -y -p ~/build_env pybind11 libsolv libsodium libarchive "libcurl=7.76.1=*_0" nlohmann_json cxx-compiler ccache cmake gtest gmock cpp-filesystem reproc-cpp yaml-cpp pyyaml cli11 -c conda-forge
+          micromamba create -y -p ~/build_env pybind11 libsolv libsodium libarchive "libcurl=7.75.0=*_0" nlohmann_json cxx-compiler ccache cmake gtest gmock cpp-filesystem reproc-cpp yaml-cpp pyyaml cli11 -c conda-forge
         env:
           PYTHON_VERSION: ${{ matrix.python-version }}
       - name: build micromamba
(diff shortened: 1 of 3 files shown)


wolfv:
thanks @syslaila I am not sure this is the right thing to do. Also this would only affect the test builds in the CI here while we'd need to change the "feedstock" on conda-forge (https://github.com/conda-forge/mamba-feedstock). Can you share the format of your basic auth?

syslaila:
Hey @wolfv 
The channel has the format "https://:<JWT token>@<url>"
We don't have a user set (it is empty) and we pass a JWT token 
We are getting 401 unauthorized because we expect the authorization token to be set, but it is not.
Also, this seems to happen when we have multiple channels - we are getting 200 responses from some of the channels, but for some we are getting a 401 response. We use the same JWT token for all the channels. 
If I only use one channel, then I always get a 200 response.

wolfv:
what kind of characters are in the JWT token? Anything out of the ordinary?

syslaila:
not really: we use the standard structure: "which is 3 base64-encoded strings combined with two .s"
from https://en.wikipedia.org/wiki/JSON_Web_Token#Structure something like this: 
`eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJsb2dnZWRJbkFzIjoiYWRtaW4iLCJpYXQiOjE0MjI3Nzk2Mzh9.gzSraSYS8EXBxLN_oWnFSRgCzcmJmMjLiuyu5CSpyHI`
We have letters, 2 . (dots) and a - and a _ in my specific case

syslaila:
Actually , one more find - I was able to reproduce it several times so it may be the actual issue ?
It seems I always get a 401 for the last channel I have in the list (changed the order a few times for the channels and was able to reproduce it)
eg: `mamba install -vv matplotlib transforms pandas numpy -c https://:<JWT_TOKEN>@<channel_1> -c https://:<JWT_TOKEN>@<channel_2> -c https://:<JWT_TOKEN>@<channel_3>`

channel_3 always gets the 401s
Do you have any idea why that could be ?

wolfv:
@syslaila would you be able to add tests to the channel parsing for your kind of tokens? In test_channel.cpp

That could be a great first step.e

wolfv:
https://github.com/mamba-org/mamba/blob/master/test/test_channel.cpp

wolfv:
Also I added some comment in the issue: https://github.com/mamba-org/mamba/issues/1132#issuecomment-900926665

syslaila:
update: I was able to run the command you posted in the issue with curl 7.76.1 (I had to install it from source https://curl.se/docs/install.html + configure it to work with ssl: ./configure --prefix=$HOME/curl --with-ssl=<path to openssl>) and I am getting random 401s when I pass multiple repodata.json as parameters:
`curl -vv -I https://:<TOKEN>@<channel_url1>/repodata.json https://:<TOKEN>@<channel_url2>/repodata.json https://:<TOKEN>@<channel_url3>/repodata.json`

response:
```
server: nginx
date: Wed, 18 Aug 2021 12:59:09 GMT
content-type: application/json
cache-control: max-age=0, must-revalidate
...

HTTP/2 401 
server: nginx
date: Wed, 18 Aug 2021 12:59:10 GMT
content-type: text/html;charset=UTF-8
...

HTTP/2 200 
server: nginx
date: Wed, 18 Aug 2021 12:59:10 GMT
content-type: application/json
...
```

I also got the generated c code from the curl call - let me know if I should attach it here - although I don't see any difference between the 200 request and the 401 request.

I would open an issue on curl side, what do you think ? Also, until the issue is unblocked, do you think we can downgrade to libcurl 7.75.0 ?

wolfv:
are you confident it's not an issue on the webserver side?

wolfv:
also you should eb able to regularly install curl `mamba install curl -c conda-forge` and then just use it :)

wolfv:
Hi @syslaila do you know wether the issue still persists with the latest curl 7.79.1?

If you have a bit of time, it would be great if you can check.

syslaila:
Apologies for the delay, I was OOTO, yes the issue persists when using mamba with the latest version of libcurl (7.79.1). 
The workaround I found was setting the context.max_parallel_downloads to 0. 
I am a little bit stuck on how to debug this further, I was unable to reproduce the issue directly via curl 7.79.1 (curl'ing multiple channel urls with the same settings that mamba sets for curl).

wulmer:
I have the same `Multi-download failed` issue with mamba 0.19.1. We are using credentials in the channel URLs for authenticating against an internal mirror and mamba always responds with 401 errors.

I just added some print-debug statements (sorry, I am no C++ guy) to the libmamba code to verify that the wrong URLs (without credentials) are passed to curl.  I also tried to set `max_parallel_downloads=1`; no improvement.
(BTW thanks a lot for the perfectly easy setup for building and testing libmamba/mamba. Took me just a few minutes to set up CMake and VS Code and everything compiled immediately! Good job!)

So my conclusion is that the credentials are not passed from mamba to libmamba which will then always result in 401s.
I'll further do my best to understand the mamba code and try to find the place where the credentials are removed and where they should be added again... but if anyone else could help me find the cause of this issue I'd be very glad.

wolfv:
You can also use -vvv to get lots of info from curl

wulmer:
Thanks, @wolfv , I already did this before, but wanted to be sure where those messages came from (from libmamba or from curl), that's why I edited the libmamba code.

wolfv:
Turns out this was a bug in curl, and it's fixed now thanks to @johnhany97 !

Issue #1123, opened 2021-08-11
Resolving from cache fails when cache dir rights are applied with ACL's


FeatCrush:
OS : RHEL7
Mamba version : 0.15.2
Conda version : 4.10.3

## How to replicate :

1. Create cache dir
```
mkdir -p /tmp/cache/conda
```
2. Add ACL's for your user
```
 setfacl -m d:u:<user>:rwx /tmp/cache/conda/
 setfacl -m u:<user>:rwx /tmp/cache/conda/
```
3. Change folder owner to root to rely solely on ACL's
```
chown root:root /tmp/cache/conda/
```

## Expected behavior 

`conda create -n test python`
The command works fine and the cache is populated with downloaded packages.

## Actual behavior
`mamba create -n test python`
Resolving breaks while accessing cache registry
```
RuntimeError: Operation not permitted: '/tmp/cache/conda/cache/f7c32d1b.json'
```

In our use case, package cache has to be accesible to mutiple groups.
We cannot rely solely on user:group rights.

wolfv:
Hmm, I haven't used ACL's myself. Any idea if we'd need to change the C++ code for this to work? I would have thought that a regular std::ofstream("myfile"); works fine...

Do you have the permission to create folders etc. in that cache location?

FeatCrush:
I was able to replicate the behavior in a centos7 docker container 
This is my new test process :
1.  Create two users (toto and titi in my case)
2. Create conda env with `toto` user 
3. Check that cache is populated (OK) 
4. Create same env with mamba, cache is hit (OK)
5. Switch to user `titi`
6. Create conda env with conda, cache is hit (OK)
7. Create same env with mamba (KO)

It seems that the cache files **have to be owned by the exec user**

ACL's for the problematic file 
```
# file: var/cache/conda/cache/28dc370b.json
# owner: toto
# group: toto
user::rwx
user:toto:rwx
user:titi:rwx
group::r-x
mask::rwx
other::r--
```

Stacktrace : 

```
# >>>>>>>>>>>>>>>>>>>>>> ERROR REPORT <<<<<<<<<<<<<<<<<<<<<<

    Traceback (most recent call last):
      File "/home/titi/.conda/envs/mamba/lib/python3.9/site-packages/conda/exceptions.py", line 1079, in __call__
        return func(*args, **kwargs)
      File "/home/titi/.conda/envs/mamba/lib/python3.9/site-packages/mamba/mamba.py", line 920, in exception_converter
[...]

PR #1137, opened 2021-08-21
feat(fs): Replace 'last_write_time' by underlying 'utimesat' to suppo…

FeatCrush:
#1123

### DIFF
--- a/include/mamba/core/fsutil.hpp
+++ b/include/mamba/core/fsutil.hpp
@@ -8,6 +8,8 @@
 #define MAMBA_CORE_FS_UTIL
 
 #include <string>
+#include <fcntl.h>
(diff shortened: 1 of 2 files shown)


wolfv:
Thanks for investigating this and proposing this fix. I am doubtful that this works on Windows. Also I think I'd prefer to wrap this in a function, and maybe only use it on Linux.

FeatCrush:
Hello @wolfv  i'll try to take some time to work on the issue

wolfv:
Ping @FeatCrush are you on this?

adriendelsalle:
@wolfv I'm not sure any change is required, please see https://github.com/mamba-org/mamba/issues/1123#issuecomment-959739126

PR #1140, opened 2021-08-24
Continue init when can't create prefix script dir

### LINKED ISSUE #1139: [micromamba] Init fails when prefix is not writable and profile.d is missing
I'm trying to run
```bash
micromamba shell init -vvv --shell=bash --prefix=/opt/conda
```
as a user without write permissions to `/opt/conda` when `/opt/conda/etc/profile.d` does not yet exist. This results in
```bash
ERROR   Permission denied: '/opt/conda/etc/profile.d'
```

I would expect that the init procedure should continue since writing to `profile.d` is unnecessary for activation.

Looking at the source, I suspect that the problem is with `fs::create_directories()`:

https://github.com/mamba-org/mamba/blob/81a490a046b1e1c0093c98f76803b75375e17628/src/core/shell_init.cpp#L404-L419

maresb:
Closes #1139

Please review carefully since my C++ skills are not the best.

### DIFF
--- a/src/core/shell_init.cpp
+++ b/src/core/shell_init.cpp
@@ -356,8 +356,15 @@ namespace mamba
     {
         fs::path exe = get_self_exe_path();
 
-        fs::create_directories(root_prefix / "condabin");
-        fs::create_directories(root_prefix / "Scripts");
+        try
+        {
+            fs::create_directories(root_prefix / "condabin");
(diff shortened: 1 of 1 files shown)


maresb:
Each failed run has 6 tests failing due to repo.mamba.pm being down.

PR #1149, opened 2021-09-02
expose methods for virtual packages functions

madhur-tandon:


src/mamba/py_interface.cpp
```diff
@@ -235,7 +236,8 @@ PYBIND11_MODULE(mamba_api, m)
     py::class_<PrefixData>(m, "PrefixData")
         .def(py::init<const std::string&>())
         .def_readwrite("package_records", &PrefixData::m_package_records)
-        .def("load", &PrefixData::load);
+        .def("load", &PrefixData::load)
+        .def("add_virtual_packages", &PrefixData::add_virtual_packages);
```
madhur-tandon, line 240: Not sure about this, my hunch is that something more needs to be done to expose member functions.

wolfv:
hmm, it looks fine. let me check locally

wolfv:
OK, I now realize that there is some interdependency that we should probably clear up for the virtual packages. I think it relies on `Config` from micromamba. 

@adriendelsalle could we add the necessary values to `context`?

src/mamba/py_interface.cpp
```diff
@@ -20,6 +20,8 @@
 #include "mamba/core/transaction.hpp"
 #include "mamba/core/url.hpp"
 #include "mamba/core/util.hpp"
+#include "mamba/core/virtual_packages.hpp"
+#include "../core/virtual_packages.cpp"
```
adriendelsalle, line 24: I think you need to remove that include and add the translation unit in the setup.py if you want to use it from mamba https://github.com/mamba-org/mamba/blob/fe4ecc5061a49c5b400fa7e7390b679e983e8456/setup.py#L121

src/mamba/py_interface.cpp
madhur-tandon: I think what I did works for member data, but not member functions

adriendelsalle (approved):
LGTM

PR #1152, opened 2021-09-06
Implements support for fish shell

### LINKED ISSUE #713: fish shell support
fish is referenced as one of the supported shells in the help text (`micromamba shell --help`) although there is no activator for it.

soraxas:
This PR implements all the necessary fish scripts to support fish shell in mamba (along with fish shell autocompletes).

This closes #713.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -274,11 +274,16 @@ jobs:
           then
             exit 1;
           fi
-      - name: install zsh and xonsh
+      - name: install zsh, xonsh and fish in linux
         if: matrix.os == 'ubuntu-latest'
         shell: bash -l -eo pipefail {0}
         run: |
-          sudo apt-get install zsh xonsh -y
+          sudo apt-get install zsh xonsh fish -y
+      - name: install fish in mac
+        if: matrix.os == 'macos-latest'
+        shell: bash -l -eo pipefail {0}
+        run: |
+          brew install fish
       - name: micromamba python based tests
         shell: bash -l -eo pipefail {0}
         run: |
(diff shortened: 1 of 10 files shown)


wolfv:
Hi @soraxas this is huge! Thanks!

I just wanted to note that this is only going to affect `micromamba` but I assume you are aware of this :)

soraxas:
Ahh yes indeed I do! 

That is also one thing that I want to mention, as currently some of the activation logics actually reside in `src/core` (rather than `src/micromamba`), but AFAIK it seems like `mamba` is not affected by them, only `micromamba`.

Does that means `mamba` would eventually implement more of `conda`'s interface, e.g., the `mamba init` command?

wolfv:
yeah, we haven't quite figured out where we're going with all this :) For the time being it could make sense to move these files to `micromamba`.

Maybe micromamba is going to replace `mamba` in the fuutre as it's quite nice to have a single static executable that doesn't depend on the "base env".

wolfv:
@soraxas it looks like the fish-activation tests are failing for micromamba. could you have a look? Do we need to install the `fish` shell somehow?

Also could you run `pre-commit run --all` to fix all linting issues? If not, I can do that for you.

soraxas:
Yep @wolfv thanks for the kind offer and instructions, I will fix the linting.

And yes it does looks like the failing test in `test_activation` is due to fish not being installed (which, for obvious reason, it didn't occur on my machine)
```sh
...
>               raise child_exception_type(errno_num, err_msg, err_filename)
E               FileNotFoundError: [Errno 2] No such file or directory: 'fish'
```
I will look into it to install fish as part of the github action.

wolfv:
awesome, thanks for your work on this!

PR #1153, opened 2021-09-06
Use beta.mamba.pm server

adriendelsalle:
Description
---

Fix CI by using `beta.mamba.pm` server


wolfv:
this is never going to work because the layout of the web backend requires us to either add `/get/...` which is ugly, or use `repo.mamba.pm`.

wolfv:
ie. the link woudl need to be `beta.mamba.pm/get/conda-forge` instead.

PR #1160, opened 2021-09-09
env creation with pip: do not change cwd

### LINKED ISSUE #1154: [micromamba] relative path in environment.yml not handled properly
Within a recipe I'm performing a development installation using pip like this:

The recipe resides within a subdirectory, e.g. the top level directory contains a setup.py
``` yaml
dependencies:
  - pip
  - pip:
    - -e ../
```

When I create a environment using this file, I receive the following error:
```
Installing pip packages: -e ../
2021-09-06T14:01:47.1013804Z ERROR: File "setup.py" or "setup.cfg" not found. Directory cannot be installed in editable mode: /home/runner/micromamba/envs
```
It seems that some process changed the working directory from the source/working directory to the envs dir of micromamba. But I could not figure out where this happens exactly. Note that the environment is being created by provision-with-micromamba. So there is no external, or explicit changing of directories.

Here you can have a look at the complete setup: https://github.com/marscher/weldx/blob/8b4213692ca81427deade4ebe6c55f558373672c/.github/workflows/docs.yml

marscher:
should fix #1154

### DIFF
--- a/src/api/install.cpp
+++ b/src/api/install.cpp
@@ -53,8 +53,6 @@ namespace mamba
 
         reproc::options options;
         options.redirect.parent = true;
-        std::string cwd = ctx.target_prefix;
-        options.working_directory = cwd.c_str();
 
         std::cout << "\n"
                   << termcolor::cyan << "Installing " << pkg_mgr
--- a/src/api/install.cpp
+++ b/src/api/install.cpp
@@ -27,12 +27,15 @@
 
 namespace mamba
 {
-    static std::vector<std::tuple<std::string, std::vector<std::string>>> other_pkg_mgr_specs;
     static std::map<std::string, std::string> other_pkg_mgr_install_instructions
         = { { "pip", "pip install -r {0} --no-input" } };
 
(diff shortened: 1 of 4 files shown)


src/api/install.cpp
```diff
@@ -53,8 +53,6 @@ namespace mamba
 
         reproc::options options;
         options.redirect.parent = true;
-        std::string cwd = ctx.target_prefix;
-        options.working_directory = cwd.c_str();
```
wolfv, line 56: I think it would still be better to explicitly set the current work directory (as they do in the conda code I linked).

marscher: Right, currently I have a hard time finding a good spot to set the cwd option. We need to set it in the config instance, when the environment.yaml file is being parsed.

adriendelsalle:
I proposed to:
- use configuration to store other pkg mgrs specs (and remove the static variable `other_pkg_mgr_specs` for storage)
- add an `other_pkg_mgr_spec` structure to store those specs
- store the directory of the yaml file as the cwd for subprocess call
- update and extend tests

wolfv:
LGTM!

marscher:
Thanks for the proper implementation @adriendelsalle

PR #1164, opened 2021-09-14
make mamba support "mamba update --all one_package==x.x.x" like conda

zhangfeiran:
This is useful to update all packages while excluding one (several) specific package(s).

### DIFF
--- a/mamba/mamba.py
+++ b/mamba/mamba.py
@@ -442,15 +442,9 @@ def install(args, parser, command="install"):
         prefix_data = PrefixData(prefix)
         for s in args_packages:
             s = MatchSpec(s)
-            if s.name == "python":
-                specs.append(s)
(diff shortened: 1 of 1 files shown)


jjerphan:
Hello,

Unfortunately the code-base of mamba has drastically changed since `1.x` and this PR is not applicable anymore. Also I do not think we want to support all the features of conda.

Hoeze:
> Also I do not think we want to support all the features of conda.

@jjerphan While I agree in general, this specific feature is quite annoying not to have.

zhangfeiran:
I'm actually ok without this feature now, since adding package to `myenv/conda-meta/pinned` file is a more persistent way.

jjerphan:
I invite you to continue the discussions on this issue https://github.com/mamba-org/mamba/issues/1231, specifying your needs.

PR #1168, opened 2021-09-16
Improve outputs

adriendelsalle:
Description
---

update `termcolor` header to fix colorized static variable
make `ConsoleStream` colorized in constructor
add initial support for emojis
add warning emoji when using experimental mode

### DIFF
--- a/src/core/output.cpp
+++ b/src/core/output.cpp
@@ -10,17 +10,19 @@
 #include <sys/ioctl.h>
 #endif
 
-#include <algorithm>
-#include <cstdlib>
-#include <iostream>
-#include <regex>
-
 #include "mamba/core/output.hpp"
 #include "mamba/core/thread_utils.hpp"
 #include "mamba/core/url.hpp"
 #include "mamba/core/util.hpp"
 
 #include "thirdparty/termcolor.hpp"
+
+#include <algorithm>
+#include <cstdlib>
+#include <iostream>
+#include <map>
+#include <regex>
+#include <string>
 
 namespace mamba
 {
@@ -279,6 +281,10 @@ namespace mamba
     /*****************
      * ConsoleStream *
      *****************/
+    ConsoleStream::ConsoleStream()
+    {
+        termcolor::colorize(*this);
+    }
 
     ConsoleStream::~ConsoleStream()
     {
--- a/include/mamba/core/output.hpp
+++ b/include/mamba/core/output.hpp
@@ -209,7 +209,7 @@ namespace mamba
     class ConsoleStream : public std::stringstream
     {
     public:
-        ConsoleStream() = default;
+        ConsoleStream();
         ~ConsoleStream();
     };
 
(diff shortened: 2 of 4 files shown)


wolfv:
Hmm, I wonder if we need the `emojize` function or if typing an emoji directly "just works" :tm:

Could lead to problems on Windows :)

adriendelsalle:
I need to try on Windows.

>  or if typing an emoji directly "just works" tm

what do you mean? I only introduced a simple mapping of names and unicode emojis

wolfv:
I mean literally inserting `std::string = "✅"` in the source code.

adriendelsalle:
What string would you like to pass?  something like `std::string check = ":warning:"`?
Sorry I don't see really what you describe

wolfv:
I mean you can _literally_ write `std::cout << "🎄" << std::endl;` (instead of looking it up in a `std::map`).

The other option would be to make a `#define christmastree 🎄` and use that.

adriendelsalle:
Yeah thanks I figured out that.
I don't expect that to make a difference since it's only copy-pasting some UTF-8 encoded string rendered in almost all text editors or terminals. It's just more explicit for developers to see what's the emoji.

The most painful is that terminals are rendering emojis with variable widths. In bash it varies from 1char width (copyright) to about 3chars. Which makes pretty not easy to have nice alignments in tables or text.
I'll work on this further.

adriendelsalle:
@wolfv I tested on Windows and both are not well supported, even in pwsh.
I'll drop this idea for now, thanks for your ideas!

wolfv:
sorry to hear that! :/ could be interesting to have a look at how other package managers / fancy javascript tools do it.

for powershell this seems to work? https://www.powershellcenter.com/2021/02/19/add-emojis-to-powershell-output/

wolfv:
Maybe you need to run `Windows Terminal` & not cmd.exe?

adriendelsalle:
I also tested `Windows Terminal` but I didn't managed to make it work..
I think it's better to open an issue and see later if we find something

src/core/output.cpp
```diff
@@ -279,6 +281,10 @@ namespace mamba
     /*****************
      * ConsoleStream *
      *****************/
+    ConsoleStream::ConsoleStream()
+    {
+        termcolor::colorize(*this);
```
wolfv, line 286: Maybe we should only do that if `stdout` `isatty` is true?

adriendelsalle: > Maybe we should only do that if `stdout` `isatty` is true?

I implemented this because `ConsoleStream` destructor calls `Console::instance().print(str())` which is *always* printing to `stdout`, but the intermediate stream will not be detected as a tty (because it's not), preventing `termcolor` to add ansi colors

> There is also the env var CLICOLOR_FORCE=1 and if it's set we should always force colors :)

do you want to add some `CLICOLOR_FORCE` flag in `libmamba`?

PR #1174, opened 2021-09-20
Parallel extraction

jonashaag:
This PR is meant as a sketch for an idea that I'd like to suggest.

Package decompression (e.g. bz2) can take a lot of time. In some scenarios (eg. if you have fast network and fast disk) it can even take most of the time of a `mamba install` run.

Unfortunately libarchive does not support threading, which is why decompression if currently single-threaded behind a mutex, IIUC.

As a workaround we could simply do decompression in a subprocess. I've implemented a sketch here by just calling into `tar`, but we should probably implement it in the mamba codebase using libarchive, eg. using a "secret" mamba invocation such as `mamba _untar foo.tar.bz2`.

With the changes in this PR I've had install speedups of 30–80%. As an example, installing the mamba dev environment on Apple M1 drops from 15 s to 10 s.

Curious to hear your thoughts.

### DIFF
--- a/src/core/package_handling.cpp
+++ b/src/core/package_handling.cpp
@@ -7,6 +7,8 @@
 
 #include <archive.h>
 #include <archive_entry.h>
+#include <sys/wait.h>
+#include <unistd.h>
 
 #include <sstream>
 
@@ -52,33 +54,6 @@ namespace mamba
     private:
         const fs::path& m_file;
     };
-
-    static int copy_data(archive* ar, archive* aw)
-    {
-        int r;
-        const void* buff;
-        std::size_t size;
-        la_int64_t offset;
-
-        while (true && !is_sig_interrupted())
-        {
-            r = archive_read_data_block(ar, &buff, &size, &offset);
-            if (r == ARCHIVE_EOF)
-            {
-                return ARCHIVE_OK;
-            }
-            if (r < ARCHIVE_OK)
-            {
-                throw std::runtime_error(archive_error_string(ar));
-            }
-            r = archive_write_data_block(aw, buff, size, offset);
-            if (r < ARCHIVE_OK)
-            {
-                throw std::runtime_error(archive_error_string(aw));
-            }
-        }
-        return r;
-    }
 
     // Bundle up all files in directory and create destination archive
     void create_archive(const fs::path& directory,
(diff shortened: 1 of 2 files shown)


wolfv:
Hi Jonas,

this is great and very much in line with the ideas that I had! Perfect! 

What my plan was is to add a new subcommand to `mamba` (and `micromamba`), e.g. `micromamba extract mypkg.tar.bz` (or `mypkg.conda`). Maybe we should also give some additional arguments like the target directory or so. And then we just call `mamba extract ...` or `micromamba extract ...`.

The benefit would be that we can keep our current implementation for extracting packages which is somewhat "battle tested" and supports both formats.

Other possibilities that I've partially explored:

- use `fork()` to spawn new processes --> this is a similiar idea to subprocesses and doesn't work well on Windows IIRC so subprocesses is probably easier & preferred
- use a proper C++-implementation for extraction (basically what would be great is to use boost iostreams "filters" https://www.boost.org/doc/libs/1_77_0/libs/iostreams/doc/classes/bzip2.html but ... afaik they don't have a `zip` and `zstd` filter necessary for the new `.conda` format and also boost is a huge intertwined dependency that I don't really want)
- figure out if we can get `libarchive` to extract with multiple threads concurrently -- I attempted this at least two times and always got weird segfaults or badly extracted data -- I might have done some things wrong but basically couldn't ever get it to work and haven't found any good docs


So yeah, all things considered I think this is the best approach!

wolfv:
PS: your time measurements are very encouraging! We should also figure out how to only start N or N-1 processes at the same time, with N being the number of available processors.

wolfv:
Just as another datapoint regarding streams (and as a mental note for myself) it looks like a bz2 / zip / zstd stream might not be incredible hard to do:

https://github.com/headupinclouds/bzip2stream/blob/master/bzip2stream/bzip2stream.hpp

But I still think the subprocess option is better for now.

jonashaag:
Thanks for the detailed feedback!

I started working on a proper implementation today and came across the following problem. How are we going to figure out the right `mamba` command to run?

For Micromamba it is pretty straightforward because it's a single binary so in `package_handling.cpp` we could just find the current invocation path (eg. `myenv/bin/mamba`) and run another copy of that.

For Mamba it is more involved because Python invocations tend to be fairly complex, for example the `python` executable may be in a Conda environment and may not work correctly unless you enter that particular environment before running `python`. Or someone could run Mamba with something like `mamba run -n xxx mamba` or even `mamba run python -c 'import mamba; mamba.mamba.main()'`.

Besides, I'm concerned that starting lots of `python` interpreters (one for each package) may be very slow, although I haven't done any benchmarking.

Maybe it is easier to use `fork()` in `package_handling.cpp`; I've tried that before opening this PR and got lots of weird crashes, but maybe we can make it work. For Windows we could just not do parallelisation if it's unstable.

Another option would be to use `tar xzf` like in this PR.

Yet another option would be to switch to thread safe decompression libraries.

wolfv:
How about we create another small executable called `mamba-extractor` that's pure C++ but links dynamically to the necessary libraries? We can ship that with `mamba`. 

For micromamba, we could just call micromamba itself.

If you want, you could start with micromamba by just adding a new subcommand (`extract`, or `extract-package`) and later I can take care of adding a standalone executable for use in `mamba`.

wolfv:
For micromamba, it's easy to find the "self exe" using `get_self_exe_path()` by the way :)

wolfv:
@jonashaag do you want to coordinate how to proceed here? I would be happy to help out -- e.g. by writing the simple `mamba-extract` binary / subcommand as described above. But I also want to avoid duplicate work so please let me know how things are going :)

jonashaag:
I haven’t had the time to work on this in the past days. Feel free to start working on anything.

Issue #1169, opened 2021-09-17
Mamba send request with empty user-agent and cause some mirror site refuse to provide files


TTTPOB:
mamba request header looks like 

![image](https://user-images.githubusercontent.com/19909103/133821645-5b760404-fcf5-48b2-9089-455c0bcb5799.png)

and conda request header looks like 
![image](https://user-images.githubusercontent.com/19909103/133822515-3a51a925-1de4-4db4-af9f-7cd9b86c91d1.png)

instead of providing the file, the mirror site just recognize mamba as mobile/spam device and returns
```html
<!DOCTYPE html>
<html>
<head>
  <meta charset="utf-8">
  <meta http-equiv="X-UA-Compatible" content="IE=edge">
  <meta name="viewport" content="width=device-width, initial-scale=1">
  <title>清华大学 TUNA 镜像站</title>
  <link href="//cdn.jsdelivr.net/npm/bootstrap@3.3.7/dist/css/bootstrap.min.css" rel="stylesheet">
  <style>body { padding-top: 40px; }</style>
</head>
<body>
  <div class="container">
    <div class="well">
      <p lang="zh-cn">尊敬的访问者，</p>
      <p lang="en">Dear visitor,</p>
      <p lang="fr">Chère cliente,</p>
      <p lang="ja">訪問者へ 、</p>
      <p>&nbsp;</p>
      <p lang="zh-cn">您好！</p>
      <p lang="en">Hello!</p>
      <p lang="fr">Bonjour!</p>
      <p lang="ja">こんにちは ！</p>
      <p>&nbsp;</p>
      <p lang="zh-cn">我们检测到您正在使用移动设备下载镜像站上较大的二进制文件，为防止消耗过多流量，我们建议您改用计算机下载。</p>
      <p lang="en">Your attempt to download a large binary file from a mobile device has been detected. To avoid excessive consumption of data transfer, you are recommended to download it from a computer.</p>
      <p lang="fr">Une tentative de téléchargement d'un fichier binaire de grande taille depuis un appareil mobile a été détectée. Afin d'éviter une consommation excessive de données mobile, nous vous conseillons de recommencer cette action à partir d'un ordinateur.</p>
      <p lang="ja">容量の大きいバイナリファイルが今モバイルデバイスに転送しようとしております、大量のトラフィックが発生するため、コンピュータからのダウンロードをお勧めします。</p>
      <p>&nbsp;</p>
      <p lang="zh-cn">
        如有疑问，请联系
[...]

PR #1176, opened 2021-09-22
Add User-Agent when making http requests

shankerwangmiao:
Fixes: #1169

### DIFF
--- a/src/core/fetch.cpp
+++ b/src/core/fetch.cpp
@@ -13,6 +13,7 @@
 #include "mamba/core/thread_utils.hpp"
 #include "mamba/core/util.hpp"
 #include "mamba/core/url.hpp"
+#include "mamba/core/version.hpp"
 
 namespace mamba
 {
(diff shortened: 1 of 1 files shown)


src/core/fetch.cpp
```diff
@@ -226,6 +227,9 @@ namespace mamba
                 m_handle, CURLOPT_ACCEPT_ENCODING, "gzip, deflate, compress, identity");
             m_headers = curl_slist_append(m_headers, "Content-Type: application/json");
         }
+        m_headers = curl_slist_append(m_headers,
+                                      "User-Agent: Mamba/" MAMBA_VERSION_STRING
```
wolfv, line 231: Would you mind using `mamba/` (lowercase?)

Also Not sure how commmon it is to add a link to the website to the user agent. I think we can remove that. It would be cool though to add the cURL version using

`curl_version_info(CURLVERSION_NOW).version` (that returns a `char*`).

wolfv:
Thanks for working on this!

PR #1186, opened 2021-09-29
Add missing space at the end of printed text

wkusnierczyk:
The text to be printed to the terminal is merged from two separate strings.
Neither the first string ends, nor the second one starts with whitespace, hence the tail word of the first line merges with the head word of the second line, giving `closeand`:

```
==> For changes to take effect, closeand re-open your current shell. <==
```


wolfv:
Perfect, thanks!

Issue #612, opened 2020-11-26
micromamba and caching pkgs-folder


kmuehlbauer:
For work with CI it is often desirable to cache the `pkgs` folder. Normally only the archives (tar.bz2) are cached.

When doing this and trying to create a new environment micromamba breaks. Packages resolving works nice, it also detects that no package needs to be downloaded, but then it breaks with the below error.

```
Transaction starting
INFO    transaction.cpp:379 Found python version in packages to be installed 3.8.6
Linking libgfortran5-9.3.0-he4bcb1c_17
INFO    link.cpp:917 Executing install for "/home/data/kai/projects/micromamba_test/pkgs/libgfortran5-9.3.0-he4bcb1c_17"
WARNING link.cpp:919 Opening: "/home/data/kai/projects/micromamba_test/pkgs/libgfortran5-9.3.0-he4bcb1c_17/info/paths.json"
terminate called after throwing an instance of 'std::system_error'
  what():  failed to open /home/data/kai/projects/micromamba_test/pkgs/libgfortran5-9.3.0-he4bcb1c_17/info/files: No such file or directory
Abgebrochen (Speicherabzug geschrieben)
```
Copy from gitter chat @wolfv:

> in mamba, we actually rely on conda to detect this ...
> since we're not using conda with micromamba, we need to launch an unpack process if it isn't yet unpacked
> we could also do more validation of the unpacked contents (which we are also not doing right now), such as verifying file size and MD5 sum of the contents vs. what's expected from the package
> obviously PR's welcome :)

As a workaround for caching one can remove the tar.bz2 archives before cache creation. This will save at least some space and time for now.

adriendelsalle:
Hi!

I can confirm that in others situations such as a process interruption when downloading and unpacking, when you will try to resume by launching again micromamba the same errors occurs: status of packages are `cached` and the linking step will just fail silently in my case.

I will add more details to reproduce soon.

wolfv:
We've been making micromamba much more robust against these sorts of failures in the past couple of weeks, so I am confident that this is not an issue anymore.

PR #1187, opened 2021-09-29
Extract package subcommand

wolfv:
@jonashaag for reference, this is a simple command implemented as `micromamba package extract myarchive.{tar.bz2,conda}`.

### DIFF
--- a/src/micromamba/package.cpp
+++ b/src/micromamba/package.cpp
@@ -1,0 +1,67 @@
+// Copyright (c) 2019, QuantStack and Mamba Contributors
+//
+// Distributed under the terms of the BSD 3-Clause License.
+//
+// The full license is in the file LICENSE, distributed with this software.
+
+#include "common_options.hpp"
+
+#include "mamba/core/util.hpp"
+#include "mamba/core/package_handling.hpp"
+
+using namespace mamba;  // NOLINT(build/namespaces)
+
+void
+set_package_command(CLI::App* subcom)
+{
+    static std::string infile, dest;
+    static int compression_level = -1;
+
+    auto extract_subcom = subcom->add_subcommand("extract");
+    extract_subcom->add_option("archive", infile, "Archive to extract");
+    extract_subcom->callback([&]() {
+        std::cout << "Extracting " << fs::absolute(infile) << std::endl;
+        extract(fs::absolute(infile));
+    });
+
+    auto compress_subcom = subcom->add_subcommand("compress");
+    compress_subcom->add_option("folder", infile, "Folder to compress");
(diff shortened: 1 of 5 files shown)


jonashaag:
Thanks for tagging me!

I'm not sure if `mamba::extract` is the right level of abstraction that a subcommand should be calling into:

- It disables the `extraction_guard` (although I'm not entirely sure what the guard protects)
- It doesn't make too much sense for `.conda` packages IMO, because those are likely to be very quick to extract

In my WIP work (that I didn't push to GitHub so far) I have extracted the body of `extract_archive` to a `extract(archive_file, dest_folder)` routine. The subcommand then calls that one directly, and `extract_archive` calls into the subcommand (after acquiring the guard, if necessary).

wolfv:
`extract(archive_file, dest_folder)` this is coming in #612 as well :)

I think it's still worth to do this for `.conda` files as they might still be big (e.g. if you have `cudatoolkit` and `pytorch` etc.

Locking could be done at the level of the subprocess calling. Or in here. Need to think about it – it's a good point! :)

I am also adding functions to create a archive from a folder and to `transmute` (convert from `.tar.bz2` to `.conda`) a package.

src/micromamba/package.cpp
```diff
@@ -0,0 +1,55 @@
...
+        std::cout << "Extracting " << fs::absolute(infile) << std::endl;
+        extract(fs::absolute(infile));
+    });
+
+    auto compress_subcom = subcom->add_subcommand("compress");
+    compress_subcom->add_option("folder", infile, "Folder to compress");
+    compress_subcom->add_option("dest", dest, "Destination (e.g. myfile-3.1-0.tar.bz2 or .conda)");
+    compress_subcom->add_option("-c,--compression-level",
+                                compression_level,
+                                "Compression level from 0 to 50 (default is 10)");
```
jonashaag, line 32: Where does that range come from? In the command-line utilities the compression range for bzip2, gzip, zip is 0-9 and the range for zstd is 0-19 IIRC (plus some extra strong compression options that are hidden behind a flag)

wolfv: The library supports regular compression levels from 1 up to ZSTD_maxCLevel(),
  which is currently 22. Levels >= 20, labeled `--ultra`, should be used with
  caution, as they require more memory. The library also offers negative
  compression levels, which extend the range of speed vs. ratio preferences.
  The lower the level, the faster the speed (at the cost of compression).

wolfv: The above is for zstd. I am actually not sure if compression level is implemented for `tar.bz2` :) let me check

wolfv: you're right wrt to bzip2 -- that supports between 0-9 -- I added it as an option.

PR #1193, opened 2021-09-30
Improve lockfiles

adriendelsalle:
Description
---

- Handle files and directories locks
  - API is now `Lock` and takes the path *to* lock instead of the name of the supposed lock file
    - this allows to disambiguate a file path between the lock file or the file to be locked
  - `LockFile` is the holder of the file descriptor
    - it allows throwing in `Lock` constructor with a clean handling of the file descriptor (e.g. closing it)
- Handle multiple locks acquired on the same path
  - from the same process -> raise a `std::logic_error`
  - from the another process: try lock in a blocking way and raise a `std::runtime_error` is a timeout is set
    - on Windows, implement a timeout calling [_locking](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/locking?view=msvc-160) with `_LK_LOCK` every seconds
      - which is similar to `_LK_NBLOCK` but with control over the timeout
      - allows infinite loops as on Unix systems
    - on Unix, call `fcntl` with `F_SETLKW` in a separate thread and cancel it if timeout is met
    - we could change the default behavior to avoid infinite loops later (setting the timeout to some large number)
- Improve logs
- Add tests
  - add another testing CMake target `mamba_test_lock` to have a CLI for testing `is_locked` and `lock` paths for subprocesses
  - check creation of one or multiple locks on the same path from both current and foreign processes
- Add `micromamba` CLI options/flags
  - `-l,--locks` to `micromamba clean` to remove lock files in caches
  - `--lock-timeout` to set the lock file timeout

### DIFF
--- a/src/core/util.cpp
+++ b/src/core/util.cpp
@@ -4,15 +4,16 @@
 //
 // The full license is in the file LICENSE, distributed with this software.
 
-#include <cerrno>
-#include <iomanip>
-#include <iostream>
-#include <mutex>
-#include <thread>
 
 #if defined(__APPLE__) || defined(__linux__)
-#include <csignal>
+#include <stdio.h>
+#include <sys/types.h>
+#include <sys/stat.h>
 #include <unistd.h>
+#include <fcntl.h>
+#include <pthread.h>
+
+#include <csignal>
 #include <sys/file.h>
 #endif
 
@@ -30,6 +31,13 @@ extern "C"
 }
 
 #endif
+
+#include <cerrno>
+#include <iomanip>
+#include <iostream>
+#include <mutex>
+#include <thread>
+#include <condition_variable>
 
 #include "mamba/core/context.hpp"
 #include "mamba/core/util.hpp"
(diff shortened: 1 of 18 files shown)


src/core/util.cpp
```diff
@@ -754,31 +754,47 @@ namespace mamba
...
+            LOG_DEBUG << "Could not lock non existing path '" << path.string() << "'";
+            m_fd = -1;
+            return;
+        }
 
-        m_fd = open(path.c_str(), O_RDWR | O_CREAT, 0666);
+        LOG_DEBUG << "Locking '" << path.string() << "'";
+        if (fs::is_directory(path))
+            m_lock = m_path / "mamba.lock";
+        else
```
wolfv, line 767: does this really work? because we're checking the existence of the file before hand ... but then we're also trying to create the file in the next step.

The logic seems not entirely correct.

src/core/util.cpp
```diff
@@ -754,31 +754,47 @@ namespace mamba
...
         else
         {
+            LOG_TRACE << "Lockfile created at '" << m_lock.string() << "'";
             m_pid = getpid();
             int ret = try_lock(m_fd, m_pid, false);
             if (ret != m_pid)
             {
-                LOG_WARNING << "Cannot lock file " << path
-                            << "\nWaiting for other mamba process to finish.\n";
+                LOG_WARNING << "Cannot lock '" << m_path << "'"
```
wolfv, line 783: usually the quoting is done automatically by `fs::path`.

src/core/util.cpp
```diff
@@ -754,31 +754,47 @@ namespace mamba
     LockFile::LockFile(const fs::path& path)
         : m_path(path)
     {
-        LOG_INFO << "Locking " << path;
+        if (!fs::exists(path))
```
wolfv, line 757: should we only check existence if the it's a directory, or check if the parent directory exists?

src/core/util.cpp
adriendelsalle: I just try to have homogeneous logs, using the same syntax everywhere

src/core/util.cpp
adriendelsalle: don't totally get it, I'll have a look

src/core/util.cpp
adriendelsalle: I don't really see. Here I want to check existence for both directory or file

src/core/util.cpp
wolfv: if `m_path` is a `fs::path` then the `operator<<` calls `std::quoted`. So you don't need to add `'`

src/core/util.cpp
```diff
@@ -643,142 +651,363 @@ namespace mamba
...
+
+        std::error_code ec;
+        LOG_TRACE << "Removing file '" << m_path.string() << "'";
+        fs::remove(m_path, ec);
+
+        if (ec)
         {
-            ret = _locking(fd, LK_NBLCK, 1 /*lock_file_contents_length()*/);
+            LOG_ERROR << "Removing lock file '" << m_path.string() << "' failed\n"
+                      << "You may need to remove it later";
```
wolfv, line 678: remove it manually would sound better, no?

src/core/util.cpp
```diff
@@ -643,142 +651,363 @@ namespace mamba
...
+        close_fd();
+
+        std::error_code ec;
+        LOG_TRACE << "Removing file '" << m_path.string() << "'";
+        fs::remove(m_path, ec);
+
+        if (ec)
         {
-            ret = _locking(fd, LK_NBLCK, 1 /*lock_file_contents_length()*/);
+            LOG_ERROR << "Removing lock file '" << m_path.string() << "' failed\n"
```
wolfv, line 677: Just printing path should automatically `quote` it. Don't need to call `string` on it.

src/core/util.cpp
adriendelsalle: yep

src/core/util.cpp
adriendelsalle: yeah I saw your comment, I would prefer to follow what's done everywhere else, which is currently printing `'/my/path'` rather than having both `'/my/path'` and `"/my/path"`

src/core/util.cpp
```diff
@@ -643,142 +651,363 @@ namespace mamba
...
+            m_fd = -1;
         }
-        if (ret == 0)
+        return ret;
+    }
+
+    bool LockFile::unlock() const
+    {
+        int ret = 0;
+
```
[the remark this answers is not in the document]
wolfv: but unlock here doesn't close the file descriptor, does it?

test/testing/lock.cpp
```diff
@@ -27,7 +27,7 @@ is_locked(const fs::path& path)
     // From another process than the lockfile one, we can open/close
     // a new file descriptor without risk to clear locks
     int fd = open(path.c_str(), O_RDWR | O_CREAT, 0666);
-    bool is_locked = mamba::LockFile::is_locked(fd);
+    bool is_locked = mamba::Lock::is_locked(fd);
```
wolfv, line 30: I would like to keep the LockFile name -- because that's what it is. A "Lock" is a bit ambigous.

PR #1195, opened 2021-10-01
Extract packages in subproc

jonashaag:
Based on the work in https://github.com/mamba-org/mamba/pull/1187. Refs https://github.com/mamba-org/mamba/pull/1174.

For some reason, on my machine reproc always reports an exit code of 32 (Broken pipe), but the child processes run just fine, which is why the code currently skips error checking. I have spent quite some time debugging this but have not succeeded in identifying the root cause. Things I've tried: Replacing `mamba` with `echo`; putting a mutex around `extract_subproc`; removing any output from the `extract` subcommand; others.


wolfv:
Cool! Will give it a try. 
The error code ... I'll need to look into it. It didn't occur to me before -- maybe we're not returning properly from `main`?

jonashaag:
Interesting, in a Docker Ubuntu (arm64) the problem doesn't occur.

wolfv:
On an M1 macOS PC it was fine as well ... 

Let's check for the exit code and see if the CI complains.

src/core/package_handling.cpp
```diff
@@ -391,26 +394,52 @@ namespace mamba
...
     }
 
+    fs::path extract_subproc(const fs::path& file)
+    {
+        std::vector<std::string> args = { get_self_exe_path(), "package", "extract", file };
+        std::string out, err;
+        auto [status, ec] = reproc::run(
+            args, reproc::options{}, reproc::sink::string(out), reproc::sink::string(err));
+        /*
+        LOG_INFO << "Exit status " << args[0] << " " << args[1] << " " << args[2] << " " << args[3] << " " << status << " " << ec.value() << " " << ec.message() << " stdout: " << out << "stderr: " << err;
```
wolfv, line 434: Keeping in some LOG_INFO might be good (to help with debugging)

src/core/package_handling.cpp
```diff
@@ -391,26 +394,52 @@ namespace mamba
...
+        if (ec)
```
wolfv, line 435: I think we should keep this and see what the CI says.

src/core/package_handling.cpp
```diff
@@ -391,26 +394,52 @@ namespace mamba
...
         return dest_dir;
     }
 
+    fs::path extract_subproc(const fs::path& file)
+    {
+        std::vector<std::string> args = { get_self_exe_path(), "package", "extract", file };
+        std::string out, err;
+        auto [status, ec] = reproc::run(
+            args, reproc::options{}, reproc::sink::string(out), reproc::sink::string(err));
+        /*
```
jonashaag, line 433:
```suggestion
```

src/core/package_handling.cpp
```diff
@@ -391,26 +394,52 @@ namespace mamba
...
+        LOG_INFO << "Exit status " << args[0] << " " << args[1] << " " << args[2] << " " << args[3] << " " << status << " " << ec.value() << " " << ec.message() << " stdout: " << out << "stderr: " << err;
+        if (ec)
+        {
+            throw std::runtime_error("Error extracting " + file.string() + " with 'mamba extract': " + err);
+        }
+        */
```
jonashaag, line 439:
```suggestion
```

jonashaag:
Calling into the subprocess from mamba (not micromamba) currently fails because `get_self_exe_path()` returns `python`. What would be the best way to deal with this?

wolfv:
@jonashaag I think the best way will be to create & install an additional executable (`mamba-extractor`) that is pure-c++ (no python whatsoever) and contains a slimmed down version of the `micromamba package extract` command, only using `argv` etc.

We can install that as a regular binary (with dynamic linking) and drop it in `condabin/mamba-extractor` and it should be found by [`shutil.which`](https://docs.python.org/3/library/shutil.html#shutil.which) or other means (as `$PATH` has to be correctly set anyways for the dynamically linked mamba to function).

But maybe it's easier to start with `micromamba` and move on to `mamba` once we've validated this works well? Did you already implement a limit of parallel extraction processes? I think we should probably limit the number to either `CPU_COUNT` or `CPU_COUNT - 1`?

jonashaag:
We'll have to deal with `mamba` because it uses the same code as `micromamba`. That's also why the tests fail right now.

Can you give some background why we don't want to have `mamba` call into `micromamba` and instead have another binary?

wolfv:
So micromamba and mamba do share a lot of code but there is also a compile time flag that we could use to `#ifdef UMAMBA_STATIC`.

The builds are different in teh following ways:

- mamba links dynamically to `libarchive`, `libcurl`, etc...
- micromamba links statically (single file executable)

The `micromamba` binary is ~12 Mb unpacked, and we do not ship it as part of `mamba`. If we would ship it with `mamba` it would increase the size of the `mamba` package quite a bit.

In the future we might make `micromamba` to `mamba` and have a lesser featured `micromamba` be the new `micromamba`. 

So for now we could keep the old code for `mamba` and only call into this when `#ifdef UMAMBA_STATIC` is true. Or actually we can / should add another flag to our CMake build system because actually I am not sure if we compile with static dependencies in the CI. 

Does this explanation make sense?

wolfv:
Hi @jonashaag I am having a look now. I am going to rebase the PR if you don't mind.

PR #1208, opened 2021-10-12
improve tarball and directory checks if no size known

wolfv:
This changes two things:

- we ignore a file size of `0` coming from the used `PkgInfo` as that indicates an _unknown_ value. This can happen if we use an explicit environment spec where a URL each package entry is of the form: `https://conda.anaconda.org/conda-forge/linux-64/_libgcc_mutex-0.1-conda_forge.tar.bz2#d7c89558ba9fa0495403155b64376d81`. In this case we _do know_ the MD5 sum of the tar ball, but not the file size
- However, if we're running from `constructor` we add the file size to the `repodata_record.json` by getting it from the disk (using fs::file_size). This will improve cache re-usability and we still validate using the supplied MD5 sum.

### DIFF
--- a/src/core/package_cache.cpp
+++ b/src/core/package_cache.cpp
@@ -129,7 +129,13 @@ namespace mamba
             }
             else
             {
-                if (Context::instance().safety_checks == VerificationLevel::kEnabled)
+                if (Context::instance().safety_checks == VerificationLevel::kWarn)
+                {
+                    LOG_WARNING << "Could not validate package '" + tarball_path.string()
+                                       + "': md5 and sha256 sum unknown.\n"
+                                         "Set safety_checks to disabled to override this warning.";
+                }
+                else if (Context::instance().safety_checks == VerificationLevel::kEnabled)
                 {
                     // we cannot validate this archive, but we could also not validate a downloaded
                     // archive since we just don't know the sha256 or md5 sum
@@ -137,12 +143,6 @@ namespace mamba
                         "Could not validate package '" + tarball_path.string()
                         + "': md5 and sha256 sum unknown.\n"
                           "Set safety_checks to warn or disabled to override this error.");
-                }
-                if (Context::instance().safety_checks == VerificationLevel::kWarn)
-                {
-                    LOG_WARNING << "Could not validate package '" + tarball_path.string()
-                                       + "': md5 and sha256 sum unknown.\n"
-                                         "Set safety_checks to disabled to override this warning.";
                 }
             }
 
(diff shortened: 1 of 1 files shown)


src/core/package_cache.cpp
```diff
@@ -117,8 +117,9 @@ namespace mamba
         {
             fs::path tarball_path = m_path / s.fn;
             // validate that this tarball has the right size and MD5 sum
-            valid = validate::file_size(tarball_path, s.size);
-            valid = (valid || s.size == 0) && validate::md5(tarball_path, s.md5);
+            // we handle the case where s.size == 0 (explicit packages) or md5 is unknown
+            valid = s.size == 0 || validate::file_size(tarball_path, s.size);
+            valid = valid && (s.md5.empty() || validate::md5(tarball_path, s.md5));
```
adriendelsalle, line 122: I'm not sure we should authorize loose validation. Here missing file size and MD5 would return valid, which is not the case (it's `unknown`).

We could/should then have a `tarball_safety_checks`, with possible values `all`, `one_of`, `size`, `md5`, `none`/`disabled`, and explicitly set it to a given value in corner cases (explicit spec installation, etc.) with probably a warning log message

for now, maybe something like the following would be preferable?
```cpp
bool unknown_size = s.size == 0;
bool valid_size = validate::file_size(tarball_path, s.size);

bool unknown_md5 = s.md5.empty();
bool valid_md5 = validate::md5(tarball_path, s.md5);

valid = (valid_size || unknown_size) && (valid_size || unknown_size)  && (valid_size || valid_md5);
```

src/core/package_cache.cpp
```diff
@@ -170,45 +179,42 @@ namespace mamba
...
-                            }
-                            else if (s.size == 0)
-                            {
-                                // for explicit env, we have no size, nor sha256 so we need to
-                                // set valid true here
-                                valid = true;
-                            }
+                            // in case we have no s.size
+                            // set valid true here
+                            valid = true;
```
adriendelsalle, line 194: looks like this condition is not used anymore

src/core/package_cache.cpp
```diff
@@ -170,45 +179,42 @@ namespace mamba
...
                             LOG_WARNING << "Extracted package cache '" << extracted_dir.string()
-                                        << "' has no checksum";
+                                        << "' has invalid MD5 checksum";
                             valid = false;
                         }
+                        else if (s.size == 0)
+                        {
+                            // for explicit env, we have no size, nor sha256 so we need to
+                            // set valid true here
+                            valid = true;
```
adriendelsalle, line 209: same here?

src/core/package_cache.cpp
wolfv: this condition is still used if the repodata_record.json doesn't contain a value (== "").

adriendelsalle: I don't get it, default value is already set to `true`

PR #1280, opened 2021-11-19
Fix `mamba` print when `json` output is required

adriendelsalle:
Description
---

Fix `mamba` print when `json` output is required


mamba/mamba/utils.py
```diff
@@ -174,7 +174,7 @@ def load_channels(
             # ignore non-loaded subdir if channel is != noarch
             continue
 
-        if context.verbosity != 0:
+        if context.verbosity != 0 and not context.json:
```
wolfv, line 177: why not print to stderr?

PR #1281, opened 2021-11-19
Allow `mamba` to set max extract threads

adriendelsalle:
Description
---

Allow `mamba` to set max extract threads:
- expose `DownloadExtractSemaphore::set_max` from `libmamba`
- use `MAMBA_EXTRACT_THREADS` env var if existing to set the max extract threads value


mamba/mamba/utils.py
```diff
@@ -211,6 +211,11 @@ def init_api_context(use_mamba_experimental=False):
     api_ctx.channels = context.channels
     api_ctx.platform = context.subdir
 
+    if "MAMBA_EXTRACT_THREADS" in os.environ:
+        max_threads = int(os.environ["MAMBA_EXTRACT_THREADS"])
```
wolfv, line 215: Why not put this straight into `context.hpp` as is done for a couple other default-to-env var options?

mamba/mamba/utils.py
```diff
@@ -211,6 +211,11 @@ def init_api_context(use_mamba_experimental=False):
...
+        api_ctx.extract_threads = max_threads
+        api.DownloadExtractSemaphore.set_max(max_threads)
```
wolfv, line 217: Ideally we don't have to call this function and use `Context` as the imperative configuration value. If we really have to we could call this function inside the Context constructor.

mamba/mamba/utils.py
adriendelsalle: we can but the `Context` value is not used except from the `Configuration` to set the semaphore max

mamba/mamba/utils.py
adriendelsalle: I understand, having to call `set_verbosity` or other methods to apply values is then also not correct.
I'll have a look at what could be done without changing the API

mamba/mamba/utils.py
wolfv: maybe we shoudl use it? When starting the downloads.

mamba/mamba/utils.py
wolfv: Verbosity is special because it also needs to call some functions for libcurl and libsolv to set their verbosities

adriendelsalle:
I simplified by setting/updating the max extract threads in `MTransaction::fetch_extract_packages`.
I prefer getting the env var from `mamba` instead of having to use `std::stoi` and some error handling around in the `Context` constructor, which would be still an exception to other configuration not initialized from env vars.

wolfv:
sounds good!

PR #1284, opened 2021-11-20
Use cpp-terminal

jonashaag:
Resurrected https://github.com/mamba-org/mamba/pull/541.

Seems like cpp-terminal development has moved back to https://github.com/jupyter-xeus/cpp-terminal. I've asked what's their recommended version to use at https://github.com/jupyter-xeus/cpp-terminal/issues/130.

### DIFF
--- a/libmamba/CMakeLists.txt
+++ b/libmamba/CMakeLists.txt
@@ -384,6 +384,7 @@ macro(libmamba_create_target target_name linkage deps_linkage output_name)
     if (${linkage_upper} STREQUAL "STATIC")
         find_package(nlohmann_json CONFIG REQUIRED)
         find_package(Threads REQUIRED)
+        find_package(cpp-terminal REQUIRED)
 
         target_link_libraries(${target_name} PUBLIC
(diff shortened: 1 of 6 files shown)


libmamba/CMakeLists.txt
```diff
@@ -384,6 +384,7 @@ macro(libmamba_create_target target_name linkage deps_linkage output_name)
     if (${linkage_upper} STREQUAL "STATIC")
         find_package(nlohmann_json CONFIG REQUIRED)
         find_package(Threads REQUIRED)
+        find_package(cpp-terminal REQUIRED)
```
jonashaag, line 387:
```suggestion
```

wolfv:
Thanks for taking a stab at this!

jjerphan:
Hi Jonas,

Mamba has evolved significantly since then, we do not need or want to depend on another dependency like cpp-terminal.

Feel free to comment or reopen this PR if you think this can bring any benefits. :slightly_smiling_face:

Issue #1283, opened 2021-11-20
Make releasing easier


wolfv:
Now that we've got all these sub-projects in a monorepo, we need to make releasing easier. It's super error prone right now as we have to edit many files.

I propose the following:

- Use a central changelog file in the root of this repository
- In that changelog file, we prefix every item of the changelog with `[libmamba]` or `[mamba, micromamba]` etc.
- The first line will indicate what we are releasing, e.g. `libmamba 0.19.0, libmambapy 0.19.0` or similar
- Then we have a Python script that generates the release: updates version numbers where appropriate, creates the right tags and adds the changelog items to the sub-changelogs in the sub-projects in this repo

cc @adriendelsalle

PR #1287, opened 2021-11-22
Update changelogs

adriendelsalle:
Description
---

Update changelogs
Delete top level changelog, split into projects


wolfv:
please lets do it this way: https://github.com/mamba-org/mamba/issues/1283

wolfv (changes requested):
I think it will be better to handle changelog and version numbers in one place.

Issue #1288, opened 2021-11-22
Regression in activation


adriendelsalle:
Description
---

https://github.com/mamba-org/mamba/pull/1252 introduced a bug when activating an env named exactly the same as a subdir of the current working directory.

```bash
cd ~
micromamba create -r ~/micromamba -n foo xtensor
mkdir foo
micromamba activate foo
```
tries to activate `~/foo` instead of `~/micromamba/envs/foo`.

Previous behavior was:
- `micromamba activate foo` activates `~/micromamba/envs/foo`
- `micromamba activate ./foo` activates `~/foo`

PR #1289, opened 2021-11-22
Fix shell activation regression

adriendelsalle:
Description
---

Fix shell activation
- fix shell activate command when the current working directory contains a subdir with the same name as the named env

Closes https://github.com/mamba-org/mamba/issues/1288


wolfv:
looks good to me!

Issue #1285, opened 2021-11-21
Check that `mamba clean --locks` works fine


wolfv:
I had a situation where I closed the terminal window while mamba was extracting. 
After running mamba again, it told me that the "same PID has already locked the file" and then exited. 

I ran `mamba clean --lock` but it didn't seem to clean up that `mamba.lock` file.

I think we shoudl also print the full path to the lock file and some more explanation.

I was also confused because I don't believe that the PID of the next mamba process would be the same as the previous one.

ManifoldFR:
I started getting the same error as you did, today. Indeed the `mamba.lock` lockfile wasn't being cleaned up.

adriendelsalle:
@wolfv did you checked if the lock file was orphaned (e.g. the lock file name was not corresponding to any existing file to lock)?

To avoid deletion of environment *lock* files (in user or packages' files), I put a check about existence of a file having the same name without the `.lock` extension.
https://github.com/mamba-org/mamba/blob/acc44f8cef95808dd2689f62629ff0fad28b682c/libmamba/src/api/clean.cpp#L69-L70

The error is probably more about having the same PID, any clue how to reproduce this? I'm not able to reproduce locally.

PR #1290, opened 2021-11-22
Fix dangling `LockFile`s

adriendelsalle:
Description
---

Fix dangling `LockFile`s cleaning when the PID of the file doesn't match a running process on Unix systems:
- set the variable storing already existing lock file to false when a dangling lock file is detected
- improve debuggability by adding the path of the dangling lock file to the error log message
- change lock file name for directories locks to make the filename more obvious to understand to the user when a dangling one is found
- set the root prefix from `conda` context if `MAMBA_ROOT_PREFIX` is set from a `micromamba` installation to make sure the correct package cache is cleaned
- improve debuggability of micromamba install Python tests suite by redirecting subprocess stderr to system one and printing the ran sub-command

Closes https://github.com/mamba-org/mamba/issues/1285

### DIFF
--- a/libmamba/src/api/clean.cpp
+++ b/libmamba/src/api/clean.cpp
@@ -67,10 +67,13 @@ namespace mamba
                     for (auto& p : fs::directory_iterator(pkg_cache->path()))
                     {
                         if (p.exists() && ends_with(p.path().string(), ".lock")
-                            && fs::exists(rstrip(p.path().string(), ".lock")))
+                            && (fs::exists(rstrip(p.path().string(), ".lock"))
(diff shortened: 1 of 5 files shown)


wolfv:
Can you check that we clean `pkgs/mamba.lock` files properly?

wolfv:
actually it's simple to recreate the issue from the other thread:

`touch ~/micromamba/mamba.lock` 

`mamba create -n xyz xtensor` 

and I get `error    Path already locked by the same PID` (even though there is no PID in the mamba.lock file.

wolfv:
and running `mamba clean --lock` prints `Cleaning lock files..` but the `mamba.lock` file stays.

PR #1300, opened 2021-11-24
Release 0.18.2

wolfv:


CHANGELOG.md
```diff
@@ -1,3 +1,14 @@
+2021.11.24
+==========
+
+Releases: libmamba 0.18.2, libmambapy 0.18.2, mamba 0.18.2, micromamba 0.18.2
+
+Bug fixes
+- [mamba, libmamba] Fix use of a read-only cache (@adriendelsalle) #1294
+- [mamba, libmamba] Fix dangling LockFiles (@adriendelsalle) #1290
+- [micromamba] Fix CMake config for micromamba fully statically linked on Windows (@adriendelsalle) #1297
+- [micromamba] Fix shell activation regression (@adriendelsalle) #1289
```
adriendelsalle, line 10: I think this last one also concerns `libmamba`

libmambapy/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+libmambapy 0.18.2 (November 24, 2021)
+=====================================
+
+
```
adriendelsalle, line 4: maybe it would be better to not add an empty section?

adriendelsalle: @wolfv it would have been nice to modify the script to remove this artifact

the output name is also not necessary since the changelog is located in the sub-project directory

wolfv: we can fix that later down the line. I think it's OK to have it like this.

PR #1303, opened 2021-11-25
Remove libmamba from install_requires for libmambapy

duncanmmacleod:
This PR removes `libmamba` from the `install_requires` metadata for `libmambapy` - the former isn't a python package, so doesn't provide python metadata (`.egg-info/.dist-info`), meaning `pip check` complains loudly that the requirement isn't fulfilled after installing `libmambapy`:

```console
+ python -m pip check
libmambapy 0.18.2 requires libmamba, which is not installed.
```


wolfv:
Thanks!

PR #1304, opened 2021-11-25
add autocompletion to init, remove useless auto-activate-base

wolfv:
This removes `--auto-activate-base` which was useless.
~~This adds `--autocomplete` to initialize autocompletion alongside the rest of "shell init" which makes this more robust. ~~

This adds a config option `autocomplete: true/false`.


wolfv:
- [ ] need to remove the completion action

Issue #1201, opened 2021-10-05
Unicode-containing prefixes on Windows make `api.PrefixData` fail loading prefix state


jaimergp:
`conda` has quite high expectations for Unicode and spaces on paths, and the test suite creates prefixes with all these niceties. While Linux passes the tests alright, Windows [fails](https://github.com/conda/conda/pull/10881/checks?check_run_id=3793321756#step:5:797) with:

```
______________ IntegrationTests.test_clean_tarballs_and_packages ______________
Traceback (most recent call last):
  File "C:\conda_src\tests\test_create.py", line 2461, in test_clean_tarballs_and_packages
    run_command(Commands.REMOVE, prefix, "bzip2", "--yes", "--json")
  File "C:\conda_src\tests\test_create.py", line 293, in run_command
    result = do_call(args, p)
  File "C:\conda_src\conda\cli\conda_argparse.py", line 83, in do_call
    return getattr(module, func_name)(args, parser)
  File "C:\conda_src\conda\cli\main_remove.py", line 90, in execute
    txn = solver.solve_for_transaction()
  File "C:\conda_src\conda\core\solve.py", line 129, in solve_for_transaction
    unlink_precs, link_precs = self.solve_for_diff(update_modifier, deps_modifier,
  File "C:\conda_src\conda\core\solve.py", line 172, in solve_for_diff
    final_precs = self.solve_final_state(update_modifier, deps_modifier, prune, ignore_pinned,
  File "C:\conda_src\conda\core\solve.py", line 1094, in solve_final_state
    state = self._setup_state()
  File "C:\conda_src\conda\core\solve.py", line 1191, in _setup_state
    mamba_prefix_data.load()
RuntimeError: [json.exception.parse_error.101] parse error at line 1, column 1: syntax error while parsing value - unexpected end of input; expected '[', '{', or a literal
```

The last debugging lines are:

```
DEBUG conda.gateways.logging:set_verbosity(231): verbosity set to 3


EXECUTING COMMAND >>> $ conda install --dev -p "D:\Users\jrodriguez\AppData\Local\Temp\1\pytest-of-jrodriguez\pytest-4\test_update_deps_flag_absent0\tmpqznsiqz6\1036 áγђß家固êôōçñ한" flask -y -q


INFO conda.common.io:captured(246): overtaking stderr and stdout
[...]

PR #1306, opened 2021-11-26
Fix prefix unicode

wolfv:
Partial success! Mamba can now install into a weird prefix on Windows :)

Micromamba needs some additional fixes, though.

@jaimergp I think this will also fix the PrefixData loading issue you described https://github.com/mamba-org/mamba/issues/1201

Screenshot 2021-11-26 at 16 47 25

### DIFF
--- a/libmamba/include/mamba/core/fsutil.hpp
+++ b/libmamba/include/mamba/core/fsutil.hpp
@@ -71,7 +71,7 @@ namespace mamba
                     }
                 }
                 // directory exists, now create empty file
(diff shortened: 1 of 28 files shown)


wolfv:
this was also taking a lot of inspiration from vcpkg :)

wolfv:
there are a few places where we call `open` directly, usually with `path.c_str()`. 

I think we need to convert them, too, to use `wfopen_s` and `path.native()` or something along those lines.

PR #1309, opened 2021-11-30
close json repodata file after opening

wolfv:
@jaimergp this will hopefully fix the remaining error.


jaimergp:
Thanks! We annotated the tests as xfail for now but we will check once there's a new patch release!

PR #1318, opened 2021-12-01
add better update-all and repoquery, factor out load channels

wolfv:
Messy PR that refactors micromamba a bit.

Factored out a `load_channels` function. 

Adding repoquery.
Better `update --all`.


libmamba/include/mamba/api/update.hpp
```diff
@@ -8,14 +8,16 @@
...
 
 #include <string>
 #include <vector>
 
 
 namespace mamba
 {
-    void update(bool update_all = false);
+    void update(bool update_all = false, bool prune = false);
+    void repoquery(QueryType type, const std::string& query);
```
wolfv, line 20: needs to go

PR #1319, opened 2021-12-01
use remove_or_rename to fix deleting files in use on Windows

wolfv:
This implements similar logic as conda has.

We need to go a bit further and also implement this via script (calling out to cmd.exe).
And we need to implement something to scan for trash or to index trash and then take it out on a later day.


wolfv:
- [ ] still need to implement locking of the environment (in conda-meta folder) as well as taking out the trash

Issue #1325, opened 2021-12-03
Support [package>=version]


zxdawn:
Hi, mamba and micromamba are so quicker than conda! Thanks for the contribution.

I test it with `mamba create -n test "matplotlib>=3.5.0 cartopy" -c conda-forge`, and got this error:
```
Looking for: ["matplotlib[version='>=3.5.0',build=cartopy]"]

conda-forge/linux-64     Using cache
conda-forge/noarch       Using cache
pkgs/main/noarch         [====================] (00m:00s) No change
pkgs/r/linux-64          [====================] (00m:00s) No change
pkgs/main/linux-64       [====================] (00m:00s) No change
pkgs/r/noarch            [====================] (00m:00s) No change
Encountered problems while solving:
  - nothing provides requested matplotlib >=3.5.0 cartopy
```

Hope this function will be supported, as something like `>=` is useful for many applications.

wolfv:
that already works. use `mamba create -n test "matplotlib>=3.5.0" cartopy -c conda-forge`

zxdawn:
Thanks @wolfv Is it better to add this example to [User Guide](https://mamba.readthedocs.io/en/latest/user_guide/mamba.html#repoquery). I suppose users usually use `matplotlib>=3.5.0 cartopy` like in conda.

wolfv:
yes, that would be great. Do you want to make a PR?

zxdawn:
Yes, I can create one soon.

PR #1326, opened 2021-12-03
Add example of install specific version of package

zxdawn:
Add a simple example of how to install specific package (#1325)


wolfv:
thanks!

PR #1334, opened 2021-12-06
Fix curl progress callback

adriendelsalle:
Description
---

Fix curl progress callback
- use static method and cast the first argument (`void*`) as data `DownloadTarget*`
- fixes badly passed hidden this pointer, shifting *total to download* to *now downloaded*
- remove unnecessary patch reversing *total to download* and *now downloaded*
- remove delay added for the previously mentioned patch


adriendelsalle:
@wolfv any idea why the CI is failing?

wolfv:
yes #1328

PR #1337, opened 2021-12-07
perf: add fast path for hide_secrets

baszalmstra:
Adds a fast path to the `Console::hide_secrets` function. `std::regex_replace` is rather slow as can be seen in this flamegraph of the `LinkPackage::execute` method:

flamegraph_unoptimized

I added a fast path by checking if `/t/` is contained within the string and only then use regex to replace to contents of the string. The same goes for the http_basicauth regex. After this change the framegraph looks like this:

flamegraph_optimized

In real life this shaves off about ~1sec of a 10sec install. 

Note that the other changes in the file are due to clang-format! I assume they are correct since there is a .clang-format file in the repo.

### DIFF
--- a/libmamba/src/core/output.cpp
+++ b/libmamba/src/core/output.cpp
@@ -320,8 +320,17 @@ namespace mamba
     std::string Console::hide_secrets(const std::string_view& str)
     {
         std::string copy(str);
-        copy = std::regex_replace(copy, token_re, "/t/*****");
-        copy = std::regex_replace(copy, http_basicauth_re, "://$1:*****@");
+
+        if (contains(str, "/t/"))
+        {
+            copy = std::regex_replace(copy, token_re, "/t/*****");
+        }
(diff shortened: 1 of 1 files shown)


wolfv:
Awesome! 

I think the problem with the clang-format changes is that we are currently on a relatively old version (11.1.0). You could install that specific version with `clang-tools=11.1.0`

https://github.com/mamba-org/mamba/blob/master/.github/workflows/linters.yml

baszalmstra:
Could we also add clang-format to the environment.yml files? Then we don't run into this problem! :)

baszalmstra:
I installed the tools as you suggested and ran the pre-commit hooks. Linting should be correct now. :)

wolfv:
Great! We can update clang-format to a later version at some point, but I think that should happen in a dedicated PR as the diff will be quite large.

PR #1338, opened 2021-12-07
perf: use the original sha256 hash if a file doesnt change

baszalmstra:
This PR implements using the original path data's sha256 hash during linking if the file that is copied is not modified.

Profiling this change results in the following flamegraph, before this change a large chunk if time is spend computing sha's:

flamegraph_unoptimized

After this change computing sha hashes takes a lot less time (the highlighted block is the sha256sum function):

flamegraph_optimized

In practice, a stupid test showed on an install of 14seconds a ~2sec improvement (taking only 12sec).

I also removed some code that doesn't seem to do anything in the first place? Or maybe there should be a return in front of it?

### DIFF
--- a/libmamba/src/core/link.cpp
+++ b/libmamba/src/core/link.cpp
@@ -859,10 +859,6 @@ namespace mamba
                     }
                     return std::make_tuple(validate::sha256sum(dst), rel_dst);
                 }
-                else
-                {
-                    std::make_tuple(validate::sha256sum(dst), rel_dst);
-                }
 
 #else
                 std::size_t padding_size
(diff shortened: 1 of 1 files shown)


wolfv:
nice!

Yes, I removed that part of code as well in #1319 :)

libmamba/src/core/link.cpp
```diff
@@ -978,8 +974,7 @@ namespace mamba
             throw std::runtime_error(std::string("Path type not implemented: ")
                                      + std::to_string(static_cast<int>(path_data.path_type)));
         }
-        // TODO we could also use the SHA256 sum of the paths json
-        return std::make_tuple(validate::sha256sum(dst), rel_dst);
+        return std::make_tuple(path_data.sha256, rel_dst);
```
wolfv, line 977: I think we should change this line to check if `path_data.sha256` has a value. Old packages might not have shipped a `paths.json` file and thus there might not be a value available.

Something like `path_data.sha256.empty() ? validate::sha256sum(dst) : path_data.sha256` should do the trick.

wolfv (changes requested):
small change requested, otherwise awesome :)

libmamba/src/core/link.cpp
baszalmstra: good catch! would it also make sense to change this to an optional then in the PathData to make this explicit? Or is that a bit overkill?

wolfv: Yeah we currently treat strings as optional by abusing the empty property.
I think we could start using optional more, but maybe we should do that more consistently in more places.

baszalmstra: Cool! I changed it! :)

PR #1343, opened 2021-12-10
Implement --prune flag for mamba env update

### LINKED ISSUE #688: Support --prune flag in 'mamba install'
Matching changes done to channel/package specs to an environment could be done efficiently by supporting the prune flag properly; however conda env --prune [...] takes in account the history of conda install commands, and while a fix is in the works, it is taking a very long time (the issue was first reported in 2018 : https://github.com/conda/conda/issues/7279 ).
The current workaround is deleting and recreating the environment. Given how fast mamba is at solving plus caching, this works for now. In fact the only package I recall ever taking a long time to setup is jupyter, with its build step.
Either way, I wonder if it would be possible to support --prune directly in mamba install. Perhaps with a different name to not cause confusion ('-sync'?).

adament:
This fixes #688 and the similar issue for conda conda/conda#7279.

The prune flag is implemented by running the solver as if it was given
an empty prefix and then computing the diff between the current prefix
and the newly solved target and executing those resulting links and
unlinks.

This assumes that it is always possible to transform a given prefix to
any other consistent set of packages, since the solver never sees the
list of currently installed packages.

Furthermore a small fix for leaking temporary files is included.

This pull request is not compatible with the current conda implementation but instead implements the feature as documented in `conda env update --help` and as requested in the related conda issue (as far as I understand it). As it was suggested in #688 it might thus be preferred to implement this as a separate argument to mamba env update but with my very limited understanding of conda internals it seems a much greater task.

### DIFF
--- a/mamba/mamba/mamba_env.py
+++ b/mamba/mamba/mamba_env.py
@@ -5,6 +5,7 @@
 
 from __future__ import absolute_import, print_function
 
+import os
 import sys
 import tempfile
 
@@ -16,7 +17,14 @@ from conda_env.installers import conda
 
 import libmambapy as api
 from mamba.linking import handle_txn
-from mamba.utils import get_installed_jsonfile, init_api_context, load_channels, to_txn
+from mamba.utils import (
+    compute_final_precs,
+    get_installed_jsonfile,
+    init_api_context,
+    load_channels,
+    to_txn,
+    to_txn_precs,
+)
 
 
 def mamba_install(prefix, specs, args, env, dry_run=False, *_, **kwargs):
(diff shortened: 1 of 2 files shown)


adament:
Off topic question but is there a way I can run the python formatters locally?

baszalmstra:
> Off topic question but is there a way I can run the python formatters locally?

I simply run `pre-commit run --all`, but you can also simply install and run `black=19.10b`

adament:
> > Off topic question but is there a way I can run the python formatters locally?
> 
> I simply run `pre-commit run --all`, but you can also simply install and run `black=19.10b`

Thank you!

wolfv:
Thanks!

Issue #1335, opened 2021-12-06
No error output with micromamba 0.18 --json


baszalmstra:
We have been using `micromamba --json` internally and found a regression where error messages are no longer displayed.

The output of `micromamba 0.16.0`:
```
❯ ./micromamba.0.16.0 env create -p test xtensor "xtl <0.1" -c conda-forge --json
WARNING 'test' does not contain any filesystem separator.
        It will be handled as env name, resulting to the following
        'target_prefix': '/home/bzalmstra/micromamba/envs/test'
        If 'target_prefix' is expressed as a relative directory to
        the current working directory, use './some_prefix'
Encountered problems while solving:
  - nothing provides requested xtl <0.1

ERROR   Could not solve for environment specs
```

The output of `micromamba 0.18.0`:
```
❯ ./micromamba.0.18.1 env create -p test xtensor "xtl <0.1" -c conda-forge --json
```

Neither is perfect: `0.16.0` outputs non-json (but on stderr so that's fine for me). `0.18.1` outputs nothing. 

I would expect the error to be outputted as json as well, so I can consume it in another tool.

wolfv:
I just tested and I get some beautiful json for the command line input you provided :)

PR #1344, opened 2021-12-10
feat: log json if solving failed

baszalmstra:
This implements the logic to output solver problems when running with `--json` as discussed in #1335 

I also solved some `const` correctness issues in `MSolver`

### DIFF
--- a/libmamba/src/api/install.cpp
+++ b/libmamba/src/api/install.cpp
@@ -522,7 +522,7 @@ namespace mamba
             if (ctx.json)
             {
                 Console::instance().json_write(
-                    { { "success", false }, { "solver_problems", solver.problems_to_json() } });
+                    { { "success", false }, { "solver_problems", solver.all_problems() } });
                 Console::instance().json_print();
             }
 
--- a/libmamba/src/api/install.cpp
+++ b/libmamba/src/api/install.cpp
@@ -519,6 +519,13 @@ namespace mamba
             if (ctx.freeze_installed)
                 Console::print("Possible hints:\n  - 'freeze_installed' is turned on\n");
 
+            if (ctx.json)
+            {
+                Console::instance().json_write(
+                    { { "success", false }, { "solver_problems", solver.problems_to_json() } });
+                Console::instance().json_print();
+            }
+
             Console::stream() << "The environment can't be solved, aborting the operation";
             LOG_ERROR << "Could not solve for environment specs";
             throw std::runtime_error("UnsatisfiableError");
--- a/libmamba/src/core/solver.cpp
+++ b/libmamba/src/core/solver.cpp
@@ -315,7 +315,7 @@ namespace mamba
         }
     }
 
-    bool MSolver::is_solved()
+    bool MSolver::is_solved() const
     {
         return m_is_solved;
     }
@@ -354,7 +354,7 @@ namespace mamba
         return success;
     }
 
-    std::string MSolver::all_problems_to_str()
+    std::string MSolver::all_problems_to_str() const
     {
         std::stringstream problems;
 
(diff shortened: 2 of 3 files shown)


libmamba/src/api/install.cpp
```diff
@@ -519,6 +519,13 @@ namespace mamba
             if (ctx.freeze_installed)
                 Console::print("Possible hints:\n  - 'freeze_installed' is turned on\n");
 
+            if (ctx.json)
+            {
+                Console::instance().json_write(
+                    { { "success", false }, { "solver_problems", solver.problems_to_json() } });
+                Console::instance().json_print();
```
baszalmstra, line 526: This is required to actually print the above json. Should this be somewhere "higher" near the main function? Maybe just always `Console::instance().json_print()` at the end of the main function?

libmamba/src/core/solver.cpp
```diff
@@ -402,6 +402,22 @@ namespace mamba
         return "Encountered problems while solving:\n" + problems.str();
     }
 
+    nlohmann::json MSolver::problems_to_json() const
```
wolfv, line 405: Maybe it would be better to have a function `problems(bool all)` that returns a `std::vector<std::string>` with all problems?

Then in another level we can just use that vector of strings to inject it into a nlohmann_json, or do other post-processing (e.g. in Python).

At some point it would also be good to have programmatic access to the underlying data of the problems (just as a FYI). Each problem comes with `type, source, target, dep` where source is the source pacakge, target is the target package an  dep is the dependency connection (as far as I understand).

libmamba/src/api/install.cpp
adriendelsalle: The current implementation clearly makes it hard to ensure that we don't print multiple times JSON outputs.
In the same time, this `libmamba` code can be used from other interfaces than `micromamba` so having a clean API to get/print the JSON from the buffer would be pretty nice. Then rely on the client interfaces (cli etc) to effectively display, write to a file, ...

notes: 
- we need to check all the places where `exit` is invoked from `libmamba` and remove them. it's up to the caller to handle that
- we catch errors thrown to nicely exit with a `critical` log message, we could also use that hook to return a JSON message there

libmamba/src/core/solver.cpp
adriendelsalle: Agree, having a more refined API and corresponding bindings could be pretty useful!

baszalmstra: Makes sense, all change it!

baszalmstra: I changed the function to `all_problems` which returns a `std::vector<std::string>`. I imagine that once the error reporting is a bit more fleshed out we can change that to and `std::vector<Problem>` or something like that. I can take a look at that later but this at least unblocks us from adopting the latest version of micromamba.

libmamba/src/api/install.cpp
baszalmstra: So should I leave it like this for now?

libmamba/src/core/solver.cpp
wolfv: yes, this looks great. For `std::vector<Problem>` -- yep, I think we need to do some additional R&D on how to represent the problems. Libsolv internally uses `Id` to reference everything, but we might want to convert to a type that is easier to access or have some helper functions.

libmamba/src/api/install.cpp
wolfv: Yeah, I think so. 
In the future, we might want to emulate what vcpkg-tool does. They have some "special" error handling functions, such as `exit_with_message` that we could use instead of plain `exit`, and that could take care of teardown functions such as logging json to stdout.

https://github.com/microsoft/vcpkg-tool/blob/8553f677a56aff07b4d5485fc3f1539546f28964/include/vcpkg/base/checks.h#L12-L19

PR #1345, opened 2021-12-10
feat: expose all keys of PackageInfo

baszalmstra:
I'd like to move from calling `micromamba .. --json` to simply using the python API. However, I'm running into things that are missing from the API. This PR adds missing keys for `PackageInfo` that are available in C++.


wolfv:
cool!

Issue #1051, opened 2021-07-01
Flaky progress bar


adriendelsalle:
Description
---

Using latest `micromamba` build from master on linux-64 gives broken progress bar output:
- mixed `Extracted`, `Finished`, `Transaction starting`, `Extracting`, `Downloading`, Linking` states making it not informative
- frozen extraction count and bars

cc @JohanMabille any thoughts on that?

<details>
 <summary>
output..
 </summary>

```
$ micromamba install xtensor
(...)
Confirm changes: [Y/n] 
Extracted _libgcc_mutex                 
Extracted libstdcxx-ng                  
Extracted ca-certificates               
Extracted libgcc-devel_linux-64         
Finished cmake                                (00m:00s)              15 MB     13 MB/s
Extracted libstdcxx-devel_linux-64      
Extracted libgomp                       
Extracted _openmp_mutex                 
Extracted libgcc-ng                     
Extracted zlib                          
Extracted xz                            
Extracted rhash                         
Extracted libuv                         
Extracted ncurses                       
Extracted bzip2                         
Extracted openssl                       
Extracted expat                         
Extracted make                          
Extracted zstd                          
Extracted tk                            
Extracted kernel-headers_linux-64       
Extracted sysroot_linux-64              
Extracted binutils_impl_linux-64        
Extracted gcc_impl_linux-64             
Extracted binutils_linux-64             
Extracted gxx_impl_linux-64             
Extracted gcc_linux-64                  
Extracted gxx_linux-64                  
Downloading  [====================================================================================================] (00m:11s)   13.09 MB/s
Extracting   [===>                                                                                                ] (00m:00s)       1 / 28


Transaction starting
Downloading  [====================================================================================================] (00m:11s)   13.09 MB/s
Extracting   [===>                                                                                                ] (00m:00s)       1 / 28
Linking _libgcc_mutex-0.1-conda_forge
Downloading  [====================================================================================================] (00m:11s)   13.09 MB/s
Extracting   [===>                                                                                                ] (00m:00s)       1 / 28
Linking libstdcxx-ng-9.3.0-h6de172a_19
Downloading  [====================================================================================================] (00m:11s)   13.09 MB/s
Extracting   [===>                                                                                                ] (00m:00s)       1 / 28
Linking ld_impl_linux-64-2.35.1-hea4e1c9_2
Downloading  [====================================================================================================] (00m:11s)   13.09 MB/s
[...]

PR #1350, opened 2021-12-14
Improve progress bars

adriendelsalle:
Description
---

Improve the progress bars of `libmamba`:
- fix the `finished` status not always set making the progress bars updated after the operation is finished
- use a `Chrono` class to handle time related operation such as `start`, `stop`, `pause`, etc.
- make `ProgressBar` and `ProgressBarManager` inherit from `Chrono` to handle time but also state/status (`started`, `stopped`, etc.)
- add a `ProgressBarManager::watch_print` method to periodically print/refresh the progress bar
  - run the update loop `ProgressBarManager::run` in a detached thread
  - make sure the thread has exited in destructor
- add `Console` and `MessageLogger` buffers to store console and log outputs when a `ProgressBarManager` is active
  - flush the buffers between updates of the progress bars using registered hooks (see `ProgressBarManager::register_print_hook` and `Console::init_progress_bar_manager`)
  - for the moment start with console buffer then replay log messages without timestamp consideration (console and log messages will not be sorted by timestamp)
- refactor `ProgressBar` representation and create a dedicated `ProgressBarRepr` for this purpose
  - also create `FieldRepr` for each field of the `ProgressBar` representation
  - make a `ProgressBarRepr` composed by the following `FieldRepr`s: `<prefix> <progress> [<current>]  [<separator>]  [<total>]  [<speed>]  [<postfix>] [<elapsed>]`
- improve the UX of progress bars
  - display the global time
  - display the time of each progress bar, taking into account the status of the bar
  - flush the bars when finished, with a possible delay to let see the finished status
  - flush the download bar without delay to have it replaced by the extraction bar (there is always an extraction bar after the download bar when fetching packages, but not for repodata files but it's ok because the global operation is much faster)
  - improve the spinner
  - improve the progress bar to use colors (relying on `fmt`)
  - handle multiple console size by resizing or deactivating fields of the representation
    1. resize and deactivate some fields
      - reduce bar width
      - shrink time elapsed
      - shrink the prefix
      - remove the total value and the separator
      - remove the speed
      - remove the postfix
      - truncate the prefix if too long (keep a minimal width of 25)
      - display progress without a bar (on 4 characters to handle a spinner or a progress percentage as `100%`)
      - remove the total value
      - remove the elapsed time
    2. redistribute possible available space (from previous deactivations)
      - start with the prefix if it was shrinked
      - give the remaining free space to the progress bar

```
[+] 16.5s
 - cudatoolkit  ━━━━━━━━━━╺━━━━━╸          extracting   3.5s
Download        ━━━━━━━━━━━━━━━━━ 815.3MB              13.1s
Extract         ━━━━━━━━━╺━━━━━╸━       0               3.4s
```
```
[+] 12.9s
 - occt             100%           2.9s
 - cudatoolkit       ⣽            12.9s
Download            100%  30.8MB   0.7s
Extract              ⣾         3  12.9s
```

Closes https://github.com/mamba-org/mamba/issues/1051


adriendelsalle:
@wolfv could you please have a look?

wolfv:
One issue I just found is that with non-CDN (cached) channels I get this output:

```
conda-forge/osx-arm64    Using cache
conda-forge/noarch       Using cache
robostack-experimental/osx-arm64  ??.?MB @  ??.?MB/s No change  0.7s
robostack-experimental/noarch  ??.?MB @  ??.?MB/s No change  0.7s
```

wolfv:
I even get this when downloading by forcing a new ROOT prefix: 

```
conda-forge/osx-arm64   1.3MB @   2.1MB/s downloaded  0.7s
robostack-experimental/noarch  ??.?MB @  166.0B/s downloaded  0.8s
robostack-experimental/osx-arm64  ??.?MB @  27.3kB/s downloaded  0.8s
conda-forge/noarch     6.4MB @   4.3MB/s downloaded  1.8s
```

adriendelsalle:
I'll have a look thanks!

adriendelsalle:
I did a couple of improvements since:
- do not display sub bars to have something less flickering and more simple
- compute and display average speed for aggregated download bar (over a period of 500ms)
- compute and display currently executing tasks on aggregated bars using the postfix (update every seconds to keep it readable and not flickering)
- fix the display of `total` and `speed` when the repodata didn't changed
- remove the `downloaded` postfix when printing the final status to the console
- fix error related to multithreaded progress bars without mutexes

Hereafter an example with ~200 packages (and higher download threads):
mamba-progress-bars02

I still have a minor fix to do on printing the repodata for really small files that don't have their total sizes updated

adriendelsalle:
The computation of the speed for small files/fast downloads should be fixed now.
When `cURL` info `CURLINFO_SPEED_DOWNLOAD_T` is not yet available and a progress bar is associated to the operation, use the progress bar average speed to display relevant info instead of `??.?MB/s`
```
adrien@tower:~$ micromamba create -n foo xtl -y --no-pyc --dry-run -c robostack -c pytorch -c cosapp

                                           __
      /_/

robostack/noarch                      1.1kB @   1.7kB/s  0.6s
pytorch/noarch                        5.0kB @   8.1kB/s  0.6s
cosapp/linux-64                      138.0B @  214.0B/s  0.6s
cosapp/noarch                         1.1kB @   1.5kB/s  0.7s
pytorch/linux-64                    119.0kB @ 129.0kB/s  0.9s
robostack/linux-64                  590.5kB @ 362.9kB/s  1.6s
conda-forge/noarch                    7.0MB @   3.1MB/s  2.3s
conda-forge/linux-64                 20.4MB @   2.8MB/s  7.5s
```

adriendelsalle:
Latest update:
- the in-progress status is now also display using a different color (cyan)
  - this is mainly useful for aggregated bars
- computation of progress bars representation (spinner and normal progress) has been improved to avoid rounding issues making the bar step back
- broken tests due to API changes are fixed

mamba-progress-bars03

libmamba/include/mamba/core/util.hpp
```diff
@@ -59,7 +59,7 @@ namespace mamba
 
     bool is_package_file(const std::string_view& fn);
 
-    void to_human_readable_filesize(std::ostream& o, double bytes, std::size_t precision = 0);
+    // void to_human_readable_filesize(std::ostream& o, double bytes, std::size_t precision = 0);
```
wolfv, line 62: remove?

libmamba/src/core/fetch.cpp
```diff
@@ -640,8 +697,33 @@ namespace mamba
...
     {
+        auto& ctx = Context::instance();
+
+        if (m_targets.empty())
+        {
+            LOG_INFO << "All targets to download are cached";
+            return true;
+        }
+
+        std::sort(
```
wolfv, line 708: is this to optimize the download order?

libmamba/src/core/output.cpp
```diff
@@ -399,35 +416,31 @@ namespace mamba
 
     ProgressProxy Console::add_progress_bar(const std::string& name, size_t expected_total)
     {
-        return p_progress_manager->add_progress_bar(name, expected_total);
-    }
-
-    void Console::init_multi_progress(ProgressBarMode mode)
-    {
-        p_progress_manager = make_progress_bar_manager(mode);
+        if (Context::instance().no_progress_bars)
```
wolfv, line 419: can we turn this into an assert and make sure that `add_progress_bar` is never called when this is true?

wolfv (commented):
Some initial comments

libmamba/include/mamba/core/output.hpp
```diff
@@ -206,8 +115,8 @@ namespace mamba
 
     // The next two functions / classes were ported from the awesome indicators
     // library by p-ranav (MIT License) https://github.com/p-ranav/indicators
-    std::ostream& write_duration(std::ostream& os, std::chrono::nanoseconds ns);
-    int get_console_width();
+    // std::ostream& write_duration(std::ostream& os, std::chrono::nanoseconds ns);
```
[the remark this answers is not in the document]
adriendelsalle: done!

libmamba/include/mamba/core/util.hpp
adriendelsalle: yes!

libmamba/src/core/fetch.cpp
adriendelsalle: yes, it is a basic attempt to optimize download speed by starting large downloads first.
this could be part of another PR if you prefer, and should be probably improved by implementing several other strategies. I'm thinking about a more global optimization of installation process where some package (some packages are downloaded while others are extracting or (un-)linking)

libmamba/src/core/output.cpp
adriendelsalle: Just discussed together it can still make sense for caller code to create a `ProgressProxy` and associate a `ProgressBar` later or just rely on its default constructor (see https://github.com/mamba-org/mamba/blob/d4acbf7dc3d4743c805e8b9085033703f823b9d6/libmamba/include/mamba/core/fetch.hpp#L105).

It makes sense to add `operator bool` to allow syntax like `if(proxy)`, which I added since

libmamba/src/core/progress_bar.cpp
```diff
@@ -342,206 +441,1521 @@ namespace mamba
...
+                                                      + p_progress_bar->in_progress())
+                                  / static_cast<double>(p_progress_bar->total()) * width));
+                }
+
+                auto spinner_half_width = width / 8;
+                for (std::size_t i = 0; i < width; ++i)
+                {
+                    if ((i < current_pos - 1 && current_pos > 0) || (current_pos == width))
+                    {
+                        sstream << fmt::format(
```
wolfv, line 655: I don't like that we're calling `fmt::format` inside a for loop to format single characters.
It would be enough to set the color once and then print all characters at once, no?

libmamba/src/core/progress_bar.cpp
```diff
@@ -342,206 +441,1521 @@ namespace mamba
...
+        }
+
+        progress.set_value(sstream.str());
+    }
+
+    /*****************
+     * ProgressProxy *
+     *****************/
+
+    ProgressProxy::ProgressProxy(ProgressBar* ptr)
```
wolfv, line 731: I am wondering if we still need the ProgressProxy ... in fact I forgot why we need it in the first place :)

libmamba/src/core/util.cpp
```diff
@@ -70,19 +70,21 @@ namespace mamba
         auto status = fs::symlink_status(path);
         return status.type() != fs::file_type::not_found || status.type() == fs::file_type::symlink;
     }
-
-    void to_human_readable_filesize(std::ostream& o, double bytes, std::size_t precision)
-    {
-        const char* sizes[] = { "  B", " KB", " MB", " GB", " TB" };
-        int order = 0;
-        while (bytes >= 1024 && order < (5 - 1))
+    /*
```
wolfv, line 73: remove?

libmamba/src/core/progress_bar.cpp
adriendelsalle: you basically don't know if the next char will be blue, magenta or grey. 
there is probably room for improvement, for example with groups of same colors.
I can at least compute each character once and reuse it

libmamba/src/core/progress_bar.cpp
adriendelsalle: I don't know either, the pattern was implemented by @JohanMabille  :)

libmamba/src/core/util.cpp
adriendelsalle: yep!

libmamba/src/core/output.cpp
```diff
@@ -26,53 +26,59 @@
 
 namespace mamba
 {
+    /*
```
wolfv, line 29: can / should we remove this?

libmamba/src/core/output.cpp
```diff
@@ -26,53 +26,59 @@
...
-#endif
-
-        return -1;
+        oss << s.count() << "." << std::round(static_cast<double>(ms.count()) / 100) << "s";
+        os << oss.str();
+        return os;
     }
+    */
 
+    /*
```
wolfv, line 65: remove?

libmamba/src/core/progress_bar.cpp
wolfv: I think the idea was to use the fmt fill character stuff here: 

```c++
int count = 10;
fmt::format("{:━{}}", "", count);
```

but we'd need to compute `count` before hand which I'd hope is possible?

wolfv:
Just the linters now! :)
And removing of the indicated comments

adriendelsalle:
done, sorry for that

mamba/mamba/mamba.py
```diff
@@ -552,23 +555,30 @@ def install(args, parser, command="install"):
...
+            if (
+                newenv
+                and not isdir(Path(prefix) / "conda-meta")
+                and not context.dry_run
+            ):
+                mkdir_p(Path(prefix) / "conda-meta")
+
+            transaction.execute(prefix_data)
     else:
+        if not transaction.prompt():
```
davidbrochart, line 577: For some reason `transaction` is not defined here on Windows (see #1459).

PR #1355, opened 2021-12-16
Rename conflicting mamba.sh to micromamba.sh

### LINKED ISSUE #1322: Micromamba and Mamba create conflicting mamba.sh files
The directory `$MAMBA_ROOT_PREFIX/etc/profile.d/` contains various scripts which are used during shell activation.

Micromamba writes [this standalone script](https://github.com/mamba-org/mamba/blob/bab7ae50d75f067c493b8d0cd065ea29bd3f9e8c/libmamba/data/mamba.sh) as `mamba.sh`. This script defines the Bash shell function `micromamba()`.

Similarly, Mamba writes [this different script](https://github.com/mamba-org/mamba/blob/bab7ae50d75f067c493b8d0cd065ea29bd3f9e8c/mamba/mamba/shell_templates/mamba.sh) as `mamba.sh`. This script defines `mamba()` and depends on `conda.sh` having been sourced beforehand (it invokes `__conda_activate()`).

Having two scripts with the exact same filename but disjoint functionality leads to errors.

For example, if `mamba.sh` exists, then `micromamba shell hook --shell=bash` uses it as the init script. However, because Mamba's `mamba.sh` doesn't define `micromamba()`, you can't subsequently activate with `micromamba activate xyz`. Doing so results in
```text
The following arguments were not expected: xyz activate
```
(see also #1323)

This issue is the direct cause of issues #1358 and https://github.com/mamba-org/micromamba-docker/issues/57

### LINKED ISSUE #1358: Micromamba linking fails when Mamba is also installed
During the linking stage near the end of `micromamba install`, if `$MAMBA_ROOT_PREFIX/etc/profile.d/mamba.sh` was created by Mamba, one observes the following error (the first two lines only being visible with `--log-level=trace`) repeated at each linking step: 
```
debug    Running wrapped python compilation command /opt/conda/bin/python3.9 -Wi -m compileall -q -l -i /tmp/mambaf2022jMdTKB -j0
/tmp/mambafMqKgw4Fwir
The following arguments were not expected: /opt/conda activate
```

As explanation, the Micromamba linker is creating a temporary script with contents
```bash
eval "$('/bin/micromamba' 'shell' 'hook' '-s' 'bash' '-p' '/opt/conda')"
micromamba activate "/opt/conda"

/opt/conda/bin/python3.9 -Wi -m compileall -q -l -i /tmp/mambaf2022jMdTKB -j0
```
and executing it. This then triggers #1322, leading to the error.

maresb:
`mamba` creates [`mamba.sh`](https://github.com/mamba-org/mamba/blob/master/mamba/mamba/shell_templates/mamba.sh) under `$MAMBA_ROOT_PREFIX/etc/profile.d/mamba.sh`.

`micromamba` creates a different [`mamba.sh`](https://github.com/mamba-org/mamba/blob/master/libmamba/data/mamba.sh) in the same location.

Bad things #1322 happen when the two get mixed up. This PR tries to rename `micromamba`'s `mamba.sh` to `micromamba.sh` so that they can coexist.

Closes #1322
Closes #1358 
Closes mamba-org/micromamba-docker#57


wolfv:
Hi, thanks for digging into this! I think you got to the bottom of it and sorry for clobbering this up!

maresb:
This is one instance of several where Micromamba names conflict with Mamba stuff. For instance, there's `MAMBA_ROOT_PREFIX` and the `# >>> mamba initialize >>>` section of `.bashrc`.

This makes me wonder: is the end goal for Micromamba to replace Mamba? If that's the case, then I can understand the logic, and maybe it makes sense to persevere with these conflicting names (unless it leads to actual problems like here). Otherwise, should we try to fix the other conflicts?

adriendelsalle:
few thoughts about that:
- `micromamba` was started to only bootstrap envs
- `libmamba` is now the shared code base of `mamba` and `libmamba` (but also potential other downstream projects)
- I'm not sure that setting a different default root prefix for `mamba` and `micromamba` is a common case, so sharing env var `MAMBA_ROOT_PREFIX` is ~ fine
- I don't think `micromamba` will replace `mamba` in the coming  weeks/months so it would worth to fix such issues
- we have a bunch of shell scripts deeply inspired by `conda` in `libmamba` with hardcoded ref to `micromamba` and I'm not sure it's totally correct
- the references to a downstream project in `libmamba/src/core/shell_init.cpp` and similar are probably also not correct

maresb:
@adriendelsalle, thanks so much for the info!

Would it make sense to fix this current conflict with `mamba.sh` and leave any remaining issues to fix (e.g. `# >>> mamba initialize >>>` → `# >>> micromamba initialize >>>`) to a subsequent PR?

My motivation here is to bootstrap `mamba` with `micromamba` in order to make Miniforge and relatives obsolete. I feel like this goal is very close! (It feels much more elegant to use `micromamba` to install the base environment directly from the latest conda-forge repodata.)

wolfv:
Thanks @maresb this is helpful!

PR #1357, opened 2021-12-16
Fix conda init clobbering micromamba init

### LINKED ISSUE #1356: conda bash init clobbers micromamba's init
Micromamba's init adds something like this to `~/.bashrc`:

```bash
# >>> mamba initialize >>>
export MAMBA_EXE="/bin/micromamba";
export MAMBA_ROOT_PREFIX="/opt/conda";
__mamba_setup="$('/bin/micromamba' shell hook --shell bash --prefix '/opt/conda' 2> /dev/null)"
if [ $? -eq 0 ]; then
    eval "$__mamba_setup"
else
    if [ -f "/opt/conda/etc/profile.d/mamba.sh" ]; then
        . "/opt/conda/etc/profile.d/mamba.sh"
    else
        export PATH="/opt/conda/bin:$PATH"
    fi
fi
unset __mamba_setup
# <<< mamba initialize <<<
```

If Conda's init is subsequently run, it comments out the `export PATH=…` line:

```bash
# export PATH="/opt/conda/bin:$PATH"  # commented out by conda initialize
```

Because there is no command between `else` and `fi`, this leads to a syntax error in Bash:

```
bash: /home/micromamba/.bashrc: line 128: syntax error near unexpected token `fi'
```

maresb:
Closes #1356

### DIFF
--- a/libmamba/src/core/shell_init.cpp
+++ b/libmamba/src/core/shell_init.cpp
@@ -242,7 +242,8 @@ namespace mamba
                 << " ]; then\n";
         content << "        . " << (env_prefix / "etc" / "profile.d" / "mamba.sh") << "\n";
         content << "    else\n";
-        content << "        export PATH=\"" << env_bin.c_str() << ":$PATH\"\n";
+        content << "        export  PATH=\"" << env_bin.c_str() << ":$PATH\""
+                << "  # extra space after export prevents interference from conda init\n";
         content << "    fi\n";
         content << "fi\n";
         content << "unset __mamba_setup\n";


wolfv:
actually we should very seldomly end up in that if/else branch anyways. But thanks!

maresb:
@wolfv, I agree, but it leads to a syntax error. (Sometimes I really hate Bash...)

By the way, thank you soooo much for all the merges!!! :pray:  I really appreciate it! :smile: Willkommen zurück und guten Rutsch!

PR #1360, opened 2021-12-17
Implement activate as a micromamba subcommand

### LINKED ISSUE #1323: Confusing micromamba error when shell not initialized
Running `micromamba activate base` when the shell is not initialized results in
```
The following arguments were not expected: base activate
```

This can be very confusing for new users. Or for more experienced users who are struggling with various types of Bash shells in Docker. :wink:

I would expect that the output should look something like:
```
Cannot activate an environment when the shell has not been initialized. Please run

    micromamba shell init --shell=basename($SHELL) --prefix=/path/to/conda/prefix

and restart the shell, or run

    eval "$(micromamba shell hook --shell=bash)"

to activate the current shell.
```

### LINKED ISSUE #1359: `activate` is not a subcommand of `micromamba`
When I run `bin/micromamba activate my-custom-env`, micromamba raises an error that `activate` and `my-custom-env` are not expected subcommands.
This is an image of the `bin/micromamba --help`. It is possible to see that indeed `activate` is not listed in the subcommands list.
I would appreciate any help with this :)
image

maresb:
This way it shows up under `micromamba --help`, and it prints a useful error message when running `./micromamba activate`.

Closes #1359 
Closes #1323

### DIFF
--- a/micromamba/src/activate.cpp
+++ b/micromamba/src/activate.cpp
@@ -1,0 +1,49 @@
+// Copyright (c) 2019, QuantStack and Mamba Contributors
+//
+// Distributed under the terms of the BSD 3-Clause License.
+//
+// The full license is in the file LICENSE, distributed with this software.
+
+#include "common_options.hpp"
+
+#include "mamba/api/configuration.hpp"
+
+#include "termcolor/termcolor.hpp"
+
+
+using namespace mamba;  // NOLINT(build/namespaces)
+
+void
+init_activate_parser(CLI::App* subcom)
+{
+    auto& config = Configuration::instance();
+
+    auto& env_name = config.insert(Configurable("activate_env_name", std::string(""))
+                                       .group("cli")
+                                       .description("The name of the environment to activate"));
+    subcom->add_option("env_name", env_name.set_cli_config(""), env_name.description());
+}
+
+void
+set_activate_command(CLI::App* subcom)
+{
+    init_activate_parser(subcom);
+
+    subcom->callback([&]() {
+        std::stringstream message;
+
+        message
+            << "\nIn order to use activate and deactivate you need to initialize your shell.\n"
+            << "Either call shell init ... or execute the shell hook directly:\n\n"
+            << "    $ eval \"$(./micromamba shell hook -s bash)\"\n\n"
+            << "and then run activate and deactivate like so:\n\n"
+            << "    $ micromamba activate  " << termcolor::white
+            << "# notice the missing dot in front of the command\n\n"
+            << "To permanently initialize your shell, use shell init like so:\n\n"
+            << "    $ ./micromamba shell init -s bash -p /home/$USER/micromamba \n\n"
+            << "and restart your shell. Supported shells are bash, zsh, xonsh, cmd.exe, powershell, and fish."
+            << termcolor::reset;
+
+        throw std::runtime_error(message.str());
+    });
+}
--- a/micromamba/src/activate.cpp
+++ b/micromamba/src/activate.cpp
@@ -33,15 +33,16 @@ set_activate_command(CLI::App* subcom)
         std::stringstream message;
 
         message
-            << "\nIn order to use activate and deactivate you need to initialize your shell.\n"
-            << "Either call shell init ... or execute the shell hook directly:\n\n"
-            << "    $ eval \"$(./micromamba shell hook -s bash)\"\n\n"
-            << "and then run activate and deactivate like so:\n\n"
-            << "    $ micromamba activate  " << termcolor::white
-            << "# notice the missing dot in front of the command\n\n"
-            << "To permanently initialize your shell, use shell init like so:\n\n"
-            << "    $ ./micromamba shell init -s bash -p /home/$USER/micromamba \n\n"
-            << "and restart your shell. Supported shells are bash, zsh, xonsh, cmd.exe, powershell, and fish."
+            << "\n"
+            << "In order to use activate and deactivate you need to initialize your shell. "
+            << "(Micromamba is running as a subprocess, so it cannot modify the parent shell.)\n"
+            << "To initialize the current (bash) shell, run:\n\n"
+            << "    $ eval \"$(micromamba shell hook --shell=bash)\"\n\n"
+            << "and then activate or deactivate with:\n\n"
+            << "    $ micromamba activate\n\n"
+            << "To automatically initialize all future (bash) shells, run:\n\n"
+            << "    $ micromamba shell init --shell=bash --prefix=/home/$USER/micromamba \n\n"
+            << "Supported shells are bash, zsh, xonsh, cmd.exe, powershell, and fish."
             << termcolor::reset;
 
         throw std::runtime_error(message.str());
--- a/micromamba/src/umamba.cpp
+++ b/micromamba/src/umamba.cpp
@@ -79,19 +79,6 @@ set_umamba_command(CLI::App* com)
     CLI::App* env_subcom = com->add_subcommand("env", "List environments");
     set_env_command(env_subcom);
 
-    std::stringstream footer;  // just for the help text
-
-    footer
-        << "In order to use activate and deactivate you need to initialize your shell.\n"
-        << "Either call shell init ... or execute the shell hook directly:\n\n"
-        << "    $ eval \"$(./micromamba shell hook -s bash)\"\n\n"
-        << "and then run activate and deactivate like so:\n\n"
-        << "    $ micromamba activate  " << termcolor::white
-        << "# notice the missing dot in front of the command\n\n"
-        << "To permanently initialize your shell, use shell init like so:\n\n"
-        << "    $ ./micromamba shell init -s bash -p /home/$USER/micromamba \n\n"
-        << "and restart your shell. Supported shells are bash, zsh, xonsh, cmd.exe, powershell, and fish."
-        << termcolor::reset;
-
-    com->footer(footer.str());
+    CLI::App* activate_subcom = com->add_subcommand("activate", "Activate an environment");
+    set_activate_command(activate_subcom);
 }
--- a/micromamba/CMakeLists.txt
+++ b/micromamba/CMakeLists.txt
@@ -21,6 +21,7 @@ set(MICROMAMBA_LINKAGE "DYNAMIC" CACHE STRING "micromamba linkage against librar
 # ============
 
 set(MICROMAMBA_SRCS
+    ${CMAKE_CURRENT_SOURCE_DIR}/src/activate.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/src/clean.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/src/config.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/src/completer.cpp
--- a/micromamba/src/umamba.hpp
+++ b/micromamba/src/umamba.hpp
@@ -59,6 +59,9 @@ void
 set_env_command(CLI::App* subcom);
 
 void
+set_activate_command(CLI::App* subcom);
+
+void
 get_completions(CLI::App* app, int argc, char** argv);
 
 #endif


maresb:
Before 107b737, the tests were mysteriously failing, even though they shouldn't have been.

After some investigation, I noticed that without applying `->required()` in 107b737, for reasons which I don't understand, when `action` coincides with a top-level command name, validation will succeed, even though the CLI::IsMember check fails. For example,
```bash
micromamba shell list
```
results in a call to libmamba's
```python
shell(action="", shell_type="", prefix="", stack=0)
```

On the other hand, `micromamba shell xyz` results in the expected
```
action: xyz not in {activate,deactivate,hook,init,reactivate}
```

I don't understand CLI11, and I really don't understand the discrepancy in validation. It makes me suspect that there is a further bug.

micromamba/src/activate.cpp
```diff
@@ -0,0 +1,50 @@
...
+
+
+using namespace mamba;  // NOLINT(build/namespaces)
+
+void
+init_activate_parser(CLI::App* subcom)
+{
+    auto& config = Configuration::instance();
+
+    auto& env_name = config.insert(Configurable("activate_env_name", std::string(""))
```
adriendelsalle, line 21: maybe here we could reuse an existing configurable instead of adding a new one (`prefix` or `env_name`)

micromamba/src/activate.cpp
```diff
@@ -0,0 +1,50 @@
...
+            << "In order to use activate and deactivate you need to initialize your shell. "
+            << "(Micromamba is running as a subprocess, so it cannot modify the parent shell.)\n"
+            << "To initialize the current (bash) shell, run:\n\n"
+            << "    $ eval \"$(micromamba shell hook --shell=bash)\"\n\n"
+            << "and then activate or deactivate with:\n\n"
+            << "    $ micromamba activate\n\n"
+            << "To automatically initialize all future (bash) shells, run:\n\n"
+            << "    $ micromamba shell init --shell=bash --prefix=/home/$USER/micromamba \n\n"
+            << "Supported shells are bash, zsh, xonsh, cmd.exe, powershell, and fish."
+            << termcolor::reset;
```
adriendelsalle, line 46: why `termcolor::reset` is needed if you don't use any color/style before?

micromamba/src/umamba.cpp
```diff
@@ -79,19 +79,6 @@ set_umamba_command(CLI::App* com)
...
-        << "    $ micromamba activate  " << termcolor::white
-        << "# notice the missing dot in front of the command\n\n"
-        << "To permanently initialize your shell, use shell init like so:\n\n"
-        << "    $ ./micromamba shell init -s bash -p /home/$USER/micromamba \n\n"
-        << "and restart your shell. Supported shells are bash, zsh, xonsh, cmd.exe, powershell, and fish."
-        << termcolor::reset;
-
-    com->footer(footer.str());
+    CLI::App* activate_subcom = com->add_subcommand("activate", "Activate an environment");
+    set_activate_command(activate_subcom);
```
adriendelsalle, line 83: I'm still not 100% sure it will work fine on all platforms/shells (e.g. Windows/cmd.exe) if the executable is in the current working directory or found in the path

micromamba/src/shell.cpp
```diff
@@ -48,13 +48,13 @@ init_shell_parser(CLI::App* subcom)
                                                       "activate",
                                                       "deactivate",
                                                       "hook",
-                                                      "reactivate",
-                                                      "deactivate"
```
adriendelsalle, line 52: I would expect this to break https://github.com/mamba-org/mamba/blob/bab7ae50d75f067c493b8d0cd065ea29bd3f9e8c/libmamba/data/mamba.sh#L69

micromamba/src/umamba.cpp
adriendelsalle: this probably also needs an update of the shell completer subcommand https://github.com/mamba-org/mamba/blob/bab7ae50d75f067c493b8d0cd065ea29bd3f9e8c/micromamba/src/completer.cpp#L87

adriendelsalle (commented):
Thanks @maresb 
I'm not 100% everything works fine, I'll take a look!

micromamba/src/shell.cpp
maresb: Hmm, here I am deleting a duplicate `deactivate` entry in a `std::set`, because it looks like a typo. Based on the usual definition of "set" I wouldn't expect this to have any effect. (In a mathematical set, an object is either a member or not, so it shouldn't matter how many times it occurs.)

micromamba/src/activate.cpp
maresb: Sounds like a good idea to me, I just don't know how to do it.

micromamba/src/activate.cpp
maresb: Because I just copied over the previous message. :see_no_evil: So no good reason.

maresb:
Hi @adriendelsalle, thanks so much for the code review!!! My inexperience in C++ really shows. :laughing: 

I see this PR as merely a cosmetic improvement. If possible, could you prioritize #1355? That one is causing breakage.

Since I really don't know what I'm doing, please feel free to commit directly in my branch, or to simply start over. (I'm just trying to be a good citizen by attempting PRs for my issues.)

micromamba/src/umamba.cpp
maresb: > I'm still not 100% sure it will work fine on all platforms/shells...

I'm just trying to modify the error messages as per #1323... I don't understand yet how this could lead to breakage. I just want `./micromamba activate` to print a more sensible error.

> this probably also needs an update of the shell completer subcommand

Sounds good! Would you be able to help with that?

micromamba/src/shell.cpp
adriendelsalle: oh sorry I didn't saw the duplicate :)

adriendelsalle:
@maresb we are super happy with your PRs!
I'll take a deeper look into this PR later today or tomorrow. I'll also take a look to the other one :)

micromamba/src/activate.cpp
adriendelsalle: you can get an existing configurable with:
``` cpp
auto& env_name = config.at("env_name").get_wrapped<std::string>();
```
this syntax is required because of type erasure

micromamba/src/activate.cpp
adriendelsalle: no problem ;)

micromamba/src/umamba.cpp
adriendelsalle: > I don't understand yet how this could lead to breakage
I'm just not sure that calinng `micromamba activate` on `Windows/cmd.exe` always prioritize to the shell function instead of the executable. For other subcommands it's not a problem because the args are just forwarded to the executable, but for activation it could just make the shell function not reachable (masked) by the executable

> Sounds good! Would you be able to help with that?
yes I'll take a look

maresb:
@adriendelsalle, great! I just hope I saved you some small amount of work in getting started. (It probably takes me literally 20× longer than you to code anything in C++.)

I have enabled "Allow edits by maintainers" in my PRs. Would it be acceptable for me to resign from making any further commits for this PR and #1355 and leave the rest to you?

micromamba/src/activate.cpp
adriendelsalle: my bad I didn't saw the `termcolor::white`

adriendelsalle:
Let's have a look together :)
I'm working on this right now, sorry for the delay. @wolfv already took care of the other PR

maresb:
Excellent!

One thing I didn't get around to yet is to add `--stack` as an argument to `./micromamba activate` so that it's shown as an option when running `./micromamba activate --help`.

adriendelsalle:
I'll do that!
I'm investigating the weird behavior with the subcommand when `required` is not set but can't really figure out what's going on for now..

adriendelsalle:
That was a *simple* `com->require_subcommand(/* min */ 0, /* max */ 1);` missing at the top level `CLI::App`, making collisions between options and subcommands.
For instance, `micromamba info list` was doing both `info` and `list` subcommands, some `micromamba shell activate foo` was trying to do `micromamba shell` then `micromamba activate foo`. It was not caught before because it's probably the first time we have a collision.

maresb:
Awesome!!! I'm very pleased that you managed to track this down. Seems very subtle!

adriendelsalle:
You're welcome ;). I'll do a last check on Windows to be sure the newly created `activate` subcom doesn't collides with the shell function.
It should break the CI but I prefer to double check, especially on Windows :/.

adriendelsalle:
Looks like it works as expected as well.

I'm also implementing a more specific error message depending on the current shell, which gives
- on `bash` (other posix shells will give similar messages)
``` bash
$ micromamba activate

'micromamba' is running as a subprocess and can't modify the parent one.
In order to use activate and deactivate you need to initialize your shell.

To initialize the current bash shell, run:
    $ eval "$(micromamba shell hook --shell=bash)"
and then activate or deactivate with:
    $ micromamba activate

To automatically initialize all future (bash) shells, run:
    $ micromamba shell init --shell=bash --prefix=~/micromamba

Supported shells are {bash, zsh, xonsh, cmd.exe, powershell, fish}.
```

- on `powershell`
``` bash
$ ./micromamba.exe activate

'micromamba' is running as a subprocess and can't modify the parent one.
In order to use activate and deactivate you need to initialize your shell.

To initialize the current powershell shell, run:
    $ .\micromamba.exe shell hook -s powershell | Out-String | Invoke-Expression
and then activate or deactivate with:
    $ micromamba activate

To automatically initialize all future (powershell) shells, run:
    $ micromamba shell init --shell=powershell --prefix=~/micromamba

Supported shells are {bash, zsh, xonsh, cmd.exe, powershell, fish}.

critical Shell not initialized
```

- on `cmd.exe` the shell hook is not working thus the corresponding instructions are removed
``` bash
$ ./micromamba.exe activate

'micromamba' is running as a subprocess and can't modify the parent one.
In order to use activate and deactivate you need to initialize your shell.

To automatically initialize all future (cmd.exe) shells, run:
    $ micromamba shell init --shell=cmd.exe --prefix=~/micromamba

Supported shells are {bash, zsh, xonsh, cmd.exe, powershell, fish}.

critical Shell not initialized
```

maresb:
This is looking really great!!! Thanks so much @adriendelsalle  for pushing this forward, it's much more than I could have managed. :rocket: 

As a second pair of eyes, I have a few suggestions/observations:

1. The message
    > Defines if the activation has to be stacked or not

    doesn't really explain what stacking does. Perhaps this would be more helpful?

    > Activate the specified environment without first deactivating the current one

2. Another minor suggested rewording:
    >'micromamba' is running as a subprocess and can't modify the parent one.
    In order to use activate and deactivate you need to initialize your shell.

    →

    >'micromamba' is running as a subprocess and can't modify the parent shell.
    Thus you must initialize your shell before using activate and deactivate.

3. Is the `.\` prefix really necessary for Powershell? (If it's like in Linux, then it seems to me like we shouldn't assume that `micromamba.exe` is in the current directory; it might just be on the PATH.)

4. If we want to be 100% technically correct, in Windows the message should be changed:

    `'micromamba' is running as a subprocess` → `'micromamba.exe' is running as a subprocess`.

    but I feel like this suggestion is extremely pedantic. :laughing: 

5. I noticed that these two lines of code are identical; is that intentional?  https://github.com/mamba-org/mamba/blob/96564e7ec32e00544650dfb2335ef9f01098f919/micromamba/src/activate.cpp#L32-L34

6. I notice that `--prefix=~/micromamba` is hardcoded in all current cases.
    - Does this syntax even work in Windows where everything (especially the slashes) is backwards? :smile:
    - Maybe it would be possible to determine the correct prefix here from `MAMBA_ROOT_PREFIX` and the value of `--prefix`? But this seems like some additional work, so maybe that could be put into a different issue or PR in order to not delay the current progress?

adriendelsalle:
Hey @maresb thanks for the suggestions! Feel free to push those improvements!
1. & 2.: sure!
3. you're right 
4. I think it's my main fear about collision of the executable and the shell function if the executable is called first. I'll double check on linux as well
5. shell_hook_command is not used for cmd.exe, it's a typo!
6. I think it's ok that way
- it should work because we expand the home directory, etc.
- we don't really care about that here, it's just for the example to be correct and work when used by the user because in all case we just raise an error about the fact that the shell has to be activated + the shell api already transforms the prefix to catch incorrect prefix or prefix which are env name (related to a root prefix). See https://github.com/mamba-org/mamba/blob/aa0365c1ce78caddd354a324e2b27a9c0183f8c9/libmamba/src/api/shell.cpp#L87

maresb:
Great, I pushed another commit. The numbering in your reply got a bit messed up, so I was a little confused. I wasn't sure what you wanted with `.\micromamba` → `micromamba` for Powershell. I changed it so please change it back in case it doesn't work. (I have no way to test Windows.)

adriendelsalle:
Ah sorry, I was just saying that changing `.\micromamba.exe` (don't forget the `.exe` on Windows!) to `micromamba.exe` LGTM.

But now I'm not sure that having the `micromamba` executable in the path on unix systems would not collide with the shell functions specifically because there is no specific extension. I'll check that next week!

maresb:
@adriendelsalle, in Bash the shell function always has priority over `PATH`.

adriendelsalle (approved):

wolfv:
Thanks!

PR #1363, opened 2021-12-21
misc: experimental is logged twice

baszalmstra:
I noticed that in my `--json` runs the line:

```
Experimental mode enabled
```

was outputted. I simply removed this line because its also logged as a warning. 

I only use the experimental mode to be able to use authentication tokens. Any idea when that will become stable?


wolfv:
Thanks! 

We haven't worked any further on the token bits in a while. I think we can move them out of the experimental flag! The main thing I was worrying about is token leakage to the log output.

Issue #1375, opened 2022-01-13
micromamba: make max_parallel_downloads configurable


wolfv:
And set it to a higher value on CI with fast internet connection.

PR #1377, opened 2022-01-14
Make max download threads configurable

adriendelsalle:
Description
---

Makes the maximum download threads configurable.
@wolfv do you want to automatically set the default value to something higher (10?) when `on_ci` is true?

Closes https://github.com/mamba-org/mamba/issues/1375


libmamba/include/mamba/api/configuration.hpp
```diff
@@ -317,6 +317,28 @@ namespace mamba
         };
 
 
+        template <>
+        struct cli_config<std::uint16_t>
```
wolfv, line 321: in order to keep binary size smaller, why not use std::size_t or another type that's already being used?

libmamba/include/mamba/api/configuration.hpp
```diff
@@ -317,6 +317,28 @@ namespace mamba
...
+        {
+            using value_type = std::uint16_t;
+            using storage_type = std::int16_t;
+
+            cli_config(const storage_type& value)
+                : m_value(value){};
+
+            bool defined()
+            {
+                return m_value > -1;
```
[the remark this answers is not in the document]
adriendelsalle: not the storage ;)

libmamba/include/mamba/api/configuration.hpp
adriendelsalle: yes we can do that

PR #1380, opened 2022-01-14
Add environment variable to disable low speed limit

### LINKED ISSUE #1361: Add option to disable slow connection aborts
We are using `mamba` in a setting behind a corporate proxy that also scans the content for viruses. This is done in large chunks. While this process is not very performant at all times, this can sometimes lead the connection to be stalling for 1-3 minutes. Currently, requests get aborted due to  
https://github.com/mamba-org/mamba/blob/bab7ae50d75f067c493b8d0cd065ea29bd3f9e8c/libmamba/src/core/fetch.cpp#L172-L173 While this is a sane default behaviour, it would be nice to have a switch to disable it.

xhochy:
Fixes #1361


wolfv:
Thanks!

wolfv:
Thanks!

PR #1406, opened 2022-01-27
add live stream and use prefix_command to get all remaining args

wolfv:
@JohanMabille this does what we need -- extracting all remaining arguments after the first unknown one.
I also added the `--live-stream` option (default to true) to get the `stdout` and `stderr` on the screen. We might want to check the options here and see how to map them best to `reproc`.

### DIFF
--- a/micromamba/src/run.cpp
+++ b/micromamba/src/run.cpp
@@ -12,30 +12,43 @@ set_run_command(CLI::App* subcom)
 {
     init_prefix_options(subcom);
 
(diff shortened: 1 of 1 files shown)


jonashaag:
Actually what you've implemented is more similar to `--no-capture-output`. The `--live-stream` option doesn't make any sense to me in the first place.

Compare impls of https://github.com/conda/conda/pull/9646 and https://github.com/conda/conda/pull/10270

PR #1412, opened 2022-01-28
fix error printing when computing sha sum of broken symlink

wolfv:


wolfv:
I've seen "error" printed sometimes when computing the SHA256 sum of a broken symlink.

Let's check if fs::exists is true first before attempting to compute the SHA256 and otherwise just default to the empty SHA256.

Issue #1393, opened 2022-01-25
Micromamba: Support environment variables in `.condarc`


jonashaag:
Eg. if you have `${PASSWORD}` in your `.condarc`, works with Conda/Mamba but not Micromamba.

wolfv:
We could pre-process the rc files and find & replace all instances of ${...} with the given env var.

jonashaag:
We currently use `envsubst < condarc-with-vars > condarc-without-vars`

wolfv:
makes sense. I think it wouldn't be to hard to come up with a std::regex and then match all occurences. we do have the `getenv` machinery already ...

jonashaag:
For reference, here is the source code of `envsubst`. It's really really simple. https://github.com/autotools-mirror/gettext/blob/1f1c6330c688a00d8f1b8b41b16a11ba4037e241/gettext-runtime/src/envsubst.c#L464

jonashaag:
I've had a look at implementing this. We could do it in a very simplistic manner by changing `YAML::Load(f)` to `YAML::Load(substitute_env_vars(read_file(f)))`. But I think we should do it a little bit more sophisticated by introducing something like `expandvars` in Conda: https://github.com/conda/conda/blob/fc56fdeb2931370845edfd14453ff738d06662dc/conda/base/context.py#L212. It was unable to identify the right place to put this logic though.

wolfv:
@jonashaag I think if you make a C++ function that uses a regex to parse `some test ... ${MYVAR} blabla` and properly finds `${MYVAR}` and replaces it from an env var that would be a great first step. It doesn't yet have to be fully integrated as long as you can add a test e.g. from the C++ side.

jonashaag:
I built a clone of `os.path.expandvars` (which is used by Conda). Turned out to be much more complicated than I thought, but it should be (almost) identical to Conda, whereas a regex-based solution couldn't achieve that. Will clean the code but here's a sneak peek:

```c++
std::string expandvars(std::string s)
{
    auto res = std::vector<std::string>();

    while (s.size())
    {
        const char c = s[0];
        bool is_percent = ON_WIN && c == '%';
        if (ON_WIN && c == '\'') {
            // No expansion within single quotes (Windows only)
            auto j = s.find("'", 1);
            if (j == std::string::npos) {
                res.push_back(s);
                break;
            } else {
                res.push_back(s.substr(0, j + 1));
                s = s.substr(j + 1);
            }
[...]

PR #1416, opened 2022-01-28
Bunch of fixes

jonashaag:
Closes https://github.com/mamba-org/mamba/issues/1393

@wolfv Feel free to use this as a starting point to integrate parsing properly.


micromamba/tests/test_config.py
```diff
@@ -198,28 +201,28 @@ def test_list_with_sources(self, rc_file, source_flag):
         home_folder = os.path.expanduser("~")
         src = f"  # '{str(rc_file).replace(home_folder, '~')}'"
         assert (
-            config(
-                "list", "--no-env", "--rc-file=" + str(rc_file), source_flag
-            ).splitlines()[:-1]
+            config("list", "--no-env", "--rc-file", rc_file, source_flag).splitlines()[
```
jonashaag, line 204: Oops didn't mean to commit those black fmt

jonashaag:
Test failure on Windows is correct because we are applying the parsing logic on the entire file rather than the parsed strings currently.

wolfv:
I think this looks great! 

I am not so happy that we have a different potential format for Windows (using `%...%`) when I think this should ideally always be cross-platform. I guess this is the way it is in Python / conda? I would be happy to advise everyone to just use `${ ... }` which would work on all platforms and drop `%...%` support?

I'll have a deeper look at this + the tests on Monday :)

jonashaag:
Yes this is how it works in Conda (and `os.path.expandsvars`). It's not actually unit-tested for there I think.

We could also not implement Windows-style env vars and throw a helpful error message if we find any. But the parsing code works well with Windows-style env vars so we might as well keep it.

I'll add a test for Posix-style env vars that covers what I've described in https://github.com/mamba-org/mamba/pull/1416#issuecomment-1024908628

jonashaag:
Added test case to fails on Posix. It's a really edgy edge case though, it requires a crazy env var name like `x", "y`

jonashaag:
I think we should be safe to keep the approach I've taken to substitute over the entire raw YAML file (rather than applying to specific parsed options only, as done by Conda) if we disallow YAML special character inside `${...}`. As per `os.path.expandvars`, any character inside `${...}` is allowed except for `}`.

libmamba/src/api/configuration.cpp
```diff
@@ -25,6 +25,113 @@
...
+                auto j = win_percent ? s.find("%", 1) : s.find("}", 2);
+                if (j == std::string::npos)
+                {
+                    res.push_back(s);
+                    break;
+                }
+                else
+                {
+                    auto var = s.substr(win_percent ? 1 : 2, j - (win_percent ? 1 : 2));
+                    auto val = std::getenv(var.c_str());
```
wolfv, line 79: I think it might be nice to do a preprocessing before calling `std::getenv` to make sure that `var` only contains ASCII chars and no weird things like space or `"`. If that is the case we can just return the un-modified version.

libmamba/src/api/configuration.cpp
```diff
@@ -25,6 +25,113 @@
...
+            }
+            else
+            {
+                // s.size() >= 2
+                size_t j;
+                for (j = 1; j < s.size() && isalpha(s[j]); ++j)
+                    ;
+                auto var = s.substr(1, j - 1);
+                auto val = std::getenv(var.c_str());
+                if (val == NULL)
```
wolfv, line 100: `NULL` is C, in C++ `nullptr` is slightly nicer :)

libmamba/src/api/configuration.cpp
```diff
@@ -25,6 +25,113 @@
...
+                s = s.substr(j + 1);
+            }
+            else
+            {
+                // s.size() >= 2
+                size_t j;
+                for (j = 1; j < s.size() && isalpha(s[j]); ++j)
+                    ;
+                auto var = s.substr(1, j - 1);
+                auto val = std::getenv(var.c_str());
```
wolfv, line 99: I think we have a function in `env::get("...")` to retrieve environment variables

wolfv: https://github.com/mamba-org/mamba/blob/26a5b678b740090c5804f45fa5bb9c33b806de34/libmamba/include/mamba/core/environment.hpp#L49

wolfv (commented):
Looks great, couple small suggestions!

wolfv:
I am really wondering if we should just remove support for `%%var%%` and `$$var$$` as well as `%var%` and _only_ support `${var}`.

The reasoning is that it's always easy to add features, but hard to remove them. And we can much easier prove correctness if we keep the "surface" small.
And IMHO people should really not use a platform specific syntax anyways.

jonashaag:
@wolfv Thanks for the review, feel free to make more C++ coding style suggestions, it's one of my weakest languages :)

Re: vars style, currently `%var%` is supported on Windows, and `$var` and `${var}` on all systems. `$$var$$` is not supported. `$$/%%` means literal `$/%` (escaping). I'll make a version that drops `%var%` support.

libmamba/src/api/configuration.cpp
jonashaag: Note that Posix supports variables using non-ASCII chars, and as a consequence so do `os.path.expandsvars` and Conda. Should not be something that's widely used though.

jonashaag: Would you prefer throwing an error on non-ASCII chars or dealing with them the way we deal with unset vars? (= Keep the original string)

jonashaag:
@wolfv the code advances the position in the string using `s = s.substr(off)`. We could also do it by keeping an index and always working relative to that index, ie. `index += off` and then `s.find("x")` -> `s.find("x"< index)`. What's nicer?

wolfv:
What do you think about only supporting `${...}`? We might also have fewer issues with escaping and less surprised users?
Regarding ASCII characters -- yeah, not sure. We could still support them but filter a couple of weird cases like space or `'"` or `\t`? Or we let `getenv` handle this for us :)

jonashaag:
@xhochy any insights into whether supporting only `${var}` (not `$var`) has any drawbacks?

xhochy:
> @xhochy any insights into whether supporting only `${var}` (not `$var`) has any drawbacks?

Not that I know of.

PR #1421, opened 2022-01-31
Include credentials for default channel URLs

### LINKED ISSUE #1364: Mamba doesn't use auth info for default channels
Hi, I had already commented on an [older issue with libcurl](https://github.com/mamba-org/mamba/pull/1131#issuecomment-990294246), but it seems that my issue is slightly different, although it shows the same symptoms.

## My problem

I faced several download errors, showing up like this:

```
      File "c:\home\hub\mamba\mamba\mamba\mamba.py", line 486, in install
        index = load_channels(pool, channels, repos)
      File "c:\home\hub\mamba\mamba\mamba\utils.py", line 122, in load_channels
        index = get_index(
      File "c:\home\hub\mamba\mamba\mamba\utils.py", line 103, in get_index
        is_downloaded = dlist.download(True)
    RuntimeError: Multi-download failed.
```

It took me a while to figure out whether the problem lies on the libmamba/libcurl side or in mamba/conda.
Eventually, I was able to track it down to a **miscommunication between mamba and conda**.

In `utils.py`, mamba retrieves the default channels from conda:
https://github.com/mamba-org/mamba/blob/bab7ae50d75f067c493b8d0cd065ea29bd3f9e8c/mamba/mamba/utils.py#L249

Unfortunately, the `x.base_url` here doesn't include the `x.auth` info (in case there is any), so any credentials in the default channels are dropped. This eventually leads to the multi download errors, originating from 401 HTTP errors.

## How to reproduce

To reproduce the issue, just use a channel URL in your default channels which requires authentication. Even when the credentials are given in the default channels list, the server will not receive any credentials in the request.

## Possible solution

I was able to fix this broken behavior by manually adding the auth info into the base URL, but I am not very confident that my code is the best way to do it... I also looked for a method on the channel object returned from conda, but it seems that there is currently no method to retrieve the `base_url` with credentials (unlike for example the `.url(with_credentials)` method of the channel object).

Below is my patch which should replace the line referenced above (see the last line of the following snippet).

```python
    # The channels from `conda.base.context.default_channels` may contain
    # `.auth` info. But mamba doesn't use the `.auth` info, but simply takes
    # the `.base_url` and ignores the `.auth` info.
    # This leads to errors if default channels have credentials in the URL.
    def get_base_url_with_auth(c):
        if not c.auth:
            return c.base_url
        else:
            from urllib.parse import urlparse
            parts = urlparse(c.base_url)
            return parts._replace(netloc=c.auth + '@' + parts.netloc).geturl()
    api_ctx.default_channels = [get_base_url_with_auth(x) for x in context.default_channels]
    #api_ctx.default_channels = [x.base_url for x in context.default_channels]
```

## Next steps

How should I continue? I assume that my fix above needs some polishing, so should I prepare a PR for my fix? Or do you have another idea on how to fix the issue?

wulmer:
Fixes #1364.

In our organization, we are using default channels from an internal Artifactory mirror, which requires that default channel URLs have credentials.

This PR adds the missing credential information to the default channel URLs.

For recreating the bug, set your default channel to a URL with credentials and check to what URL mamba connects. Before this PR, mamba would connect to the URL without credentials, resulting in fetching errors. With this PR, the correct URL including the credentials is used.

### DIFF
--- a/mamba/mamba/utils.py
+++ b/mamba/mamba/utils.py
@@ -246,7 +246,9 @@ def init_api_context(use_mamba_experimental=False):
                 )
     api_ctx.custom_multichannels = additional_custom_multichannels
 
-    api_ctx.default_channels = [get_base_url(x.url(with_credentials=True)) for x in context.default_channels]
+    api_ctx.default_channels = [
+        get_base_url(x.url(with_credentials=True)) for x in context.default_channels
+    ]
 
     if context.ssl_verify is False:
         api_ctx.ssl_verify = "<false>"
--- a/mamba/mamba/utils.py
+++ b/mamba/mamba/utils.py
@@ -246,7 +246,7 @@ def init_api_context(use_mamba_experimental=False):
                 )
     api_ctx.custom_multichannels = additional_custom_multichannels
 
-    api_ctx.default_channels = [x.base_url for x in context.default_channels]
+    api_ctx.default_channels = [get_base_url(x.url(with_credentials=True)) for x in context.default_channels]
 
     if context.ssl_verify is False:
         api_ctx.ssl_verify = "<false>"


wolfv:
Thanks for the contribution!

Issue #1148, opened 2021-08-31
0.14.1 (and 0.15.3) selects incorrect package


as-ec:
When resolving dependencies mamba appears to select the first match rather than the most specific match when a local version label is used. 

```
$ conda  search -c https://artifactory.epoch.capital/artifactory/api/conda/epoch/gtp/release infra=6178
Loading channels: done
# Name                       Version           Build  Channel             
infra                    6178+LABEL 7fbc86c99a385335151c8392aa1ac195fc330294  artifactory/api/conda/epoch/gtp/release
infra                           6178 7fbc86c99a385335151c8392aa1ac195fc330294  artifactory/api/conda/epoch/gtp/release
```
In the above `mamba search` generates the same.

When installing mamba picks the local labelled version, not the version 6178.
```
mamba install -q -n md-proxy -c https://artifactory.epoch.capital/artifactory/api/conda/epoch/gtp/release -c https://artifactory.epoch.capital/artifactory/api/conda/conda-forge infra=6178
[.... ascii noise removed ...]
  + infra        6178+LABEL  7fbc86c99a385335151c8392aa1ac195fc330294  artifactory.epoch.capital/artifactory/api/conda/epoch/gtp/release/linux-64         61 MB
```

with conda 6178 is selected:

```
conda install -q -n md-proxy-c https://artifactory.epoch.capital/artifactory/api/conda/epoch/gtp/release -c https://artifactory.epoch.capital/artifactory/api/conda/conda-forge infra=6178

[.... ascii noise removed ...]
  infra              artifactory/api/conda/epoch/gtp/release/linux-64::infra-6178-7fbc86c99a385335151c8392aa1ac195fc330294
```

We're using the `+` here as conda "acnowledges" [PEP-440](https://www.python.org/dev/peps/pep-0440) for versions which allows for a local version label: `<public version identifier>[+<local version label>]`

wolfv:
@as-ec the code that is supposed to do these comparisons is here: https://github.com/openSUSE/libsolv/blob/3d88b0ca5188e8d4a180ff30e7c183f71a367d0c/src/conda.c#L206-L211

Would be great if you could tackle this in libsolv! 

In general you could also set the build number to something higehr and that should work?

wolfv:
maybe your problem is that only one of the two packages has a local version, and that that's handled wrongly.

as-ec:
> maybe your problem is that only one of the two packages has a local version, and that that's handled wrongly. 

[...]

Issue #1206, opened 2021-10-10
Add flag to *not* compile noarch packages at install time


dhirschfeld:
As discussed on [gitter](https://gitter.im/conda-forge/conda-forge.github.io?at=6162b280cd4972068b3c1312), this could potentially speed up CI by only compiling what you use at runtime.

dhirschfeld:
This may be particularly relevant on Windows as I think IO on Windows is particularly slow

PR #1422, opened 2022-01-31
Faster pyc compilation

### LINKED ISSUE #1413: Installation is notably slower when running in an active environment
While working on the better `pyc` compilation patch I discovered that installation is slowed by being in an environment.

This is caused by [`micromamba acitvate` being called repeatedly](https://github.com/mamba-org/mamba/blob/32e5056cbf80aa8fae745be102044fd996152036/libmamba/src/core/util.cpp#L1415-L1423). In combination with a relatively slow set of deactivate scripts (~300ms) this compounds to being a large effect (see reproducer).

I'm not sure what might trigger this, maybe environments with many postlink scripts?

This will be made much better by the batched `pyc` compilation but perhaps it should still be addressed by cleaning the environment for subprocess?

### Reproducer

```bash
# Create a test environment with the compilers package from conda-forge
# It contains relatively heavy deactivation scripts
mamba create --name test compilers
# Populate the caches with a package which contains lots of noarch: python
time /path/to/micromamba create --prefix /tmp/test-env celery --yes
```

Without an environment active:

```
$ time /path/to/micromamba create --prefix /tmp/test-env celery --yes
...
________________________________________________________
Executed in    4.56 secs    fish           external
   usr time    3.82 secs  553.00 micros    3.82 secs
   sys time    0.67 secs    0.00 micros    0.67 secs
```

With the test environment active:

```
$ conda activate test
$ time /path/to/micromamba create --prefix /tmp/test-env celery --yes
...
________________________________________________________
Executed in    9.83 secs    fish           external
   usr time    8.76 secs  505.00 micros    8.76 secs
   sys time    1.70 secs   50.00 micros    1.70 secs
```

chrisburr:
This PR included all but one of the changes I've mentioned on gitter. My C++ knowlege is lacking so I'm not sure how idiomatic this PR is, feel free to sugest major changes. I'm also happy for someone else to take over this PR if it's easier.

Fixes #1413

## Performance comparision

I picked a few packages I'm familiar with to use for benchmarking here:

```
micromamba create --prefix /tmp/test-env --yes celery pandas tqdm sqlalchemy uproot scikit-hep jupyterlab
```

I've ran the command several times to make sure the caches were populated.

I've also tried various other sets of packages, installers generated using conda-constructor and running on slower filesystems (NFS, Ceph volumes, ...). All of my test cases have been significantly faster.

### Lastest micromamba release from inside an exisiting environment 

Slowed by calling `deactivate` (see #1148)

```
________________________________________________________
Executed in   45.73 secs    fish           external
   usr time   41.04 secs  610.00 micros   41.04 secs
   sys time    8.31 secs  193.00 micros    8.31 secs
```

### Lastest micromamba release outside an exisiting environment 

```
________________________________________________________
Executed in   19.87 secs    fish           external
   usr time   16.14 secs    1.67 millis   16.13 secs
   sys time    3.48 secs    0.00 millis    3.48 secs
```

### This PR

```
________________________________________________________
Executed in    9.57 secs    fish           external
   usr time   11.92 secs  800.00 micros   11.92 secs
   sys time    1.81 secs    0.00 micros    1.81 secs
```


libmamba/src/core/transaction_context.cpp
```diff
@@ -3,13 +3,37 @@
...
 #include "mamba/core/transaction_context.hpp"
 #include "mamba/core/output.hpp"
 #include "mamba/core/util.hpp"
 
 namespace mamba
 {
+    void compile_python_sources(std::ostream& out)
+    {
+        out << "from compileall import compile_file\n";
+        out << "from concurrent.futures import ProcessPoolExecutor\n";
```
wolfv, line 21: Hmm, this was introduced in python 3.2

I am not sure we should break compatibility with older pythons, especially 2.7 might still be used (at least to recreate historical envs).

wolfv:
Super exciting! I will give this a proper review tomorrow :)

libmamba/src/core/transaction_context.cpp
chrisburr: Definitely (I'm still forced to Python 2.7 myself).

I have a fallback to use `-m compileall` for `Python < 3.5` (basically the same condition that used to be used for adding `-j0` before).

Worst case scenario is that the pyc files fail to be installed and will be generated on first import (most of the time at least, permissions or read only filesystems might mess things up).

jonashaag: Why not simply use `multiprocessing`?

I wonder if we should add batching because the number of pyc files may easily grow to 100k and this will create as many jobs as pyc files.

jonashaag: We might also want to expose this Python code and not hide it here in the C++ code so that people can use it outside the install process.

libmamba/src/core/package_handling.cpp
```diff
@@ -461,6 +461,13 @@ namespace mamba
         }
     }
 
+    fs::path extract_subproc(const fs::path& file)
```
jonashaag, line 464: Dead code?

libmamba/src/core/transaction_context.cpp
chrisburr: > Why not simply use multiprocessing?

I started from the [implementation in cpython](https://github.com/python/cpython/blob/ee0ac328d38a86f7907598c94cb88a97635b32f8/Lib/compileall.py#L102-L113) which uses `concurrent.futures` and saw not reason not to use it.

> I wonder if we should add batching because the number of pyc files may easily grow to 100k and this will create as many jobs as pyc files.

With [`max_workers=None`](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ProcessPoolExecutor) `ProcessPoolExecutor` uses the number of cores and takes care of the batching for sending data the subprocesses.

> We might also want to expose this Python code and not hide it here in the C++ code so that people can use it outside the install process. We could even make it compatible to the command-line interface of compileall

It already exists in cpython however it:
* Waits for stdin to be closed before compilation starts
* It only parallelises when recursing into directories

I'm inclined to improve upstream rather than focusing on it to much here so that `-m compileall` can be used for Python 3.11+ if they approve of the change.

libmamba/src/core/package_handling.cpp
chrisburr: Yup, this is part of my less-mergable follow-up PR and got caught in the wrong commit. I'll amend the commit tomorrow.

libmamba/src/core/transaction_context.cpp
jonashaag: Yes that sounds like a much better plan

jonashaag: FYI with `multiprocessing` you don't get the batching for free, you'll have to do it manually.

dhirschfeld:
*Aside:* `plotly` is a good package to test `pyc` compilation times as it has historically taken a long time (xref: #1206)

wolfv:
I tried running this but in verbose mode I see the following output:

```
      File "/Users/wolfvollprecht/micromamba/envs/testpyc/lib/python3.10/multiprocessing/popen_spawn_posix.py", line 42, in _launch
        raise RuntimeError('''
    RuntimeError:
            An attempt has been made to start a new process before the
            current process has finished its bootstrapping phase.

            This probably means that you are not using fork to start your
            child processes and you have forgotten to use the proper idiom
            in the main module:

                if __name__ == '__main__':
                    freeze_support()
                    ...

            The "freeze_support()" line can be omitted if the program
```

And no `pyc` files are compiled. 

Any ideas? :)

wolfv:
I modified the script slightly:

```
from compileall import compile_file
from concurrent.futures import ProcessPoolExecutor
import sys

def main():
    results = []
    with sys.stdin:
        with ProcessPoolExecutor(max_workers=None) as executor:
            while True:
                name = sys.stdin.readline().strip()
                if not name:
                    break
                results.append(executor.submit(compile_file, name, quiet=1))
            success = all(r.result() for r in results)
    return success

if __name__ == "__main__":
    success = main()
    sys.exit(int(not success))
```

And that seems to improve the situation (testing on M1 osx-arm64)

chrisburr:
Ah yes, I forgot macOS and windows not properly supporting `fork` ([bpo-33725](https://bugs.python.org/issue33725)). Thanks for fixing it.

wolfv:
@chrisburr did you have a chance to test this PR on Windows? Not sure, but it seems to hang indefinitely.

chrisburr:
I have no access to a windows machine unfortunately. I might be able to spin up a VM it’ll probably take me a while to figure out how to get an environment up and running.

Does it work if you try installing a python 2.7 environment? (i.e. triggering the fallback to using `-m compileall`)

wolfv:
OK, no worries, I can have a look later on the conda-forge Windows VM.

wolfv:
@chrisburr I was under the impression that the python process should terminate when it receives an "empty" file (e.g. `""`).

However, I can't find a place where we would send such an empty string via stdin. Am I looking at it wrong? 

The Windows process seems stuck waiting at the "draining" point.

wolfv:
@chrisburr you also used `options.env.behavior = reproc::env::empty;` which is usually really nice. The downside is that a dynamically linked micromamba doesn't (at least on Windows) because it can't find DLLs without the proper `%PATH%`. 

I wanted to see if that influencees the tests here.

wolfv:
I did actually find this isseu on reproc which sounds similar to what we're experiencing here: https://github.com/DaanDeMeyer/reproc/issues/41

You may well be right, btw. I need to debug further to see what's going on on Windows.

wolfv:
Locally on Windows this seems to run fine, btw! Let's see if we get more enlightening info out of this CI run

micromamba/src/constructor.cpp
```diff
@@ -97,6 +97,7 @@ construct(const fs::path& prefix, bool extract_conda_pkgs, bool extract_tarball)
                 std::string pkg_name = index["name"];
 
                 index["fn"] = entry.path().filename();
+                bool found_match = false;
```
wolfv, line 100: Is this also part of the other PR?

wolfv:
Looks like the removal of the environment cleaning has fixed this. I think that's only because we are linking against `dll's` in the tests and if micromamba dependencies are not found at runtime, the `micromamba activate` calls don't work.

I guess there are two workarounds:

- Put everything necessary for activation inside the script generated by wrapped_subprocess call so that micromamba doesn't even need to be runnable for that call -- then we should be able to run with a completely cleaned env
- keep the current behavior with non-cleaned env on Windows and only when we're in a dynamic link mode (controlled by a flag from CMake)

It looks like this is working really great. I did some checking if all files are compiled and "it looks like it" but maybe we shoudl add tests?

And is this also doign the proper bookkeeping in `conda-meta/my-pkg-123.321-hash.json`? Adding all compiled files there and SHA256 sums?

Actually, in the future we could put the compiled files as part of the `pkgs` cache folder and that would make recreating environments with the same Python version even faster (or at least less CPU hungry!).

chrisburr:
> I guess there are two workarounds:

Would inheriting only inheriting the `PATH` variable be good enough? 

> It looks like this is working really great. I did some checking if all files are compiled and "it looks like it" but maybe we shoudl add tests?

Definitely, I can try to look at this. Actually I've already found a bug, non-cpython bytecode isn't handled correctly: https://github.com/mamba-org/mamba/blob/5961d76afdd8b0f070bf0f2da396ef25289c965c/libmamba/src/core/link.cpp#L81-L82

> And is this also doign the proper bookkeeping in conda-meta/my-pkg-123.321-hash.json? Adding all compiled files there and SHA256 sums?

I don't think they're computed for these files?

```json
$ cat ~/mambaforge/conda-meta/sphinx-4.3.1-pyh6c4a22f_0.json
...
      {
        "_path": "lib/python3.9/site-packages/sphinx/writers/__pycache__/text.cpython-39.pyc",
        "path_type": "pyc_file"
      },
      {
        "_path": "lib/python3.9/site-packages/sphinx/writers/__pycache__/xml.cpython-39.pyc",
        "path_type": "pyc_file"
      },
...
```

chrisburr:
Actually should these files always be added to `conda-meta/my-pkg-123.321-hash.json` even if compilation fails or is disabled? The reasoning being that they'll be automatically generated regardless in most cases and we want to make sure they're cleaned up.

wolfv:
true, i just checked conda's behavior and you're right, they don't record the SHA256 for those files (so we don't have to, either).

Regarding recording them even if compilation doesn't succeed -- tough question. I wish there was a specification :)
I think historically we haven't recorded them if compilation fails, but I would need to double check also conda's behavior there.

wolfv:
Also, yes, I think inheriting `PATH` only should be good enough!

chrisburr:
> Regarding recording them even if compilation doesn't succeed -- tough question. I wish there was a specification :)
I think historically we haven't recorded them if compilation fails, but I would need to double check also conda's behavior there.

I checked the code and it appears to always register them.

To check for sure, try installing for ppc64le:

```bash
$ CONDA_SUBDIR=linux-ppc64le conda create --name test python celery -v
...
pyc file failed to compile successfully (run_command failed)
python_exe_full_path: /home/cburr/miniconda3/envs/test/bin/python3.9
py_full_path: /home/cburr/miniconda3/envs/test/lib/python3.9/site-packages/prompt_toolkit/layout/utils.py
pyc_full_path: /home/cburr/miniconda3/envs/test/lib/python3.9/site-packages/pip/_vendor/chardet/__pycache__/jpcntx.cpython-39.pyc
compile rc: 255
compile stdout:
compile stderr: qemu-ppc64le-static: Could not open '/lib64/ld64.so.2': No such file or directory
...


$ ls /home/cburr/miniconda3/envs/test/lib/python3.9/site-packages/pip/_vendor/chardet/__pycache__/jpcntx.cpython-39.pyc
ls: cannot access '/home/cburr/miniconda3/envs/test/lib/python3.9/site-packages/pip/_vendor/chardet/__pycache__/jpcntx.cpython-39.pyc': No such file or directory


$ rg -C 3 jpcntx.cpython-39.pyc /home/cburr/miniconda3/envs/test/conda-meta/
/home/cburr/miniconda3/envs/test/conda-meta/pip-22.0.2-pyhd8ed1ab_0.json
721-    "lib/python3.9/site-packages/pip/_vendor/chardet/__pycache__/gb2312prober.cpython-39.pyc",
722-    "lib/python3.9/site-packages/pip/_vendor/chardet/__pycache__/hebrewprober.cpython-39.pyc",
723-    "lib/python3.9/site-packages/pip/_vendor/chardet/__pycache__/jisfreq.cpython-39.pyc",
724:    "lib/python3.9/site-packages/pip/_vendor/chardet/__pycache__/jpcntx.cpython-39.pyc",
725-    "lib/python3.9/site-packages/pip/_vendor/chardet/__pycache__/langbulgarianmodel.cpython-39.pyc",
726-    "lib/python3.9/site-packages/pip/_vendor/chardet/__pycache__/langgreekmodel.cpython-39.pyc",
727-    "lib/python3.9/site-packages/pip/_vendor/chardet/__pycache__/langhebrewmodel.cpython-39.pyc",
--
5484-        "path_type": "pyc_file"
5485-      },
5486-      {
5487:        "_path": "lib/python3.9/site-packages/pip/_vendor/chardet/__pycache__/jpcntx.cpython-39.pyc",
5488-        "path_type": "pyc_file"
5489-      },
5490-      {
```

micromamba/src/constructor.cpp
chrisburr: It's mostly there as it would have made debugging https://github.com/conda/constructor/pull/487 much easier.

This can be either PR or forgotten about entirely, as you prefer.

maresb:
This is very exciting! I wonder to what extent this helps with the lack of speedup here https://github.com/jupyter/docker-stacks/pull/1213#issuecomment-761517281 which have always baffled me.

wolfv:
@maresb with "regular" mamba we're using the same "linking" methods that conda would use so I think in that particular case nothing should change (for now). But maybe `docker-stacks` wants to switch to micromamba and then .. hopefully they'd see a speed increase :)

maresb:
@wolfv, great points! My recent contributions to micromamba-docker and micromamba have been partially motivated by wanting a docker-stacks replacement. With 0.20.0 we squished some bugs which made bootstrapping mamba with micromamba infeasible, so indeed now may be the time for another attempt.

jonashaag:
FYI `python -c 'import site; print(site.getsitepackages()[0])'`

libmamba/src/core/transaction_context.cpp
```diff
@@ -180,13 +179,8 @@ namespace mamba
         m_pyc_process = std::make_unique<reproc::process>();
 
         reproc::options options;
+#ifndef _WIN32
```
wolfv, line 182: we might be able to not do this with static micromamba builds. We can add a variable or something like that later.

wolfv:
Hmm, looks like the env cannot be solved on Windows because `vc 9.*` is needed for python 2.7. Maybe we used to get that from the `defaults` channel?

wolfv:
I think one last thing we might want to add here is to be able to optionally set a number of threads to be used. It does currently max out the CPU and on some systems that's not great (e.g. if they are constrained cloud workers etc).

We could use the same computation as for `extract_threads`

wolfv:
Let's see how this holds up in real life! :)

baszalmstra:
Once this is released we’ll start testing it right away!

PR #1423, opened 2022-01-31
Micromamba: Substitute env vars in .condarc

jonashaag:
New version of https://github.com/mamba-org/mamba/pull/1416 with only `${var}` support (and `$$` support)

### DIFF
--- a/libmamba/src/api/configuration.cpp
+++ b/libmamba/src/api/configuration.cpp
@@ -21,13 +21,47 @@
 #include "spdlog/spdlog.h"
 
 #include <algorithm>
+#include <regex>
 #include <stdexcept>
 
 #include "termcolor/termcolor.hpp"
 
-
 namespace mamba
 {
+    static std::string expandvars(std::string s)
+    {
+        if (s.find("$") == std::string::npos)
+        {
+            // Bail out early
+            return s;
+        }
+        // Match $$ and ${var} where var does not contains YAML special charaters
+        std::regex env_var_re(R"(\$(?:\$|\{([^\}'\"\s]+)\}))");
+        for (auto matches = std::sregex_iterator(s.begin(), s.end(), env_var_re);
+             matches != std::sregex_iterator();
+             ++matches)
+        {
+            std::smatch match = *matches;
+            auto var = match[1].str();
+            std::string val;
+            if (var.empty())
+            {
+                // $$
+                val = "$";
+            }
+            else
+            {
+                // TODO: change this so that empty but existing env vars are replaced with ""
+                val = env::get(var);
+            }
+            if (!val.empty())
+            {
+                s.replace(match[0].first, match[0].second, val);
+            }
+        }
+        return s;
+    }
+
     namespace detail
     {
         void ssl_verify_hook(std::string& value)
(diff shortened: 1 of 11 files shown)


jonashaag:
@wolfv this should be ready to go with one minor todo where my C++ skills weren't good enough. How do you make a function that can return a `std::string` or "None"? I guess you make it return a pointer, but then where do you allocate the `std::string`? Does `std::string x` allocate the string on the stack or on the heap? If on the heap, how is it deallocated? What's the "modern" approach to heap allocation in C++?

wolfv:
@jonashaag there are two options -- either to use `std::optional<std::string>` or `std::unique_ptr<std::string>`. Both will destroy the string when they go out of scope.

wolfv:
@jonashaag can you have a look to see if this is what you had in mind?

Now when the env var exists, but is empty, it will add an empty string (`""`).

jonashaag:
@wolfv I was about to push my changes at this very moment :D

wolfv:
Nooooooooo! Sorry :/

jonashaag:
https://github.com/jonashaag/mamba/tree/envsubst2-2

Can you review this instead?

libmamba/src/api/configuration.cpp
```diff
@@ -51,8 +51,17 @@ namespace mamba
...
             else
             {
-                // TODO: change this so that empty but existing env vars are replaced with ""
-                val = env::get(var);
+                auto env_var = env::get(var);
+                if (env_var)
+                {
+                    val = env_var.value();
+                    if (val.empty())
+                        val = "\"\"";
```
jonashaag, line 59: Sorry wasn't clear here, should be replaced by empty string

jonashaag:
But I'm happy to see that you made essentially the same changes :) Are you ok with the change in behaviour that empty env vars are now not considered absent anymore? (You also seem to have made that decision in your changes)

wolfv:
I am happy with your version, too. Do you want to force push to this branch to erase my changes and use yours?
I just read them and it's really super similar!

Sorry again, I should've coordinated :)

jonashaag:
No worries. Force pushed.

jonashaag:
@wolfv I don't understand the remaining Windows test failure. It seems that Micromamba thinks that `${foo}` is missing (not only empty!) and thus doesn't replace it. But the `env::get` code should be taking proper care of empty vs missing.

wolfv:
I think 
it's a Windows limitation. Setting empty strings to environment variables seems unsupported

So IMHO we can skip that test on Windows.

Screenshot 2022-02-01 at 21 43 12

jonashaag:
Oh wow, well that's unfortunate but then we can simplify some of the code and skip that test.

micromamba/tests/test_config.py
```diff
@@ -14,19 +16,21 @@ def config(*args):
     cmd = [umamba, "config"] + [arg for arg in args if arg]
     res = subprocess.check_output(cmd)
     if "--json" in args:
-        j = json.loads(res)
+        j = yaml.load(res, yaml.FullLoader)
```
wolfv, line 19: I think you've filed a bug for this already, right?

jonashaag: Yes

wolfv:
Thanks for tackling this!

Issue #1405, opened 2022-01-27
Improve error message if conflicting packages are requested


jonashaag:
```
$ mamba install 'snowflake-connector-python==2.7.2' 'snowflake-connector-python==2.7.3'

Looking for: ['snowflake-connector-python==2.7.2', 'snowflake-connector-python==2.7.3']

Pinned packages:
  - python 3.9.*


Encountered problems while solving:
  - package snowflake-connector-python-2.7.2-py38h5196458_0 requires python >=3.8,<3.9.0a0, but none of the providers can be installed
```

This error message is correct, but a more useful one would be this one that you get if you don't have `python=3.9` pinned:

```
  - cannot install both snowflake-connector-python-2.7.3-py310h1afff1e_0 and snowflake-connector-python-2.7.2-py38he943702_0
```

wolfv:
Yeah, providing better error messages is definitly on our wish list for this year!
I recently found a function in the libsolv API to retrieve _all_ errors from the solver: https://github.com/mamba-org/mamba/blob/4cbfcf1cb403d8919b2d6c0ada39ee26a48524e2/libmamba/src/core/solver.cpp#L409-L423

This function should also return an error for "requesting conflicting specs".

jonashaag:
Thanks for the pointer. Actually it seems that we are getting the "wrong" result from libsolv here. The `solver_problem2str` call does not get the "conflicting requests" error message. It's one of the `j`s here though: https://github.com/mamba-org/mamba/blob/4cbfcf1cb403d8919b2d6c0ada39ee26a48524e2/libmamba/src/core/solver.cpp#L362-L392

Do you think this is something that should be improved in libsolv (better select the "primary" error message)?

wolfv:
Yeah i don't actually know how libsolv selects the error message it prints. 

What I was thinking is that we could merge many error messages and print the more unique ones more prominently.
By merging I mean -- I thought we might have a lot of:

```
Cannot install A together with B 1.1
Cannot install A together with B 1.2
Cannot install A together with B 1.3
...
```

and instead we could print `Cannot install A together with B [1.1, 1.2, 1.3]`

Another thing I thought was that when we're in interactive mode we could ask the user if she/he wants to see additional error messages via `[Y/n]` prompt.

Besides "finding" this function I haven't had a chance to play around with it much more yet. I am very curious what you find!

jonashaag:
See https://github.com/mamba-org/mamba/pull/1424 for some work I did on this

[...]

PR #1424, opened 2022-01-31
Add subproblems printing code

jonashaag:
Lightweight refactoring to be able to print all unique "subproblems". Code is not actually used at the moment. Breaks the (undocumented?) `Solver` Python interface by renaming a function.

### DIFF
--- a/libmamba/include/mamba/core/solver.hpp
+++ b/libmamba/include/mamba/core/solver.hpp
@@ -53,9 +53,11 @@ namespace mamba
         void set_postsolve_flags(const std::vector<std::pair<int, int>>& flags);
         bool is_solved() const;
         bool solve();
-        std::string problems_to_str() const;
         std::vector<std::string> all_problems() const;
-        std::string all_problems_to_str() const;
+        std::vector<std::vector<std::string>> all_subproblems() const;
+        std::string problems_to_str() const;
+        std::string all_subproblems_to_str() const;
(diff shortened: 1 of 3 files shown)


jonashaag:
For the requirements in #1405 `unique_subproblems_to_str` will give something like this


```
  - cannot install both snowflake-connector-python-2.7.3-py39h1790f5e_0 and snowflake-connector-python-2.7.2-py39h1790f5e_0
  - cannot install both snowflake-connector-python-2.7.3-py39h1790f5e_0 and snowflake-connector-python-2.7.2-py39h25baaa7_0
... lots more ...
  - cannot install both snowflake-connector-python-2.7.3-py39h4648c9b_0 and snowflake-connector-python-2.7.2-py39h4648c9b_0
  - conflicting requests
  - package snowflake-connector-python-2.7.2-py310h1afff1e_0 requires python >=3.10,<3.11.0a0, but none of the providers can be installed
  - package snowflake-connector-python-2.7.2-py310h67ee329_0 requires python >=3.10,<3.11.0a0, but none of the providers can be installed
... lots more ...
  - package snowflake-connector-python-2.7.3-py38he943702_0 requires python >=3.8,<3.9.0a0, but none of the providers can be installed
```

jonashaag:
For this particular case we could come up with a useful list of conflicts by removing the build number from the messages before building the unique list.

PR #1427, opened 2022-02-01
Parallel extraction for constructor

chrisburr:
Modified `micromamba constructor --extract-tarball` to use a thread pool to extract the packages in parallel. This probably more of a proof of concept than a ready-to-merge implementation.

I've seen a few mentions on GitHub about importing a thread pool implementation from elsewhere but I don't think this has been done yet? Alternatively this could avoid using threads just manage a collection of `reproc::process` objects directly.

## Benchmarking

Testing on Linux with a local build of Miniforge3 using https://github.com/conda/constructor/pull/487. I've seen even bigger differences on the installer I build for work.

### My desktop:

* `conda-standalone`: 7.47 secs
* Current `master`: 8.35 secs
* #1422: 7.01 secs
* #1422 + this PR: 5.64 secs

### A VM with relatively slow ceph based storage:

* `conda-standalone`: 13.89 secs
* Current `master`: 17.38 secs
* #1422: 14.21 secs
* #1422 + this PR: 9.94 secs

### DIFF
--- a/libmamba/include/mamba/core/package_handling.hpp
+++ b/libmamba/include/mamba/core/package_handling.hpp
@@ -35,6 +35,7 @@ namespace mamba
     void extract(const fs::path& file, const fs::path& destination);
     fs::path extract(const fs::path& file);
     void extract_subproc(const fs::path& file, const fs::path& dest);
+    fs::path extract_subproc(const fs::path& file);
     bool transmute(const fs::path& pkg_file, const fs::path& target, int compression_level);
     bool validate(const fs::path& pkg_folder);
 }  // namespace mamba
--- a/libmamba/src/core/package_handling.cpp
+++ b/libmamba/src/core/package_handling.cpp
@@ -461,6 +461,13 @@ namespace mamba
         }
     }
 
+    fs::path extract_subproc(const fs::path& file)
+    {
+        fs::path dest_dir = extract_dest_dir(file);
+        extract_subproc(file, dest_dir);
+        return dest_dir;
+    }
+
     bool transmute(const fs::path& pkg_file, const fs::path& target, int compression_level)
     {
         TemporaryDirectory extract_dir;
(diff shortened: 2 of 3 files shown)


chrisburr:
I'd missed that entirely while trying this out. That sounds great! I'm going to do some reading about what a semaphore is 😅

wolfv:
@chrisburr do you know wether the constructor has a "explicit env" spec? Especially wether the packages are in the correct order?
We may be able to re-use code for explicit environments for this then and get "everything" for free.

chrisburr:
I think I'm not completely following the question but hopefully this answers it 🤞 

This PR is about parallelising this command
```bash
$PREFIX/conda.exe constructor --prefix "$PREFIX" --extract-conda-pkgs
```

The install is later done with:
```
$PREFIX/conda.exe install --offline --no-pyc --file $PREFIX/pkgs/env.txt -yp "$PREFIX"
```

where `pkgs/env.txt` looks like:
```
python=3.9.10=h85951f9_1_cpython
_libgcc_mutex=0.1=conda_forge
ca-certificates=2021.10.8=ha878542_0
font-ttf-dejavu-sans-mono=2.37=hab24e00_0
font-ttf-inconsolata=3.000=h77eed37_0
font-ttf-source-code-pro=2.038=h77eed37_0
font-ttf-ubuntu=0.83=hab24e00_0
ld_impl_linux-64=2.36.1=hea4e1c9_2
libgfortran5=11.2.0=h5c6108e_12
libstdcxx-ng=11.2.0=he4da1e4_12
mysql-common=8.0.28=ha770c72_0
tzdata=2021e=he74cb21_0
```

wolfv:
Ok, makes sense! A constructor is pretty similar to a "explicit" environment where the packages would be found using (relative) `file://...` URLs so I was hoping we could re-use the same code.

However, they do have a couple of different steps in there that makes it necessary for us to write additional code :)

I had a quick look the other day but also didn't find a way to simplify code further here -- yet!

PR #1430, opened 2022-02-01
Get clang-format from PyPI and version pre-commit hook

chrisburr:
Currently the `clang-format` version used by pre-commit isn't versioned so the output isn't stable. I quite like https://github.com/ssciwr/clang-format-wheel for this as it's a <2MB download and users don't need to worry having `clang-format` installed, it's all automatically managed by pre-commit.

Feel free to close if you'd prefer to do it another way 😃

### DIFF
--- a/.github/workflows/linters.yml
+++ b/.github/workflows/linters.yml
@@ -32,7 +32,7 @@ jobs:
     - name: Install dependencies
       shell: pwsh
       run: |
-        conda install clang clang-tools=11.1.0 pre-commit
+        conda install pre-commit
     - name: Conda info
       shell: pwsh
       run: |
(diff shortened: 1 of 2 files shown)


wolfv:
While I don't think that "clang-format" on PyPI is the proper way to package it ... I can see the benefit here :) We should add micromamba support to pre-commit, hehe!

xhochy:
> While I don't think that "clang-format" on PyPI is the proper way to package it ... I can see the benefit here :) We should add micromamba support to pre-commit, hehe!

pre-commit has micromamba support! Use it here 😎

chrisburr:
> pre-commit has micromamba support! Use it here 😎

I want to agree with you but unfortunately clang-format from conda-forge uses 604MB vs pip+venv which only uses 18MB. Even with a fast home directory it takes more than twice and long to initialise (and with NFS it takes 10+ times longer).

(obviously this could be improved on the conda-forge side)

wolfv:
Wow, nice @xhochy didn't know that! 

Yeah, packaging is hard! :)

PR #1433, opened 2022-02-02
WIP ccache improvements

jonashaag:


wolfv:
I'd be happy to hear about your findings :)

jonashaag:
Ok so I've tried two things:

First, use https://github.com/hendrikmuhs/ccache-action because it updates the cache after each test run, thus you get incremental cache updates. Doesn't make much of a difference though.

Second, using it on Windows, which would give a huge speedup. But ccache does not support MCSV. It has experimental support in the upcoming (unreleased) version. I'll re-try when that is released.

wolfv:
Yeah, I also did spend time to actually get `ccache` to build on conda-forge on Windows, only to be disappointed that there is no MSVC support (yet)!

Really looking forward for it to get there, though!

PR #1434, opened 2022-02-02
play around with detaching of process

wolfv:
This allows detaching the `micromamba run` from the terminal. It works nicely, however we might have to deal with signals better or figure something out.

```
micromamba create -n jupyter notebook
micromamba run --daemon -n jupyter jupyter notebook
```

We print the correct PID fo the `micromamba run` command, but the process started with `reproc` has another PID.
When running `kill -s TERM PID` the `micromamba run` process is properly killed. However, the `reproc` process continues (and e.g. jupyter continues to run).
I am not sure how we should solve this. Would need some investigation into `reproc` and UNIX process management :)

It would also be good to write the PID of the started process into a global PID file. Then we could also implement another command like `micromamba ps` that prints the running processes and allows to easily kill them.

Also, I haven't looked into Windows at all, yet.

### DIFF
--- a/micromamba/CMakeLists.txt
+++ b/micromamba/CMakeLists.txt
@@ -62,6 +62,9 @@ endif ()
 
 message(STATUS "Micromamba linkage: ${MICROMAMBA_LINKAGE}")
 
+find_package(Threads REQUIRED)
+target_link_libraries(micromamba PRIVATE Threads::Threads)
+
 if (${MICROMAMBA_LINKAGE} STREQUAL "FULL_STATIC")
     target_link_libraries(micromamba PRIVATE libmamba-full-static)
     if (WIN32)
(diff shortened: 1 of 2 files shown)


wolfv:
Another way to "tag" a process on Linux / UNIX to later find it again: https://serverfault.com/a/274613 (alternative to PID files).

wolfv:
Very detailed stack overflow post: https://stackoverflow.com/questions/17954432/creating-a-daemon-in-linux/17955149#17955149

Another linux daemon starting point: https://gist.github.com/alexdlaird/3100f8c7c96871c5b94e

Something about double forking to truly daemonize a process: https://stackoverflow.com/questions/881388/what-is-the-reason-for-performing-a-double-fork-when-creating-a-daemon

PR #1438, opened 2022-02-02
run arguments

JohanMabille:
By default, `micromamba run` redirects all the streams to the parent process.

- To select the stream to redirect: `micromamba run -a stdout -a stderr command args` or `micromamba run -a stdout,stderr command args`
- To disable redirection: `micromamba run -a command args`


JohanMabille:
Do not merge yet, I will push the cwd shortly

PR #1441, opened 2022-02-03
fix regex segfault in history parsing

wolfv:
One of the regex's we were using caused a segfault on Linux when running with very long strings. This switches to a simpler, more performnat parser implementation.

Really we should move to JSON for the History file, though.


mariusvniekerk:
maybe JSON-line instead of just JSON since it makes for appendable history

PR #1445, opened 2022-02-03
Incremental ccache updates

jonashaag:
- Use separate ccache for each of the libmamba_static downstream tasks
- Use incremental ccache, not only a single cache update per day
- Drop ccache from Windows setup


jonashaag:
I think this should be good to have independent from https://github.com/mamba-org/mamba/pull/1447

wolfv:
Thanks!

Issue #1446, opened 2022-02-03
libmamba's filelock fails on filesystems that do not support F_WRLCK


meson800:
Synopsis
----------
Programs using libmamba's filelock (e.g. `micromamba` and recent versions of mamba past when #1238 merged) fail when run on filesystems that do not support F_WRLCK, unlike `conda` and older versions of `mamba`.

Which filesystems don't support `F_WRLCK`? I use mamba on a shared computing cluster, where the main filesystem is a networked filesystem that I don't control (fstab entry: `type lustre (rw,noflock,lazystatfs)`) which has locking disabled. There is a node-local filesystem that supports locking, but the quotas on the local filesystem are too small to support a package cache (e.g. you run out of space on the node-local filesystem when trying to install a decently-sized environment).

It looks like [Conda's lock](https://github.com/conda/conda/blob/master/conda/lock.py#L55) is quite simple. Is it feasible to have a fallback in case `fcntl` returns `ENOSYS` (function not implemented) that tries to spin-lock like Conda? I can try my hand at a PR if there isn't a better solution.

Reproduction steps
----------------------
1. Be on a filesystem with locks disabled (not sure how to reproduce this 😢, perhaps with some shared-library injection hacking I could have a local testcase)
2. Make sure the root directory is set to.
3. Start any operation that tries to open the package cache (such as `micromamba create -p ./test python -c conda-forge`)
4. Get LockFile error `Could not set lock (Function not implemented)`

An example `strace` log looks like this
```
openat(AT_FDCWD, "/path/to/this/filesystem/micromamba/pkgs/cache/cache.lock", O_RDWR|O_CREAT, 0666) = 7
getpid()                                = 11367
lseek(7, 0, SEEK_SET)                   = 0
read(7, "", 20)                         = 0
fcntl(7, F_SETLK, {l_type=F_WRLCK, l_whence=SEEK_SET, l_start=21, l_len=1}) = -1 ENOSYS (Function not implemented)
clock_gettime(CLOCK_REALTIME, {tv_sec=1643927299, tv_nsec=895008863}) = 0
write(2, "error   ", 8error   )                 = 8
write(2, " Could not set lock (Function no"..., 47 Could not set lock (Function not implemented)
) = 47
```

wolfv:
Hi @meson800 thanks for the detailed analysis. I am not sure wether it's worth it to implement another file locking mechanism or to just disable file locking (maybe with a warning?).

In any case, a PR would be very much appreciated!

[...]

PR #1448, opened 2022-02-04
Locking on lock-less filesystems.

meson800:
When ENOSYS/Function not implemented is returned by fcntl when trying to use advisory locks, switch to a simple locking system where the file is assumed to be successfully locked if the lockfile did not initially exist (e.g. we were the first process to create the lock).

Draft pull request to address #1446. Will add tests/etc.


samuelchodur:
Unable to conduct initial installation of Mambaforge-23.1.0-4-Linux-x86_64.sh due to receiving this error. Not seeing a way to disable lock files with the installer.

PR #1449, opened 2022-02-07
Implemented -f option for the clean command

### LINKED ISSUE #1398: micromamba clean missing `-f` flag
We are missing the `-f, --force-pkgs-dirs` option in micromamba clean, which is used by repo2docker.

JohanMabille:
Fixes #1398

### DIFF
--- a/libmamba/src/api/clean.cpp
+++ b/libmamba/src/api/clean.cpp
@@ -28,9 +28,10 @@ namespace mamba
         bool clean_tarballs = options & MAMBA_CLEAN_TARBALLS;
         bool clean_locks = options & MAMBA_CLEAN_LOCKS;
         bool clean_trash = options & MAMBA_CLEAN_TRASH;
+        bool clean_force_pkgs_dirs = options & MAMBA_CLEAN_FORCE_PKGS_DIRS;
 
         if (!(clean_all || clean_index || clean_pkgs || clean_tarballs || clean_locks
-              || clean_trash))
+              || clean_trash || clean_force_pkgs_dirs))
         {
             Console::stream() << "Nothing to do." << std::endl;
             return;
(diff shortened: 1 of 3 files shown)


JohanMabille:
The current implementation is missing a confirmatino question.

libmamba/src/api/clean.cpp
```diff
@@ -292,6 +293,14 @@ namespace mamba
             }
         }
 
+        if (clean_force_pkgs_dirs)
+        {
+            for (auto& cache: ctx.pkgs_dirs)
```
wolfv, line 298: maybe better to use `            for (auto* pkg_cache : caches.writable_caches())`?

micromamba/src/clean.cpp
```diff
@@ -70,6 +76,16 @@ set_clean_command(CLI::App* subcom)
             options = options | MAMBA_CLEAN_LOCKS;
         if (config.at("clean_trash").compute().value<bool>())
             options = options | MAMBA_CLEAN_TRASH;
+        if (config.at("clean_force_pkgs_dirs").compute().value<bool>())
+        {
+            std::cout << "Remove all contents from the package caches ([y]/n)?" << std::endl;
```
wolfv, line 81: Can you use `Console::prompt(...)`?

wolfv: And mabye we should move thsi to the actual place where things are removed? And print what package dirs are going to be erased.

gsakkis:
Currently the -y/--yes flag is not respected with -f:
```
micromamba clean  -f -y
Remove all contents from the package caches?: [y/n] 
```

wolfv:
@gsakkis please file a new issue :)

gsakkis:
[Done!](https://github.com/mamba-org/mamba/issues/1562)

PR #1450, opened 2022-02-07
Configurable progress bar frequency

jonashaag:


wolfv:
Do you have this problem with the new progress bars or with the old ones?
Unfortunately the old ones had the throttle removed for the past two versions which might have caused undesirable behavior.

jonashaag:
The current `master` is much better but I have one weird setup with a very slow terminal where I want something 10 Hz update frequency.

wolfv:
There is a `void watch_print(const duration_t& period = std::chrono::milliseconds(100));` that should govern how often the progress bar is printed.

jonashaag:
Current frequency is ok for us.

PR #1451, opened 2022-02-07
Set progress bar frequency to 10 Hz

jonashaag:
Closes https://github.com/mamba-org/mamba/pull/1450


jonashaag:
Current frequency is ok for us.

PR #1477, opened 2022-02-10
Reverting env variable wipe before running pip

### LINKED ISSUE #1468: [micromamba] 0.21.0 breaks `PIP_` config env variables
In ~Mamba~ micromamba 0.21.0, environment variables starting with `PIP_` (used to configure pip for things like extra indexes, etc.) are being stripped. I think this is part of #1422, in particular this change: https://github.com/wolfv/mamba/commit/61c5de793c756271ec2daf5462b685686d6bd999#diff-ebac1cb6982f8e508ff9d8203d56f74aa3da80115d7ca84cc525a131ef23535aR61

There may be good reason to suppress some environment variables, but this line's change only applies to running pip (or in the future some other third party package manager) so I don't think it makes sense to empty the environment variables. At least, env vars which start with `PIP_` should be passed through.

austin1howard:
Resolves #1468.

### DIFF
--- a/libmamba/src/api/install.cpp
+++ b/libmamba/src/api/install.cpp
@@ -58,7 +58,6 @@ namespace mamba
         auto [wrapped_command, tmpfile] = prepare_wrapped_call(ctx.target_prefix, install_args);
 
         reproc::options options;
-        options.env.behavior = reproc::env::empty;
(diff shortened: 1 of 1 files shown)


wolfv:
great. I was just preparing a PR with the same fix :)

PR #1482, opened 2022-02-11
adjust cache url hashing function to match conda better

wolfv:
@chrisburr do you think this will be enough? This should match `conda` behavior better -- though not exactly because I didn't want to change the signature of our function. We don't automatically try to add a `/`, however, that case should basically _never_ happen.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,3 +1,4 @@
+exclude: libmamba/tests/repodata_json_cache*
 repos:
 -   repo: https://github.com/psf/black
     rev: 21.12b0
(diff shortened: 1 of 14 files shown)


chrisburr (approved):
Yup I think this should do the trick :+1:

wolfv:
Looks like we uncovered another small incompatibility now that the `HASH.json` names are the same :) I'm having a look.

wolfv:
Another "problem" is that in our logic we always expect 4 keys (_mod, _etag, _url, _cache_control) to be present, but `conda` only writes out `_url` and `_mod` for files that were e.g. downloaded from a local file system. I'll add some tests and so on for that, too.

PR #1491, opened 2022-02-14
micromamba run process names

Klaim:
`micromamba run` now runs unix-like systems programs with an associated name. (see #1491)
By default that name is made unique either by using a prefix adjective to the name of the executable program, or by using a randomly generated prefix if no prefix+program name is available.
Users can specify a name using `--pname <name>`.

To guarantee unique names, we randomly chose a prefix to generate the name then check against the currently used names launched by other `micromamba run` commands, retry generating until we find a name that is not being used and if we don't have alternative prefixes to add we simply generate a random one and try again. If the user specified a name using `--pname`, we prevent execution of the program if the name is already used, we do not generate a new one.

This feature is not supported on  Windows. `--pname` will not appear in the possible options.

### DIFF
--- a/micromamba/src/run.cpp
+++ b/micromamba/src/run.cpp
@@ -28,11 +28,11 @@ namespace mamba {
 
     bool is_process_name_running(std::string_view name);
 
-    std::string generate_unique_process_name(std::string_view process_name)
+    std::string generate_unique_process_name(std::string_view program_name)
     {
-        assert(!process_name.empty());
+        assert(!program_name.empty());
 
-        static std::vector prefixes_bag = {
+        static const std::vector prefixes = {
             "curious",
             "gentle",
             "happy",
@@ -70,6 +70,18 @@ namespace mamba {
             // TODO: add more here
         };
 
+        static std::vector alt_names{
+            "program",
+            "application",
+            "app",
+            "code",
+            "blob",
+            "binary",
+            "script",
+        };
+
+        static std::vector prefixes_bag = prefixes;
+        std::string selected_name{ program_name };
         while(true)
         {
             std::string selected_prefix;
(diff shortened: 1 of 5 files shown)


micromamba/src/run.cpp
```diff
@@ -213,6 +221,12 @@ set_run_command(CLI::App* subcom)
 
         std::tie(pid, ec) = proc.pid();
 
+        {
+            auto lock = lock_proc_dir();
+            fs::fstream pid_file{ proc_dir / "pid.json" };
```
wolfv, line 226: the filename should be `fmt::format("{}.json", getpid());` :)

micromamba/src/run.cpp
```diff
@@ -100,16 +285,45 @@ set_run_command(CLI::App* subcom)
...
+            {
+                const auto unique_name = generate_unique_process_name(*exe_name_it);
+                command.insert(exe_name_it, { {"-a"}, unique_name });
+                return unique_name;
+            }
+            else
+            {
+                if(is_process_name_running(specific_process_name))
+                {
+                    LOG_ERROR << fmt::format("Another process with name '{}' is currently running.", specific_process_name);
```
wolfv, line 314: maybe if you throw a `std::runtime_error` here you don't need the static lockfile? Since later you catch the error.

micromamba/src/run.cpp
```diff
@@ -165,13 +379,18 @@ set_run_command(CLI::App* subcom)
...
+        int status;
+        PID pid;
         std::error_code ec;
 
         ec = proc.start(wrapped_command, opt);
 
         std::tie(pid, ec) = proc.pid();
 
+#ifndef _WIN32
+        static ScopedProcFile scoped_proc_file{ process_name, pid, std::move(proc_dir_lock) }; // Note: this object is static only so that calls to `std::exit()` will call the destructor.
```
wolfv, line 391: I think it might be simpler to only store the parent process ID? Everything may be much simpler & faster with locking etc.

The OS should have a way to find the child processes if we need & want to.

micromamba/src/run.cpp
Klaim: Yes that would work, I'll change that.

There are other exit calls after we finish using the lock (after line 391 basically) but that shouldnt impact the present code.

I'm considering replacing all the exit calls in this lambda by throws except the successful one to help future modifications to not rely on exit and break RAII usage here. I could then move the whole code in an inner scope and the last successful exit call out of that scope, that would basically remove the statics in this lambda - but it might make the code look weird haha

micromamba/src/run.cpp
wolfv: But the process name is something we determine, right? I don't think I understand :)

We want to store `process_name` in a file that is called <pid>.json -- and then later we can map the pid's to process names.

Klaim:
Following discussion, one thing is missing but coming: generating names with `{prefix}_{name_from_a_list}` as a first failsafe when no `{prefix}_{program_name}` are available, then use the random prefix solution only if all other cases failed.

PR #1494, opened 2022-02-15
Dead code?

marimeireles:


marimeireles:
Ok, this is a mistake. I don't think any of these should be removed.
I'll just add a little test for `print_context_only` and add documentation for the others.

wolfv:
I don't think we need to document these as they are really only used in tests -- at least the print-context-only.

JohanMabille:
A few comments in the code won't hurt ;)

wolfv:
yep, 100% agree. need to get better at this :)

marimeireles:
Seems good!

marimeireles:
But is the `--debug` flag really only used in the tests?
If it is then I think `print_context_only` is dead code cause it's not used in the tests, it's only use it's in [here](https://github.com/mamba-org/mamba/blob/b3aef48677b08e540e7ec4b08c03d6099ff48969/libmamba/include/mamba/api/configuration.hpp#L24-L37).

PR #1500, opened 2022-02-16
Config refactoring

JohanMabille:
The main changes are the followings:
- use `std::optional` to store the CLI values. This allows to get rid of sentinel values to check if an option was defined..
- merged the template classes `Configurable` and `Wrapper` into `ConfigurableImpl`.
- renamed the `WrapperBase` class into `ConfigurableImplBase`. This class holds all the data members that do not depends on the template parameter.
- renamed the `ConfigurableInterface` class into `Configurable`. This is now the unique interface to use for manipulating configurable objects.
- Removed all the over-duplicated API. The only remaining methods are those implicitly depending on the template parameter but where we cannot deduce it (yaml parsing, json dumping, etc ...)
- Fixed const-correctness issues.
- Removed suboptimal copies of Configurable.


JohanMabille:
@wolfv I think this ready to go.

PR #1503, opened 2022-02-16
initial powerloader integration

wolfv:
cc @JohanMabille this is some very initial work on integrating with powerloader.

What's missing:

- [ ] mirror configuration
- [ ] auth config?
- [ ] subdirdata also via powerloader
- [ ] replace mamba::url with powerloader::url
- [ ] Decide if we want to use std::function or raw function pointers for callbacks
- [ ] Add mod / etag functions to powerloader to retrieve cached information
- [ ] Figure out if we should enhance the callback on powerloader side with throttling and average speed computation

lots of other things likely, too :)


wolfv:
Also needs this powerloader PR: https://github.com/mamba-org/powerloader/pull/79

PR #1507, opened 2022-02-18
improve linebreak handling for libsolv debug logs

wolfv:


libmamba/src/core/pool.cpp
```diff
@@ -12,22 +12,26 @@ namespace mamba
...
         {
-            logger->warn(s);
+            dbg->first->warn(dbg->second);
         }
         else if (Context::instance().verbosity > 2)
         {
-            logger->info(s);
+            dbg->first->info(dbg->second);
         }
+        dbg->second.clear();
```
AntoinePrv, line 34: @wolf what was the idea behind the `std::string` callback data member? it is cleared afterwards and never used again. Was it for reusing the alloc?
Wouldn't wrapping in a `std::string_view` suffice here?

wolfv: Only if we don't return before (line 17)

wolfv: But I am sure this can be improved!

PR #1510, opened 2022-02-18
Fixed repoquery not working outside of activated environment

JohanMabille:
This makes repoquery work outside of an activated environment, and fixes a segfault when no spec is passed.


wolfv:
Nice!

PR #1511, opened 2022-02-18
add noarch recompilation step

wolfv:
@chrisburr this is some initial work to fix this noarch-recompilation issue. Will need to polish it up more.

The idea is to post-process the transaction and add all `noarch: python` packages that need recompilation.

### DIFF
--- a/libmamba/include/mamba/core/package_info.hpp
+++ b/libmamba/include/mamba/core/package_info.hpp
@@ -49,6 +49,7 @@ namespace mamba
         std::string name;
         std::string version;
         std::string build_string;
+        std::string noarch;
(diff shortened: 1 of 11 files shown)


wolfv:
- [ ] do not remove / modify the noarch packages that are actually updated or changed in the transaction

chrisburr:
@wolfv Thanks for putting this together, I really like this solution to the problem.

wolfv:
I hope that soon I can tackle a bit more proper rewrite of some transaction modifications (including force-reinstall, no-deps and only-deps). but yes, this bug was due to be fixed!

PR #1516, opened 2022-02-21
fake-install repo

wolfv:
This code makes it possible to create a new "installed" repo from a transaction. Then we can re-solve an environment based on this "fake installed" packages. This will be useful for pseudo-installing a package (to keep the environment compatible while not installing all packages).


jonashaag:
Is this for the same purpose as https://github.com/mamba-org/mamba/pull/1103?

wolfv:
Yep, I think it would be the low level part of that. Happy to discuss this further!

jonashaag:
cc @FelixHoehleQC

wolfv:
@JohanMabille I adjusted this for the new `MRepo` interface and generally cleaned it up.

This seems to be working quite nicely, actually. The next step is to make a proper transaction-differ so that we only install the remaining packages.

PR #1518, opened 2022-02-21
Fix `history.parse` tests generating expenentionally bigger test data file

Klaim:
The issue, I believe, was that the code was trying to backup, modify and then restore the history file. However the path to the history files were wrong in the last bit. This lead to the history file size used in that test to double size at each run (because the test needs to re-inject it's content in the history), instead of being restored to it's initial size each end of run. Note that this issue is only visible if you run the tests more than once locally, the CI will re-configure each run which overwrite the history file and prevent from making the issue visible.

These changes mainly refactor the test for clarity and future maintenance, which lead to fixing the issue.
(please tell me if my understanding of the test is wrong)
- Made sure the history files are always the same, no way to make the same mistake again.
- Some manual file content copy/pasting were replaced by explicit file copies.
- Added some RAII for making sure the history file is restored even if an operation fails in the test (except if file access fails, but then a reconfigure would fix that).
- Added some comments to help figuring out what this test was testing.
- Also made some paths absolute for helping with debugging (it is not always clear which files are being used when running `make test` vs when running the test executable manually).

### DIFF
--- a/libmamba/src/core/history.cpp
+++ b/libmamba/src/core/history.cpp
@@ -14,7 +14,7 @@ namespace mamba
 {
     History::History(const fs::path& prefix)
         : m_prefix(prefix)
-        , m_history_file_path(m_prefix / "conda-meta" / "history")
+        , m_history_file_path(fs::absolute(m_prefix / "conda-meta" / "history"))
     {
     }
 
@@ -29,7 +29,7 @@ namespace mamba
             return res;
         }
 
-        static std::regex head_re("==>\\s*(.+?)\\s*<==");
+        static const std::regex head_re("==>\\s*(.+?)\\s*<==");
         std::ifstream in_file = open_ifstream(m_history_file_path, std::ios::in);
 
         std::string line;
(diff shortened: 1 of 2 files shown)


jonashaag:
This should also apply to `parse_segfault`

libmamba/src/core/history.cpp
```diff
@@ -14,7 +14,7 @@ namespace mamba
 {
     History::History(const fs::path& prefix)
         : m_prefix(prefix)
-        , m_history_file_path(m_prefix / "conda-meta" / "history")
+        , m_history_file_path(fs::absolute(m_prefix / "conda-meta" / "history"))
```
wolfv, line 17: the `fs::absolute` shouldn't really change anything, right?

wolfv:
I think for parse_segfault we're not duplicating / appending to the history file, right?

wolfv:
Awesome, thanks!

libmamba/src/core/history.cpp
Klaim: Sorry for the delay.
Yes this doesn't change the behavior of the code, however it makes the log in the constructor clearer as to which location the file is. The relative path was misleading me when I was debugging and felt it would help later maintenance to have it absolute so that any logging will always be precise.

PR #1519, opened 2022-02-22
Docs

marimeireles:
Adding docs:

- command ref. to micromamba
- glossary
- how to build documentation docs
- `micromamba package` command line

I think I should add these changes to the changelog also?

I tried to add a command. ref to mamba too. I actually wrote a big part of it, but it just stroke me as very weird because all of the time it's referencing conda. I don't know how to go about that, let me know what you think @wolfv.

Thanks!


wolfv:
Awesome!

wolfv:
Hi @marimeireles thanks for putting this together! Did you automatically generate the `.rst` files from micromamba output? I think that would be awesome, if that was possible.

marimeireles:
No I didn't, that was the reason why it was super boring.
I tried to, but that's what I said the other day in the meeting. The way I figured out to be able to do it is when you have a python codebase and it's attached to `argparse`.
This is bad, because every time we change something we have to manually change the docs too...
According to Sylvain the way to fix it would be using the breath + doxygen thing.

marimeireles (commented):
Hey Wolf! Could u please help me to check my syntax on this?
Or maybe you know a tooll that helps me to do it?
The onlly way I know is by committing literallly a thousand times.
Thank you!

wolfv:
Hey @marimeireles I just pushed after running the linter and moving the `API-gen.py` file to `docs/api_gen.py`.

I don't actually see where the file is used? Are you already running it in the CI?
I think I'd suggest to run this more like a linter -- if someone changes in the micromamba CLI and the developer forgets to update the docs (by running the generation tool) then there should be a step that complains about this (e.g. by generating the docs and running `git diff` to see if there are differences between what's there and what's missing).

wolfv:
For the linting, you ca install `pre-commit` locally (e.g. using `mamba install pre-commit` and then just run `pre-commit run --all` to fix up all files wrt to linting :)

doc_linter.py
```diff
@@ -0,0 +1,36 @@
...
+            existing_doc_files_tuple = tuple(file for file in existing_doc_files)
+
+    if micromamba_files == existing_doc_files:
+        pass
+    #Raise error in case the doc files weren't generated
+    else:
+        missing_doc_files = set(micromamba_files) - set(existing_doc_files)
+        print('The following files require an up to date documentation:')
+        print(missing_doc_files)
+        print('Check the Mamba documentation about how to generate these files.')
```
marimeireles, line 36: Should we do smth other than printing?

marimeireles:
I think it makes sense for me to generate the docs and commit them here to start with, oder? @wolfv 
Also... It's pretty easy to create docs to mamba too, besides the micromamba docs. Should I do it?
The reason why I didn't want to do it it's because it's essentially conda docs.

wolfv:
yes, that makes sense! :)

wolfv:
Ah, for mamba, I think we can hold off with it. Would be good to make some more custom docs with prose though.

wolfv:
The remaining problem in the linter is that you just need to run `pre-commit run --all` to make sure that line sizes are adjusted etc. Otherwise I think this action is green!

wolfv:
Cool! I think before we can merge, we need all the CI jobs back :)

marimeireles:
@wolfv I removed the comments, I think the problem is some kind of invalid syntax on the CI.
```
        if: matrix.os == 'ubuntu-latest'

      - name: run docs linter
        uses: ammaraskar/sphinx-action@master
        with:
          build-command: "python docs/API-gen.py"
          docs-folder: "docs/"

        shell: bash -l {0}
        run: |
            git add docs/source/user_guide/commands_micromamba
            git diff --name-status master > diff
            python doc_linter.py
```
Can you spot what's wrong here? I have no idea... It's not a YAML syntax error and I can't spot what's wrong in the github actions side.

wolfv:
I think you have two steps in one:

```yaml
      - name: run docs linter
        uses: ammaraskar/sphinx-action@master
        with:
          build-command: "python docs/API-gen.py"
          docs-folder: "docs/"

     - name: check diff
        shell: bash -l {0}
        run: |
            git add docs/source/user_guide/commands_micromamba
            git diff --name-status master > diff
            python doc_linter.py
```

jonashaag:
@marimeireles what's the state of this PR? Does it only need fixing CI?

PR #1530, opened 2022-02-23
Renamed `micromamba run` option `--pname` to `--label`

Klaim:
The directory was accessed before use by a log, changed to be sure the directory exists even if the command falis.

### DIFF
--- a/micromamba/src/run.cpp
+++ b/micromamba/src/run.cpp
@@ -276,7 +276,7 @@ set_run_command(CLI::App* subcom)
     subcom->add_option(
         "--label",
         specific_process_name,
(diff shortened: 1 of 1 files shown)


micromamba/src/run.cpp
```diff
@@ -274,7 +274,7 @@ set_run_command(CLI::App* subcom)
 #ifndef _WIN32
     static std::string specific_process_name;
     subcom->add_option(
-        "--pname",
+        "--label",
         specific_process_name,
         "Specifies the name of the process. If not set, a unique name enerated by deriving from the executable name will be generated.");
```
wolfv, line 279: I think the sentence also needs a bit of love. `enerated` is no word I recognize :)

Klaim: woups, will improve that XD

PR #1532, opened 2022-02-23
Tollerate PATH variable being unset

chrisburr:
I've found when running a installer created with constructor I've ran into this error:

```
Linking setuptools-60.9.1-py39hf3d152e_0
Linking pango-1.48.10-h54213e6_2
Linking rrdtool-1.7.2-h80f2daa_2
Linking wheel-0.37.1-pyhd8ed1ab_0
error    Program not found. Make sure it's available from the PATH. No such file or directory
critical pyc compilation failed with program not found. Aborting.
```

This first bug here is that `pyc` compilation should be allowed to fail (which was my fault, sorry about that 😅).

The underlying issue was that the machine didn't have `$PATH` set which means when `micromamba` looks for `bash` [here](https://github.com/mamba-org/mamba/blob/42f2cd88ba7e50d696dc6d7376a1cfa7d02315a2/libmamba/src/core/util.cpp#L1463-L1467) the lookup fails and defaults to an empty string [here](https://github.com/mamba-org/mamba/blob/master/libmamba/include/mamba/core/environment.hpp#L115-L139). This then results in the `Program not found. Make sure it's available from the PATH. No such file or directory.`.

The same failure can be reproduced using:

```bash
PATH= micromamba create --prefix /tmp/test-env six
```

### DIFF
--- a/libmamba/include/mamba/core/environment.hpp
+++ b/libmamba/include/mamba/core/environment.hpp
@@ -112,10 +112,10 @@ namespace mamba
 #endif
         }
 
-        inline fs::path which(const std::string& exe)
+        inline fs::path which(const std::string& exe, const std::string& override_path = "")
         {
             // TODO maybe add a cache?
-            auto env_path = env::get("PATH");
+            auto env_path = override_path == "" ? env::get("PATH") : override_path;
             if (env_path)
             {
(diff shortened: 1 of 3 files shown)


wolfv:
Cool! Thanks for the PR!!

PR #1533, opened 2022-02-24
Passthrough QEMU_LD_PREFIX to subprocesses with empty environments

chrisburr:
When running with emulation I've hit a couple of cases where mamba's subprocesses fail due to `QEMU_LD_PREFIX` not being set causing. I don't particularly like this solution but I don't see a cleaner option other than maybe centralising the creation of subprocesses and having a way of configuring the environment of subprocesses (e.g. `MAMBA_ENV_xxxx` environment variables).

What do you think?

### DIFF
--- a/libmamba/src/core/link.cpp
+++ b/libmamba/src/core/link.cpp
@@ -419,6 +419,11 @@ namespace mamba
         if (activate)
         {
             options.env.behavior = reproc::env::empty;
+            auto qemu_ld_prefix = env::get("QEMU_LD_PREFIX");
+            if (qemu_ld_prefix)
+            {
+                envmap["QEMU_LD_PREFIX"] = qemu_ld_prefix.value();
+            }
         }
         else
         {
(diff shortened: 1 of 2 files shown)


wolfv:
I think we should have a list of environment variables that we consider necessary even in a clean environment and propagate all of them. Unfortunately I haven't found such list for Unix / Linux (only for Windows: https://github.com/microsoft/vcpkg-tool/blob/aee79d78ec25ad162197a1f7c6f5a394a9741837/src/vcpkg/base/system.process.cpp#L285 )

chrisburr:
Yeah, the more I think about it the more I agree.

I have a list that I use for some of my pacakges at work (see below) though I think the list might still need to be configurable as a couple of these are quite domain specific.


```python
    r"DISPLAY",
    r"EDITOR",
    r"HOME",
    r"HOSTNAME",
    r"KRB5.*",
    r"LANG",
    r"LC_.*",
    r"OMP_.*",
    r"QEMU.*",
    r"TERM",
    r"TMPDIR",
    r"TZ",
    r"USER",
    r"VISUAL",
    r"VOMS_.*",
    r"X509_.*",
    r"XRD_.*",
```

PR #1544, opened 2022-02-28
Stores channel into msubdirdata and libsolv repo appdata

JohanMabille:
~~@wolfv this is not exactly what we want in the end, but storing th MRepo in repo->appdata instead of the channel leads to a weird bug (repo->appdata has an empty channel when retrieving it in the channel_match function).~~

### DIFF
--- a/libmamba/include/mamba/core/subdirdata.hpp
+++ b/libmamba/include/mamba/core/subdirdata.hpp
@@ -13,6 +13,7 @@
 
 #include "nlohmann/json.hpp"
 
+#include "mamba/core/channel.hpp"
 #include "mamba/core/context.hpp"
 #include "mamba/core/fetch.hpp"
 #include "mamba/core/mamba_fs.hpp"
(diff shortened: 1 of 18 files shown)


libmamba/include/mamba/core/subdirdata.hpp
```diff
@@ -44,20 +44,6 @@ namespace mamba
...
-        /**
-         * Constructor.
-         * @param name Name of the subdirectory (<channel>/<subdir>)
-         * @param repodata_url URL of the repodata file
-         * @param repodata_fn Local path of the repodata file
-         * @param possible packages caches
-         * @param is_noarch Local path of the repodata file
-         */
-        MSubdirData(const std::string& name,
-                    const std::string& repodata_url,
```
wolfv, line 55: I think we may want to keep one where we can give explicitly the repodata url (that might also be used by the conda integration).

JohanMabille: I've added an argument to specify the repodata file name (and there is already an argument to pass the "base" URL)

PR #1547, opened 2022-03-01
Building MRepo objects through static function that register the repo…

JohanMabille:
… into the pool

@wolfv I still have some cleanup to do, but that gives you the idea.


wolfv:
Lgtm

PR #1550, opened 2022-03-02
remove prefixdata from solver interface

wolfv:
- [ ] check if we need to add the `real_solvable_url` thing for `mamba` to make force-reinstall work there, too.


libmamba/src/core/package_info.cpp
```diff
@@ -110,7 +110,7 @@ namespace mamba
             build_number = n;
         }
 
-        Id real_repo_key = pool_str2id(pool, "solvable:real_repo_url", 1);
+        static Id real_repo_key = pool_str2id(pool, "solvable:real_repo_url", 1);
```
JohanMabille, line 113: Notice that this would make this constructor not threadsafe (I don't know if that can be a problem in the future).

JohanMabille (commented):
Nice!

libmamba/src/core/package_info.cpp
wolfv: I think we should really define these variables in the MPool constructor so that we have them available as keys everywhere. I haven't figured out how to do it properly yet.

wolfv:
OK, so we need to find a way to merge this info with the differently loaded install data in mamba.

wolfv:
- [ ] need to use the same function inside `mamba_env.py` or factor it out

libmamba/src/core/package_info.cpp
wolfv: I think we can ignore this for now since I think any operations on the Pool / Repo in libsolv aren't really threadsafe.

PR #1554, opened 2022-03-02
slight refactoring

JohanMabille:


wolfv:
We shoudl do the same for `PrefixData`.

JohanMabille:
Ok let's do it

PR #1555, opened 2022-03-02
remove prefixdata.load()

wolfv:


wolfv:
I like the simplification of this, but it takes the opportunity away to do some error handling in the `load` function (and return something that's e.g. a `cpp::result<T>` from `load` instead of void...

Since we don't have it right now, I guess it's fine. In the future we might want to have a helper function or a static constructor? What do you think @Klaim @JohanMabille ?

Klaim:
I was wondering what `load()` does exactly, but note that with the present PR, if `load()` returned an result in error in the constructor, then should the object be invalid? if so it means you would still have to throw an exception, because it's a constructor. The alternative is to allow the object to be invalid (not cool ,because forces a lot of checking on the user's side, but might make sense, I'm not sure because I don't have the whole context) - or make the object fallback into a state where it's ok if the load failed.
In any way making `load()` return a result make sense, standalone, and the code using it have to decide what to do.
That's where my thoughts stop for now. :)

wolfv:
@Klaim `load()` loads a couple of JSON files from the `$PREFIX/conda-meta/*json` folder. There is a chance that some of these files is corrupted so it would be good to have some more proper error handling (in this case, nlohmann::json will currently throw an error).

Klaim:
I see, so it would make sense to make it a factory which returns the object or error indeed.
And definitely not call it in the constructor. It should also help the user code clarify that the object's state comes from loading files.

JohanMabille:
I like the idea of @Klaim, since we're going to release a non backward compatible version, I would be in favor to implement this before the release, what do you think?

wolfv:
sounds like a plan!

PR #1563, opened 2022-03-05
Plugged tl::expected and refactored MSubdirData construction

JohanMabille:


libmamba/src/core/subdirdata.cpp
```diff
@@ -160,17 +160,53 @@ namespace mamba
...
+        catch(...)
+        {
+            return tl::make_unexpected("Unkown error when trying to load subdir data " + url);
+        }
+    }
+
     MSubdirData::MSubdirData(const Channel& channel,
                              const std::string& platform,
                              const std::string& url,
                              MultiPackageCache& caches,
```
wolfv, line 201: should we make this a shared ptr?

libmambapy/src/main.cpp
```diff
@@ -226,11 +226,22 @@ PYBIND11_MODULE(bindings, m)
...
+                         const std::string& repodata_fn) -> MSubdirData
+            {
+                auto sres = MSubdirData::create(channel, platform, url, caches, repodata_fn);
+                if (sres.has_value())
+                {
+                    return std::move(sres.value());
+                }
+                else
+                {
+                    throw std::runtime_error(sres.error().what());
```
wolfv, line 242: can we just throw the subdirdata_error?

https://pybind11.readthedocs.io/en/stable/advanced/exceptions.html#registering-custom-translators

libmamba/src/core/subdirdata.cpp
JohanMabille: I don't think there is a need for a `shared_ptr` here, the subdirdata is not a shared ressource (meaning we know when to delete it and who should do it).

libmambapy/src/main.cpp
JohanMabille: Sure

libmamba/src/core/subdirdata.cpp
wolfv: I was thinking we could make the `MultiPackageCache` a `shared_ptr` so we know that the pointer doesn't become invalid.

JohanMabille:
I need to propagate the tl::expected mechanism, but I prefer to do it in an dedicated PT to keep this one short

jonashaag:
I think `cpp-expected` is missing in `environment-static-dev.yml`

PR #1564, opened 2022-03-07
micromamba run: add test runnning python from a conda env

Klaim:


wolfv:
maybe you need to `strip()` the result string before comparison? Or is the encoding wrong?

Klaim:
> maybe you need to `strip()` the result string before comparison? Or is the encoding wrong?

I think it's just python's `print()` that adds a `\n` at the end of whatever is printed, the comparison was failing because of that.
```
    def test_classic_specs(self, temp_env_prefix):
        res = umamba_run(
            "-p", temp_env_prefix, "python", "-c", "import sys; print(sys.prefix)"
        )
>       assert res == temp_env_prefix.strip()
E       AssertionError: assert '/home/joel/t.../R5VVKDLEDK\n' == '/home/joel/t...vs/R5VVKDLEDK'
```

Using `res.strip()` does solve this, will push it.

wolfv:
On windows it adds \r\n ;)

PR #1566, opened 2022-03-09
add structured problems extraction and python interface

wolfv:
cc @jonashaag I've added a small interface to access the problems in a structured way from Python. 

I hope this can help for figuring out how we can extract something more legible :)


jonashaag:
Rel https://github.com/mamba-org/mamba/pull/1424

PR #1141, opened 2021-08-24
Repro libcurl 401s

johnhany97:
Reproing https://github.com/mamba-org/mamba/issues/1132

Running this locally after setting up the dev environment will lead to this snippet:
```
127.0.0.1 - - [24/Aug/2021 15:01:29] "GET /noarch/repodata.json HTTP/1.1" 401 -
Got Basic KG5pbCk6dGVzdA==
Got Basic KG5pbCk6dGVzdA==
127.0.0.1 - - [24/Aug/2021 15:01:31] "GET /osx-64/repodata.json HTTP/1.1" 401 -
127.0.0.1 - - [24/Aug/2021 15:01:31] "GET /osx-64/repodata.json HTTP/1.1" 401 -
Got Basic KG5pbCk6dGVzdA==
127.0.0.1 - - [24/Aug/2021 15:01:31] "GET /osx-64/repodata.json HTTP/1.1" 401 -
Got Basic KG5pbCk6dGVzdA==
127.0.0.1 - - [24/Aug/2021 15:01:31] "GET /noarch/repodata.json HTTP/1.1" 401 -
Got Basic KG5pbCk6dGVzdA==
127.0.0.1 - - [24/Aug/2021 15:01:31] "GET /noarch/repodata.json HTTP/1.1" 401 -
Got
127.0.0.1 - - [24/Aug/2021 15:01:31] "GET /noarch/repodata.json HTTP/1.1" 401 -
Got Basic KG5pbCk6dGVzdA==
127.0.0.1 - - [24/Aug/2021 15:01:33] "GET /osx-64/repodata.json HTTP/1.1" 401 -
Got Basic KG5pbCk6dGVzdA==
127.0.0.1 - - [24/Aug/2021 15:01:33] "GET /osx-64/repodata.json HTTP/1.1" 401 -
Got Basic KG5pbCk6dGVzdA==
127.0.0.1 - - [24/Aug/2021 15:01:33] "GET /osx-64/repodata.json HTTP/1.1" 401 -
Got Basic KG5pbCk6dGVzdA==
127.0.0.1 - - [24/Aug/2021 15:01:33] "GET /noarch/repodata.json HTTP/1.1" 401 -
Got Basic KG5pbCk6dGVzdA==
127.0.0.1 - - [24/Aug/2021 15:01:33] "GET /noarch/repodata.json HTTP/1.1" 401 -
Got
[...]

PR #1569, opened 2022-03-09
Bump libcurl to 7.82.0 and add a test that catches the bug

johnhany97:
Originally reproed & written in https://github.com/mamba-org/mamba/pull/1141.

We found that reusing the connections through libcurl where the url had the credentials in basic auth form would yield 401s.

This was fixed by curl in https://github.com/curl/curl/pull/8451/

This PR bumps libcurl to the version with the fix (7.82.0) and adds the test that was used to repro the multi-use scenario.

### DIFF
--- a/libmamba/environment-dev.yml
+++ b/libmamba/environment-dev.yml
@@ -10,7 +10,7 @@ dependencies:
   - libsolv >=0.7.18
   - libarchive
   - libsodium
-  - libcurl 7.76.1 *_0
+  - libcurl 7.82.0 *_0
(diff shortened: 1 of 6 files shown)


wolfv:
Awesome!

nicornk:
hey @wolfv & @johnhany97 
Thanks for the fix :) When do you plan to release the next version of mamba, containing this fix?

wolfv:
For mamba it should be enough to just update curl. For micromamba there is a bit more work involved.

PR #1577, opened 2022-03-14
environment lockfile parsing & usage in `create` and `install` sub-commands

Klaim:
Adds the handling of environment lockfile (`*-lock.yaml` files) specifying specific dependency versions (see `conda lock`) when using the subcommands `create` and `install`.


Klaim:
There is one test (or set of tests) missing that I'm working on, but it could be added after.

libmamba/include/mamba/core/env_lockfile.hpp
```diff
@@ -0,0 +1,126 @@
...
+    {
+        unknown_failure,     /// Something failed while parsing but we can't identify what.
+        unsuported_version,  /// The version of the file does not matched supported ver.
+        parsing_failure,     /// The content of the file doesnt match the expected format/language
+                             /// structure or constraints.
+        invalid_data,        /// The structure of the data in the file is fine but some fields have
+                             /// invalid values for our purpose.
+    };
+
+    class EnvLockFileError : public mamba_error
```
JohanMabille, line 36: I added a "m_data" member of type `any` to `mamba_error` so that you can store additional information without having to subclass it. Do you think it would make sense to use it to store the parsing error code?

Klaim: I used it to store the YAMLcpp error type_index, but I could store the code there too. This would mean storing a struct instead then.

Klaim: I pushed the change, also improved the interface. Storing extra data using this technique with inheriting error types seems to work well.

Klaim:
After discussion with Johan I will remove the additional error type and move  it's data into our `mamba_error` type. Wait for it before merging ^^

Klaim:
@JohanMabille I did the changes we talked about regarding error handling, is it fine this way?

JohanMabille:
Yes, that's perfect!

JohanMabille:
@wolfv I think this one can be merged.

Klaim:
Agreed. I will put the tests in another MR once I fix the issue that this PR introduces with `--json` (basically it seems the json output is malformed but only with lockfiles, I see why but didn´t find a good solution yet)

Klaim:
The issue preventing me to add tests here is fixed (#1600) so I added the tests here.

libmamba/src/core/env_lockfile.cpp
```diff
@@ -0,0 +1,214 @@
...
+                    if (const auto& optional_node = package_node["optional"])
+                        return optional_node.as<bool>();
+                    return false;
+                }(),
+                /* .category = */ package_node["category"].as<std::string>(),
+                /* .manager = */ package_node["manager"].as<std::string>(),
+                /* .platform = */ package_node["platform"].as<std::string>(),
+            };
+
+            package.info.version = package_node["version"].as<std::string>();
```
wolfv, line 38: Wouldn't this throw a YAML Exception if no version is available? Just wondering if we should catch the exception and transform to an expected?

Klaim: So far for any Yamlcpp exception, I assumed it's fine to catch it in `read_environment_lockfile` (below) instead of every functions, but I could catch them earlier and return expected indeed. In `read_environment_lockfile` the exception is already converted to an expected, though it's not visible from here so no problem if you prefer local catching instead.

PR #1590, opened 2022-03-24
Headers cleanup

JohanMabille:


wolfv:
The failure on macOS is due to a faulty OpenSSL3 build afaik.

PR #1613, opened 2022-04-05
libmamba: correct macOS header casing

l2dy:
macOS can be installed on case-sensitive file systems, where libmamba fails to build.

```
.../mamba-2022.02.11/libmamba/src/core/util_os.cpp:15:10: fatal error: 'libProc.h' file not found
```


wolfv:
Cool, I didn't know. Thanks for the contribution!

Issue #1611, opened 2022-04-04
Micromamba sporadically hangs  (QEMU + ARM)


maresb:
Over at [Jupyter Docker Stacks](https://github.com/jupyter/docker-stacks/pull/1594#issuecomment-1087895999), we have been trying create the initial Conda environment for our `base-notebook` by using Micromamba instead of Miniforge. We have been plagued by sporadic hangs in our multi-arch Docker build process on GH Actions at a rate of ~20% of all runs when for `aarch64` we `RUN` the command

```bash
./micromamba install --root-prefix=/opt/conda --prefix=/opt/conda -vv --yes python=3.9 mamba notebook jupyterhub jupyterlab
```

The hang occurs sometime between the end of extraction and the beginning of linking. The last message printed during an unsuccessful run is something like

```
  #29 188.0 debug    libmamba Extracted to '/opt/conda/pkgs/mistune-0.8.4-py39h14843e3_1005'
```

without proceeding to a line like

```
#29 178.0 Linking pandoc-2.17.1.1-h8af1aa0_0
```

Execution hangs without any further output until the 6h limit is reached.

# References
(in chronological order)

* Mamba [sporadically hangs](https://github.com/jupyter/docker-stacks/issues/1539) when building jupyter/docker-stacks arm64 image
* Original [PR](https://github.com/jupyter/docker-stacks/pull/1594) in jupyter/docker-stacks to replace MiniForge with Micromamba.
* Partially [understood/isolated](https://github.com/jupyter/docker-stacks/pull/1594#issuecomment-1087895999) the problem
* This bug report
* @jonashaag [patches my jupyter/docker-stacks PR](https://github.com/jonashaag/docker-stacks/pull/1) to disable concurrency, and I confirm that it appears to work
* @jonashaag [patches libmamba](https://github.com/mamba-org/mamba/pull/1616) to disable subprocesses when `extract_threads=1`
* Discussion returns to this thread

jonashaag:
Is it possible for you to try to reproduce the error with concurrency = 1 for everything that involves concurrency (extract, download, …)?

maresb:
Hi @jonashaag, thanks so much for helping! I am looking into the config, and I see `download_threads` and `extract_threads`. Is this what you mean? Are there any other values you suggest?

[...]

PR #1616, opened 2022-04-08
Extract in subprocess only with concurrency > 1

jonashaag:
cc @maresb

### DIFF
--- a/libmamba/include/mamba/core/thread_utils.hpp
+++ b/libmamba/include/mamba/core/thread_utils.hpp
@@ -147,6 +147,7 @@ namespace mamba
         inline counting_semaphore(std::ptrdiff_t max = 0);
         inline void lock();
         inline void unlock();
+        inline std::ptrdiff_t get_max();
         inline void set_max(std::ptrdiff_t value);
 
     private:
@@ -182,6 +183,11 @@ namespace mamba
             }
         }
         m_cv.notify_all();
+    }
+
+    inline std::ptrdiff_t counting_semaphore::get_max()
+    {
+        return m_max;
     }
 
     inline void counting_semaphore::set_max(std::ptrdiff_t value)
--- a/libmamba/include/mamba/core/transaction.hpp
+++ b/libmamba/include/mamba/core/transaction.hpp
@@ -98,6 +98,7 @@ namespace mamba
     class DownloadExtractSemaphore
     {
     public:
+        static std::ptrdiff_t get_max();
         static void set_max(int value);
 
     private:
--- a/libmamba/src/core/transaction.cpp
+++ b/libmamba/src/core/transaction.cpp
@@ -47,18 +47,14 @@ namespace mamba
 
     counting_semaphore DownloadExtractSemaphore::semaphore(0);
 
+    std::ptrdiff_t DownloadExtractSemaphore::get_max()
+    {
+        return DownloadExtractSemaphore::semaphore.get_max();
+    }
+
     void DownloadExtractSemaphore::set_max(int value)
     {
         DownloadExtractSemaphore::semaphore.set_max(value);
-    }
-
-    static std::mutex lookup_checksum_mutex;
-    std::string lookup_checksum(Solvable* s, Id checksum_type)
-    {
-        Id unused;
-        std::lock_guard<std::mutex> lock(lookup_checksum_mutex);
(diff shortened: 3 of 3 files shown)


maresb:
If I am reading this correctly, the change is:

Before: When extracting, spawn `extract_threads` subprocesses for extracting.

After: When extracting, spawn `extract_threads` subprocesses for extracting, unless `extract_threads=1` in which case use the main process.

I just checked the status of [my experiments](https://github.com/jonashaag/docker-stacks/pull/1#issuecomment-1092571485) and the runs went as expected. I did however notice one unexpected detail in [run #105](https://github.com/maresb/docker-stacks/runs/5881499257?check_suite_focus=true) which is hanging after

```
 #36 43.88 info     libmamba Transfer finalized, status 200 [https://conda.anaconda.org/conda-forge/linux-aarch64/python-3.9.12-h2eada40_1_cpython.tar.bz2] 13382366 bytes
#36 43.88 info     libmamba Download finished, validating '/opt/conda/pkgs/python-3.9.12-h2eada40_1_cpython.tar.bz2'
```

Do we need to implement the same logic for `download_threads`?

maresb:
# References
(in chronological order)

* Mamba [sporadically hangs](https://github.com/jupyter/docker-stacks/issues/1539) when building jupyter/docker-stacks arm64 image
* Original [PR](https://github.com/jupyter/docker-stacks/pull/1594) in jupyter/docker-stacks to replace MiniForge with Micromamba.
* Partially [understood/isolated](https://github.com/jupyter/docker-stacks/pull/1594#issuecomment-1087895999) the problem
* [Bug report](https://github.com/mamba-org/mamba/issues/1611) to mamba-org/mamba
* @jonashaag [patches my jupyter/docker-stacks PR](https://github.com/jonashaag/docker-stacks/pull/1) to disable concurrency, and I confirm that it appears to work
* This

jonashaag:
> Do we need to implement the same logic for `download_threads`?

I don't think so; parallel downloading has been there for much longer than parallel extraction and it doesn't seem to cause trouble for you (v0.17).

maresb:
It's not like the logs would stop, only the micromamba logs. So either GH is doing some sort of really clever filtering or the log limit is in micromamba itself.

wolfv:
Thanks both for the thorough debugging of this!
I would love to understand the root cause. Are you using emulation for aarch64? E.g. I think buildx does some form of emulation?
Could it be a memory problem? (e.g. parallel extraction using up all available memory?)

maresb:
Hi @wolfv! Apparently `buildx` uses `QEMU`. It would be interesting to try and reproduce this locally. I have not yet tried, since it's just in the last few days that I started to understand the nature of the problem.

I can't exclude it being a memory problem, but it seems unlikely to me because before @jonashaag's patch I tried `extract_threads=1` and it didn't help. (I wouldn't expect one thread to require significantly more memory, but maybe I'm wrong for some reason.)

maresb:
> parallel downloading has been there for much longer than parallel extraction and it doesn't seem to cause trouble for you (v0.17)

I'm not sure what to make of [this log](https://pipelines.actions.githubusercontent.com/serviceHosts/ee2723ef-0d2c-4021-a502-75bbd74b9474/_apis/pipelines/1/runs/96/signedlogcontent/2?urlExpires=2022-04-08T20%3A53%3A40.1651231Z&urlSigningMethod=HMACV1&urlSignature=XkZrd%2BCWrFMgo8Ljotdu48pHoyC34FYtVZZr2m6bi30%3D). It certainly doesn't cause trouble to the same level as parallel extraction, but I wonder if it could still be occasionally (although rarely) causing trouble.

maresb:
I managed to reproduce locally by cloning maresb/docker-stacks, checking out the `micromamba` branch, and running

```bash
cd base-notebook; docker buildx build . --platform "linux/arm64" --no-cache 
```

For preparatory steps to configure qemu with buildx, see [this comment](https://github.com/jupyter/docker-stacks/pull/1594#issuecomment-1028316739).

jonashaag:
Wow that’s great, will do some debugging later!

Do you have debug logs that you can share in the meantime?

jonashaag:
Also shall we move discussion to a different ticket?

maresb:
> Also shall we move discussion to a different ticket?

Yes, let's end discussion here and continue [there](https://github.com/mamba-org/mamba/issues/1611).

# END

PR #1625, opened 2022-04-14
User agent

wolfv:
cc @jezdez @jaimergp 

What do you think? Need to start with `User-Agent: ` and end with an `" "`.


wolfv:
OK, switched the implementation. You can now set anything as the first part of the user agent, no need to worry about starting with `User-Agent: ` and ending with space.

What do yuo think? Is that enough?

jaimergp:
Yup, I think that should be more than enough! Thanks Wolf!

jaimergp:
An extra nice bit would be to allow the libcurl version to be added anywhere by using `fmt::arg` (named placeholder?) instead of a positional arg there, but as you wish!

wolfv:
@jaimergp I think with User-Agents it's usually just a collection of space delimited strings and you're supposed to parse them anyways so tbh I don't think it matters that much ... and then we'd have to pass the dynamic format string as arg to the context ...  although good to know about the fmt named args!

jaimergp:
Thanks!

jezdez:
@wolfv this looks good, thank you for the quick effort!

PR #1629, opened 2022-04-14
add more explanatory error message

wolfv:
@wholtz what do you think about this?


wholtz:
I like it. Thanks @wolfv !

Issue #1584, opened 2022-03-22
[micromamba] post-link script of qt on Windows not executed?


wolfv:
@marcelotrevisani reports that this line https://github.com/conda-forge/qt-feedstock/blob/d6d71b77424020e07f371d0aeb4aaf3bd455e130/recipe/bld.bat#L188 isn't executed (at least not when explicitly installing).

marcelotrevisani:
Try the following on windows (I just tested that on windows):
```
micromamba create -p SOME_FOLDER pyqt qt python -c conda-forge -y
```
after that try to activate that env and import pyqt ``import PyQt5``
it will fail :/
If I do the same with conda or mamba that will be fine

jonmmease:
I'm also seeing this issue.  During installation, I also see the following warning/error while linking qt.

```
...
Linking qt-5.12.9-h556501e_6
info     libmamba ****************** Backtrace Start ******************
debug    libmamba Loading configuration
trace    libmamba Compute configurable 'create_base'
trace    libmamba Compute configurable 'no_env'
trace    libmamba Compute configurable 'no_rc'
trace    libmamba Compute configurable 'rc_files'
trace    libmamba Compute configurable 'root_prefix'
trace    libmamba Get RC files configuration from locations up to HomeDir
critical libmamba Cannot determine HOME (checked USERPROFILE, HOMEDRIVE and HOMEPATH env vars)
info     libmamba ****************** Backtrace End ********************
```

I've tried explicitly setting the `USERPROFILE` environment variable, but that doesn't seem to make a difference.

jonmmease:
I probably won't get a chance to dig into the micromamba code itself, but let me know if there's anything else that would be helpful here!

marcelotrevisani:
Hi @wolfv , I saw that you fixed this problem. Can you generate a new version please?

marcelotrevisani:
I just saw that you already released it. I was checking in the release part, but I saw it in the tags :) thanks!

PR #1630, opened 2022-04-14
do not clear env for post-link scripts

wolfv:
@chrisburr unfortunately I think I'll revert this change for now since we had reports where the Qt post-link script doesn't work on Windows because some env vars are missing. 

We could still define a sensible subset of env vars to pass through but not sure if it's worth it. 
Woudl be better to continue discouraging / removing these scripts everywhere possible :)


wolfv:
Fixes https://github.com/mamba-org/mamba/issues/1584

PR #1637, opened 2022-04-20
Replace thread detaching by thread joining before main's end

Klaim:
Goal: make sure threads do not outlive main() - remove usage of `std::thread::detach` to solve issues related to order of execution.

This fixes undefined behavior when threads are running after `main()`
is finished. At the end of `main()`, static objects are destroyed in
an unspecified order which can also lead easilly to undefined behaviors.
This change should fix both issues.

For now the implementation of `MainExecutor` is simple: it spawn a
new thread for each task scheduled and does not provide means to track
the progress and/or end of tasks. This reflects the previous behavior.
The intent is to enabled changing the implementation of `MainExecutor`
in the future to improve execution speed and/or execution resources
usage.

### DIFF
--- a/libmamba/include/mamba/core/execution.hpp
+++ b/libmamba/include/mamba/core/execution.hpp
@@ -37,7 +37,7 @@ namespace mamba
         MainExecutor();
 
         // Closes (see `close()`) and unregister itself as the main executor.
-        // Could be
+        // Blocks until all scheduled tasks are done and all resources are released (threads joined).
         ~MainExecutor();
 
         // Returns a reference to the current main executor.
@@ -53,7 +53,7 @@ namespace mamba
         static void stop_default();
 
         // Schedules a task for execution.
-        // The task must be a callable which takes either the provided arugments or none.
+        // The task must be a callable which takes either the provided arguments or none.
         // If this executor is open, the task is scheduled for execution and will be called
         // as soon as execution resources are available. The call to the task is not garanteed
         // to have been done at the end of the execution of this function, nor before.
--- a/libmamba/include/mamba/core/execution.hpp
+++ b/libmamba/include/mamba/core/execution.hpp
@@ -37,7 +37,8 @@ namespace mamba
         MainExecutor();
 
         // Closes (see `close()`) and unregister itself as the main executor.
-        // Blocks until all scheduled tasks are done and all resources are released (threads joined).
+        // Blocks until all scheduled tasks are done and all resources are released (threads
+        // joined).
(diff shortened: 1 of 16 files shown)


Klaim:
Tests are in progress but coming soon

Klaim:
I think I should add a few tests tomorrow but otherwise it should be ok to review.

Klaim:
Tests are probably complete enough now.

libmamba/include/mamba/core/execution.hpp
```diff
@@ -0,0 +1,117 @@
...
+    // mechanisms.
+    class MainExecutor
+    {
+    public:
+        // Set itself as the main executor.
+        // Throws `MainExecutorError` if another instance already exists.
+        MainExecutor();
+
+        // Closes (see `close()`) and unregister itself as the main executor.
+        // Could be
```
wolfv, line 40: This sentence looks unfinished

libmamba/include/mamba/core/execution.hpp
```diff
@@ -0,0 +1,117 @@
...
+        // undefined behaviors.
+        static MainExecutor& instance();
+
+        // If the default (global) main executor is being used, close and destroy it.
+        // Do nothing otherwise.
+        // This is mostly used for testing and libraries using the default main executor.
+        static void stop_default();
+
+        // Schedules a task for execution.
+        // The task must be a callable which takes either the provided arugments or none.
```
wolfv, line 56: tiny typo in `arugments`.

wolfv (approved):
Looks great to me. One typo and one unfinished sentence was all I could find.

wolfv:
Once tests are done, I'll merge!

jonashaag:
Not sure if you were aware or this change was even partly motivated by #1611 (since I suggested this in https://github.com/mamba-org/mamba/issues/1611#issuecomment-1104406156). But this seems to fix #1611. I will run a few more trials but haven't seen any deadlocks so far.

wolfv:
Oh wow, that's great news. I didn't think that could have an impact on #1611 !

Klaim:
I'm surprised too! From what I read in #1611 its possible that some better ordering of the threads helped, though it is also possible that it just made less likely to hit the deadlock (or livelock Im not sure). I guess further testing will clarify that indeed ^^

Klaim:
@jonashaag Nice! Thanks for checking!

wolfv:
We will try to have a bugfix release for mamba / micromamba in the coming few days!

PR #1648, opened 2022-04-26
Curl correct init & cleanup

Klaim:
Adds correct usage of libcurl which should be initialized and cleanup at the beginning and end, respectively, of `main()`.

WARNING: This specific change relies on the assumption that threads using libcurl are done (aka joined) before the end of `main()`, which is only guaranteed after PR #1637 will be merged, otherwise the behavior is not defined.

### DIFF
--- a/libmamba/src/core/validate.cpp
+++ b/libmamba/src/core/validate.cpp
@@ -22,6 +22,23 @@
 
 namespace mamba
 {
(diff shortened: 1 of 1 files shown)


wolfv:
I've merged the other PR so is this one good to go as well? :)

PR #1651, opened 2022-04-27
Set compiler optimization level

### LINKED ISSUE #1644: Binary size / compiling with `-O0`
I was wondering why the macOS binary size of micromamba is so large.

When I compile with the default settings I get a binary size of 25 MiB, same size that you get from micro.mamba.pm.

When compiling with `-Os` I get 3.7 MiB (3.9 with `-O2` and 4.0 with `-O3`). When compiling with `-O0` explicitly I get 25 MiB. So it seems like the release build of micromamba is compiled with `-O0` (= the default settings).

Shall we use `-O2` as default optimization level?

jonashaag:
Closes #1644

I chose `-Os` assuming that no performance critical code exists in Mamba.

Also change the order of default settings to it's possible to overwrite them.


jonashaag:
Closing, see #1644

PR #1652, opened 2022-04-27
Show reason for Multi-download failure

syslaila:
Whenever mamba fails to download repodata/packages, we will get a generic "Multi download" failure. In order to get the reason for the failure, we need to re-run the mamba command with increased verbosity. This PR will throw the reason for a multi download failure.


wolfv:
Hi @syslaila thanks a ton for this PR. I like it a lot!

Do you want to squash your commits? And you might also use the correct email address in the commit to associate the commits with your user account.

Issue #1645, opened 2022-04-26
Provide high level Python API


martinRenou:
It would be nice if mamba provided high-level Python APIs:

```python
from mamba import install, create

create('myenv')
install('python=3.10', 'ipython', 'jupyterlab', channels=['conda-forge'], env='myenv')
```

There are such APIs in https://github.com/mamba-org/mamba/blob/master/mamba/mamba/mamba.py but they assume being ran under the CLI (args being populated).

PR #1656, opened 2022-05-03
Add high-level Python API

martinRenou:
Fix https://github.com/mamba-org/mamba/issues/1645

cc. @DerThorsten 

This is adding two functions `from mamba.api import install, create` with the following signatures:

```python
install(env_name: str, specs: list = [], channels: list = [], target_platform: str = None)
create(env_name: str, specs: list = [], channels: list = [], target_platform: str = None)
```

### DIFF
--- a/mamba/mamba/api.py
+++ b/mamba/mamba/api.py
@@ -114,7 +114,11 @@ class MambaSolver:
 
 
 def install(
-    env_name: str, specs: tuple = (), channels: tuple = (), target_platform: str = None, base_prefix: str = None
+    env_name: str,
+    specs: tuple = (),
+    channels: tuple = (),
+    target_platform: str = None,
+    base_prefix: str = None,
 ):
     """Install packages in a given environment.
 
@@ -153,7 +157,11 @@ def install(
 
 
 def create(
-    env_name: str, specs: tuple = (), channels: tuple = (), target_platform: str = None, base_prefix: str = None
+    env_name: str,
+    specs: tuple = (),
+    channels: tuple = (),
+    target_platform: str = None,
+    base_prefix: str = None,
 ):
     """Create a mamba environment.
 
--- a/mamba/mamba/api.py
+++ b/mamba/mamba/api.py
@@ -130,17 +130,19 @@ def install(
     RuntimeError :
         If the solver did not find a solution or if the installation failed.
     """
-    prefix = "/home/martin/micromamba/envs/{}".format(env_name)
-    (pathlib.Path(prefix) / "conda-meta").mkdir(parents=True, exist_ok=True)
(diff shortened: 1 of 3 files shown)


mamba/mamba/api.py
```diff
@@ -0,0 +1,167 @@
...
+
+        if pkg_cache_path is None:
+            # use values from conda
+            pkg_cache_path = self.context.pkgs_dirs
+
+        package_cache = libmambapy.MultiPackageCache(pkg_cache_path)
+        return libmambapy.Transaction(api_solver, package_cache)
+
+
+def install(env_name: str, specs: list = [], channels: list = [], target_platform: str = None):
```
martinRenou, line 120: Not sure what to do if `target_platform` is not provided? Is there a way to know the target platform for which the env was previously created? Also I guess a meaningful default target platform for creating envs should be the current platform?

mamba/mamba/api.py
```diff
@@ -0,0 +1,168 @@
...
+    channels : tuple of str
+        The channels from which to pull packages e.g. ['default', 'conda-forge'].
+    Raises
+    ------
+    RuntimeError :
+        If the solver did not find a solution or if the installation failed.
+    """
+    prefix = pathlib.Path(os.environ["CONDA_PREFIX"]) / "envs" / env_name
+    (prefix / "conda-meta").mkdir(parents=True, exist_ok=True)
+    (prefix / "pkgs").mkdir(parents=True, exist_ok=True)
```
wolfv, line 135: Ideally we re-use the one from the "base" prefix here.

martinRenou: So something like `os.environ["CONDA_PREFIX"]) / "pkgs"`?

wolfv: Yeah, unfortunately it's not that easy to find the base prefix. 

With micromamba, it's the `$MAMBA_ROOT_PREFIX` environment variable.

With conda it's the `import sys; sys.prefix`, but that assumes that conda is running from the base environment.

martinRenou: Yeah I was under the impression it wouldn't be easy.

So you're suggesting like:
```python
base_prefix = None
if "MAMBA_ROOT_PREFIX" in os.environ:
    base_prefix = os.environ["MAMBA_ROOT_PREFIX"]
else:
    import sys
    base_prefix = sys.prefix
prefix = base_prefix / "envs" / env_name
(prefix / "conda-meta").mkdir(parents=True, exist_ok=True)
(base_prefix / "pkgs").mkdir(parents=True, exist_ok=True)
```
?

wolfv: yeah, and maybe make it override-able from a function argument

martinRenou: thanks!

martinRenou:
Ready for another review :)

PR #1665, opened 2022-05-10
Use standard C++ filesystem API only

Klaim:
See rationale in libmamba/include/mamba/core/mamba_fs.hpp

### DIFF
--- a/libmamba/include/mamba/api/configuration.hpp
+++ b/libmamba/include/mamba/api/configuration.hpp
@@ -376,10 +376,10 @@ namespace mamba
         using grouped_config_type = std::pair<std::string, std::vector<Configurable*>>;
         std::vector<grouped_config_type> get_grouped_config();
 
-        std::vector<fs::path> sources();
-        std::vector<fs::path> valid_sources();
+        std::vector<fs::u8path> sources();
+        std::vector<fs::u8path> valid_sources();
 
-        void set_rc_values(std::vector<fs::path> possible_rc_paths, const RCConfigLevel& level);
+        void set_rc_values(std::vector<fs::u8path> possible_rc_paths, const RCConfigLevel& level);
 
         void load();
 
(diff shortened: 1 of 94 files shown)


Klaim:
Here is where I am at:
- I fixed all conversion errors on my local build with visual studio;
- I didn't address yet the linking issues visible on current CI build, I don't see them locally
- For some reason I can't link a Release build because of symbols missing from fmt, I suspect it's because I'm not using the same compiler as the fmt's dll (I used VS2022 IDE instead of building from cli, so that's understandable);
- With the debug build, I get runtime errors coming from the YAML-CPP implementation, so it is opaque, but it might be due to the same mismatch of build than for fmt;
- I suspect that the path->string conversions will fail on runtime on Windows (with no experimental UTF-8 mode) whith unicode paths, but I didn't reach the point where I could test that yet. Hopefully the `pytest mamba/tests/` should catch these. There are several solutions if it fails, we should discuss that once this at least build on all platforms.

libmamba/include/mamba/api/configuration.hpp
```diff
@@ -569,6 +569,24 @@ namespace mamba
             node[name] = m_value;
         }
 
+        template <>
+        inline void ConfigurableImpl<fs::path>::dump_json(nlohmann::json& node,
```
Klaim, line 573: Not sure if it's the right way to do it. I suspect `nlohmann` to have a version that supports `std::filesystem::path` values, if there is, switching to it would make this change unnecessary, I believe.

Klaim:
This version compiles and run on my Windows which is set to have UTF-8 by default in everywhere. However only the libmamba tests can be run, for some reason the `pytest` based tests all fail to find some executable.

wolfv:
@Klaim it looks like one last Unicode test fails on `mamba`.

Klaim:
> @Klaim it looks like one last Unicode test fails on `mamba`.

Yes, it's the one revealing the need for wrapping the path type. I was doing that but I'm starting to think it might be more complicated than just using a `std::string to_utf8(const fs::path& path)` function that would do the work where we need it.
From what I tried, wrapping the type is a bit brittle because it means every time we use a free function or operator from `std::filesystem::path` we end up with that type instead of `fs::path` and that can lead to bad surprises.
I'm trying to figure out what would be best without disturbing too much of the code.

Klaim:
Rebased over current master (it was not trivial 😬 )

Klaim:
Situation update: this is not merged yet because we hit this issue with MSVC:
https://github.com/microsoft/STL/pull/1559

Upgrading to a recent Visual Studio version will work (see https://github.com/mamba-org/mamba/pull/1665/commits/4a4c65747ec8707951c0ddb7c0c0cd6fb7345610) but it requires toolchains to be available for recent Visual Studio versions, which is taking a lot of time and might not happen (as we are currently stuck with a 6 years old version of VS...)

Now I will attempt a workaround to not require upgrading VS, so I reverted the commit doing that.

Klaim:
(rebased over CI not reporting issues that should have been reported, which I expect to fail because I added a test that should fail before I will push more commits).

Klaim:
OK Now the test fails as it should.
I will now smash the commits for the test together and add my solution (that cannot run locally)

Klaim:
looks like my workaround for `fs::remove`and `fs::remove_all` when VS is old works (at least on the CI)

cleanup incoming

wolfv:
Thanks for pushing this through until the end!

PR #1668, opened 2022-05-11
Fix the PowerShell activator a better way to unset envs

chawyehsu:
It should work based on my previous pr to conda, I didn't test it though.


wolfv:
thanks!

PR #1677, opened 2022-05-16
fix up explicit spec installation -- use correct subdir and filename

wolfv:


jonashaag:
https://github.com/conda-forge/yaml-cpp-feedstock/pull/13?

jonashaag:
Hmm no it already uses the latest version

PR #1679, opened 2022-05-17
Update Micromamba download URL and use cURL

jonashaag:
I think cURL comes preinstalled on more systems than wget.


wolfv:
thanks!

PR #1684, opened 2022-05-17
fix curl write callback to not exit

wolfv:
cc @Klaim

### DIFF
--- a/libmamba/src/core/fetch.cpp
+++ b/libmamba/src/core/fetch.cpp
@@ -344,6 +344,30 @@ namespace mamba
 
     bool DownloadTarget::can_retry()
     {
+        switch (result)
+        {
+            case CURLE_ABORTED_BY_CALLBACK:
+            case CURLE_BAD_FUNCTION_ARGUMENT:
+            case CURLE_CONV_REQD:
+            case CURLE_COULDNT_RESOLVE_PROXY:
+            case CURLE_FILESIZE_EXCEEDED:
+            case CURLE_INTERFACE_FAILED:
+            case CURLE_NOT_BUILT_IN:
+            case CURLE_OUT_OF_MEMORY:
+            // See RhBug: 1219817
+            // case CURLE_RECV_ERROR:
+            // case CURLE_SEND_ERROR:
+            case CURLE_SSL_CACERT_BADFILE:
+            case CURLE_SSL_CRL_BADFILE:
+            case CURLE_WRITE_ERROR:
+            case CURLE_OPERATION_TIMEDOUT:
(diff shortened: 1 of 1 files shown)


libmamba/src/core/fetch.cpp
```diff
@@ -344,6 +344,30 @@ namespace mamba
...
+        {
+            case CURLE_ABORTED_BY_CALLBACK:
+            case CURLE_BAD_FUNCTION_ARGUMENT:
+            case CURLE_CONV_REQD:
+            case CURLE_COULDNT_RESOLVE_PROXY:
+            case CURLE_FILESIZE_EXCEEDED:
+            case CURLE_INTERFACE_FAILED:
+            case CURLE_NOT_BUILT_IN:
+            case CURLE_OUT_OF_MEMORY:
+            // See RhBug: 1219817
```
Klaim, line 357: Is this a reference from librepo?

libmamba/src/core/fetch.cpp
```diff
@@ -391,7 +415,7 @@ namespace mamba
             {
                 LOG_ERROR << "Could not open file for download " << s->m_filename << ": "
                           << strerror(errno);
-                exit(1);
+                return size * nmemb + 1;
```
Klaim, line 418: This is equivalent to `(size * nmemb) + 1`, is this what you intended? (it's not clear for me what the `1` is).

libmamba/src/core/fetch.cpp
```diff
@@ -400,7 +424,7 @@ namespace mamba
         if (!s->m_file)
         {
             LOG_ERROR << "Could not write to file " << s->m_filename << ": " << strerror(errno);
-            exit(1);
+            return size * nmemb + 1;
```
Klaim, line 427: Same question

Klaim (approved):
I mostly have minor questions, LGTM otherwise.

libmamba/src/core/fetch.cpp
wolfv: yep

libmamba/src/core/fetch.cpp
wolfv: we just need to return a "size different from size * nmemb". We could always return `0` but it's possible (unlikely though) that we download an empty file with `0` bytes (the curl manual mentions this). I chose to just randomly add `1`. Maybe I should add a comment ...

libmamba/src/core/fetch.cpp
wolfv: same answer :)

PR #1721, opened 2022-06-02
MSVC Improvements

Klaim:
- Better warning levels: made sure we mostly see warnings coming from this project's code source except if a dependency's header have a severe warning.
- Fixed `__cplusplus` not be set at the right value (if no flag is passed to msvc).
- Tried to fix the most obvious flags. Note that there was interesting undefined behaviors and there is still warnings that are difficult to decide a solution for.
- Force parallel build by default when using msvc/MSBuild


Klaim:
Also note that some of the commits here are extracted from https://github.com/mamba-org/mamba/pull/1665

wolfv:
Nice improvements!

Issue #1694, opened 2022-05-22
Allow spdlog to be used with external fmt lib


gdolle:
Hi all,

The current source code include some fmt headers bundled in spdlog, e.g
https://github.com/mamba-org/mamba/blob/master/libmamba/src/core/progress_bar_impl.cpp#L5

However, spdlog may be configured to use fmt as an external library;  SPDLOG_FMT_EXTERNAL is uncommented in `spdlog/tweakme.h`.
In that case, the fmt lib should be linked with micromamba target, and `<spdlog/bundled/fmt/*>` should be conditionnaly replaced with `<fmt/*>` whether SPDLOG_FMT_EXTERNAL is defined.

You may refer to this patch:
https://aur.archlinux.org/cgit/aur.git/tree/mamba-2022.05.20.patch?h=micromamba

wolfv:
Hi @gdolle would you be willing to send a PR with these changes? Happy to merge them!

gdolle:
@wolfv ok ! I created a PR. It should be a bit better than the path I suggested initally.  Feel free to modify.

PR #1732, opened 2022-06-06
Fix issue #1694 external fmt lib

gdolle:
This PR handle issue #1694 to allow fmt to be used as an external library.

- Update headers to use spdlog.h instead of fmt.h (=> include tweakme.h machinery).
- Modify cmake to link with fmt libraries if SDPLOG_FMT_EXTERNAL macro is set/uncommented in tweakme.h.


wolfv:
I just fixed the linting. Let's see if the tests are still failing ... I don't quite understand why they would! :)

wolfv:
Thanks!

PR #1736, opened 2022-06-10
Address a couple of compilation warnings

nstinus:


wolfv:
Hi, thanks for this! I just merged another (earlier) PR that also addresses some compilation warnings. Maybe you can rebase on master? Thanks!

libmamba/src/core/progress_bar_impl.cpp
```diff
@@ -765,14 +765,14 @@ namespace mamba
                     spinner = { "|", "/", "-", "|", "\\", "|", "/", "-", "|", "\\" };
 
                 constexpr int spinner_rounds = 2;
-                auto pos = static_cast<std::size_t>(
-                               std::round(progress * ((spinner_rounds * spinner.size()) / 100.0)))
-                           % spinner.size();
+                const std::size_t pos
```
Klaim, line 768: `llround` will return a `long long` but `size_t` isnt guaranteed to be a that (it can be `unsigned int` or `unsigned long long`).
Suggestion: use `auto` here instead, or static_cast the result of `llround` to `size_t`

Issue #1737, opened 2022-06-10
micromamba create shows channels with platform in --json --dry-run mode


nstinus:
Added a test which shows:

```
E               AssertionError: assert 'https://cond...orge/linux-64' == 'https://cond...g/conda-forge'
E                 - https://conda.anaconda.org/conda-forge
E                 + https://conda.anaconda.org/conda-forge/linux-64
E                 ? 
```

PR #1738, opened 2022-06-10
micromamba create: test --dry-run --json channels returned

nstinus:
In both cases it is wrong as we add the arch:

E               AssertionError: assert 'https://cond...orge/linux-64' == 'https://cond...g/conda-forge'
E                 - https://conda.anaconda.org/conda-forge
E                 + https://conda.anaconda.org/conda-forge/linux-64
E                 ?

Issue #1737


nstinus:
No, I am not.

PR #1755, opened 2022-06-22
Improvement: Log the exception that caused a failure to parse configuration

johnhany97:
### Before this PR

When I run
```
MAMBA_LOG_PATTERN='%v' micromamba create -n base numpy --dry-run -y --json --log-level info -vvv
```
I get 
```
trace    libmamba Compute configurable 'log_pattern'
error    libmamba Bad conversion of configurable 'log_pattern' from environment variable 'MAMBA_LOG_PATTERN' with value '%v'
critical libmamba bad conversion
```
but there's no way for me to find out why it failed to parse the pattern.

### After this PR

The exception will be logged so the client can adjust their configuration.


wolfv:
nice, thanks!

Issue #1753, opened 2022-06-21
Fully statically linked build broken on Mac OS


AdrianFreundQC:
I tried to create a fully statically linked micromamba build following the documentation (https://mamba.readthedocs.io/en/latest/developer_zone/build_locally.html#fully-statically-linked) on a M1 Mac and ran into a few problems:

- [ ] `./micromamba/environment-static-dev.yml` doesn't exist. The documentation calls for installing this environment, but it doesn't exist. `./micromamba/environment-dev.yml` and `./libmamba/environment-static-dev.yml` exist and installing them both works.
- [ ] Undefined symbol:
```
Undefined symbols for architecture arm64:
  "_SCDynamicStoreCopyProxies", referenced from:
      _Curl_resolv in libcurl.a(libcurl_la-hostip.o)
ld: symbol(s) not found for architecture arm64
```
Linking Framework SystemConfiguration to libmamba fixes this. I'm currently working on a PR for this one.
- [ ] libcurl SecureTransport backend not available
```
warning  libmamba Could not use preferred SSL backend (Linux: OpenSSL, OS X: SecureTransport, Win: SChannel)
warning  libmamba Please check the cURL library configuration that you are using.
```
Looks like libcurl-static on conda-forge doesn't enable this option. libmamba falls back to using the OpenSSL backend
- [ ] `SSL certificate problem: unable to get local issuer certificate` statically linked OpenSSL doesn't find the certificate(s). Running with `--ssl-verify false` makes my micromamba build work.

wolfv:
Hi @AdrianFreundQC great! 
Yes, unfortunately the libcurl build is a very specific (older) one from conda-forge. The last working build was 

```
    - libcurl 7.76.1 *_0           # [osx]
    - libcurl-static  7.76.1 *_0   # [osx]
```

Later builds removed the `darwin-ssl` feature from `libcurl` because it requires a newer macOS version. conda-forge tries to stay compatible with old versions of macOS.

We should create a variant in the curl-feedstock that emits a version of libcurl & libcurl-static that has darwin-ssl enabled, or we should build `libcurl` as part of the micromamba feedstock.

alanhoyle:
The underlying issue appears to be that the micro.mamba.pm Let's Encrypt certificate has expired.  This affects the mac and linux installation instructions at https://mamba.readthedocs.io/en/latest/installation.html

```
[...]

PR #1759, opened 2022-06-23
Fix fully static build failing on Mac OS

AdrianFreundQC:
The build was failing at the linking stage as recent versions of curl-static still need to be dynamically linked to the SystemConfiguration framework. This fixes a part of #1753


wolfv:
closing as this seems to work on main :)

PR #1762, opened 2022-06-24
Implement --use-index-cache

### LINKED ISSUE #74: Add use_index_cache behavior
~~index cache only tells conda to use the cache _even if_ it has expired. Therefore the current logic needs to be adjusted.~~

AdrianFreundQC:
This makes the --use-index-cache option work in mamba.

Fixes #74.

Note that this still isn't quite the same as conda's behavior:
In case no index cache is present conda returns an empty index. Mamba will still download the index.

### DIFF
--- a/libmamba/src/core/subdirdata.cpp
+++ b/libmamba/src/core/subdirdata.cpp
@@ -332,7 +332,8 @@ namespace mamba
 
                     auto cache_age_seconds
                         = std::chrono::duration_cast<std::chrono::seconds>(cache_age).count();
-                    if ((max_age > cache_age_seconds || Context::instance().offline))
+                    if ((max_age > cache_age_seconds || Context::instance().offline
+                         || Context::instance().use_index_cache))
                     {
                         // valid json cache found
                         if (!m_loaded)


wolfv:
Hey! Thanks for the PR!
Looks good to me :)

dufferzafar:
@wolfv Is this released?

```
❯ micromamba --use-index-cache -v search doxygen
The following argument was not expected: --use-index-cache

❯ micromamba --version
0.25.1
```

jonashaag:
It is released. Maybe it doesn’t work for `search`. In that cause please open an issue.

jonashaag:
Or maybe you have to put `search` before the option in the command line.

dufferzafar:
The doesn't work either. Gives the same error.

AdrianFreundQC:
I just checked the code and it looks like `--use-index-cache` is currently only supported in mamba, not in micromamba. Adding it to micromamba should be relatively easy, as the code handling it internally is shared by mamba and micromamba, so only the code for parsing the command-line argument would have to be added.

dufferzafar:
@AdrianFreundQC It sounds simple enough that I'd like to take a stab at this myself. Since you have understanding of the code, could you point me to the file/function where this change would need to go? I'll try to build mamba from source locally.

AdrianFreundQC:
There's two places you need to look at:
* `libmamba/src/api/configuration.cpp` defines all the configurables and how they can be set (rc-file, env-variables, ...) for both mamba and micromamba.
* `micromamba/src/common_options.cpp` defines the micromamba cli options and connects them to libmamba configurables.

Note that in mamba CLI options are actually parsed by conda, then passed on to libmamba instead of being parsed by mamba itself (mamba reuses some parts of conda, while micromamba doesn't)

AdrianFreundQC:
Oh. And also `libmamba/include/mamba/core/context.hpp defines the Context::use_index_cache property that you actually need to set based on the configurable option.

PR #1764, opened 2022-06-28
make clobber warnings less annoying

wolfv:


wolfv:
need to decide what to do with `create_application_entry_point`

PR #1767, opened 2022-06-28
Fix shell init --dry-run plus some cleanup

jonashaag:


wolfv:
just to make sure – you removed the `reverse` for now?

jonashaag:
yes, can also bring it back, it seemed confusing because it's not implemented anywhere

wolfv:
I think the `reverse` is implemented in the conda shell init which is why I also originally planned for it... Now we have the PR that implements shell deinit in #1781 (which unfortunately will conflict with this PR, I think).

jonashaag:
Yep, happy to rebase when the other one is merged

jonashaag:
This should be good to merge but needs squashing.

PR #1773, opened 2022-07-01
Add support for nodefaults channel

### LINKED ISSUE #1107: Investigate if we handle special nodefaults correctly
Conda has a special nodefaults channel that's used to deactivate the defaults channel. I'm not sure if we handle it correctly with the new channel stuff

AdrianFreundQC:
Closes #1107

This PR adds support for the nodefaults channel to micromamba.

In contrast to conda this implementation also allows passing the `nodefaults` channel on the command line. This is because the internal design of libmamba's configuration handler doesn't allow differentiating between channels passed on the command line and channels specified in the environments.yml file.

Because the configuration handler only allows adding a single hook to an option I had to remove the channels hook from micromamba to add it to libmamba. Instead of directly handling `--override-channels` micromamba now adds `nodefaults` to the channels when it encounters `--override-channels` and lets libmamba handle the rest.

### DIFF
--- a/libmamba/src/api/install.cpp
+++ b/libmamba/src/api/install.cpp
@@ -719,17 +719,17 @@ namespace mamba
         void channels_hook(std::vector<std::string>& channels)
         {
             auto& config = Configuration::instance();
+            auto& config_channels = config.at("channels");
+            std::vector<std::string> cli_channels;
 
-            if (find(channels.begin(), channels.end(), "nodefaults") != channels.end())
+            if (config_channels.cli_configured())
             {
-                if (config.at("channels").cli_configured())
+                cli_channels = config_channels.cli_value<std::vector<std::string>>();
+                auto it = find(cli_channels.begin(), cli_channels.end(), "nodefaults");
+                if (it != cli_channels.end())
                 {
-                    channels = config.at("channels").cli_value<std::vector<std::string>>();
-                    channels.erase(find(channels.begin(), channels.end(), "nodefaults"));
-                }
-                else
-                {
-                    channels.clear();
+                    cli_channels.erase(it);
+                    channels = cli_channels;
                 }
             }
         }
--- a/libmamba/src/api/install.cpp
+++ b/libmamba/src/api/install.cpp
@@ -715,5 +715,23 @@ namespace mamba
                 }
             }
         }
+
+        void channels_hook(std::vector<std::string>& channels)
+        {
+            auto& config = Configuration::instance();
+
+            if (find(channels.begin(), channels.end(), "nodefaults") != channels.end())
+            {
+                if (config.at("channels").cli_configured())
+                {
+                    channels = config.at("channels").cli_value<std::vector<std::string>>();
+                    channels.erase(find(channels.begin(), channels.end(), "nodefaults"));
+                }
+                else
+                {
+                    channels.clear();
+                }
+            }
+        }
     }  // detail
 }  // mamba
--- a/micromamba/src/common_options.cpp
+++ b/micromamba/src/common_options.cpp
@@ -170,7 +170,7 @@ init_channel_parser(CLI::App* subcom)
     auto& config = Configuration::instance();
 
     auto& channels = config.at("channels");
-    channels.set_post_merge_hook(channels_hook).needs({ "override_channels" });
+    channels.needs({ "override_channels" });
     subcom
         ->add_option("-c,--channel", channels.get_cli_config<string_list>(), channels.description())
         ->type_size(1)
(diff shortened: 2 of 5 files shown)


jonashaag:
I guess we should add a test for this.

micromamba/src/common_options.cpp
```diff
@@ -256,6 +238,17 @@ override_channels_hook(bool& value)
             << "'override_channels' disabled by 'override_channels_enabled' set to 'false' (skipped)";
         value = false;
     }
+
+    if (value)
```
wolfv, line 242: Do we really need this? We actually do not have "`defaults`" at all in micromamba. If we have to add `nodefaults` here, it would mean that the `defaults` channel has been configured in the `.mambarc` file. However, that should already be taken care of by using the `--override-channels` option, right?

libmamba/src/api/install.cpp
```diff
@@ -715,5 +715,23 @@ namespace mamba
...
+            if (find(channels.begin(), channels.end(), "nodefaults") != channels.end())
+            {
+                if (config.at("channels").cli_configured())
+                {
+                    channels = config.at("channels").cli_value<std::vector<std::string>>();
+                    channels.erase(find(channels.begin(), channels.end(), "nodefaults"));
+                }
+                else
+                {
+                    channels.clear();
```
wolfv, line 732: Is that correct, to clear the entire `channels` vector?

micromamba/src/common_options.cpp
```diff
@@ -224,30 +224,12 @@ init_channel_parser(CLI::App* subcom)
     channel_priority.needs({ "strict_channel_priority", "no_channel_priority" });
 }
 
-void
-channels_hook(std::vector<std::string>& channels)
-{
-    auto& config = Configuration::instance();
-    bool override_channels = config.at("override_channels").value<bool>();
```
wolfv, line 231: was this line lost?

wolfv:
Hey, thanks for your PR! I finally found the time to review it and added some questions.

libmamba/src/api/install.cpp
```diff
@@ -715,5 +715,23 @@ namespace mamba
...
+        void channels_hook(std::vector<std::string>& channels)
+        {
+            auto& config = Configuration::instance();
+
+            if (find(channels.begin(), channels.end(), "nodefaults") != channels.end())
+            {
+                if (config.at("channels").cli_configured())
+                {
+                    channels = config.at("channels").cli_value<std::vector<std::string>>();
+                    channels.erase(find(channels.begin(), channels.end(), "nodefaults"));
```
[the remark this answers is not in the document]
AdrianFreundQC: `channels` has a different value at the second `find`. At the first `find` it contains Channels from all sources (.condarc, ENV variables, cli options, environment.yml), while at the second `find` it only contains the options from the cli and environment.yml (due to a quirk in libmamba's environment.yml parsing the channels just get appended to the channels passed on the cli, so they also get returned by `cli_value()`.

libmamba/src/api/install.cpp
AdrianFreundQC: Clearing the `channels` vector is what the previous `override_channels` hook did, so it should still be correct.
`nodefaults` means only channels passed explicitly (cli and environment.yml) should be used. If no channels are passed explicitly and we won't use non-explicitly passed channels then we don't have any channels.


`override_channels` isn't used anymore, as the `override_channels` hook now appends `nodefaults` to the channels. `--override-channels` on the cli and `nodefaults` in environment.yml` do the same thing.

micromamba/src/common_options.cpp
AdrianFreundQC: No this isn't needed anymore. See commenty above. Also we can't use `override_channels` from the new hook location in libmamba, as the `override_channels` configurable is only defined in micromama, so doesn't exist if libmamba is used from mamba.

micromamba/src/common_options.cpp
AdrianFreundQC: Despite it's name `nodefaults` has nothing to do with `defaults`.

`defaults` tells conda to use the default anaconda channels. `nodefaults` tells conda to not use any channels from .condarc, but only use channels from the current environment.yml file.

libmamba/src/api/install.cpp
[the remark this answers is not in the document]
AdrianFreundQC: Although I could probably get the `cli_value()` first and then do both searches on it. Having `nodefaults` in `.condarc` doesn't make any sense and also isn't allowed by conda.

AdrianFreundQC:
I checked out the failing tests. The only tests that are failing are the tests that are currently also failing on master and are unrelated to these changes.

PR #1776, opened 2022-07-05
Remove password from log

AntoinePrv:
There was use of `std::endl` in one message and not in the other. Not sure where it's needed.

### DIFF
--- a/libmamba/src/core/context.cpp
+++ b/libmamba/src/core/context.cpp
@@ -222,14 +222,15 @@ namespace mamba
                         if (pass)
                         {
                             info.value = concat(user, ":", pass.value());
-                            LOG_INFO << "Found credentials for user " << user << " for host" << host
-                                     << " in ~/.mamba/auth/authentication.json";
+                            LOG_INFO << "Found credentials for user " << user << " for host "
+                                     << host << " in ~/.mamba/auth/authentication.json";
                         }
                         else
                         {
-                            LOG_ERROR << "Found credentials for user " << user << " for host"
-                                      << host << " in ~/.mamba/auth/authentication.json"
-                                      << "but could not decode base64 password" << std::endl;
+                            LOG_ERROR
+                                << "Found credentials for user " << user << " for host " << host
+                                << " in ~/.mamba/auth/authentication.json but could not decode base64 password"
+                                << std::endl;
                         }
                     }
                     m_authentication_info[host] = info;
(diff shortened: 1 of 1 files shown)


libmamba/src/core/context.cpp
```diff
@@ -211,23 +211,27 @@ namespace mamba
...
                         {
-                            info.value = concat(el.value("user", ""), ":", pass.value());
+                            info.value = concat(user, ":", pass.value());
+                            LOG_INFO << "Found credentials for user " << user << " for host" << host
+                                     << " in ~/.mamba/auth/authentication.json";
                         }
                         else
                         {
-                            LOG_ERROR << "Could not decode base64 password" << std::endl;
+                            LOG_ERROR << "Found credentials for user " << user << " for host"
```
wolfv, line 230: here as well

libmamba/src/core/context.cpp
```diff
@@ -211,23 +211,27 @@ namespace mamba
...
+                                      << host << " in ~/.mamba/auth/authentication.json"
```
wolfv, line 231: here we need another space :)

wolfv:
Looks good, just some spaces missing

AntoinePrv:
`Run server auth test` are failing at `  × git checkout -q master did not run successfully.`
Am I missing some permissions?

wolfv:
Looks like they renamed the branch to `main`... https://github.com/conda/conda-content-trust

Do you want to fix that in this PR? Happy to merge right away – or we make a seperate PR to fix the CI.

jonashaag:
I think this particular test failure comes from env caching. It caches `build_env/envs`. I made a fix here but I think it would be better to revert this fix and add the `envs/` cleanup to the test suite instead. https://github.com/mamba-org/mamba/blob/f751daf8039008aea945e88044d6c503fa9da046/.github/workflows/main.yml#L45

AntoinePrv:
Yes, it look like the "Cleanup" steps does not get run unconditionally (which happened in the previous failure).

Perhaps we could extend `cache-env` could accept a list of environment to cache/ignore so that it gets always run?
Otherwise it could be nice to have a generic `Finally` action that let you customize some `post` command.

jonashaag:
You can have `if: always()` in GHA.

I think the problem should be fixed in the Mamba test suite because it's not nice to leave traces behind oneself in tests :)

PR #1781, opened 2022-07-07
Add micromamba deinit shell command

### LINKED ISSUE #1763: [micromamba] Add option to reverse `micromamba shell init`
It would be nice to have a command that reverses everything that `micromamba shell init` does. Something like `micromamba shell deinit`.

pavelzw:
Closes #1763 

This PR adds the option to revert the changes that are performed by `micromamba shell init`.

### DIFF
--- a/libmamba/src/core/shell_init.cpp
+++ b/libmamba/src/core/shell_init.cpp
@@ -29,11 +29,11 @@ namespace mamba
 {
     namespace
     {
-        std::regex CONDA_INITIALIZE_RE_BLOCK("# >>> mamba initialize >>>(?:\n|\r\n)?"
+        std::regex CONDA_INITIALIZE_RE_BLOCK("\n# >>> mamba initialize >>>(?:\n|\r\n)?"
                                              "([\\s\\S]*?)"
                                              "# <<< mamba initialize <<<(?:\n|\r\n)?");
 
-        std::regex CONDA_INITIALIZE_PS_RE_BLOCK("#region mamba initialize(?:\n|\r\n)?"
+        std::regex CONDA_INITIALIZE_PS_RE_BLOCK("\n#region mamba initialize(?:\n|\r\n)?"
                                                 "([\\s\\S]*?)"
                                                 "#endregion(?:\n|\r\n)?");
     }
@@ -42,27 +42,38 @@ namespace mamba
     {
         std::string parent_process_name = get_process_name_by_pid(getppid());
 
-        if (contains(parent_process_name, "bash"))
+        LOG_DEBUG << "Guessing shell. Parent process name: " << parent_process_name;
+
+        std::string parent_process_name_lower = to_lower(parent_process_name);
+
+        if (contains(parent_process_name_lower, "bash"))
         {
             return "bash";
         }
-        if (contains(parent_process_name, "zsh"))
+        if (contains(parent_process_name_lower, "zsh"))
         {
             return "zsh";
         }
-        if (contains(parent_process_name, "xonsh"))
+        // xonsh in unix, Python in macOS
+        if (contains(parent_process_name_lower, "python"))
+        {
+            Console::stream() << "Your parent process name is " << parent_process_name
+                              << ".\nIf your shell is xonsh, please use \"-s xonsh\"." << std::endl;
+        }
+        if (contains(parent_process_name_lower, "xonsh"))
         {
             return "xonsh";
         }
-        if (contains(parent_process_name, "cmd.exe"))
+        if (contains(parent_process_name_lower, "cmd.exe"))
         {
             return "cmd.exe";
         }
-        if (contains(parent_process_name, "powershell") || contains(parent_process_name, "pwsh"))
+        if (contains(parent_process_name_lower, "powershell")
+            || contains(parent_process_name_lower, "pwsh"))
         {
             return "powershell";
         }
-        if (contains(parent_process_name, "fish"))
+        if (contains(parent_process_name_lower, "fish"))
         {
             return "fish";
         }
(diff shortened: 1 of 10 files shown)


libmamba/src/core/shell_init.cpp
```diff
@@ -663,7 +835,63 @@ namespace mamba
...
+        else
+        {
+            throw std::runtime_error("Support for other shells not yet implemented.");
+        }
+
+        deinit_root_prefix(shell, conda_prefix);
+
+#ifdef _WIN32
+        if (Console::prompt("Do you want to disable long path support?", 'n')) {
+            set_long_paths_support(0, false);
```
wolfv, line 893: This is tough - I think a lot of people aren't aware of this registry setting and we do not know at all if mamba was setting it or if it was previously configured. I also think it doesn't do any harm to have it enabled. I am not sure if we should even ask to remove it...

pavelzw: I guess it doesn't do any harm when leaving it enabled... Let's leave this line out then.

pavelzw:
Fixes #1785

pavelzw:
Fixes #1800

Adds xonsh shell init testing.

libmamba/src/core/shell_init.cpp
```diff
@@ -49,7 +51,8 @@ namespace mamba
         {
             return "zsh";
         }
-        if (contains(parent_process_name, "xonsh"))
+        // xonsh in unix, Python in macOS
+        if (contains(parent_process_name, "xonsh") || contains(parent_process_name, "Python"))
```
jonashaag, line 55: Hmm, is this safe? What if we run `subprocess.run(["micromamba", "shell", "init"])`, then the parent process is Python.

I think it would be ok to require explicit `-s xonsh` instead.

libmamba/src/core/shell_init.cpp
```diff
@@ -69,6 +72,66 @@ namespace mamba
...
+            std::wstring_view stripped = wstrip(*it);
+            // delete if stripped version matches to hook_string
+            if (stripped == hook_string)
+            {
+                autorun_list.erase(it);
+
+                std::cout << "Removing from cmd.exe AUTORUN: " << termcolor::green;
+                std::wcout << hook_string;
+                std::cout << termcolor::reset << std::endl;
+                break;
```
jonashaag, line 123: Is there a reason to limit this to one replacement?

libmamba/src/core/shell_init.cpp
```diff
@@ -69,6 +72,66 @@ namespace mamba
...
+            }
+        }
+
+        // join the list back into a string
+        return join(L" & ", autorun_list);
+    }
+
     void init_cmd_exe_registry(const std::wstring& reg_path,
                                const fs::path& conda_prefix,
                                bool reverse)
```
jonashaag, line 137: Looks like this is the only place where we use `reverse` instead of having another function, shouldn't we make it consistent?

libmamba/src/core/shell_init.cpp
```diff
@@ -439,6 +525,44 @@ namespace mamba
...
+        for (auto& f : { micromamba_bat,
+                         _mamba_activate_bat,
+                         condabin_activate_bat,
+                         scripts_activate_bat,
+                         mamba_hook_bat })
+        {
+            if (fs::exists(f))
+            {
+                fs::remove(f);
+                LOG_INFO << "Removed " << f << " file.\n";
```
jonashaag, line 545: I think (but not sure) `LOG_XXX <<` doesn't need `\n`

libmamba/src/core/shell_init.cpp
```diff
@@ -572,6 +761,84 @@ namespace mamba
...
+
+        Console::stream() << "Removing the following in your " << profile_path << " file\n"
+                          << termcolor::colorize << termcolor::green
+                          << "#region mamba initialize\n...\n#endregion\n"
+                          << termcolor::reset;
+
+        profile_content = std::regex_replace(profile_content, CONDA_INITIALIZE_PS_RE_BLOCK, "");
+        LOG_DEBUG << "Profile content:\n" << profile_content << "\n";
+
+        if (!Context::instance().dry_run)
```
jonashaag, line 783: You can save 1 layer of nesting by having an early `return` here.

libmamba/src/core/shell_init.cpp
```diff
@@ -572,6 +761,84 @@ namespace mamba
...
+                reproc::sink::string(out),
+                reproc::sink::string(err));
+            if (ec)
+            {
+                throw std::runtime_error(ec.message());
+            }
+            return std::string(strip(out));
+        }
+        catch (...)
+        {
```
jonashaag, line 837: Do we want to have `LOG_DEBUG <<` the exception msg here?

libmamba/src/core/util.cpp
```diff
@@ -246,6 +246,24 @@ namespace mamba
         return strip(input, WHITESPACES);
     }
 
+    // todo one can deduplicate this code with the one in strip() by using C++ 20 with concepts
+    std::wstring_view wstrip(const std::wstring_view& input)
+    {
+        return wstrip(input, WHITESPACES_WSTR);
+    }
+
+    std::wstring_view wstrip(const std::wstring_view& input, const std::wstring_view& chars)
```
jonashaag, line 255: You didn't find a way to use templating for `strip`?

jonashaag (approved):
Awesome!

libmamba/src/core/shell_init.cpp
pavelzw: My python parent process is called `python3.10`. Since micromamba looks for `Python` with a capital P, everything should be fine. 🤡🤪 

I think a good solution would be to check for python (case insensitive) and print a special info message telling the user about `-s xonsh`.

libmamba/src/core/shell_init.cpp
pavelzw: No.

libmamba/src/core/util.cpp
pavelzw: You are right, the only thing that doesn't work is templating when using `WHITESPACES` or `WHITESPACES_WSTR` because they have a fixed type... But `strip can be deduplicated`. 👍🏻

libmamba/src/core/shell_init.cpp
pavelzw: Actually, if we remove this `break`, the code runs into a memory error... I'll try to fix it now.

wolfv:
Very impressive work! Thanks a lot –– I just finished reading through the PR. Happy to merge if tests are passing :)

wolfv:
PS: I can squash all commits down to a single one. Or you can squash them interactively (and re-wording them) by running `git rebase -i origin/master`

Let me know which way you prefer.

jonashaag:
You can also enable squash merges in GitHub

pavelzw:
I think squashing the commits is probably the easiest...

pavelzw:
I rebased it onto master. Can you restart the workflows please @wolfv?

pavelzw:
Hmm `incorrect checksum\n\x00|\xf8\xd9\x1cV\x00\x00ntrib-jsmath-1.0.1-py_0. \x01\x00\x00\x00\x00\x00 extraction failed\n'` seems like a temporary CI problem… Could you run the failed tests again? 😅

wolfv:
@jonashaag yep, I can always squash from Github but sometimes people prefer to keep multiple commits :)

Issue #1720, opened 2022-06-01
`micromamba auth` with multiple channels per host


jonashaag:
Similar to https://github.com/conda-incubator/conda-lock/pull/127, we have multiple channels on the same host, at different URL prefixes.

The current design/impl of `micromamba auth login` doesn't allow for this.

PR #1784, opened 2022-07-08
Multi channel per host

AntoinePrv:
Close https://github.com/mamba-org/mamba/issues/1720

- [x] Create test case
  - [x] Extend `reposerver.py` to allow multiple channels
  - [x] Add multi channel tests to `testserver.sh`
- [x] Implement `micromamba` with multiple channel per host
  - [x] Change key written in `authentication.json`
  - [x] Search for auth with and without channel
  - [x] Remove scheme from login id
  - [x] Edit key used for matching channels
  - [x] Add relevant multi-channel tests
- [x] This PR branch is based on https://github.com/mamba-org/mamba/pull/1791

### DIFF
--- a/docs/source/user_guide/micromamba.rst
+++ b/docs/source/user_guide/micromamba.rst
@@ -20,19 +20,25 @@ Quickstarts
   $ micromamba --help
 
   Subcommands:
-    activate                    Activate a conda / micromamba environment
-    deactivate                  Deactivate a conda / micromamba environment
     shell                       Generate shell init scripts
     create                      Create new environment
     install                     Install packages in active environment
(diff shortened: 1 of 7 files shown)


AntoinePrv:
In the implementation, I've assume authentication for `conda.anaconda.org` and `conda.anaconda.org/conda-forge` are two separate  things.

Here, tests are failing (partly) because before authentication for `conda.anaconda.org` was for all channels.
Should we assume that if no authentication is found for `conda.anaconda.org/conda-forge`, we  look up `conda.anaconda.org/`? I believe that's what is done in https://github.com/conda-incubator/conda-lock/pull/127.

PR #1789, opened 2022-07-11
add some utility functions for better error reporting

wolfv:


libmamba/include/mamba/core/pool.hpp
```diff
@@ -22,6 +22,64 @@ namespace spdlog
...
+    class MQueue
+    {
+    public:
+        MQueue()
+            : p_queue(new Queue)
+        {
+            queue_init(p_queue);
+        }
+
+        ~MQueue()
```
Klaim, line 34: This deletes copy and move operators, although for future readers I think it would be best to state that explicitly, as intentional (with `= delete`).

libmamba/include/mamba/core/pool.hpp
```diff
@@ -22,6 +22,64 @@ namespace spdlog
...
+        {
+            return &p_queue->elements[0];
+        }
+
+        const Id* end() const
+        {
+            return &p_queue->elements[p_queue->count];
+        }
+
+        operator Queue*()
```
Klaim, line 69: Did you consider making it explicit? Not sure what's best in this specific case, I'm just wondering every time I see a conversion operator.

libmamba/include/mamba/core/pool.hpp
```diff
@@ -22,6 +22,64 @@ namespace spdlog
...
+        {
+            return p_queue;
+        }
+
+        std::vector<Id> as_vector()
```
Klaim, line 74: Not sure if it's necessary as a member function, could have been a generic function.

Klaim:
LGTM, I think some code could be refactored to reduce code and help once this is merged.

libmamba/include/mamba/core/pool.hpp
wolfv: It makes it simple to use the Queue in interactions with `libsolv`. I think with explicit, we'd lose the `implicit` conversion, e.g. when doing something like `queue_pop(my_mqeue, 1)`, right?

libmamba/include/mamba/core/pool.hpp
wolfv: Good idea. We can also implement them (now or later).

libmamba/include/mamba/core/pool.hpp
Klaim: Indeed. The main issue with conversion operations is unwanted implicit conversions, but I guess this one is fine as nothing is lost in the conversion.

PR #1790, opened 2022-07-13
Fix #1783: Add `micromamba env create`

### LINKED ISSUE #1783: Add `micromamba env create`
xref https://github.com/pre-commit/pre-commit/pull/2446

jonashaag:
Fixes #1783


jonashaag:
Yep but `micromamba create` is (mostly) equivalent to `conda env create`

AntoinePrv:
I agree with the title of the conda issue that having both `create` and `env create` is confusing.
Since I don't think `micromamba` CLI is supposed to be 100% compatible with `conda`, I'd argue against this unless there is a good reason (and even, perhaps find a less confusing name).

jonashaag:
There is a really good reason which is that `env create` is required for pre-commit support: there is the `PRE_COMMIT_MICROMAMBA` flag in pre-commit but it doesn't work because pre-commit expects `env create` to exist and the pre-commit project doesn't accept changes to make it work with the `env` interface.

jonashaag:
@wolfv merge this?

jonashaag:
@wolfv @AntoinePrv can someone please approve this, otherwise I'm unable to merge

wolfv (approved):

PR #1792, opened 2022-07-13
Fix #1751: Better error message if root prefix is not a directory

### LINKED ISSUE #1751: micromamba crashes when launched from home folder
To reproduce, execute
```
sudo docker run -it --rm ubuntu:latest bash
apt update ; apt install -y curl bzip2
cd ; curl -Ls https://micro.mamba.pm/api/micromamba/linux-64/latest | tar -xvj bin/micromamba ; cp -r bin/micromamba ~
./micromamba; echo $?
```
and observe that the exit code 1 is printed instead of the micromamba info page followed by exit code 0

I observed this behavior on Ubuntu 22.04 and CentOS 7.

If micromamba is placed in home folder, other copies of the executable placed in other location will crash as well.

jonashaag:
Fixes https://github.com/mamba-org/mamba/issues/1751

@adriendelsalle it seems like these lines cause the error messages I improved not to be shown. Is it on purpose that only critical errors are printed in this case? https://github.com/mamba-org/mamba/blob/f2ac46b1c478668ed09463cf5b88d4802af7661b/libmamba/src/api/configuration.cpp#L1636-L1637

### DIFF
--- a/libmamba/src/api/configuration.cpp
+++ b/libmamba/src/api/configuration.cpp
@@ -532,13 +532,20 @@ namespace mamba
                     LOG_WARNING << "'root_prefix' set with default value: " << prefix.string();
                 }
 
-                if (fs::exists(prefix) && !fs::is_empty(prefix))
+                if (fs::exists(prefix))
                 {
-                    if (!(fs::exists(prefix / "pkgs") || fs::exists(prefix / "conda-meta")
-                          || fs::exists(prefix / "envs")))
+                    if (fs::is_directory(prefix))
                     {
+                        if (!fs::is_empty(prefix) && (!(fs::exists(prefix / "pkgs") || fs::exists(prefix / "conda-meta")
+                                  || fs::exists(prefix / "envs"))))
+                        {
+                            LOG_ERROR << "Could not use default 'root_prefix': " << prefix.string();
+                            LOG_ERROR << "Directory exists, is not empty and not a conda prefix.";
+                            exit(1);
+                        }
+                    } else {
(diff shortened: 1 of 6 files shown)


adriendelsalle:
Hi @jonashaag the configuration computation starts before knowing the log level from rc file, cli flag or env var so I disabled the logger to avoid printing some logs from configurables that are loaded before the log level.
We use the backtrace to replay logs if some errors are thrown.
Since this I also implemented a log buffer needed for the progress bars display (to flush the log messages and regular prints at the right time when updating the status, between deleting the N previous lines and writing the new ones)

adriendelsalle:
The logs are printed at the end of the same function https://github.com/mamba-org/mamba/blob/f2ac46b1c478668ed09463cf5b88d4802af7661b/libmamba/src/api/configuration.cpp#L1668

jonashaag:
Thanks, but then we have stuff like this that makes the logs never show up https://github.com/mamba-org/mamba/blob/57668935ce5697616d8bc1d2e1b6b0da6589a0c5/libmamba/src/api/configuration.cpp#L542

adriendelsalle:
Indeed, it's also bad to use `exit` for bindings because it would cause `R` or `Python` interpreters to exit from interactive mode.
Could you try to throw an error instead?

@wolfv we discussed a long time ago to refactor the code base to replace those statements this way, right?

jonashaag:
Changed most `exit()` calls to C++ exception.

wolfv:
Thanks! Nice improvement

PR #161, opened 2020-04-22
add proxy support

wolfv:

PR #1814, opened 2022-07-22
Add support for configuring proxies in .condarc

### LINKED ISSUE #151: Implement ssl_verify and SOCKS network options

AdrianFreundQC:
Closes #151, based on #161.

This PR adds support for configuring proxies in the `.condarc` file.

Current progress:

- [x] Implementation
- [x] Automated test for configuration
- [x] Automated test for proxy selection
- [x] Automated test for usage of the proxy (Might be to difficult, because an actual proxy server would be needed)

Manual tests:
- [x] Proxy without ssl_verify
- [x] Proxy with ssl_verify and cert_path
- [ ] Proxy with ssl_verify without cert_path (doesn't find certs manually added to macOS Keychain. Does OpenSSL installed by conda use it's own keychain? Seems unrelated to this PR.)
- [x] Proxy with authentication if authentication is in .condarc (Not tested yet. Should just be handled by curl)
- [ ] Proxy with authentication if authentication isn't in .condarc (Conda documentation claims it prompts the user for authentication, but in my tests it just prints an error. Mamba just fails to connect and also prints an error currently)

Open questions:
~~Should we log the proxy being used?
https://github.com/mamba-org/mamba/pull/1814/files#diff-bb45c384b0499833356c5e8fb5f6ee142bdbc44187536c1c801d2f986647beffR267~~

~~This could be very useful for debugging, however the proxy strings could contain usernames and passwords if they are configured in .condarc.~~
Password is now redacted in the log.


libmamba/src/core/fetch.cpp
```diff
@@ -260,18 +260,34 @@ namespace mamba
             curl_easy_setopt(handle, CURLOPT_SSL_OPTIONS, CURLSSLOPT_NO_REVOKE);
         }
 
+        const char* proxy = proxy_match(url);
```
jonashaag, line 263: I guess it would be preferable to use `std::string` whenever possible.

jonashaag:
For automated tests I think it would be good to at least cover simple cases with a proper proxy like mitmproxy. IMO we don't have to have full integration tests for all cases if it's too hard to set up.

wolfv:
Yes, locally i was also using mitmproxy and found it to work quite well.

libmamba/src/core/fetch.cpp
AdrianFreundQC: I just copied that from #161, as I thought there might be a reason for using `char*`. I can try to change it tough.
The problem is that `std::string` isn't allowed to be `nullptr` and using empty string instead feels a bit hacky. I might be able to use `std::optional` though.

AdrianFreundQC:
I also used mitmproxy for the manual tests. As it's written in python it might be convenient to integrate into the mamba or micromamba tests, which are also in python. It is however not available on conda-forge, so I would have to look into porting the pypi package to conda first.

AdrianFreundQC:
> How do the `proxy_server` settings work together with system-wide settings like `HTTP_PROXY` and `NO_PROXY`? For the latter see also [conda/conda#11411](https://github.com/conda/conda/issues/11411), which we should implement in Mamba.

The env variables are handled directly by curl, while the settings in the `.condarc` are handled by libmamba and passed to curl as options, so the behavior currently entirely depends on curl's precedence rules. I'll look into how it currently works and, if needed, manual behavior changes, such as not passing the options to curl if the env variables are set.

AdrianFreundQC:
Progress on submitting mitmproxy to conda-forge: https://github.com/conda-forge/staged-recipes/pull/19813

AdrianFreundQC:
I added some tests to micromamba. The CI currently fails as mitmproxy isn't in conda-forge yet, but the tests pass locally. Once that has been added I will try to finish up this PR.

AdrianFreundQC:
I finally got a windows VM with working development environment set up, so I can now look into the failing windows tests.

jonashaag:
Maybe you document your VM setup steps here in the ticket or in the docs?

AdrianFreundQC:
> > How do the `proxy_server` settings work together with system-wide settings like `HTTP_PROXY` and `NO_PROXY`? For the latter see also [conda/conda#11411](https://github.com/conda/conda/issues/11411), which we should implement in Mamba.
> 
> The env variables are handled directly by curl, while the settings in the `.condarc` are handled by libmamba and passed to curl as options, so the behavior currently entirely depends on curl's precedence rules. I'll look into how it currently works and, if needed, manual behavior changes, such as not passing the options to curl if the env variables are set.

According to the curl documentation setting a proxy using `CURLOPT_PROXY` overrides proxies set using the env variables, however, in contrast to `requests`, which is used by conda, curl still uses the `NO_PROXY` environment variable unless `CURLOPT_NOPROXY` is used (which is currently never used by mamba).

This means with this PR it's already possible to specify proxies in .condarc and exclude urls using the `NO_PROXY` env variable.
For `no_proxy` support in .condarc I would prefer waiting for conda to implement it to not end up with incompatibilities in .condarc

AdrianFreundQC:
Tests pass. I think this should be ready for final review and merging.

libmamba/src/core/fetch.cpp
```diff
@@ -260,18 +260,34 @@ namespace mamba
             curl_easy_setopt(handle, CURLOPT_SSL_OPTIONS, CURLSSLOPT_NO_REVOKE);
         }
 
+        std::optional<std::string> proxy = proxy_match(url);
+        if (proxy)
+        {
+            curl_easy_setopt(handle, CURLOPT_PROXY, proxy->c_str());
+            LOG_INFO << "Using Proxy " << *proxy;
```
jonashaag, line 267: Assuming that this may contain authentication info, we should strip it before logging. Not sure how it is implemented elsewhere.

libmamba/src/core/util.cpp
```diff
@@ -1487,4 +1488,33 @@ namespace mamba
         return std::string((const char*) output.data());
     }
 
+    std::optional<std::string> proxy_match(const std::string& url)
+    {
+        auto& proxies = Context::instance().proxy_servers;
+        if (proxies.empty())
```
jonashaag, line 1494: Nit: Should be ok to drop this case.

libmamba/src/core/util.cpp
```diff
@@ -1487,4 +1488,33 @@ namespace mamba
...
+        {
+            return std::nullopt;
+        }
+
+        std::size_t max_match = 0;
```
jonashaag, line 1499: Nit: Could drop this and use `match.size()` instead.

micromamba/tests/dump_proxy_connections.py
```diff
@@ -0,0 +1,20 @@
+from mitmproxy import ctx
```
jonashaag, line 1: This is great. Can you add a tiny description what this does and how to use it? Might be useful even outside of the tests.

micromamba/tests/test_proxy.py
```diff
@@ -0,0 +1,120 @@
...
+
+    def start_proxy(self, port, options=[]):
+        assert self.proxy_process is None
+        script = Path(__file__).parent / "dump_proxy_connections.py"
+        self.proxy_process = subprocess.Popen(
+            [
+                TestProxy.mitm_exe,
+                "--listen-port",
+                str(port),
+                "-s",
```
jonashaag, line 50:
```suggestion
                "--scripts",
```

micromamba/tests/test_proxy.py
```diff
@@ -0,0 +1,120 @@
...
+    @pytest.mark.parametrize("ssl_verify", (True, False))
+    def test_install(self, unused_tcp_port, with_auth, ssl_verify):
+        if with_auth:
+            options = ["--proxyauth", "foo:bar"]
+            proxy_url = "http://foo:bar@localhost:{}".format(unused_tcp_port)
+        else:
+            options = []
+            proxy_url = "http://localhost:{}".format(unused_tcp_port)
+
+        self.start_proxy(unused_tcp_port, options)
```
jonashaag, line 82:
```suggestion
            proxy_options = ["--proxyauth", "foo:bar"]
            proxy_url = "http://foo:bar@localhost:{}".format(unused_tcp_port)
        else:
            proxy_options = []
            proxy_url = "http://localhost:{}".format(unused_tcp_port)

        self.start_proxy(unused_tcp_port, proxy_options)
```

micromamba/tests/test_proxy.py
```diff
@@ -0,0 +1,120 @@
...
+        self.proxy_process.terminate()
+        try:
+            self.proxy_process.wait(3)
+        except TimeoutExpired:
+            self.proxy_process.kill()
+        self.proxy_process = None
+
+    @pytest.mark.parametrize("with_auth", (True, False))
+    @pytest.mark.parametrize("ssl_verify", (True, False))
+    def test_install(self, unused_tcp_port, with_auth, ssl_verify):
```
jonashaag, line 74: Can you add a tiny description here how this test works (checks that all expected requests are in the dump)

jonashaag (approved):
Very clean implementation, love it! Great work.

libmamba/src/core/util.cpp
AdrianFreundQC: I would prefer to leave that in for readability. It's not immediately obvious that empty `proxies` leads to match staying the default constructor value, which is empty string, which in turn fails the `.size !=0` check.

jonashaag:
@wolfv please review if you have time. Would like to merge this in 1-2 weeks.

wolfv:
I just looked through the PR and it looks wonderful to me. I'll let you merge whenever you want @jonashaag – I guess you communicated internally about the latest changes from yesterday?

Issue #1805, opened 2022-07-19
[micromamba] micromamba create fails to install pip dependencies if prefix contains whitespace (Windows)


AtR1an:
I have a very simple environment.yml looking something like this:
```
name: test_env
channels:
- conda-forge
dependencies:
- python=3.9
- pip
- pip:
  - pydantic
```

Creating an environment with this yml works fine if there are no whitespaces in the prefix: `micromamba create -p "C:/tmp/micromamba_envs/env" --file "C:/tmp/environment.yml"`.
However, if I have a whitespace in the prefix: `micromamba create -p "C:/tmp/micromamba envs/env` --file "C:/tmp/environment.yml"` the installation of the pip dependencies fails with the following error:
```
Installing pip packages: typing-extensions === 4.3.0, pydantic === 1.9.1
failed to create process.
```

Note that this happens on a Windows 10 machine with micromamba 0.24.0.

chaubold:
This issue also shows up on macOS. It seems like the path to the pip executable is not put in quotes when micromamba invokes it:

```
Installing pip packages: pydantic
/var/folders/<snip>: /Users/chaubold/Desktop/test env/bin/pip: /Users/chaubold/Desktop/test: bad interpreter: No such file or directory
critical libmamba pip failed to install packages
```

wolfv:
that would be a very welcome contribution @chaubold !

chaubold:
I was afraid that that would be the answer 😅 We'll see what we can do

chaubold:
Adding quotes did not solve the issue, but the difference seems to be that `mamba` calls `pip` using the pip executable/script, whereas conda calls `python -m pip` as seen in a conda log:

```
Installing pip dependencies: - Ran pip subprocess with arguments:
['/Users/chaubold/Desktop/conda test env/bin/python', '-m', 'pip', 'install', '-U', '-r', '/Users/chaubold/Desktop/condaenv.bvmf2kqm.requirements.txt']
```

Conda also warns about having spaces in environment paths, I haven't seen this warning in `mamba`. But in any case, running pip directly in an env with a space in its name does not work on UNIXes because the content of the pip executable script is:

```
#!/Users/chaubold/Desktop/conda test env/bin/python3.9
# -*- coding: utf-8 -*-
import re
import sys

from pip._internal.cli.main import main

if __name__ == '__main__':
    sys.argv[0] = re.sub(r'(-script\.pyw?|\.exe)?$', '', sys.argv[0])
[...]

PR #1815, opened 2022-07-22
Fix pip execution in environments with spaces in path (#1805)

chaubold:
Conda executes pip via `python -m pip install ...` when installing pip dependencies into the environment. Mamba is using `pip install` directly, but that does not work in environments with spaces in the path. This PR makes sure mamba als uses `python -m pip install ...` during the package installation process.

See #1805 for a longer discussion.


wolfv:
Thanks! nice fix.

wolfv:
I think it may be related to using `python` vs `python.exe`.

chaubold:
Oh, good catch, thanks for the fix!

PR #1816, opened 2022-07-22
Add micromamba docs

jonashaag:


jonashaag:
@marimeireles this works now but there seem to be some problems with `api_gen.py` (maybe because of changed Micromamba output?)

PR #1821, opened 2022-07-25
CI: Add fully static micromamba build

jonashaag:


jonashaag:
@wolfv I think it would be good to have this build in CI. I'm not happy with my environment.yml changes though. What would you recommend?

PR #1822, opened 2022-07-26
Small libmambapy improvements

AntoinePrv:


libmamba/src/core/pool.cpp
```diff
@@ -98,6 +98,15 @@ namespace mamba
         return id;
     }
 
+    std::optional<PackageInfo> id2pkginfo(Id id)
```
wolfv, line 101: MPool::id2pkginfo, right?

PR #1828, opened 2022-07-28
Fixed libmamba crash when using `pip`

Klaim:
Also added an overload of `env::which` taking a list of search paths for the executable to find.

### DIFF
--- a/libmamba/include/mamba/core/activation.hpp
+++ b/libmamba/include/mamba/core/activation.hpp
@@ -63,7 +63,6 @@ namespace mamba
                                         const std::string& conda_default_env,
                                         int old_conda_shlvl);
 
-        std::vector<fs::path> get_path_dirs(const fs::path& prefix);
         std::vector<fs::path> get_clean_dirs();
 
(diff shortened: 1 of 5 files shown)


chaubold:
I'm sorry that I introduced that crash on Windows with my fix in #1815, but @Klaim your fix actually re-introduces the problem that installing packages with pip in environments with spaces in their path does not work 😞 #1805. 

Now, using `micromamba=0.25.1` the Python executable used for the pip installation process is not found if the path to Python contains a space. At least on Mac. If I use `micromamba=0.25.0` then it works.

Klaim:
@chaubold Ouch, sorry about that, I wasn't aware of that other issue. I'll make a fix so that both issues work.

chaubold:
Yeah, I should have added a test in the first place... Sorry about that. 
If you set up a fix that would be awesome, please ping me in the PR, I'm happy to try it out and help reviewing.

Klaim:
@chaubold https://github.com/mamba-org/mamba/pull/1838 should fix it, also added a test

PR #1830, opened 2022-08-02
Add 'disable_lockfile' configuration to disable locking globally

danpf:
The point of this PR is to add a flag to `micromamba run` that will disable the lockfile creation before running the command.

## Reasoning

Dealing with unique filesystem problems is unfortunately all too common in my life and I have recently encountered issues with my parallel commands not running/erroring out due to my server's inability to create many lockfiles at a time.  It appears that I am not the only one that has to deal with this #1446.

Looking at the code it seems to be a relatively simple change and is easily tested, but let me know if there's something I should change.  I suppose an alternative to this would be #1448 but I think this still has its utility in cases when you know a lockfile is not necessary (like on HPCs.)

Also -- a genuine question -- I understand the reasoning behind using a lockfile when doing install/remove/etc, but I don't understand why a lock is necessary when using run (I assume that when using run the most common use-case is to run a binary in a separate environment) -- Was the command added to help with `micromamba ps` or were there other scenarios that were envisioned when locking was added to `run`?

Cheers,
~Danny


wolfv:
Hi @danpf thanks for the PR! 

The lockfile is needed to store meta information about the running process in a folder on the computer, yeah. 

How about we make a "disable_filelock: true/false" option in `.mambarc`? So that globally this can be enabled or disabled?

aghr:
Dear @danpf and @wolfv,

Such a global flag/option "disable_filelock: true/false" in `.mambarc` looks like a very good solution. By default it could be set to true. Mamba users could then switch it off on file systems that don't support the file-locking mechanism. Our HPC sys admin deems such a solution preferred over other/more complex solutions.

wolfv:
@danpf this looks almost perfect. 
Instead of special casing in the `run` command, can we change the `LockFile` to check the value in `Context::instance().disable_lockfile`? That way, it would take effect everywhere.

danpf:
@wolfv oh hey ya sorry you're too quick for me I was going to ask if that was the best way to handle this or if we should be special casing in places where locks are created.  Juggling a few things but I'll do that soon.

danpf:
Assuming tests pass this should be good.

I left the micromamba run .json "lockfiles" (ScopedProcFile ctor) in as they are just created with ofstream and I figure that is pretty unlikely to fail.  but if you think they should be removed as well I can do that too (just 1 line change) (see run.cpp)

danpf:
@wolfv Let me know if you think this PR needs anything else!

libmamba/src/api/configuration.cpp
```diff
@@ -1372,6 +1372,16 @@ namespace mamba
                         LockFile timeout for blocking mode when waiting for another process
                         to release the path. Default is 0 (no timeout))")));
 
+        insert(Configurable("disable_lockfile", &ctx.disable_lockfile)
+                   .group("Extract, Link & Install")
+                   .set_rc_configurable()
+                   .set_env_var_names()
+                   .description("Dont use a lockfile")
```
jonashaag, line 1379:
```suggestion
                   .description("Don't use a lockfile")
```

libmamba/src/api/configuration.cpp
```diff
@@ -1372,6 +1372,16 @@ namespace mamba
...
+                   .long_description(unindent(R"(
+                        Don't create a lockfile when running mamba commands. Use with caution!
+                        -- Some filesystems don't support File Locking and locks don't always
+                        make sense - like when on an HPC.  Default is False (use a lockfile)")));
```
jonashaag, line 1383: Can you add an example of what problems may arise when not using locking? What does Mamba use locks for that will now be the responsibility of the user?

jonashaag: Also, what commands are affected by this? If it's only a small selection of commands then it might be worthwhile explicitly listing them here.

libmamba/src/core/util.cpp
```diff
@@ -1133,7 +1137,7 @@ namespace mamba
         try
         {
             // Don't even log if the file/directory isn't writable by someone or doesnt exists.
-            if (path::is_writable(path))
+            if (Context::instance().disable_lockfile || path::is_writable(path))
```
jonashaag, line 1140: Either I lack context or this is wrong, shouldn't it be `!disable_lockfile && is_writable`? If so, this lacks a test.

libmamba/src/core/util.cpp
```diff
@@ -1133,7 +1137,7 @@ namespace mamba
         try
         {
             // Don't even log if the file/directory isn't writable by someone or doesnt exists.
```
jonashaag, line 1139:
```suggestion
            // Don't even lock if the file/directory isn't writable by someone or doesnt exists.
```

?

libmamba/include/mamba/core/context.hpp
```diff
@@ -197,6 +197,7 @@ namespace mamba
         bool no_env = false;
 
         std::size_t lock_timeout = 0;
+        bool disable_lockfile = false;
```
jonashaag, line 200: IMO we should negative this boolean (and option) like here https://github.com/mamba-org/mamba/blob/e0527eb061b0ce724dcb3c2f34db25c907ee7113/libmamba/include/mamba/core/transaction_context.hpp#L51 https://github.com/mamba-org/mamba/blob/e0527eb061b0ce724dcb3c2f34db25c907ee7113/micromamba/src/common_options.cpp#L341-L343

jonashaag: @wolfv WDYT?

libmamba/src/api/configuration.cpp
jonashaag: Depending on how much you want to write on this topic you might also want to link to the new troubleshooting page once it's merged #1873

micromamba/src/run.cpp
```diff
@@ -161,6 +161,8 @@ namespace mamba
                        std::unique_ptr<LockFile> proc_dir_lock = lock_proc_dir())
             : location{ proc_dir() / fmt::format("{}.json", getpid()) }
         {
+            if (Context::instance().disable_lockfile)
```
jonashaag, line 164: I think the code modified here does not do what you think. IIUC this creates a "pid file" that is used to give a list of running processes (see `~/.mamba/proc/` and `micromamba ps`)

libmamba/src/core/util.cpp
danpf: To revert this change you would need to modify this function:
https://github.com/danpf/mamba/blob/danpf-nolock/micromamba/src/run.cpp#L105 to not fail if nullptr is returned.

I also think it should not matter if `path::is_writable` or not if we are not making a lockfile anyway so I think this is correct.

micromamba/src/run.cpp
danpf: I agree and I originally did not have this line, however on my older hpc I was still getting the error:
```
mamba run' failed to open/create file...
```

I cannot explain this, but when I disable the json pid fifles from being created - everything works great.

libmamba/include/mamba/core/context.hpp
danpf: If that is what you want I will change it.  It's your choice.

libmamba/src/api/configuration.cpp
danpf: :+1:  I will update this once we decide on the name to use

libmamba/src/core/util.cpp
wolfv: I think the logic is now confusing (although I think I get it). Maybe we can return earlier when the lockfile is disable? Or do we even need that check here?

micromamba/src/run.cpp
wolfv: @danpf is the problem here with the lockfile or generally with writing to the home directory?

We do in fact lock the PID directory while the process is writing out the json file, so it's possible that it's also a lockfile issue.

Would it make more sense to put these checks in the `lock_proc_dir()` function?

libmamba/src/api/configuration.cpp
wolfv: We use lockfiles mainly to lock the package & repodata cache (bad things happen if two processes write to the same file...

We also use it in `micromamba run` to write some json files to store data about the process in `~/.mamba/proc/...`

wolfv:
Hi @danpf sorry for taking so long to look at this PR. I'd be super happy to merge this. We also have some additional fixes coming for lockfiles (#1193) (there were some logic issues with the assumption that PIDs are globally unique). 

With those two PRs in, the next mamba release should be much better in terms of file-locks! :)

libmamba/src/api/configuration.cpp
vicecea: > We also use it in `micromamba run` to write some json files to store data about the process in `~/.mamba/proc/...`

Short question, I have noticed that I can't have more than 1 `micromamba run ...` invocations running at the same time, spawning multiple ones fail with `critical libmamba Another process with name '...' is currently running.` - Similarly if a `micromamba run ...` process get's SIGKILLed, `micromamba run ...` fails with the same error until the zombie `~/.mamba/proc/*.json` get's cleaned up.

Will this PR fix these issues or should I open an issue?

wolfv: @vicecea I think this might be confusing some issues. Running two `run` commands with the same `label` is disallowed. I don't think this PR will fix that. If you can open a new issue and describe exactly what is going on, that would be nice!

danpf:
Finally got a chance to check this out and I see that it's already done! Just wanted to say thank you to everyone involved for finishing this PR!

PR #1833, opened 2022-08-03
Add missing SolverRuleinfo enum members to libmambapy

AntoinePrv:
Some enum values were not bound to Python (I encounter an unknown of them using `libmambapy`).

I would also argue for removing the `SOLVER_RULE_` prefix, which is redundant with Python (scoped) enums.
That breaks the API but I don't think anybody use it at this point.


libmambapy/src/main.cpp
```diff
@@ -609,6 +609,8 @@ PYBIND11_MODULE(bindings, m)
                SolverRuleinfo::SOLVER_RULE_PKG_IMPLICIT_OBSOLETES)
         .value("SOLVER_RULE_PKG_INSTALLED_OBSOLETES",
                SolverRuleinfo::SOLVER_RULE_PKG_INSTALLED_OBSOLETES)
+        .value("SOLVER_RULE_PKG_RECOMMENDS", SolverRuleinfo::SOLVER_RULE_PKG_RECOMMENDS)
+        .value("SOLVER_RULE_PKG_CONSTRAINS", SolverRuleinfo::SOLVER_RULE_PKG_CONSTRAINS)
```
wolfv, line 613: Windows claims that this exists twice now.

PR #1838, opened 2022-08-05
Fixed create/install requiring pip install failing when the prefix have whitespaces on Windows

### LINKED ISSUE #1805: [micromamba] micromamba create fails to install pip dependencies if prefix contains whitespace (Windows)

Klaim:
Also adds tests checking this issue in the first commit. They fail for me on Windows 11 without the fix provided in the second commit.

I believe this should fix #1805  which was re-opened after #1828 re-introduced the issue.

### DIFF
--- a/libmamba/src/api/install.cpp
+++ b/libmamba/src/api/install.cpp
@@ -28,15 +28,24 @@ namespace mamba
 {
     namespace
     {
-        tl::expected<std::string, std::runtime_error> get_other_pkg_mgr_install_instructions(
+        using command_args = std::vector<std::string>;
+
+        tl::expected<command_args, std::runtime_error> get_other_pkg_mgr_install_instructions(
             const std::string& name, const std::string& target_prefix)
         {
             const auto get_python_path
                 = [&] { return env::which("python", get_path_dirs(target_prefix)).u8string(); };
 
-            const std::unordered_map<std::string, std::string> other_pkg_mgr_install_instructions{
+            const std::unordered_map<std::string, command_args>
+                other_pkg_mgr_install_instructions{
                 { "pip",
-                  fmt::format("{} {}", get_python_path(), "-m pip install -r {0} --no-input") }
+                  { get_python_path(),
+                    "-m",
+                    "pip",
+                    "install",
+                    "-r",
+                    "{0}",
+                    "--no-input" } }
             };
 
             auto found_it = other_pkg_mgr_install_instructions.find(name);
@@ -90,7 +99,7 @@ namespace mamba
 
         const auto& ctx = Context::instance();
 
-        std::string install_instructions = [&]
+        command_args install_instructions = [&]
         {
             const auto maybe_instructions
                 = get_other_pkg_mgr_install_instructions(pkg_mgr, ctx.target_prefix);
@@ -106,10 +115,11 @@ namespace mamba
             specs_f << d.c_str() << '\n';
         specs_f.close();
 
-        replace_all(install_instructions, "{0}", specs.path());
+        for (auto& arg : install_instructions)
+            replace_all(arg, "{0}", specs.path());
 
-        std::vector<std::string> install_args = split(install_instructions, " ");
-        auto [wrapped_command, tmpfile] = prepare_wrapped_call(ctx.target_prefix, install_args);
+        auto [wrapped_command, tmpfile]
+            = prepare_wrapped_call(ctx.target_prefix, install_instructions);
 
         reproc::options options;
         options.redirect.parent = true;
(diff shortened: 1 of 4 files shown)


libmamba/src/api/install.cpp
```diff
@@ -106,10 +115,11 @@ namespace mamba
             specs_f << d.c_str() << '\n';
         specs_f.close();
 
-        replace_all(install_instructions, "{0}", specs.path());
+        for (auto& arg : install_instructions)
```
Klaim, line 118: I'm not totally sure this is the right way to do this replacement. Would it be ok to just replace the whole string it it is equal to `"{0}"`, or other simpler mechanism?
Maybe it would simpler to pass the value into `get_other_pkg_mgr_install_instructions` so that the instructions already contain the path.

Klaim:
@chaubold pinged you as this should fix #1805 that was a regression from #1828

chaubold:
Preventing the argument splitting by creating a `vector<string>` right away makes a lot of sense to me, thanks.

I have built it locally and tried the same use case that gave rise to my [comment](https://github.com/mamba-org/mamba/pull/1828#issuecomment-1206198657) in #1828 and to reopening the issue. With the changes in this PR, everything works fine 👍 Nice one!

micromamba/tests/test_create.py
```diff
@@ -641,3 +641,17 @@ def test_pyc_compilation(self, version, build, cache_tag):
...
         assert (site_packages / pyc_fn).exists()
         assert pyc_fn.name in six_meta
+
+    test_env_requires_pip_install_path = os.path.join(
+        Path(__file__).parent.resolve(), "env-requires-pip-install.yaml"
+    )
+
+    def test_requires_pip_install(self):
+        prefix = Path(TestCreate.prefix)
+        cmd = ["-p", f"{prefix}", "-f", self.test_env_requires_pip_install_path]
```
jonashaag, line 651:
```suggestion
        cmd = ["-p", prefix, "-f", self.test_env_requires_pip_install_path]
```

micromamba/tests/test_create.py
```diff
@@ -641,3 +641,17 @@ def test_pyc_compilation(self, version, build, cache_tag):
...
+        create(*cmd)
+
+    def test_requires_pip_install_prefix_spaces(self):
+        prefix = Path(f"{TestCreate.prefix} with space")
+        cmd = ["-p", f"{prefix}", "-f", self.test_env_requires_pip_install_path]
```
jonashaag, line 656:
```suggestion
        cmd = ["-p", prefix, "-f", self.test_env_requires_pip_install_path]
```

jonashaag (approved):

jonashaag:
Does this also apply if we have a environment name with a space in `environment.yml`? (Is that even allowed?) If yes, shall we add a test as well?

Klaim:
> Does this also apply if we have a environment name with a space in `environment.yml`? (Is that even allowed?) If yes, shall we add a test as well?

I didn't consider that case, I'll add a test.

Klaim:
@jonashaag For some unclear reason your changes don't pass for me (windows 11) because the type of `prefix` is `WindowsPath('...')` and can't be passed as is to that function so `f"{prefix}"` is used as a conversion to string. I'm not sure why it passes on CI, but maybe it's a python version mismatch.

Klaim:
The initial fix should cover any prefix value provided (as long as it have been correctly parsed before reaching that part of the code). I added the test.

wolfv:
Thanks @Klaim !

jonashaag:
> @jonashaag For some unclear reason your changes don't pass for me (windows 11) because the type of `prefix` is `WindowsPath('...')` and can't be passed as is to that function so `f"{prefix}"` is used as a conversion to string. I'm not sure why it passes on CI, but maybe it's a python version mismatch.

The proper way would be to use str.

chaubold:
Are there any plans to build a 0.25.2 release with this fix?

PR #1839, opened 2022-08-05
Improve micromamba installation documentation

pavelzw:


wolfv:
Thanks. There is also another (currently undocumented) way of installing micromamba.

`curl micro.mamba.pm/install.sh | bash` works on macOS and Linux :)

pavelzw:
Since `zsh` is the default shell on macOS I would suggest writing two separate installation commands, one for linux with `bash` and one for macOS with `zsh`.

wolfv:
that should also work!

jonashaag:
We already have instructions in the README and in the Gist linked there. Let’s deduplicate them somehow. 

Btw @yyyasin19 is working a Homebrew recipe for Micromamaba.

pavelzw:
Also, the download instructions for macOS in the Readme are wrong. The linux executable doesn't work on macOS... I now tried to deduplicate as much as possible and just redirect to the official documentation in the readme...

jonashaag (approved):

pavelzw:
If we want to deduplicate the installation guides, @wolfv should probably reference the official documentation in his [Gist](https://gist.github.com/wolfv/fe1ea521979973ab1d016d95a589dcde) (and maybe remove the installation instructions there completely?).

PR #1852, opened 2022-08-10
Fix Linux version regex

### LINKED ISSUE #1851: Incorrect Linux version regex `[linux version not found (virtual package skipped)]`
The regex on line 301 in `libmamba/src/core/util_os.cpp` is incorrect for detecting Linux versions.

The current regex is `([0-9]+\.[0-9]+\.[0-9]+)-.*` and expects the Linux kernel version to end with `-kernel-flavour`. Ex. `5.15.59-27-generic`. Not all kernel versions end with a flavour tag however. On NixOS the kernel versions are simply `5.15.59`, meaning the current regex fails matching. This leads to the `linux version not found (virtual package skipped)` warning on certain distros.

A fix would be using a regex such as `([0-9]+\.[0-9]+\.[0-9]+)(?:-.*)?` or `^([0-9]+\.[0-9]+\.[0-9]+)(?:-.*)?$`.

**Outputs**
```
$ uname -a
Linux pc 5.15.59 #1-NixOS SMP Wed Aug 3 10:03:56 UTC 2022 x86_64 GNU/Linux

$ uname -r
5.15.59

$ micromamba --version
0.24.0
```

kelszo:
Closes #1851 

Current PR fixes an incorrect regex for detecting Linux kernel versions leading to not being able to create a Linux virtual package on certain distros.

### DIFF
--- a/libmamba/src/core/util_os.cpp
+++ b/libmamba/src/core/util_os.cpp
@@ -298,7 +298,7 @@ namespace mamba
             return "";
         }
 
-        std::regex re("([0-9]+\\.[0-9]+\\.[0-9]+)-.*");
+        std::regex re("([0-9]+\\.[0-9]+\\.[0-9]+)(?:-.*)?");
         std::smatch m;
 
         if (std::regex_search(out, m, re))
@@ -310,6 +310,8 @@ namespace mamba
                 return linux_version.str();
             }
         }
+
+        LOG_DEBUG << "Could not parse linux version";
 
         return "";
     }


jonashaag:
Thanks for the patch. Can you add a test?

kelszo:
Shouldn't the patch be covered by the `test_virtual_packages.cpp` tests already? Or do you want more specific tests?

jonashaag:
> Shouldn't the patch be covered by the `test_virtual_packages.cpp` tests already? Or do you want more specific tests?

If we had a test for this change then CI would’ve been red before, but it was green, or do I miss something?

kelszo:
The reason why the tests didn't fail before was because in the CI they seem to run on Ubuntu. Ubuntu's kernel versions include the flavour tag. See #1851 for more information. 

The tests failed first locally on my machine (NixOS) so I guess running the tests on such a distro would cover this patch.

jonashaag:
Thanks, that makes sense. I guess it’s not worth adding another distro to the CI setup but we could add a unit test for the code that matches the reges and supply a hardcoded name in the test.

jonashaag:
It would require that we refactor the code a little bit though because as it is it’s not testable unless we override the `uname` program

PR #1854, opened 2022-08-10
Initial powerloader integration, using std filesystem

Klaim:
This is a follow-up on the first efforts to use powerloader in libmamba (https://github.com/mamba-org/mamba/pull/1503) but using C++ standard filesystem (rebased on https://github.com/mamba-org/mamba/pull/1665 - required to be merged before merging this) and rebased on current master.

From the initial PR, @wolfv said: 
> What's missing:
> 
* [ ]  mirror configuration
* [ ]  auth config?
* [x]  subdirdata also via powerloader
* [x]  replace mamba::url with powerloader::url
* [x]  Decide if we want to use std::function or raw function pointers for callbacks
* [ ]  Add mod / etag functions to powerloader to retrieve cached information
* [ ]  Figure out if we should enhance the callback on powerloader side with throttling and average speed computation
* [ ]  When subdirdata tries to read an empty file it crashes with a weird json error.


Klaim:
I had to first squash all the commits into one to be able to properly process a rebase over main, which is why there is now only one commit with all the changes. We didn't lose much as we were mostly experimenting and most of the commits were not atomic anyway. I also did a backup of these commits [on this branch of my fork](https://github.com/Klaim/mamba/tree/klaim/backup/initial_powerloader_integratoin_with_std_filesystem) in case we want them back or want to consult them.
I might have made mistakes while rebasing.
Hopefully we are back to the state where only windows tests were not passing. Whatever the results I'll try to progress on the issues. Right now I cannot build the C++ tests of libmamba because of a local issue that I already had on `main` so it is unrelated.

wolfv:
argh, sorry, I messed something up with git 🤦

Klaim:
I'm tired, I didnt mean "powershell" but "powerloader" 🤣

wolfv:
Note, while the last commit fixes the `test_login.py` tests for me locally, it still does too many retries because we create an on-the-fly mirror that does _not_ have the correct configuration. We need to improve that part.

Klaim:
That last commit works for me with https://github.com/mamba-org/powerloader/pull/152
I managed to biuld and test libmamba with that PR and it passes locally on my linux.
There is still 1 issue to fix and the static build and I think we will be close to make it work here.

Klaim:
rebased on recent main but issues not fixed yet

libmamba/src/api/channel_loader.cpp
```diff
@@ -86,6 +86,63 @@ namespace mamba
         }
     }
 
+    std::string get_system_certificate()
+    {
+        std::string res = "";
+        if (on_linux)
+        {
+            std::array<std::string, 6> cert_locations{
```
wolfv, line 94: should we make this an array of fs::path?

libmamba/src/api/channel_loader.cpp
```diff
@@ -86,6 +86,63 @@ namespace mamba
         }
     }
 
+    std::string get_system_certificate()
```
wolfv, line 89: shouldn't this return `std::optional<fs::path>`?

libmamba/src/api/channel_loader.cpp
```diff
@@ -86,6 +86,63 @@ namespace mamba
...
+        powerloader::Context& ctx = Context::instance().plcontext;
+        
+        if (value == "<false>")
+        {
+            ctx.disable_ssl = true;
+            ctx.ssl_ca_info = "";
+        }
+        else if (value == "<system>")
+        {
+            ctx.disable_ssl = false;
```
wolfv, line 132: this logic is different now, right? Before it was disableing SSL if no system certificates were found.

JohanMabille: I dont( think so, for me it was throwing if no system certificate was found on linux, and was relying on the OS to find the certificate otherwise.

libmamba/src/api/channel_loader.cpp
JohanMabille: Indeed, that would be an improvement

libmamba/src/api/channel_loader.cpp
JohanMabille: That would require some change in powerloader; otherwise, we would have to forward a missing value as an empty path. Also I'm not usre of what happens if we don't explicitly set `CURLOPT_CAINFO` to null when the certificate string is empty.

Klaim:
The powerloader logs fix will require powerloader to be updated, so this will fail until we make https://github.com/mamba-org/powerloader/pull/169 available

jpedrick:
@wolfv @Klaim 

I would love this feature to go mainstream

jpedrick:
@JohanMabille I see. That’s unfortunate, as I was hoping for S3 support in mamba sooner than later.

Is there anywhere I can see the discussion on why powerloader integration was held back? I’m also interested in the timeline and design motivations are for the mamba refactoring.

PR #1855, opened 2022-08-11
Fix the ci not reporting build failures + fix for that failure

Klaim:


Klaim:
Ok so the heart of the issue is that we have sequences of command lines in one `run` command in the CI scripts, and these sequences act like shell scripts. So if one command fails, it does not automatically fails the whole script.
Will try to fix this using `set -e` see if it can handle that but I doubt it.

The best way would be to rewrite the sequence of commands so that they each have a `run`, maybe. I'll try some variations.

Klaim:
It might be that just the stats command was hiding the issue.

Klaim:
Nope, any program at the end of the multi-line run that will set the exit error code can fail the script, not the preceding programs.

Klaim:
Now the CI reveals the issue that was a false-positive before.
Attempting a fix now.

jonashaag:
@Klaim are you sure? The commands are run via `bash some_temporary_file.sh` by GitHub and by default GitHub has `-e` enabled. Did we accidentally disable that (`bash -l` maybe, missing the `-e`)?

wolfv:
I did also see the failure that Klaim saw in our CI, so I definitely believe that it's true. Although indeed, I wouldn't have expected that from bash on Windows, either! I thought we were running in cmd.exe

Klaim:
Yes the issue was mainly visible with `cmd` based scripts, I didn't check if there was an issue in other cases as the windows cases where the ones I was seeing failing. In doubt I generalized the change so that we don't see this issue again whatever the shell.

jonashaag:
Makes sense! cmd.exe is horrible

PR #1864, opened 2022-08-18
Add explicit channels to update

### LINKED ISSUE #1846: micromamba update fails when using channels
When specifying a package with a channel, or when using a yaml file, `micromamba update` fails. This kinda looks like https://github.com/mamba-org/mamba/issues/504 

**The issue** : 
```
(server) root@2e8550c5affe:/app# micromamba update conda-forge::gunicorn

                                           __
      /_/

error    libmamba Selected channel specific (or force-reinstall) job, but package is not available from channel. Solve job will fail.

Pinned packages:
  - python 3.8.*

Transaction

  Prefix: /opt/conda/envs/server

  All requested packages already installed
```

**Expected result**. 

- Existing packages updated
- Newly added pacakges installed

AntoinePrv:
Closes #1846

This should to be enough to solve the issue.

@wolfv 
- This piece of code is duplicated from `install_specs` so there should be a better place to put it.  Perhaps we could do it in `add_channel_specifc_job`, but we do not have access to the context there; or perhaps we change the order of the instruction in `mamba::update` (not sure what an `MTransaction` is).
  It seems that
  - running `micromamba install` we go through it order `install_specs` then `add_job>add_channel_specifc_job`;
  - running `micromamba update`  we go through it order `add_job>add_channel_specifc_job` then `MTransaction>install_specs`;
- Where do you think would be a good place to add a test for this? It is a `libmamba` change, but high-level tests seems to rather be done in `mamba` or `micromamba`

### DIFF
--- a/libmamba/src/api/install.cpp
+++ b/libmamba/src/api/install.cpp
@@ -403,10 +403,9 @@ namespace mamba
         MultiPackageCache package_caches(ctx.pkgs_dirs);
 
         // add channels from specs
-        std::vector<mamba::MatchSpec> match_specs(specs.begin(), specs.end());
-        for (const auto& m : match_specs)
+        for (const auto& s : specs)
         {
-            if (!m.channel.empty())
+            if (auto m = MatchSpec{ s }; !m.channel.empty())
             {
                 ctx.channels.push_back(m.channel);
             }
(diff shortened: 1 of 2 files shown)


AntoinePrv:
```
 micromamba\tests\test_menuinst.py:15: in <module>
    import win32com.client
E   ModuleNotFoundError: No module named 'win32com'
```

JohanMabille:
Let's merge it as is, and refactor later to avoid code duplication.

PR #1867, opened 2022-08-20
Add `brew` installation option to micromamba

YYYasin19:
Hey, 

Recently, I have added a formula to `homebrew`, the beloved package manager, and wanted to include this in the docs.
I have also changed the reference from `OS X` to `macOS` here since it's the official name now, but that should probably be also done in some other place.s

I hope it's okay for documentation additions to go the direct PR route since the changes are often so straightforward.

Thanks for the review!

### DIFF
--- a/docs/source/installation.rst
+++ b/docs/source/installation.rst
@@ -146,7 +146,7 @@ Manual
 
 Otherwise, you can use the download script; here, instructions are mostly the same as with :ref:`linux<umamba-install-linux>`.
 
-However, you need to download a different `micromamba`` binary depending if you are using an Apple Silicon or an Intel Mac.need to download a different `micromamba` binary depending if you are using an Apple Silicon or an Intel Mac.
+However, you need to download a different `micromamba`` binary depending if you are using an Apple Silicon or an Intel Mac.
 
 Apple Silicon:
 
--- a/docs/source/installation.rst
+++ b/docs/source/installation.rst
@@ -128,12 +128,25 @@ Now you can activate the base environment and install new packages, or create ot
 
 .. _umamba-install-osx:
 
-OS X
-****
+macOS
+*****
+``micromamba`` has macOS support as well.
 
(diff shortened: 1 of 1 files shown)


jonashaag:
Thanks for this! Scrambled sentence in the last change.

Somewhat related to this: How do we make sure the formula is up to date? Does Home-brew have some kind of automatic update check or are we responsible to manually push new updates? In the latter case how about adding another cronjob to the CI, similar to the micro.mamba.pm one, that checks for a working and up-to-date Homebrew formula?

YYYasin19:
> Thanks for this! Scrambled sentence in the last change.
👍 

> How do we make sure the formula is up to date?

Good question and something I have thought about already a little bit:
First of all, the formula contains a `livecheck` stanza, i.e. is able to automatically check if there is a new version. Currently, this is based on the anaconda distributed version.

brew has a dedicated `bump-cask-pr` command which creates a PR to the hombrew-cask repo if there is a newer version of a package available; so we'd need to call this command from time to time to create a PR that is then merged by a maintainer.
The call is fairly simple (`brew bump-cask-pr micromamba --version={new_version}`) and there is even a [GitHub Action](https://github.com/macauley/action-homebrew-bump-cask) for it.

I'll try to include test / adjust the action and make a PR :)

jonashaag:
Good to know! Is there a bot like Dependabot that runs these live checks from time to time?

YYYasin19:
I couldn't find anything regarding something like Dependabot.

The proposed action above contains an option to only launch the command if there is a new version (retrieved by livecheck) which would fit our use case, since we'd launch it on each new release, for exampe.
Issue: The livecheck compares against the newest anaconda-version and there might be a delay.

jonashaag:
A small delay is fine. You might want to contribute a similar check to the “Test micro.mamba.pm” (see this repo’s CI) that does a check if we have a new version.

PR #1873, opened 2022-08-23
Create Troubleshooting page

jonashaag:
- Move "pitfalls" to "troubleshooting"
- Add docs on https://github.com/mamba-org/mamba/issues/1611
- I want to add more to that page soon (common issues we see in the bug tracker)

### DIFF
--- a/docs/source/user_guide/troubleshooting.rst
+++ b/docs/source/user_guide/troubleshooting.rst
@@ -1,10 +1,10 @@
-.. _pitfalls:
+.. _troubleshooting:
 
-Common pitfalls
----------------
+Troubleshooting
+===============
 
-Using long paths on Windows
-===========================
+Windows long paths
+------------------
 
 Windows API historically supports paths up to 260 characters. While it's now possible to used longer ones, there are still limitations related to that.
 
@@ -44,3 +44,13 @@ Thus, the following cases will fail:
 - partially
 
   - pre-compilation of ``noarch`` packages, with no impact on capability to use the package but removing it will let artifacts (pycache) on the filesystem
+
+
+Hangs during install in QEMU
+----------------------------
+When using Mamba/Micromamba inside a QEMU guest, installing packages may sometimes hang forever due to an `issue with QEMU and glibc <https://gitlab.com/qemu-project/qemu/-/issues/285>`_. As a workaround, set ``G_SLICE=always-malloc`` in the QEMU guest, eg.::
+
+  export G_SLICE=always-malloc
+  mamba install ...
+
+See `#1611 <https://github.com/mamba-org/mamba/issues/1611>`_ for discussion.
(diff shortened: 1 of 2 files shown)


docs/source/user_guide/troubleshooting.rst
```diff
@@ -44,3 +44,13 @@ Thus, the following cases will fail:
...
+
+
+Hangs during install in QEMU
+----------------------------
+When using Mamba/Micromamba inside a QEMU guest, installing packages may sometimes hang forever due to an `issue with QEMU and glibc <https://gitlab.com/qemu-project/qemu/-/issues/285>`_. As a workaround, set ``G_SLICE=always-malloc`` in the QEMU guest, eg.::
+
+  export G_SLICE=always-malloc
+  mamba install ...
+
+See `#1611 <https://github.com/mamba-org/mamba/issues/1611>`_ for discussion.
```
maresb, line 56: This is a horribly long thread. I scanned it to look for a summarizing comment we could link instead, but couldn't find anything. (Closest seems to me like https://github.com/mamba-org/mamba/issues/1611#issuecomment-1112970540, but not particularly conclusive.)

docs/source/user_guide/troubleshooting.rst
```diff
@@ -44,3 +44,13 @@ Thus, the following cases will fail:
 - partially
 
   - pre-compilation of ``noarch`` packages, with no impact on capability to use the package but removing it will let artifacts (pycache) on the filesystem
+
+
+Hangs during install in QEMU
+----------------------------
+When using Mamba/Micromamba inside a QEMU guest, installing packages may sometimes hang forever due to an `issue with QEMU and glibc <https://gitlab.com/qemu-project/qemu/-/issues/285>`_. As a workaround, set ``G_SLICE=always-malloc`` in the QEMU guest, eg.::
+
+  export G_SLICE=always-malloc
```
maresb, line 53: Is there still any point to suggesting `micromamba config set extract_threads 1`? (If so, we should offer the corresponding `mamba` command...) Or is that now pointless after your patch?

maresb (approved):
This is really excellent, thanks for writing it up!

docs/source/user_guide/troubleshooting.rst
jonashaag: I think the `G_SLICE` solution is much better because it doesn't slow down the process noticably.

docs/source/user_guide/troubleshooting.rst
jonashaag: Yeah... I will highlight this comment top of the PR.

docs/source/user_guide/troubleshooting.rst
corneliusroemer: G_SLICE doesn't seem to work for me. However doing `RUN micromamba config set extract_threads 1` before `RUN micromamba create -vv -y -n pangolin -f /environment.yml` _does_ work luckily!

Issue #1878, opened 2022-08-24
Improve mamba unsatisfiability error messages


AntoinePrv:
## Abstract
Using `all_problem_structured`, we can improve mamba errors when an environment is not satisfiable.
The function gives a number of problem and dependencies between packages involved in the error.
We can build a Directed Acyclic Graph (DAG) of these packages, where top level (root) packages are directly requested by the user, and bottom packages (leaves) are posing problem (conflict, missing dependency...).

Together with @syslaila we successfully build two proof of concepts (POC) for such errors:
- **POC-A**: One pinpoints the root and leaf packages to produce a succinct error message #1847
- **POC-B**: The other retrace the whole path between these root and  leaves to show a detailed explanation of the conflicts/problems in a tree-like message [AntoinePrv/mamba-error-reporting](https://github.com/AntoinePrv/mamba-error-reporting).

The goal is to incorporate both of them, using a/ as a summary and b/ as the verbose details.

The work done is quite substantial, and currently the two POCs do not share the same foundation / data structures.
The aim of this issue is to find similarities between the two POC and break down the work into small, well-defined pull-requests (PRs).

## Tasks
Please add unit tests where needed.
Python bindings are encouraged when appropriate.

### Problem data
- [ ] **T11a**: `PackageInfo` represents (almost) all the available information we have on a package, but to use it in data structure we need to make it comparable and hashable.
  - [x] **T11a.1**: Optionally, Implement a generic `hash` for `std::tuple` (e.g.* from [Python](https://github.com/python/cpython/blob/6bda5b85b53443f3467865fbf85cbe72932e7cd6/Objects/tupleobject.c#L321), [Boost](https://github.com/boostorg/container_hash/blob/e00f53a69c9e6dba0af457dc22e41aedf86fd31a/include/boost/container_hash/hash.hpp#L572)) and a `PackageInfo::to_tuple` (a generic implementation will be useful for *T31* and *T32*).
  - [x] **T11a.2**: Implement `std::hash<PackageInfo>`
  - [x] **T11a.3** Implement `PackageInfo::operator==`
- [ ] **T11b**: Alternatively, `PackageInfo` does loose the `solvable_id` used by libsolv. If we need to store it, we could.
  - [ ] **T11b.1** Build a `std::map<SolvableID, PackageInfo>` and use the ids everywhere else in the algorithm (no need for 1a. anymore).
[...]

PR #1879, opened 2022-08-24
[MER][T11a] PackageInfo Hash and Comparison

AntoinePrv:
Partial implementation for #1878

### DIFF
--- a/libmamba/src/core/package_info.cpp
+++ b/libmamba/src/core/package_info.cpp
@@ -273,6 +273,11 @@ namespace mamba
     {
     }
 
+    bool PackageInfo::operator==(PackageInfo const& other) const
+    {
+        return sha256 == other.sha256;
+    }
+
     nlohmann::json PackageInfo::json_record() const
     {
         nlohmann::json j;
@@ -370,3 +375,12 @@ namespace mamba
         return concat(channel, "::", name, "-", version, "-", build_string);
     }
 }  // namespace mamba
+
+namespace std
+{
+    std::size_t hash<mamba::PackageInfo>::operator()(mamba::PackageInfo const& p) const
+    {
+        using Hasher = std::hash<std::string>;
+        return Hasher{}(p.sha256);
+    }
+}
--- a/libmamba/include/mamba/core/package_info.hpp
+++ b/libmamba/include/mamba/core/package_info.hpp
@@ -41,6 +41,8 @@ namespace mamba
                     const std::string build_string,
                     std::size_t build_number);
 
+        bool operator==(PackageInfo const& other) const;
+
         nlohmann::json json_record() const;
         nlohmann::json json_signable() const;
         std::string str() const;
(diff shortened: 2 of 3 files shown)


libmamba/src/core/package_info.cpp
```diff
@@ -273,6 +273,11 @@ namespace mamba
     {
     }
 
+    bool PackageInfo::operator==(PackageInfo const& other) const
+    {
+        return sha256 == other.sha256;
```
syslaila, line 278: I think it might be a  low probability but there could be a case where multiple packageInfos have the same sha256. In my PR  I checked the equality of each of the fields - https://github.com/mamba-org/mamba/pull/1847/files#diff-c93060e224d433cee7ee3ad6f251e0c1577515984b880e21665e16ac5b45f3c2R72
Also, would it be possible that sha256 is empty for some of the packages and only md5 is set?

AntoinePrv: >  think it might be a low probability but there could be a case where multiple packageInfos have the same sha256.

It's considered small enough to be safe, otherwise that would be a big security problem!

>  Also, would it be possible that sha256 is empty for some of the packages and only md5 is set?

We could use both.

> In my PR I checked the equality of each of the fields

I've also written code that does the `hash` for `std::tuple` so that we can use `std::tie`, but then i saw there was a `std::set` and many complicated things so I went with this.

AntoinePrv:
@syslaila do we acually need `std::hash<PackageInfo>` and `PackageInfo::operator==`? After submitting this, I realize we don't need it to put them in `DiGraph`.
Even if we need to put them in a set or a (as a key in a) map `std::set` and `std::map` would not need the hash (only the comparison).

Can you see anywhere where we may need any of it?

syslaila:
I used a mapping` std::unordered_map<MNode, node_id> visited ` when I was constructing the graph from the list of of problems. So that if that MNode(which contains the PackageInfo or a Dependency) exists(was created by a previous problem) already (say we have problems1 where A requires D & problem2 where B requires D) then D will be a single node. If we put them in a set - yes we don't need it but in this case we don't really care about the order. 

Alternatively, we can store the ids (from sourceId, targetId, depId) if we know that a solvable will only be referenced by one id (I didn't know if that is the case with the problems we get from libsolv).

AntoinePrv:
I see, thanks for your answer. In *POC-B*, I relied on the `solvable_id` provided by `all_problem_structured` and `select_solvable` for that.

One difference, here is that we have a new `node_id` to index the node in the graph (In *POC-B* I was able to use the `sovable_id` but with this graph class they need to map to the `std::vector` indices). I think that `solvable_id` should be enough to uniquely identify packages, what do you think? In that cases we'll still need to maintain a `std::map<solvable_id, node_id>` while constructing the graph; afterwards, we should be free from `libsolv` indexing.

PR #1880, opened 2022-08-25
[MER][T21|T22|T23] Extend graph class for use in error messages

AntoinePrv:
Context: #1878

This PR:
- [x] Remove some dead code form the graph class
- [x] Add a `vector_set` vector that behaves like a set (no duplicate, no order) for proper neighborhood comparison
- [x] Add predecessors information
- [x] Add optional edge data
- [x] Add `for_each_root`/`for_each_leave`.

### DIFF
--- a/libmamba/include/mamba/core/graph_util.hpp
+++ b/libmamba/include/mamba/core/graph_util.hpp
@@ -39,6 +39,13 @@ namespace mamba
         void depth_first_search(V& visitor, node_id start = node_id(0)) const;
 
     protected:
+        DiGraphBase() = default;
+        DiGraphBase(DiGraphBase const&) = default;
+        DiGraphBase(DiGraphBase&&) = default;
+        DiGraphBase& operator=(DiGraphBase const&) = default;
+        DiGraphBase& operator=(DiGraphBase&&) = default;
+        ~DiGraphBase() = default;
+
         GraphCRTP* crtp_cast()
         {
             return static_cast<GraphCRTP*>(this);
--- a/libmamba/include/mamba/core/graph_util.hpp
+++ b/libmamba/include/mamba/core/graph_util.hpp
@@ -14,7 +14,7 @@
 namespace mamba
 {
     // Simplified implementation of a directed graph
-    template <typename Node, typename GraphCRTP>
+    template <typename Node, typename Derived>
     class DiGraphBase
     {
     public:
@@ -46,30 +46,30 @@ namespace mamba
         DiGraphBase& operator=(DiGraphBase&&) = default;
         ~DiGraphBase() = default;
 
-        GraphCRTP* crtp_cast()
+        Derived* derived_cast()
         {
-            return static_cast<GraphCRTP*>(this);
+            return static_cast<Derived*>(this);
         }
-        GraphCRTP const* crtp_cast() const
+        Derived const* derived_cast() const
         {
-            return static_cast<GraphCRTP const*>(this);
+            return static_cast<Derived const*>(this);
         }
 
     private:
-        enum class color
+        enum class visited
         {
-            white,
-            gray,
-            black
+            no,
+            ongoing,
+            yes
         };
 
-        using color_list = std::vector<color>;
+        using visited_list = std::vector<visited>;
 
         template <class V>
         node_id add_node_impl(V&& value);
 
         template <class V>
-        void depth_first_search_impl(V& visitor, node_id node, color_list& colors) const;
+        void depth_first_search_impl(V& visitor, node_id node, visited_list& status) const;
 
         node_list m_node_list;
         adjacency_list m_predecessors;
@@ -80,29 +80,29 @@ namespace mamba
     class default_visitor
     {
     public:
-        using graph_type = G;
-        using node_id = typename graph_type::node_id;
+        using graph_t = G;
+        using node_id = typename graph_t::node_id;
 
-        void start_node(node_id, const G&)
+        void start_node(node_id, const graph_t&)
         {
         }
-        void finish_node(node_id, const G&)
+        void finish_node(node_id, const graph_t&)
         {
         }
 
-        void start_edge(node_id, node_id, const G&)
+        void start_edge(node_id, node_id, const graph_t&)
         {
         }
-        void tree_edge(node_id, node_id, const G&)
+        void tree_edge(node_id, node_id, const graph_t&)
         {
         }
-        void back_edge(node_id, node_id, const G&)
+        void back_edge(node_id, node_id, const graph_t&)
         {
         }
-        void forward_or_cross_edge(node_id, node_id, const G&)
+        void forward_or_cross_edge(node_id, node_id, const graph_t&)
         {
         }
-        void finish_edge(node_id, node_id, const G&)
+        void finish_edge(node_id, node_id, const graph_t&)
         {
         }
     };
@@ -187,8 +187,8 @@ namespace mamba
     {
         if (!m_node_list.empty())
         {
-            color_list colors(m_node_list.size(), color::white);
-            depth_first_search_impl(visitor, node, colors);
+            visited_list status(m_node_list.size(), visited::no);
+            depth_first_search_impl(visitor, node, status);
         }
     }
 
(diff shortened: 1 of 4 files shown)


libmamba/include/mamba/core/graph_util.hpp
```diff
@@ -7,38 +7,47 @@
...
 {
     // Simplified implementation of a directed graph
-    // where a path exists between each node and the
-    // first one (you can think of it as a tree with
-    // potential cycles)
-
-    template <class T>
-    class graph
+    template <typename Node, typename GraphCRTP>
+    class DiGraphBase
```
JohanMabille, line 18: If the definition of `DiGraph<Node, void>` remains empty, I don't think we need a CRTP here, simple inheritance should be enough. A CRTP is usually used when you want to provide different implementation of a same interface without relying on virtual method and entity semantics, and / or when you want to avoid a catch-all template generic function, and / or you don't want to lose the final type information and / or a combination of the three.

libmamba/include/mamba/core/graph_util.hpp
```diff
@@ -7,38 +7,47 @@
...
+    protected:
+        GraphCRTP* crtp_cast()
+        {
+            return static_cast<GraphCRTP*>(this);
+        }
+        GraphCRTP const* crtp_cast() const
+        {
+            return static_cast<GraphCRTP const*>(this);
+        }
+
```
JohanMabille, line 50: Whatever the design we chose, value semantic (constructor + destructor + copy and move semantic) should be implemented in this protected section so it is clear this class is not meant for being directly instnatiated.

JohanMabille (changes requested):

libmamba/include/mamba/core/graph_util.hpp
JohanMabille: Actually the CRTP is justified here, since a downcast is required in the `depth_first_search` method before invoking the visitor (I missed it during the first review). A simple inheritance would make the downcast more fragile since it would require to use the visitor internal `graph_t` type

libmamba/include/mamba/core/graph_util.hpp
```diff
@@ -7,38 +7,47 @@
...
 
+    protected:
+        GraphCRTP* crtp_cast()
+        {
+            return static_cast<GraphCRTP*>(this);
+        }
+        GraphCRTP const* crtp_cast() const
+        {
+            return static_cast<GraphCRTP const*>(this);
+        }
```
JohanMabille, line 49: It is usually more convenient to directly return refererences instead of pointers, since most of the APIs expects references. This saves you the indirection.

Nitpicking: I find the name `derived_cast` more explicit than `crtp_cast` and it would be more consistent with the usage of CRTPs we have in other repos that we maintain.

AntoinePrv:
@syslaila here's the extension to a directed graph class.
We have (I'm following Networkx naming)
  - `DiGraph<NodeType, EdgeType>`
  - `DiGraph::add_node`
  - `DiGraph::add_edge(node_id from, node_id to, EdgeType data)`
  - `DiGraph::successors` and `DiGraph::predecessors`
  - Depth first search
  
 Do we need functions to find leaves/roots (or leaves/roots of a particular node)?

syslaila:
Nice!
Yes, I am also using the following info when explaining problems: 

```node_id_list get_roots() const;``` -> getting the top level nodes

```edge_id_list  get_leaves_from(node_id)``` -> returns a list of edges that represent the leaves

also, when I also need to update the information for a node (during the merging phase) so I have these extra 2 methods:
```
template <class V>
void update_node(node_id id, V&& value);
```

```
template <class Y>
bool update_edge_if_present(node_id from, node_id to, Y&& info);
```

File: libmamba/include/mamba/core/graph_util.hpp (hunk 1 above)
Comment (AntoinePrv) [reply libmamba/include/mamba/core/graph_util.hpp:18]: @JohanMabille We could kill the CRTP if we change the visitor to be responsible for its own capture of the graph. Would be even easier with a lambda. Note there is currently only one DFS visitor implemented in mamba right now, but there will be more with the error message algo.

Thinking about this as I am about to add other visitor patterns (*e.g.* for leaves) that don't need to have so many options.
(Ideally we'd have a range but let's keep it simple.)

Comment (JohanMabille) [reply libmamba/include/mamba/core/graph_util.hpp:18]: > We could kill the CRTP if we change the visitor to be responsible for its own capture of the graph.

Not sure to get what you mean. The base graph class would not invoke the visitor anymore? The only way I see to do this would be to have public tempate functions in the final classes that forward to a common private implementation with the appropriate class, but that's potentially a lot of boilerplate.

Comment (AntoinePrv) [reply libmamba/include/mamba/core/graph_util.hpp:18]: Now we have:
```cpp
auto graph = DiGraph();
struct Visitor {
   void start_node(node_id n, graph_t const& g);
};
graph.depth_first_search(Visitor());
```
We could have (which is kind of ugly but gets rid of CRTP)
```cpp
auto graph = DiGraph();
struct Visitor {
   graph_t& g;

   void start_node(node_id n);
};
graph.depth_first_search(Visitor(g));
```
Or better with a lambda (swapping the different callback names for a `where` enum):
```cpp
auto graph = DiGraph();
graph.depth_first_search([&graph](auto where, node_id n) {
  if (what == Search::start_node){
    ...
  }
});
```
And if we really care about having separate functions in the visitor, we could have the visitor's `operator()` be templated by the `Search` enum, but that brings back the verbosity (at least until C++20 templated lambda).

JohanMabille: I find it more verbose and dangerous, you need to pas to the visitor the same instance of graph as that used for calling the method that will invoke the visitor. Besides, it is less flexible since you cannot reuse the visitor instance for another graph instance (even if I'm not sure we will need this).

AntoinePrv:
@syslaila 

> `node_id_list get_roots() const;` -> getting the top level nodes
> `edge_id_list get_leaves_from(node_id)` -> returns a list of edges that represent the leaves

I've added all four options.
I've made all of them similar, exposing the `node_ids` (the edges needs to be manually reconstructed from `predecessors`/`successors`). It's also in a `std::for_each` fashion where you pass a callable. In the test there is an example to convert it to a vector.

There is also a `vector_set`/`node_id_list` that can safely be used as a set without worrying about duplicate or insertion order.

> `void update_node(node_id id, V&& value)`

I added a non-const overload so that we can do `graph.nodes()[node_id] = value`

> `bool update_edge_if_present(node_id from, node_id to, Y&& info);`

I added a non-const overload and `DiGraph:has_node` and `DiGraph::has_edge`, then I think we can do from here in the error message algo.

Hopefully we can merge this soon so that we can start building on it. But if you want you can already checkout the branch, the API should be stable.

libmamba/include/mamba/core/graph_util.hpp
```diff
@@ -7,303 +7,493 @@
...
+        using adjacency_list = std::vector<node_id_list>;
+
+        node_id add_node(const node_t& value);
+        node_id add_node(node_t&& value);
         void add_edge(node_id from, node_id to);
 
+        bool empty() const;
+        std::size_t number_of_nodes() const;
+        const node_list& nodes() const;
+        node_list& nodes();
```
JohanMabille, line 79: This method should be removed, since an external change in the list may lead to an inconsistent state of an instance of `DiGraphBase`.

libmamba/src/core/query.cpp
```diff
@@ -575,7 +573,7 @@ namespace mamba
 
         void start_edge(node_id from, node_id to, const graph_type& g)
         {
-            m_is_last = g.get_edge_list(from).back() == to;
+            m_is_last = *(g.successors(from).rbegin()) == to;
```
JohanMabille, line 576: The `back` syntax is lighter, it might be convenient to add the method to the `vector_set` class.

JohanMabille (changes requested):

libmamba/include/mamba/core/graph_util.hpp
```diff
@@ -7,303 +7,493 @@
...
-        using base_type::start_node;
+        void add_edge(node_id from, node_id to, edge_t const& data);
+        void add_edge(node_id from, node_id to, edge_t&& data);
 
-        void tree_edge(node_id from, node_id to, const G&);
-        void finish_edge(node_id id, const G&);
-
-        const predecessor_map& get_predecessors() const;
+        const edge_map& edges() const;
+        edge_map& edges();
```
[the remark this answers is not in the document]
AntoinePrv: Switching for
```cpp
        node_t const& node(node_id id) const;
        node_t & node(node_id id);
```
Since changing single node data is allowed and needed.

libmamba/include/mamba/core/graph_util.hpp
AntoinePrv:
```cpp
        node_t const& node(node_id id) const;
        node_t & node(node_id id);
```
Since changing single node data is allowed and needed.

libmamba/include/mamba/core/graph_util.hpp
```diff
@@ -7,303 +7,493 @@
...
+
+        vector_set() = default;
+        vector_set(std::initializer_list<value_type> il);
+        template <typename InputIterator>
+        vector_set(InputIterator first, InputIterator last);
+
+        bool contains(value_type const&) const;
+
+        void insert(value_type&& value);
+        void insert(value_type const& value);
```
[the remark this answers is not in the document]
AntoinePrv: Agreed. However it should be a `const_iterator`, because an `iterator` may break class invariants and I don't believe implementing the machinery for a non const iterator is worth the effort at this point.

Which means I should probably remove the non-const overloads of `begin` / `end`.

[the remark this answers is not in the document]
JohanMabille: Indeed

JohanMabille (approved):

JohanMabille:
The CI failure is not related to this PR and has been fixed in master, let's merge!

PR #1886, opened 2022-08-30
Use separate artifact names per OS

jonashaag:
Also rename CI internal binary to reflect that it's internal.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -497,7 +497,7 @@ jobs:
         run: tar -cvf umamba.tar build/micromamba/micromamba.exe
       - uses: actions/upload-artifact@v2
         with:
-          name: micromamba_binary
+          name: _internal_micromamba_binary
           path: umamba.tar
       - name: Cleanup
         shell: bash -l {0}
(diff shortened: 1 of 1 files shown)


jonashaag:
@adriendelsalle you see to be the author of this test that's also failing on master, care to have a look?

adriendelsalle:
Hi @jonashaag it's just a test to check the behavior of the `--no-py-pin` flag.
At this time it looks like the `python_abi` was installed for `python=3.9` thus downgraded when installing an old `setuptools`. Now it looks like the `python_abi` package is not installed anymore with `python=3.9` but still required by  `python=2.7`  then just linked and not unlinked.

But I don't see anything related to this PR: ?

pavelzw:
Let's see which one gets merged first, this one or #1886 🤔. 

The loser needs to rebase :P

jonashaag:
> But I don't see anything related to this PR: ?

That’s why I said it’s unrelated :)

adriendelsalle:
oh sorry I read this a bit too fast ;). hope this help

jonashaag:
Thanks, see #1889 

@pavelzw can you incorporate this into #1888?

Issue #1882, opened 2022-08-26
Reproducible micromamba segfault when installing from lockfile after previous install


maresb:
I can reproduce in a Dockerfile built on GitHub Actions.
* [Repo](https://github.com/maresb/micromamba-segfault)
* [Failed run](https://github.com/maresb/micromamba-segfault/runs/8042434599?check_suite_focus=true)

I put some effort into making this more minimal, but didn't get very far. I hope this makes it possible for someone to easily do a trace. :)

High-level overview:
* Select some particular dependencies
* Do `micromamba install`, with or without lockfile. (I used a lockfile for the first install command to guarantee reproducibility.)
* Do `micromamba install` from a new-style lockfile produced by `conda-lock`

jonashaag:
Thanks! Let me know if you want some guidance on debugging this yourself (with gdb or lldb)

maresb:
Cool, thanks @jonashaag!!!

Can I use the usual executable, or do I have to rebuild micromamba to be able to use xdb?

maresb:
With `gdb`:
```
Thread 1 "micromamba" received signal SIGSEGV, Segmentation fault.
0x000055c66d3882d8 in pool_id2str ()
```

Is there a downloadable version of micromamba with debug symbols?

maresb:
And with `lldb`:

```
Process 1922 stopped
* thread #1, name = 'micromamba', stop reason = signal SIGSEGV: invalid address (fault address: 0x48)
    frame #0: 0x0000555555931735 micromamba`mamba::PackageInfo::PackageInfo(s_Solvable*) + 661
micromamba`mamba::PackageInfo::PackageInfo:
->  0x555555931735 <+661>: movq   0x18(%rax), %r12
    0x555555931739 <+665>: movq   $0x0, 0x1a0(%rdi)
    0x555555931744 <+676>: movq   $0x0, 0x1a8(%rdi)
    0x55555593174f <+687>: movq   $0x0, 0x1b0(%rdi)
```

@jonashaag I don't know how to proceed from here. Any advice? Thanks!

maresb:
Looks like it relates to this code? https://github.com/mamba-org/mamba/blob/2887d0058deb26624fce0512321e8af8478a826b/libmamba/src/core/package_info.cpp#L91

jonashaag:
[...]

PR #1893, opened 2022-08-31
Fixed: crash when installing with env-lockfile (#1882)

Klaim:
`install_explicit_with_transaction()` had the transaction object
being created before calling `MRepo::create()`.
`Transaction`  gathered the libsolv `Solvable` objects from pool.
Then AFTER that, `MRepo::create()` is called, but it potentially
realloc/move-in-memory the Solvable objects which pointers
were stored in the transaction.
When this happens, the transaction holds invalid pointers to Solvables.
Which lead to Undefined Behaviors.


jonashaag:
Nice! These types of bugs are hard to test. If it takes too much effort personally I think it would be fine having no test.

Klaim:
It's ok, it's just the create then install commands with the given env-lock files, I'll complete it tomorrow. I just want to make sure I'm not doing something wrong from the pov of other platforms I don't have access to ^^;

PR #1903, opened 2022-09-05
Test Isolation 2 (TBC)

AntoinePrv:
Follow up on #1791

### DIFF
--- a/micromamba/tests/test_activation.py
+++ b/micromamba/tests/test_activation.py
@@ -94,24 +94,29 @@ possible_interpreters = {
     "unix": {"bash", "zsh", "fish", "xonsh", "tcsh"},
 }
 
-shell_files = [
-    Path(x).expanduser()
-    for x in [
-        "~/.bashrc",
-        "~/.bash_profile",
-        "~/.zshrc",
-        "~/.zsh_profile",
-        "~/.xonshrc",
-        "~/.tcshrc",
-        "~/.config/fish/config.fish",
-    ]
-]
-
-if plat == "win":
-    shell_files.append(Path(paths["win"]["powershell"]))
 
 regkey = "HKEY_CURRENT_USER\\Software\\Microsoft\\Command Processor\\AutoRun"
-bkup_winreg_value = None
+
+
+@pytest.fixture
+def winreg_value():
+    if plat == "win":
+        try:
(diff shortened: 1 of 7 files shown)


AntoinePrv:
The Windows tests are fixed, I disable cleaning external `CONDA_PREFIX` from the `PATH` for that platform. Note that it was not previously done, so this is still an improvement for the the Linux tests rather that a deterioration of the Windows tests.

AntoinePrv:
Great, this is looking better! Actually the assertion that is failing was not tested before (there was not deactivation test).
```python
            stdout, stderr = call(["micromamba deactivate"] + evars)
            res = TestActivation.to_dict(stdout)
Error>      assert find_path_in_str(tmp_root_prefix / "condabin", res["PATH"])
```
On Windows, `PATH` does not contain (actually) any path with `condabin` in it after `micromamba deactivate`.
@wolfv is that expected? Or is something missing?

wolfv:
Hmm actually that is curious! I think we should keep `condabin`, even on Windows. Would you be able to rebase the test cleanup? I just changed (and aligned with conda) how activation works, so it might accidentally fix the test failure :)

micromamba/tests/test_activation.py
```diff
@@ -646,75 +576,50 @@ def test_activation(
...
+            assert find_path_in_str(tmp_root_prefix / "condabin", res["PATH"])
+            assert not find_path_in_str(tmp_root_prefix / "bin", res["PATH"])
+            assert find_path_in_str(tmp_root_prefix / "envs" / "xyz", res["PATH"])
+            assert not find_path_in_str(tmp_root_prefix / "envs" / "abc", res["PATH"])
 
             s = [
                 "micromamba activate",
                 "micromamba activate abc",
-                "micromamba activate xyz --stack",
+                "micromamba activate --stack xyz",
```
wolfv, line 587: are we not testing `--stack xyz` vs `xyz --stack` anymore?

wolfv:
Now it's a different but related failure :D

micromamba/tests/test_activation.py
AntoinePrv: We were not testing both. We were testing one, now it;'s the other.
Did not need to be changed though...

wolfv: Now we're testing both

Issue #1834, opened 2022-08-04
Error libmamba Path already locked by the same PID


bogdancondurache:
Hello,

We're currently experiencing some user-reported errors on the micromamba side. They originate from the following line
```go
mambaCommand := common.NewCommander(BinMicromamba(), "create", "--always-copy", "--no-env", "--safety-checks", "enabled", "--extra-safety-checks", "--retry-clean-cache", "--strict-channel-priority", "--repodata-ttl", ttl, "-y", "-f", condaYaml, "-p", targetFolder)
```
so it's basically happening at the create command. It seems that once an environment has this happening, it will (at least for a while) continue to have it happening, probably because of the way Windows performs PID recycling. `micromamba clean --locks` isn't solving it either. The solution was to manually delete the lock file. So probably there are 2 issues: one that the clean --locks command doesn't work and the other one that in certain cases (presumably errors) the lock file gets to stay and then will affect future runs (only on Windows we've seen this).

Trace below:

```
[2022-08-01T18:14:18.583]: FINAL union conda environment descriptor:
[2022-08-01T18:14:18.786]: ---
[2022-08-01T18:14:18.786]: channels:
[2022-08-01T18:14:18.786]: - conda-forge
[2022-08-01T18:14:18.786]: dependencies:
[2022-08-01T18:14:18.786]: - python=3.7.5
[2022-08-01T18:14:18.786]: - pip=20.1
[2022-08-01T18:14:18.787]: - pip:
[2022-08-01T18:14:18.787]:   - rpaframework==15.4.0
[2022-08-01T18:14:18.787]: ---
[2022-08-01T18:14:18.789]: ####  Progress: 05/13  v11.13.0     0.279s  Running micromamba phase.
[2022-08-01T18:14:48.925]: Fatal [Micromamba [1/1]]: exit status 1
[2022-08-01T18:14:48.933]: ---  installation plan "394a2ee9397413bd" 2022-08-01T18:14:18Z [force: false, fresh: true| rcc v11.13.0]  ---
[2022-08-01T18:14:48.933]: 
[2022-08-01T18:14:48.933]: ---  plan blueprint @0.000s  ---
[2022-08-01T18:14:48.933]: 
[2022-08-01T18:14:48.933]: channels:
[2022-08-01T18:14:48.933]: - conda-forge
[2022-08-01T18:14:48.933]: dependencies:
[2022-08-01T18:14:48.933]: - python=3.7.5
[...]

PR #1915, opened 2022-09-12
Lockfiles don't rely on PID (#1834)

Klaim:
This should remove the possibility of #1834 to ever happen, but does not attempt to fix the "clean" issue.

This change removes reliance on pid written in a lockfile for the lockfile mechanism to know if the same process is attempting to lock the same file twice. This case is currently an error (throws `std::logic_error`) but this will probably change later. Meanwhile this fix changes how lockfiles are implemented so that they could share ownership of the lock if the process already did lock that file. I kept the check for double locking for backward compatibility.

### DIFF
--- a/libmamba/src/core/util.cpp
+++ b/libmamba/src/core/util.cpp
@@ -1021,10 +1021,10 @@ namespace mamba
 
         if (blocking)
         {
-            std::chrono::seconds timer(0);
-            bool has_timeout = m_timeout.count() > 0;
-
-            while (!has_timeout || (timer < m_timeout))
+            static constexpr auto default_timeout = std::chrono::seconds(30);
+            const auto timeout = m_timeout > std::chrono::seconds::zero() ? m_timeout : default_timeout;
+            const auto begin_time = std::chrono::system_clock::now();
+            while ((std::chrono::system_clock::now() - begin_time) < timeout)
             {
                 ret = _locking(m_fd, LK_NBLCK, 1 /*lock_file_contents_length()*/);
                 if (ret == 0)
@@ -1033,7 +1033,7 @@ namespace mamba
                 timer += std::chrono::seconds(1);
             }
 
-            if (has_timeout && (timer >= m_timeout) && (ret == -1))
+            if (ret != 0)
                 errno = EINTR;
         }
         else
--- a/libmamba/src/core/util.cpp
+++ b/libmamba/src/core/util.cpp
@@ -36,6 +36,8 @@ extern "C"
 #include <mutex>
 #include <condition_variable>
 #include <optional>
+#include <unordered_map>
+#include <memory>
 #include <openssl/evp.h>
 
 #include "mamba/core/environment.hpp"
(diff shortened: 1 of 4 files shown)


Klaim:
Not ready yet, tests don't pass locally but this is the general idea. I think there is something fishy file-concurrency-wise with how the system file lock functions are called, but the mutex lock in the registry should help with that too.

Klaim:
Also: I'm open to suggestions for a better name than `LockFileOwner` 😬

libmamba/src/core/util.cpp
```diff
@@ -1060,72 +997,214 @@ namespace mamba
...
+                               // lock the whole container
+
+            std::unordered_map<int, fs::u8path>
+                fd_to_locked_path;  // this is a workaround the usage of file descriptors on linux
+                                    // instead of paths
+            mutable std::recursive_mutex
+                mutex;  // TODO: replace by synchronized_value once available
+        };
+
+        static LockedFilesRegistry files_locked_by_this_process;
```
JohanMabille, line 1110: Do we need to declare this variable static since it is defined in an anonymous namespace?

Klaim: You are correct that this is redundant. I added it in case it was not clear to everybody, but can remove it without change.

libmamba/src/core/util.cpp
```diff
@@ -848,127 +889,23 @@ namespace mamba
...
 
         // POSIX systems automatically remove locks when closing any file
         // descriptor related to the file
 #ifdef _WIN32
-        LOG_TRACE << "Removing lock on '" << m_lock.string() << "'";
+        LOG_TRACE << "Removing lock on '" << m_lockfile_path.string() << "'";
         _lseek(m_fd, MAMBA_LOCK_POS, SEEK_SET);
         ret = _locking(m_fd, LK_UNLCK, 1 /*lock_file_contents_length()*/);
 #endif
         remove_lockfile();
```
kariharju, line 903: Hi,
This may just be "shouting from the sidelines", so take this with a heavy grain of salt 😉 

Removing the lock file seems a bit odd here. We ran into a similar problem in our development a while back.
Having the lock file in place enabled us to see what is locking it, and removing the file kind of made this hidden and caused some extra hassle.
..but if you can provide some test build we can test this change with our RCC builds (https://github.com/robocorp/rcc)

Klaim: Hi, thanks for pointing this. However I'm not sure what you find odd exactly as this is the `unlock` function called when the lock have to be released, so removing the lockfile is kind of mandatory? Or did I misunderstand something?
Also I removed the pid written in the lockfile, do you think it should be kept for debugging purpose? We don't need it anymore here but it's not clear to me if it's still useful for users.

As for the build I'm not sure how I can provide it based on the CI output yet but I'm kind of new around here, I'll ask around and let you know.

kariharju: Hi, 
As said, I might be way off base, but the way we ended up doing the file locks (a simplification as I was not the one implementing 😉) is just to have the file there, take the lock from the process, and ask the OS if the file has locks or not. 

That way, the lock file always exists, and in debugging cases you can go in and check with other tools also what is holding the lock.

Klaim: Thanks, I'll discuss about it with the others, I'll consider making a PR with that logic. We have plans to refactor/review/revamp this lockfile system anyway so that's good input. 👍🏽

PR #1917, opened 2022-09-12
Install from lockfile by category

### LINKED ISSUE #1883: Micromamba: allow selecting category for new lockfile format
Based on the lack of command-line options and skimming the source code, it appears that micromamba is hardcoded to install only packages where `category=main`. It would be convenient to, for example, specify both `main` and `dev`.

https://github.com/mamba-org/mamba/blob/e0527eb061b0ce724dcb3c2f34db25c907ee7113/libmamba/src/core/transaction.cpp#L1572

jvansanten:
Add a `--category` flag to restrict which category of package should be installed from an environment lockfile. Defaults to "main".

Closes #1883.

### DIFF
--- a/libmamba/src/api/configuration.cpp
+++ b/libmamba/src/api/configuration.cpp
@@ -1264,6 +1264,10 @@ namespace mamba
                    .group("Solver")
                    .description("Only install dependencies"));
 
+        insert(Configurable("categories", std::vector<std::string>({ "main" }))
+                   .group("Solver")
+                   .description("Package categories to consider"));
+
         insert(Configurable("retry_clean_cache", false)
                    .group("Solver")
                    .set_env_var_names()
--- a/libmamba/src/api/configuration.cpp
+++ b/libmamba/src/api/configuration.cpp
@@ -1266,7 +1266,7 @@ namespace mamba
 
         insert(Configurable("categories", std::vector<std::string>({ "main" }))
                    .group("Solver")
-                   .description("Package categories to consider"));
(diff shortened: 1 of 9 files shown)


libmamba/src/api/configuration.cpp
```diff
@@ -1264,6 +1264,10 @@ namespace mamba
                    .group("Solver")
                    .description("Only install dependencies"));
 
+        insert(Configurable("categories", std::vector<std::string>({ "main" }))
+                   .group("Solver")
+                   .description("Package categories to consider"));
```
jonashaag, line 1269:
```suggestion
                   .description("Package categories to consider when installing from a lock file"));
```

libmamba/src/core/transaction.cpp
```diff
@@ -1573,31 +1574,45 @@ namespace mamba
...
+        {
+            std::vector<PackageInfo> conda, pip;
+        } packages;
+
+        for (const auto& category : categories)
+        {
+            std::vector<PackageInfo> selected_packages;
+
+            selected_packages = lockfile_data.get_packages_for(
+                category, Context::instance().platform, default_manager);
```
jonashaag, line 1589: Let's hardcode "conda" here as well? We're already hardcoding it in the `back_inserter` call.

jonashaag (commented):
Thanks, LGTM, although I'm not familiar with this area of code. Can you add a small test?

jonashaag (approved):
Awesome!

wolfv:
Thanks!

PR #1949, opened 2022-09-27
Increase curl buffer size to 100 KiB

jonashaag:


wolfv:
I am wondering if we should we make this configurable?
If I understand correctly, this will just use a bit more RAM ... probably totally fine in these modern times :)

jonashaag:
Naah, I don't think it's worth it. I wonder if we should just go to the max of 512 KiB. Even if you have 100 parallel downloads (curl default is 50 and Mamba default is 3) that will only be 50 MiB.

PR #1952, opened 2022-09-28
align mamba shell scripts more with current conda scripts

wolfv:


jonashaag:
In the fish scripts I added special support for `micromamba activate --help`, do we need it here as well?

wolfv:
that could be nice to have for the micromamba bash scripts, yeah!

PR #1953, opened 2022-09-28
Fixes for `ps` commands

wolfv:


jonashaag:
```
$ mkdir /tmp/x
$ HOME=/tmp/x /Users/j/Downloads/micromamba ps
critical libmamba 'mamba run' failed to lock (/tmp/x/.mamba/proc) or lockfile was not properly deleted
```

Then

```
$ mkdir /tmp/x/.mamba
HOME=/tmp/x /Users/j/Downloads/micromamba ps
critical libmamba filesystem error: in directory_iterator::directory_iterator(...): Not a directory ["/tmp/x/.mamba/proc"]
```

```
# Now .proc is an empty file which will break the next invocation
$ file /tmp/x/.mamba/proc
/tmp/x/.mamba/proc: empty
```

wolfv:
thanks for checking! Will continue then

wolfv:
Hmm, I just double-checked and the first two cases are actually solved now :) I am not sure we really need additional safeguards for the `proc` is a file case ... because that's really not something that should ever happen and the error message is "OK"

jonashaag:
I'm so sorry, I took the old download :(

PR #1957, opened 2022-09-28
Update fish scripts and completion

jonashaag:
1. Changes similar to https://github.com/mamba-org/mamba/pull/1952
2. Improve fish shell completion: It now supports completing subcommands and `-n`.

### DIFF
--- a/libmamba/data/mamba.fish
+++ b/libmamba/data/mamba.fish
@@ -54,53 +54,27 @@ if not set -q MAMBA_NO_PROMPT
 end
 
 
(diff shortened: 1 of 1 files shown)


jonashaag:
`micromamba deactivate` currently not supported because it's missing from `micromamba --help`.

wolfv:
Too bad that there is no way to handle this with a callback :/

jonashaag:
How about implementing a dummy command in the C++ CLI? We can make it error and print a nice message.

PR #1963, opened 2022-09-28
Renamed config option download_threads

wholtz:
Changed to max_parallel_downloads as multithreading is not used.


jonashaag:
Thanks!

Do we care about backwards compatibility here?

PR #1976, opened 2022-10-02
Add cross-compilation builds

### LINKED ISSUE #1887: Add fully static builds for all platforms
Similarly to #1860, it would be nice to create fully static builds for all supported platforms in CI. Like this, we could release nightlies.

Still to do:

- [x] `windows-64`
- [x] `osx-arm64`
- [x] `linux-aarch64`
- [x] `linux-ppc64le`

CC @jonashaag

pavelzw:
- Adds cross-compilation builds for `linux-aarch64`, `linux-ppc64le` and `osx-arm64`
- Switches from custom compilation steps to feedstock compilation steps
- Remove `Don't run this job branches because it is very slow` in the Windows static build since the Windows static build takes about 30min (~22 with caching) which is still faster than the regular CI

I didn't manage to get the conda-build pipeline working in reasonable time and since we already have build instructions for Windows, I guess that's okay.

One issue we have is that the Linux and macOS builds don't use caching any more... But since they take less time than the regular CI pipeline even without caching, it should be okay...

Closes #1887


pavelzw:
About the windows build times: 

This is without cache:

<details>
<summary>Details</summary>

image


</details>

This is with cache:

<details>
<summary>Details</summary>

image


</details>

.github/workflows/static_build.yml
```diff
@@ -13,58 +13,64 @@ concurrency:
...
+    runs-on: ${{ matrix.OS }}
+    env:
+      CI: True
+      CONDA_BUILD_YML: ${{ matrix.CONDA_BUILD_YML }}
     strategy:
+      fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-latest]
+        include:
+          - { OS: ubuntu-latest, CONDA_BUILD_YML: linux_64_, TARGET: linux-64, BUILD: linux-64 }
```
jonashaag, line 26: In GHA it's convention to use `kebab-case`.

In conda-forge terms, `BUILD` = `host`.

jonashaag: Would either include the full file name in `CONDA_BUILD_YML` so it's clear a it's a file name: `conda-build-yml-file: linux_64_.yaml`. Or autogenerate from the `target`: `echo ${{ matrix.target }}_ | tr _ -`.

jonashaag (approved):
Wow nice!

We can deal with caching later if we want to.

Shall we add a small test run with the extracted binary? I know there's already a test in the Conda recipe but maybe it's useful to add another one that actually creates an environment and runs a program. We could even use something like https://github.com/uraimo/run-on-arch-action to run it on linux-aarch64 but that's bonus.

pavelzw:
> Actually the Conda recipe already creates an env and runs a program, so all that'd be remaining is running on aarch64 (which the Conda recipe doesn't test)

In conda-forge, I think the `linux-aarch64` builds are actually tested in an emulated environment (see the [`test: native_and_emulated` flag in conda-forge](https://conda-forge.org/docs/maintainer/conda_forge_yml.html#test)). See [here](https://conda-forge.org/docs/maintainer/conda_forge_yml.html#test) for the `aarch64` build and [here](https://conda-forge.org/docs/maintainer/conda_forge_yml.html#test) for the `ppc64le` build.
For `osx-arm64` there is no testing done in the Azure pipeline since there is no emulator available. I don't know if this is a conda-forge specific thing or if we can also add the command line flags to `conda mambabuild`. I'll investigate it soon.

jonashaag:
Good to know! We don't have to add additional testing here in that case.

pavelzw:
Okay so there is only `--no-test` in the [conda-build documentation](https://docs.conda.io/projects/conda-build/en/latest/resources/commands/conda-build.html). So I guess it's a conda-forge thing. Let's use uraimo/run-on-arch-action then 👍🏻

.github/workflows/static_build.yml
```diff
@@ -13,58 +13,77 @@ concurrency:
...
-          cache-env: true
-      - name: create build environment
-        run: micromamba install -y -n build_env -f ./libmamba/environment-static-dev.yml
-      - uses: hendrikmuhs/ccache-action@main
+          repository: conda-forge/micromamba-feedstock
+          path: micromamba-feedstock
+      - name: Patch micromamba-feedstock
+        if: ${{ matrix.platform == 'linux' }}
+        run: |
+          sed -i -E ':a; N; $!ba; s/url:.*libmamba.patch/git_url: ..\/..\//g' micromamba-feedstock/recipe/meta.yaml
```
jonashaag, line 41: What is this magic sed command btw? We should document it or find a simpler more obvious alternative.

jonashaag (approved):
This looks really nice now. Thanks a bunch! Do you consider this ready to merge?

pavelzw:
Yes 🚀

.github/workflows/static_build.yml
```diff
@@ -75,8 +75,6 @@ jobs:
             ./micromamba --version
             ./micromamba --help
             ./micromamba
-            apt update && apt install ca-certificates
```
jonashaag: I guess a simple `-y` for both commands would've fixed it

.github/workflows/static_build.yml
pavelzw: This replaces the
``` 
url: ...
sha256: ...
patches: 
  libmamba.patch
```

with a simple `git_url: ../../`

.github/workflows/static_build.yml
pavelzw: Hmm, I'll try it out again. Why is `apt` not smart enough to detect on its own that it's in a non-interactive environment?

.github/workflows/static_build.yml
pavelzw: The problem is that sed is not working that well for multiple lines... On macOS, this command doesn't work and I didn't manage to find a single sed command so I split it in four sed commands... Do you know a more elegant alternative for replacement of multiple lines by a single one?

jonashaag: Unfortunately I don't :(

Specifically I was wondering what this means: `:a; N; $!ba; s/u`

.github/workflows/static_build.yml
jonashaag: 🤷

.github/workflows/static_build.yml
pavelzw: I copied it somewhere from stack overflow where there was an explanation 😅 but I can't find it anymore :/

pavelzw: Hmm, I guess `sed -i -E -z 's/url:.*libmamba.patch/git_url: ..\/..\//g'` works too...

pavelzw:
> Do you consider this ready to merge?

If you don't have anything more to add and the CI passes, you can merge it 👍🏻

jonashaag:
`sed -z` is a nice hack to do multiline replacements with sed :)

jonashaag:
It could be that we have `gsed` in the macOS runners. I think it's from the Homebrew `coreutils` package.

pavelzw:
> It could be that we have `gsed` in the macOS runners. I think it's from the Homebrew `coreutils` package.

[Unfortunately not](https://github.com/pavelzw/ci-test/actions/runs/3181007848/jobs/5185250612).

PR #1982, opened 2022-10-03
Fixed error when attempting to get the size of directories. (fix for #1979)

### LINKED ISSUE #1979: Error `critical libmamba filesystem error: cannot get file size: Is a directory` while using 0.26.0
I got this error building my applications with the latest micromamba 0.26.0:

```bash
critical libmamba filesystem error: cannot get file size: Is a directory [/opt/conda/pkgs/pip-22.2.2-pyhd8ed1ab_0/site-packages]
```

To reproduce:

```bash
micromamba install -y -f /tmp/env.yaml
micromamba remove -y --force pip
micromamba clean -a -y
```

with the following `env.yaml`

```yaml
channels:
  - conda-forge
dependencies:
  - python=3.9
  - pip
  - pip:
    - pyyaml
```

The installation of `pyyaml` isn't necessary, but illustrates the usage I am doing (minimizing the size of my app container).

I have a feeling this is related to #1975

Klaim:
This should fix  #1979 (well it does fix it for me)

The standard filesystem library cannot guarantee that attempting to get the size of a directory with result in a value or an error, because of the wild differences in implementations. So we need to check if it's a directory and don't count sum it's size to the whole if it is.

I'll add a repro in-project.

### DIFF
--- a/libmamba/src/api/clean.cpp
+++ b/libmamba/src/api/clean.cpp
@@ -220,7 +220,7 @@ namespace mamba
             std::size_t size = 0;
             for (auto& fp : fs::recursive_directory_iterator(p))
             {
-                if (!fp.is_symlink())
+                if (!fp.is_symlink() && !fp.is_directory())
                 {
                     size += fp.file_size();
                 }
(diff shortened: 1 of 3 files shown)


wolfv:
awesome. Looks like only the linter is failing

Issue #1981, opened 2022-10-03
`warning  libmamba Could not create proc dir`


wholtz:
Transfering [mamba-org/micromamba-docker issue #215](https://github.com/mamba-org/micromamba-docker/issues/215) to here.

@shughes-uk reported 

> Using 0.26.0 I see this when using `micromamba run`
> 
> ```
> libmamba Could not create proc dir: "/home/mambauser/.mamba/proc"
> ```

I am able to reproduce this on macOS (v12.6) without using docker:
```bash
$ ./micromamba run /bin/bash
warning  libmamba Could not create proc dir: "/Users/MyUserID/.mamba/proc"
$ 
```

Probably related to #1979 and #1975.

jonashaag:
I have this as well

wolfv:
Thanks for the report! The `run` functionality seems to work though. I'll have a look right now, but might only be able to fix it by tomorrow morning.

PR #1985, opened 2022-10-03
fix error condition for log warning

wolfv:
@jonashaag @shughes-uk @wholtz thanks for reporting #1981 – it was just a bad check for the error code. This should fix it! Sorry! :)

### DIFF
--- a/micromamba/src/run.cpp
+++ b/micromamba/src/run.cpp
@@ -389,10 +389,11 @@ set_run_command(CLI::App* subcom)
 
             // Make sure the proc directory is always existing and ready.
             std::error_code ec;
-            fs::create_directories(proc_dir(), ec);
-            if (!ec)
+            bool is_created = fs::create_directories(proc_dir(), ec);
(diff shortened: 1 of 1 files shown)


micromamba/src/run.cpp
```diff
@@ -389,10 +389,11 @@ set_run_command(CLI::App* subcom)
 
             // Make sure the proc directory is always existing and ready.
             std::error_code ec;
-            fs::create_directories(proc_dir(), ec);
-            if (!ec)
+            bool is_created = fs::create_directories(proc_dir(), ec);
+            if (!is_created && ec)
```
JohanMabille, line 393: It works but wh not simply ```if (ec)``` ? Can `create_directories` succeed but still return a non zero error code?

wolfv: I am not sure what happens when the directory already exists (the manual says that it returns `true` if directories are created). But yeah, checking only `ec` would probably be enough.

PR #1996, opened 2022-10-07
Fix an infinite loop in replace_all() when the search string is empty

tsibley:
The empty string is found at the start of the string, pos = 0, and replaced over and over again.  If the replacement string is also empty, the size of the data string will not change.  If the replacement string is not empty, it will be repeatedly prepended to the data string.

While most callers are careful to ensure the search string is not empty, PosixActivator::update_prompt() lost this property in "Micromamba: Substitute env vars in .condarc (#1423)" (9e5f8fd).  Rather than restore the check in the caller, fix the root of the issue in replace_all() itself.

I ran into this bug as a Micromamba user in a particular set of circumstances where the CONDA_PROMPT_MODIFIER env var was the empty string.¹  It manifested to me the user as a `micromamba create` invocation hanging at the linking phase for any package with a post-link script (such as openmpi).  The generated post-link script wrapper invoked, indirectly, `micromamba shell --shell bash activate <path>` which was hanging during update_prompt().

¹ https://github.com/nextstrain/cli/pull/223#issuecomment-1270806302

### DIFF
--- a/libmamba/src/core/util.cpp
+++ b/libmamba/src/core/util.cpp
@@ -321,6 +321,10 @@ namespace mamba
     template <class S>
     void replace_all_impl(S& data, const S& search, const S& replace)
     {
+        if (search.empty())
+        {
+            return;
+        }
         std::size_t pos = data.find(search);
         while (pos != std::string::npos)
         {
--- a/libmamba/tests/test_string_methods.cpp
+++ b/libmamba/tests/test_string_methods.cpp
@@ -61,6 +61,10 @@ namespace mamba
         replace_all(prefix, "/I/am/a/PREFIX", "/Yes/Thats/great/");
         EXPECT_TRUE(starts_with(prefix, "/Yes/Thats/great/\n"));
 
+        std::string testbuf2 = "this is another test wow";
+        replace_all(testbuf2, "", "somereplacement");
+        EXPECT_EQ(testbuf2, "this is another test wow");
+
         std::string prefix_unicode = "/I/am/Dörteæœ©æ©fðgb®/PREFIX\n\nabcdefg\nxyz";
         replace_all(
             prefix_unicode, "/I/am/Dörteæœ©æ©fðgb®/PREFIX", "/home/åéäáßðæœ©ðfßfáðß/123123123");


tsibley:
Compiling locally I was able to confirm that I could no longer reproduce the loop.

tsibley:
My pared down reproduction, which requires a set but empty `CONDA_PROMPT_MODIFIER` and also (though I don't yet know why) a `HOME` that's not your user's actual home dir.

```bash
HOME=/ CONDA_PROMPT_MODIFIER= micromamba shell -s bash activate /tmp
```

or alternatively, `MAMBA_NO_RC=true` or the equivalent `--no-rc` option:

```bash
MAMBA_NO_RC=true CONDA_PROMPT_MODIFIER= micromamba shell -s bash activate /tmp
```

jonashaag:
Sorry for that bug!

Shall we raise an exception in replace_all? What does Boost replace_all do?

Could you add a small high level test („calling mamba with empty modifier works“) and on the low level („string replace with empty needle works“)?

wolfv:
I think we shouldn't raise an exception and just return, as is done here. Seems like the logical thing to happen with an empty search.

Although, interestingly, in Python this happens:

```
>>> "aaa".replace("", "123")
'123a123a123a123'
```

tsibley:
I don't think an exception is warranted. That would avoid the infinite loop but to end users the result is the same: mamba is broken.

With an empty search, I'd expect either Python's behaviour (which is similar to long-standing conventions that I believe predate Python) or the simpler "nothing is replaced" behaviour I submitted in this PR.

AFAICT, Boost [appears to have an infinite loop possible](https://github.com/boostorg/build/blob/b9d218b24f508dfd27e996c63181d8bfe4516bcf/src/engine/builtins.cpp#L1573-L1585) when _both_ the search string and replacement string are empty, but when the replacement string is non-empty, it appears it will produce Python's behaviour. This is due to the unconditional advancement of `pos` outside of the call to `.find()`.

```c++

    template <class S>
    void replace_all(S &str, const S &from, const S &to)
    {
        const auto from_len = from.length();
        const auto to_len = to.length();
        auto pos = str.find(from, 0);
        while (pos != S::npos)
        {
            str.replace(pos, from_len, to);
            pos += to_len;
            pos = str.find(from, pos);
        }
    }
```

> Could you add a small high level test („calling mamba with empty modifier works“) and on the low level („string replace with empty needle works“)?

Maybe. I'm not familiar with the testing harness/framework used in this repo, and I'm no C++ expert, so will have to see how to do it. If you could add the tests you're interested in, that'd certainly be easier/faster.

tsibley:
Eh, scratch that analysis of Boost's behaviour. The code I linked/excerpted doesn't seem to be the actual `boost::algorithm::replace_all()` implementation but something else. What I think is the [actual implementation](https://github.com/boostorg/algorithm/blob/6f4d74c6bbe720e7606d028ca59285bc4d9f8918/include/boost/algorithm/string/replace.hpp#L637-L656) (?) does something more complicated I haven't teased apart yet. But testing it experimentally (at least with boost 1.65), the original string is returned when the search is empty. So this matches the behaviour change made by this PR.

jonashaag:
Thanks for looking into this! If you’re familiar with pytest then a Micromamba level test would be appreciated. Otherwise I can try to find some time to do it.

wolfv:
```
    size_t pos = 0;
    while ((pos = in.find(search, pos)) != std::string::npos && pos < 100) {
        in.replace(pos, search.length(), replace);
        pos += (search.length() > 0) ? replace.length() : replace.length() + 1;
    }
```

This would mimick the Python behavior. Although I think with an empty search, we should just return the original string (it matches nothing). I'm adding a simple test.

jonashaag (approved):

PR #2002, opened 2022-10-10
Add complete environment removal in micromamba

### LINKED ISSUE #1506: Micromamba: env remove
Are there any plans for Micromamba to support `micromamba env remove --name <myenv>` in the future?

Cheers for all the great work!

Hind-M:
Fix #1506


jonashaag:
I wonder if the test failure is related. Could it be that we create a Conda prefix in a subfolder of an environment that we're deleting in the tests?

Hind-M:
> I wonder if the test failure is related. Could it be that we use a Conda prefix located in a subfolder of an environment that we're deleting in the tests?

Hum, it looks like the failure is here since the first commit (where the new env remove test hasn't been added yet). I think it's coming from adding `register_env` when creating the target directory.
When the tests are launched, `~/.conda/environments.txt` is renamed in setup and then using `create` inside the tests calls `register_env` which creates a new `~/.conda/environments.txt` file I guess... I'm not sure how we would want to handle this? Not renaming/using the bkup file in the tests?

micromamba/tests/test_env.py
```diff
@@ -124,3 +122,19 @@ def test_env_export(self):
         assert ret["name"] == env_name
         assert set(ret["channels"]) == {"https://conda.anaconda.org/conda-forge"}
         assert "micromamba=0.24.0=0" in ret["dependencies"]
+
+    def test_env_remove(self):
+        env_name = "env-create-remove"
+        # Create env with xtensor
+        res = create("xtensor", "-n", env_name, "--json", no_dry_run=True)
```
wolfv, line 129: I don't see any tests with the environments file here. I think it would be nice to also check that the environment is mentioned in `~/.conda/environments.txt` file

wolfv:
for the linter, you can install `pre-commit` and run with `pre-commit run --all` and it will fix everything at once :)

Hind-M:
> for the linter, you can install `pre-commit` and run with `pre-commit run --all` and it will fix everything at once :)

Oh nice! Thanks :)

micromamba/src/env.cpp
```diff
@@ -163,4 +165,37 @@ set_env_command(CLI::App* com)
...
+        {
+            // Remove specs if exist
+            remove(MAMBA_REMOVE_ALL);
+
+            auto& ctx = Context::instance();
+            if (!ctx.dry_run)
+            {
+                auto& prefix = ctx.target_prefix;
+                // Remove env directory and all its content
+                fs::remove_all(env::expand_user(prefix));
```
wolfv, line 184: So I just noticed that we might be forgetting one annoying case here. On Windows, if a file is in use, we cannot remove it (only rename) :/

conda / mamba solves this by renaming the files to `<filename>.mamba_trash`

Issue #2004, opened 2022-10-11
micromamba failed to create environment on nfs4 because of "cannot set permissions: Operation not permitted"


hwalinga:
## Problem

When running a `micromamba create` command, the command fails with a 

```
....
Transaction starting
Linking ca-certificates-2022.07.19-h06a4308_0
Linking _libgcc_mutex-0.1-main
Linking libstdcxx-ng-11.2.0-h1234567_1
Linking libgomp-11.2.0-h1234567_1
Linking _openmp_mutex-5.1-1_gnu
Linking libgcc-ng-11.2.0-h1234567_1
Linking tbb-2020.3-hfd86e86_0
critical libmamba filesystem error: cannot set permissions: Operation not permitted [/tudelft.net/staff-groups/ewi/insy/DBL/hwalinga/software/micromamba/envs/pathoscope-env-mamba/lib/libtbb.so.2]
```

Running strace reveals:

```
fchmodat(AT_FDCWD, "/tudelft.net/staff-groups/ewi/insy/DBL/hwalinga/software/micromamba/envs/pathoscope-env-mamba/lib/libtbb.so.2", 0700) = -1 EPERM (Operation not permitted)
```

## Consideration

The file is located on a nfs4 filesystem. I suspect something is wrong with the exact configuration of the nfs4 which makes it not possible to change the permissions of the file (even though I own it). Unfortunately, I cannot change that as I don't own the system this nfs4 connects to.

However, the attempt by micromamba to change the permissions is redundant. The specific file already has permissions 700. 

## Solution

Going trough the code, I think the error originates from:

```
            fs::permissions(dst, fs::status(src).permissions());
```

https://github.com/mamba-org/mamba/blob/main/libmamba/src/core/link.cpp#L701

Would it be possible to make this specific line a bit more robust against problematic filesystems where permissions are an issue?

jonashaag:
Thanks for the report and debugging this. 

I think it’s fine to skip changing permissions if the file already has the correct permissions. 

We could also add a command line option to ignore permissions errors and/or to not attempt to set permissions in the first place.

wolfv:
@hwalinga I think we do have to change the permissions at this place in the code, but we should make it robust to not fail. If you can make a PR that uses the overload that takes an `std::error_code` as third argument and then add a check if the error code is set (and print a warning using `LOG_WARNING << "Could not change permissions on " << path << ": " << ec.what()`) that would be very much appreciated!

jonashaag:
[...]

PR #2009, opened 2022-10-11
Do not crash when permissions cannot be changed, instead log warning (#2004)

hwalinga:
As discussed in #2004. 

Tested locally on the misbehaving filesystem, it works.


wolfv:
How many warnings do you get? I am wondering if we should print only once or check if permissions actually needed changing. Could be quite a few warnings?

PR #2012, opened 2022-10-12
Test special characters in basic auth

### LINKED ISSUE #2010: Broken basic authentication with libcurl=7.85
We have the following in `.condarc`:

```
channel_alias: https:/some%40emailaddress.com:password@somehost...
```

With Micromamba/Mamba 0.27 (= curl>=7.85) this doesn't work because the `Authorization` header sent contains `%40` while it should be `@`.

It seems to be a problem with curl 7.84 or 7.85 because Mamba with curl 7.83 it works. We don't have a release with 0.74 so I can't easily test.

Is this a regression in curl? I couldn't find anything in the curl release notes.

jonashaag:
Fixes #2010

### DIFF
--- a/libmamba/environment-dev.yml
+++ b/libmamba/environment-dev.yml
@@ -9,7 +9,7 @@ dependencies:
   - libsolv >=0.7.18
   - libarchive
   - libsodium
-  - libcurl 7.82.0 *_0
+  - libcurl >=7.86
   - gtest
   - gmock
   - cpp-expected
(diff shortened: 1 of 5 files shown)


jonashaag:
Nope, this extends the test suite, we should merge it when next curl is released

wolfv:
Do you want to rebase on main and run this again?

wolfv:
Btw. I noticed that the proxy tests are sometimes flaky, unfortunately

jonashaag:
Good to review + merge. Test failure seems unrelated.

wolfv (approved):

PR #2014, opened 2022-10-12
LockFile as a testable type - no need to try_lock or use pointers.

Klaim:
This change was discussed/designed with @JohanMabille mainly. There are alternative designs to this but we can now test this one.

The goal is to simplify most of the handling of `LockFile` objects. Instead of having factory functions that might or not catch errors, we just maek `LockFile` itself not throw on construction and let the user decide if testing if it's locked is important or not.
Basically, this version of `LockFile` acts both as a lockfile and as an `expected`, which can hold an error if locking failed.

For example, from the documentation (added in the `LockFile` definition):
```c++
LockFile some_work_on(some_path)
{
    LockFile lock{ some_path, timeout };
    if(lock) // make sure the locking happened
    {
        print("locked file {}, locking counts: {}", some_path, lock.count_lock_owners()); // success might mean we are locking the same path from multiple threads
        do_something(som_path); // locking was a success
    }
    else // locking didnt succeed for some reason
    {
        if(auto error = lock.error) print(error); // some error happened while attempting the lock, maybe some other process already locks the path
        else print("didn't attempt locking {}") // locking didn't happen for some other reason, maybe a configuration option
    }
    some_more_work(some_path); // do this that the lock failed or not
    return lock; // The locking ownership can be transfered to another function if necessary
}
```

The main alternative to this design is to keep `LockFile` throwing constructors, keep `try_lock` but make it return an expected, and rationalize more clearly what `try_lock` does allow or not. Feel free to provide some arguments and guidance for alternative designs such as this one, we are open to discuss this.

### DIFF
--- a/libmamba/include/mamba/core/util.hpp
+++ b/libmamba/include/mamba/core/util.hpp
@@ -121,6 +121,11 @@ namespace mamba
         fs::u8path path() const;
         fs::u8path lockfile_path() const;
 
+        std::size_t count_lock_owners() const
+        {
+            return impl.use_count();
+        }
+
 #ifdef _WIN32
         // Using file descriptor on Windows may cause false negative
         static bool is_locked(const fs::u8path& path);
--- a/libmamba/include/mamba/core/util.hpp
+++ b/libmamba/include/mamba/core/util.hpp
@@ -117,10 +117,32 @@ namespace mamba
         static std::unique_ptr<LockFile> create_lock(const fs::u8path& path);
         static std::unique_ptr<LockFile> try_lock(const fs::u8path& path) noexcept;
 
+        // Returns true if this LockFile is currently maintaining a lock on the target path.
+        // Returns false if this instance have been moved-from without being re-assigned,
+        // or if the lock acquisition failed.
+        bool is_locked() const
+        {
+            return impl ? true: false;
+        }
+
+        // Convenient operator to check if a lockfile is actually locking a path.
+        explicit operator bool() const
+        {
+            return is_locked();
+        }
+
+        // Returns the fd of the path being locked if `is_locked() == true`.
         int fd() const;
+
+        // Returns the path being locked if `is_locked() == true`.
         fs::u8path path() const;
+
+        // Returns the path of the lock-file being locked if `is_locked() == true`.
         fs::u8path lockfile_path() const;
 
+        // Returns the count of LockFile instances which are currently locking
+        // the same path/file from the same process.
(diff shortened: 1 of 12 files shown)


libmamba/include/mamba/core/util.hpp
```diff
@@ -131,18 +218,25 @@ namespace mamba
 
         static bool is_locked(const LockFile& lockfile)
         {
+            return lockfile.is_locked() &&
 #ifdef _WIN32
-            return is_locked(lockfile.lockfile_path());
+                   is_locked(lockfile.lockfile_path());
```
JohanMabille, line 223: Why do we need this double check, not only `lockfile.is_locked()` ?

Klaim: Because this function needs to check if the path is blocked "in general", not only blocked by this process. Checking `lockfile.is_locked()` returns true only if the current process is locking that path, if not, we need to check if some other process locks it.

libmamba/src/core/subdirdata.cpp
```diff
@@ -480,7 +480,7 @@ namespace mamba
                 LOG_DEBUG << "Copying repodata cache files from '" << m_expired_cache_path.string()
                           << "' to '" << m_writable_pkgs_dir.string() << "'";
                 fs::u8path writable_cache_dir = create_cache_dir(m_writable_pkgs_dir);
-                auto lock = LockFile::create_lock(writable_cache_dir);
+                auto lock = LockFile(writable_cache_dir);
```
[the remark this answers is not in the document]
Klaim: I didn't try to change the current usage of LockFile objects because their is many cases where failing to lock is not considered an issue. So not sure.

libmamba/src/core/subdirdata.cpp
```diff
@@ -502,13 +502,13 @@ namespace mamba
 
             {
                 LOG_TRACE << "Refreshing '" << json_file.string() << "'";
-                auto lock = LockFile::create_lock(json_file);
```
wolfv, line 505: Isn't this changing the behavior here? Before this was throwing, now it isnt?

libmamba/src/core/transaction.cpp
```diff
@@ -921,7 +921,7 @@ namespace mamba
             return true;
         }
 
-        auto lf = LockFile::create_lock(ctx.target_prefix / "conda-meta");
```
wolfv, line 924: this also changes behavior, but it is OK (I need to do the same in the self-update PR).

PR #1891, opened 2022-08-31
[MER][T11|T12|T13|T14] Creating the initial problems graph

syslaila:
Graph creation implementation for #1878  (we only implemented the graph creation for a subset of the libsolv problems)

- DependencyInfo - contains the info about a dependency name & the range
- ProblemsGraph - A directed graph of the packages involved in libsolv conflicts. A node can be: RootNode (a "virtual" node that has edges to the packages a user requested that are in conflict), PackageNode (a resolved package node), UnresolvedDependencyNode (a dependency that couldn't be found), ConstraintNode (an optional dependency for a package that is in conflict) 
- The ProblemsGraph contains a map between direct package conflicts (eg. a pair of the format: <python2.7.1, python3.6.0>).

We create the graph using the libsolv problems list. We will only add the nodes in the graph if they haven't been added already. In order to check if a specific package is in the graph or not, we store a mapping between the SolveId (a unique identifier from libsolv) and Id (the unique node identifier for a node in the graph).
Because a solvable can appear in multiple libsolv problems, we might need to update the information for the node. 
eg. 

- Problem1: pkgA requires pkgB -> we create a PackgeInfo type for pkgB
- Problem1: pkgB is not installable-> we need to update the pkgB info with the new problem type

PR #2015, opened 2022-10-12
neat pick improvements

AntoinePrv:
Neat pick improvements found during #1891
- Fix subdir json creation null json
- util_random missing headers
- Add PackageInfo rvalue string ctor

### DIFF
--- a/libmamba/include/mamba/core/util_random.hpp
+++ b/libmamba/include/mamba/core/util_random.hpp
@@ -8,10 +8,10 @@
 #define MAMBA_CORE_UTIL_RANDOM_HPP
 
 #include <algorithm>
-#include <functional>
+#include <limits>
 #include <random>
 #include <string>
-#include <string_view>
+#include <cstring>
 
 namespace mamba
 {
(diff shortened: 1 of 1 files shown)


libmamba/include/mamba/core/util_random.hpp
```diff
@@ -42,12 +45,12 @@ namespace mamba
 
     inline std::string generate_random_alphanumeric_string(std::size_t len)
     {
-        static constexpr auto chars = "0123456789"
-                                      "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
-                                      "abcdefghijklmnopqrstuvwxyz";
+        static constexpr auto chars = std::string_view("0123456789"
```
wolfv, line 48: I am slightly against this change because I don't really see any benefit and it will likely compile a little bit slower (especially also with the additional includes :)

libmamba/include/mamba/core/util_random.hpp
```diff
@@ -7,8 +7,11 @@
 #ifndef MAMBA_CORE_UTIL_RANDOM_HPP
 #define MAMBA_CORE_UTIL_RANDOM_HPP
 
+#include <algorithm>
```
wolfv, line 10: I am fine with the `cstring` include, but do we need `algorithm` and `functional`?

AntoinePrv: Indeed, I cannot find a use for `<functional>`. Two uses of `std::generate_n` for `<algorithm>`, and I should also add `<limits>`.

AntoinePrv: Squashing those commits.

wolfv:
looks like now you have to rebase :/

AntoinePrv:
Ha these changes have been merged in #1891. This is only the changes from your comments left then.

PR #2016, opened 2022-10-12
Ensure package record always has subdir

jaimergp:
Old channels do not feature this entry (see `free` channel), which causes issues with package cache records not being used because the lacking subdir prevents the match.


wolfv:
Thanks!

PR #2019, opened 2022-10-13
[MER][T31|T32|T33] ProblemsGraph compression

AntoinePrv:
Partial implementation of #1878

- [x] Based on #2033 for all `*string*` files, please write review there.

### DIFF
--- a/libmamba/include/mamba/core/package_info.hpp
+++ b/libmamba/include/mamba/core/package_info.hpp
@@ -42,6 +42,11 @@ namespace mamba
                     const std::string build_string,
                     std::size_t build_number);
 
+        PackageInfo(PackageInfo const&) = default;
+        PackageInfo(PackageInfo&&) noexcept = default;
+        PackageInfo& operator=(PackageInfo const&) = default;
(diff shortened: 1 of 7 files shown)


Klaim (approved):
LGTM 👍🏽 (didn't check the domain-specific meaning though)

libmamba/include/mamba/core/graph_util.hpp
```diff
@@ -340,6 +344,12 @@ namespace mamba
         return m_node_list.size();
     }
 
+    template <typename N, typename G>
+    inline auto DiGraphBase<N, G>::number_of_edges() const -> std::size_t
```
Klaim, line 348: `inline` is redundant here ^^

AntoinePrv: I noticed this, yes. Perhaps it is better to do it in another PR since this file is not widely modified by here.

AntoinePrv: The rest is done there https://github.com/mamba-org/mamba/pull/2039

PR #2022, opened 2022-10-14
Add missing changelog entry

AdrianFreundQC:
The changelog on the github releases page has already been updated, but the actual changelog files haven't been updated, so here's a commit for that.

@jonashaag


wolfv:
Thanks! Sorry for messing this up :/

PR #2023, opened 2022-10-14
add micromamba self-update functionality

wolfv:
- [x] rerun shell init?!
- [ ] Add update notification if newer version is available?
- [x] Test (especially on Windows)
- [x] Add Windows tricks to Bash ?!


baszalmstra:
How would you now call this? Like this?

```bash
micromamba update --self 
```

I think this is a bit confusing because it mixes two commands, one to update and environment and one to update the install itself. The help page of `update` would include arguments for both which might be confusing. I would opt to add a new sub-command for this use-case. Something along the lines of:

```bash
micromamba self-update
```

WDYT?

wolfv:
yeah, it was just a placeholder to use `update --self`. I think `self-update` could be fine.

Homebrew is using `update` to update `homebrew`, and `upgrade` to update packages. We're already using `update` to update packages so we need to come up with something else I think. Also it's confusing to make `upgrade` and `update` differnt...

libmamba/src/core/util_os.cpp
```diff
@@ -542,4 +542,24 @@ namespace mamba
...
+            silence.type = reproc::redirect::discard;
+            options.redirect.out = silence;
+            options.redirect.err = silence;
+        }
+
+        std::vector<std::string> cmd = { "/usr/bin/codesign", "-s", "-", "-f", path.string() };
+        auto [status, ec] = reproc::run(cmd, options);
+        if (ec)
+        {
+            throw std::runtime_error(std::string("Could not codesign executable") + ec.message());
```
baszalmstra, line 562:
```suggestion
            throw std::runtime_error(std::string("Could not codesign executable: ") + ec.message());
```

baszalmstra:
> Homebrew is using `update` to update `homebrew`, and `upgrade` to update packages. We're already using `update` to update packages so we need to come up with something else I think. Also it's confusing to make `upgrade` and `update` different...

Thats indeed very confusing 😅

wolfv:
The other option would be to specialcase `micromamba update micromamba` and also add a stage to `micromamba update --all` (if executed in the root prefix).
But maybe it's better to let the micromamba self-modifications live outside the normal package management...

micromamba/src/update.cpp
```diff
@@ -8,10 +8,108 @@
...
+    if (on_win)
+    {
+        fs::rename(mamba_exe, mamba_exe_bkup);
+        fs::copy_file(cache_path / "Library" / "bin" / "micromamba.exe",
+                      mamba_exe,
+                      fs::copy_options::overwrite_existing);
+    }
+    else
+    {
+        fs::copy_file(
```
jonashaag, line 103: Why no backup on Unix?

Shall we attempt to codesign before overwriting the old micromamba binary?

jonashaag:
> But maybe it's better to let the micromamba self-modifications live outside the normal package management

Strongly agree!

micromamba/src/update.cpp
wolfv: We also have the `bkup` file on Unix, but we don't need the `rename` "dance" because we can overwrite files that are currently used :)

jonashaag: Can you help me find the line that does the backup on Unix? oO

wolfv: It was in my head, I swear! :)

jonashaag:
Going to try this in a funny customer environment :)

jonashaag:
Tested on Windows with micromamba.exe in Downloads\:

```
error    libmamba Could not lock non-existing path 'C:\Users\USERNAME\conda-meta'
critical libmamba LockFile error. Aborting.
```

wolfv:
ugh :) 
I get further on Windows, but then it silently crashes... 

These are the last two log lines:

```
info     libmamba Transfer finalized, status: 304 [https://conda.anaconda.org/robostack-humble/win-64/repodata.json] 0 bytes
robostack-humble/win-64                                       No change
info     libmamba Reading cache files 'C:\Users\robostack\micromamba\pkgs\cache\5afe41e9.*' for repo index 'https://conda.anaconda.org/conda-forge/win-64'
```

micromamba/src/update.cpp
```diff
@@ -8,10 +8,121 @@
...
+    if (!exp_loaded)
+    {
+        throw std::runtime_error(exp_loaded.error().what());
+    }
+
+    pool.create_whatprovides();
+    std::string matchspec;
+    if (!version)
+    {
+        matchspec = fmt::format("micromamba>{}", umamba::version());
```
Klaim, line 50: Would this work? It would avoid unnecessary duplications.
```
const auto version_to_use = version ? version.value() : umamba::version();
matchspec = fmt::format("micromamba>{}", version);
```

micromamba/src/update.cpp
```diff
@@ -8,10 +8,121 @@
...
+        matchspec = fmt::format("micromamba={}", version.value());
+    }
+
+    auto solvable_ids = pool.select_solvables(pool.matchspec2id(matchspec), true);
+
+    if (solvable_ids.empty())
+    {
+        if (pool.select_solvables(pool.matchspec2id("micromamba")).empty())
+        {
+            throw std::runtime_error(
```
Klaim, line 63: This should be a `mamba_error`

micromamba/src/update.cpp
```diff
@@ -8,10 +8,121 @@
...
+        else
+        {
+            Console::instance().print(fmt::format(
+                "\nYour micromamba version ({}) is already up to date.", umamba::version()));
+            return 0;
+        }
+    }
+
+    std::optional<PackageInfo> latest_micromamba = pool.id2pkginfo(solvable_ids[0]);
+    LOG_WARNING << latest_micromamba.value().url;
```
Klaim, line 75: This line assumes that `latest_micromamba` have a value, or will throw. Is this the behavior we want here? Or should `latest_micromamba` be checked before logging?

micromamba/src/update.cpp
```diff
@@ -8,10 +8,121 @@
...
+
+    ctx.download_only = true;
+    MTransaction t(pool, { latest_micromamba.value() }, package_caches);
+    auto exp_prefix_data = PrefixData::create(ctx.root_prefix);
+    if (!exp_prefix_data)
+    {
+        // TODO: propagate tl::expected mechanism
+        throw std::runtime_error(exp_prefix_data.error().what());
```
Klaim, line 83: `mamba_error` should be used here

micromamba/src/update.cpp
```diff
@@ -8,10 +8,121 @@
...
+    fs::u8path cache_path = package_caches.get_extracted_dir_path(latest_micromamba.value())
+                            / latest_micromamba.value().str();
+
+    fs::rename(mamba_exe, mamba_exe_bkup);
+
+    try
+    {
+        if (on_win)
+        {
+            fs::copy_file(cache_path / "Library" / "bin" / "micromamba.exe",
```
Klaim, line 101: Not sure if this is completely relevant to this specific PR, but on my windows install, in my environments, I cannot find micromamba in `Library/bin`. Is my setup broken or is it a possibility? If it is, should that be handled here?

micromamba/src/update.cpp
wolfv: This was mostly to quickly hack things together and make sure that the functionality itself works (especially on Windows it's tricky because it's not easy to replace binaries that are in use).

micromamba/src/update.cpp
wolfv: If you check the `micromamba` package from conda forge you should find the binary file there though. It's the 'standard' layout for conda packages on Windows.

micromamba/src/update.cpp
wolfv: It would not because one is `=` and the other `>` but yeah, it could be slightly shorter although I think the gain is not that big :)

libmamba/src/core/pool.cpp
```diff
@@ -87,11 +88,25 @@ namespace mamba
...
         job.push(SOLVER_SOLVABLE_PROVIDES, matchspec);
         selection_solvables(m_pool, job, solvables);
+
+        if (sorted)
+        {
+            std::sort(solvables.begin(),
+                      solvables.end(),
+                      [this](Id a, Id b)
+                      {
+                          Solvable* sa;
```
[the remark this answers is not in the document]
wolfv: For the solvables, it's just a pointer to an existing solvable in the pool and the pool stays the owner:

```
static inline Solvable *pool_id2solvable(const Pool *pool, Id p)
{
  return pool->solvables + p;
}
```

For the entire function, the IDs are just 32-bit integers. They are only correct if used with the same `pool` that they came from, but the ownership is not an issue at this point.

micromamba/src/update.cpp
```diff
@@ -8,10 +8,133 @@
...
 
+#include "mamba/core/transaction.hpp"
+#include "mamba/core/context.hpp"
+#include "mamba/core/shell_init.hpp"
+#include "mamba/core/util_os.hpp"
+#include "version.hpp"
+#include <reproc++/run.hpp>
+
+
+extern "C"
```
wolfv, line 21: remove some includes here

AntoinePrv:
Should the documentation be updated to inform users of that feature?

wolfv:
Yes, for sure!

PR #2025, opened 2022-10-14
Test stubs

wolfv:


dholth:
I would do this with the black formatter and a text compare

wolfv:
We have the formatters only in the pre-commit environment so I chose this route as we need fewer dependencies.

AntoinePrv:
That could be a cool pre-commit hook to generate the stubs!

wolfv:
Indeed, but you do need the compiled `libmambapy.bindings` file for that...

PR #2029, opened 2022-10-17
implement jlap format for incremental repodata patches

wolfv:


libmamba/src/core/validate.cpp
```diff
@@ -194,6 +194,33 @@ namespace validate
         return ::mamba::hex_string(hash, MAMBA_MD5_SIZE_BYTES);
     }
 
+    std::string blake2sum(const fs::u8path& path)
+    {
+        size_t sum, n;
+        // blake2b_state S;
+        constexpr std::size_t BUFSIZE = 32768;
```
Klaim, line 201: Prefer `static constexpr` for these, see [there](https://stackoverflow.com/a/13867690) for details

libmamba/src/core/validate.cpp
```diff
@@ -194,6 +194,33 @@ namespace validate
...
+        std::vector<char> buffer(BUFSIZE);
+
+        constexpr std::size_t outbytes = 32;
```
Klaim, line 204: `static constexpr`

libmamba/include/mamba/core/subdirdata.hpp
```diff
@@ -32,6 +32,19 @@ namespace decompress
...
 
+    // Represents a repodata update "jlap" document
+    struct Jlap
+    {
+        std::string start_hash;
+        std::string end_hash;
+        nlohmann::json metadata;
+        std::vector<nlohmann::json> patches;
+
+        static tl::expected<Jlap, std::runtime_error> parse(const fs::u8path& path);
```
Klaim, line 43: I think the error type should be `mamba_error` and depending on which `mamba_error_code` are available, maybe another one need to be added, not sure.

libmamba/src/core/subdirdata.cpp
```diff
@@ -695,4 +695,109 @@ namespace mamba
...
+                    // std::cout << " " << line << std::endl;
+                }
+            }
+        }
+        return jlap_doc;
+    }
+
+    tl::expected<void, std::runtime_error> Jlap::apply_patches_to(const fs::u8path& path)
+    {
+        std::string current_hash = validate::blake2sum(path);
```
Klaim, line 761: `const`

libmamba/src/core/subdirdata.cpp
```diff
@@ -695,4 +695,109 @@ namespace mamba
...
+        if (current_hash == metadata["latest"])
```
Klaim, line 762: Note that this is expensive, if the key is not found there is a node allocated anyway here, but not being used.
Using `.find` followed by a comparison if found would avoid that.

libmamba/src/core/subdirdata.cpp
```diff
@@ -695,4 +695,109 @@ namespace mamba
...
+            {
+                return tl::unexpected(std::runtime_error("Hash not matching"));
+            }
+        }
+
+        {
+            TemporaryFile temp_file;
+            std::ofstream of = open_ofstream(temp_file.path());
+            of << repodata.dump(2) << "\n" << std::flush;
+            if (validate::blake2sum(temp_file.path()) != metadata["latest"])
```
Klaim, line 793: Same node allocation issue.

libmamba/src/core/subdirdata.cpp
```diff
@@ -695,4 +695,109 @@ namespace mamba
...
+            }
+            catch (...)
+            {
+                return tl::unexpected(std::runtime_error("Could not parse metadata in JLAP file"));
+                // std::cout << " " << line << std::endl;
+            }
+
+            // parse patches in between
+            auto it = lines.begin() + 1;
+            for (; it != lines.end() - 2; ++it)
```
Klaim, line 740: Prefer `std::prev(linex.end(), 2)`, same reason than `std::next`

libmamba/src/core/subdirdata.cpp
```diff
@@ -695,4 +695,109 @@ namespace mamba
...
+        std::vector<std::string> lines;
+        std::string line;
+        while (std::getline(jlap_file, line))
+        {
+            lines.push_back(line);
+        }
+
+        if (lines.size() >= 3)
+        {
+            jlap_doc.start_hash = lines[0];
```
Klaim, line 717: `lines.front()` and `lines.back()` in the following line would be clearer.

libmamba/src/core/subdirdata.cpp
```diff
@@ -695,4 +695,109 @@ namespace mamba
...
+        Jlap jlap_doc;
+        std::ifstream jlap_file = open_ifstream(path);
+        if (!jlap_file.is_open())
+        {
+            return tl::unexpected(std::runtime_error("No such file or directory"));
+        }
+
+        std::vector<std::string> lines;
+        std::string line;
+        while (std::getline(jlap_file, line))
```
Klaim, line 710: BTW we don't have already some utility function to do this? (read a file as lines into a vector)

wolfv:
I think the main difficulty (similar to zchunk) will be to have a proper "state-machine" for the different error cases. It is complicated by the fact that the idea of the `jlap` format is to be able to continue writing to the jlap file from a given offset.

- no repodata json at all locally -> download repodata.json
- repodata.json locally -> check if repodata.jlap exists on server
- jlap does not exist on server -> update json
- jlap does exist, but our local jlap / offset does not match server -> redownload entire jlap file
- jlap exists, but <from>-hash of repodata.json is not found in `jlap` file -> redownload repodata.json (& remove jlap file?)
- jlap exists, from-hash exists, but at some point the hashes diverge -> redownload repodata.json (incorrect jlap file)
- parse error of jlap file (e.g. missing required fields)

Maybe there are more cases but these are the ones that come to my mind right now. Probably we should have a dedicated error class for each case.

dhirschfeld:
TIL the `jlap` format from a post on discourse:
* https://conda.discourse.group/t/how-we-reduced-condas-index-fetch-bandwidth-by-99/257

It sounds very interesting so I'll make sure to follow progress here...

jonashaag:
What’s the status of this PR?

jjerphan:
Closing because of [discontinuation of the JLAP updates](https://conda-forge.zulipchat.com/#narrow/channel/457337-general/topic/Discontinue.20.2Ejlap.20on.20conda-forge.2C.20main.20channels.2E/with/596374425).

PR #2033, opened 2022-10-19
[MER][T31] join_trunc and string_util improvements

AntoinePrv:
``join_trunc`` for #1878 and other ``string_util`` improvements.

The diffs are not easy to follow, basically this PR do
- Rename `test_string_methods.cpp` > `test_util_string.cpp` to align with header name;
- Add a number of tests for previously untested functions (`strip`, `concat`, `starts_with`, `contains`, `join`, ...)
- Refactor `join`
  - for reusing it in `join_trunc` (new)
  - for calling `reserve` when possible.
- Add a new `join_trunc`, similar to `join` but truncates the middle of the string if too long.

### DIFF
--- a/libmamba/include/mamba/core/util_string.hpp
+++ b/libmamba/include/mamba/core/util_string.hpp
@@ -131,9 +131,10 @@ namespace mamba
         std::size_t size(const wchar_t* s);
         std::size_t size(const char c);
         template <class T>
-        inline std::size_t size(T const& s)
+        std::size_t size(T const& s)
         {
-            return s.size();
+            using std::size;
+            return size(s);
         }
     }
 
@@ -236,11 +237,17 @@ namespace mamba
         };
 
         auto const [show_head, show_tail] = show;
-        join_for_each_func(first, first + show_head, sep);
-        func(sep);
+        if (show_head > 0)
+        {
+            join_for_each_func(first, first + show_head, sep);
+            func(sep);
+        }
         func(etc);
-        func(sep);
-        join_for_each_func(last - show_tail, last, sep);
+        if (show_tail)
+        {
+            func(sep);
+            join_for_each_func(last - show_tail, last, sep);
+        }
         return func;
     }
 
@@ -279,13 +286,12 @@ namespace mamba
         return out;
     }
 
-
     void replace_all(std::string& data, const std::string& search, const std::string& replace);
 
     void replace_all(std::wstring& data, const std::wstring& search, const std::wstring& replace);
 
     template <typename T>
-    auto without_duplicates(std::vector<T> values)
+    std::vector<T> without_duplicates(std::vector<T> values)
     {
         const auto end_it = std::unique(values.begin(), values.end());
         values.erase(end_it, values.end());
@@ -297,7 +303,7 @@ namespace mamba
     std::string to_lower(const std::string_view& input);
 
     template <typename... Args>
-    inline std::string concat(const Args&... args)
+    std::string concat(const Args&... args)
     {
         std::string result;
         result.reserve((details::size(args) + ...));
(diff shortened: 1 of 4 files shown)


libmamba/tests/test_util_string.cpp
```diff
@@ -0,0 +1,194 @@
...
+        EXPECT_EQ(prefix_unicode, "/home/åéäáßðæœ©ðfßfáðß/123123123\n\nabcdefg\nxyz");
+    }
+
+    TEST(util_string, to_upper_lower)
+    {
+        std::string a = "ThisIsARandomTTTeeesssT";
+        EXPECT_EQ(to_upper(a), "THISISARANDOMTTTEEESSST");
+        // std::wstring b = "áäáœ©gþhëb®hüghœ©®xb";
+        // EXPECT_EQ(to_upper(b), "ÁÄÁŒ©GÞHËB®HÜGHŒ©®XB");
+        // EXPECT_EQ(to_lower(a), "thisisarandomttteeessst");
```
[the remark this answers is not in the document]
AntoinePrv: This looks like an addition but it's just the diff showing wrongly, it was already there.
There is a comment on `to_upper` stating it does not work with unicode.

libmamba/include/mamba/core/util_string.hpp
```diff
@@ -112,86 +114,146 @@ namespace mamba
...
+        };
+        template <class T>
+        struct has_reserve<T, std::void_t<decltype(std::declval<T>().reserve(std::size_t()))>>
+            : std::true_type
+        {
+        };
+        template <typename T>
+        inline constexpr bool has_reserve_v = has_reserve<T>::value;
+
+        std::size_t size(const char* s);
```
Klaim, line 130: Not sure why these are necessary, but you can move the definition here as inline, it's literally only calls to the standard library which will be inlined and using the cpp for that prevents the inlining.

libmamba/include/mamba/core/util_string.hpp
```diff
@@ -112,86 +114,146 @@ namespace mamba
...
+        std::size_t size(const wchar_t* s);
+        std::size_t size(const char c);
+        template <class T>
+        inline std::size_t size(T const& s)
```
Klaim, line 134: The `inline` here is unnecessary as any template function is implicitly `inline`.

By the way, why not just reuse `std::size` here?

libmamba/include/mamba/core/util_string.hpp
```diff
@@ -112,86 +114,146 @@ namespace mamba
...
+        auto out_joiner = [&](auto&& val) { joiner(out, std::forward<decltype(val)>(val)); };
+        join_for_each(container.begin(), container.end(), out_joiner, sep);
+        return out;
+    }
 
-        template <class T, class... Rest>
-        inline void concat_foreach(std::string& result, const T& rhs, const Rest&... rest)
+    // TODO(C++20) Take an input range and return a range
+    template <typename InputIt, typename UnaryFunction, typename Value>
+    UnaryFunction join_trunc_for_each(InputIt first,
```
Klaim, line 176: I have trouble understanding this function and the following `join_trunc`, could you add some comments/documentation?

libmamba/include/mamba/core/util_string.hpp
AntoinePrv: It's for generic code below (*e.g.* in `concat` to call it over a parameter pack that may contain, `std::string`s, `char`s, `const char*`...).
It's a modification of what was already there.

libmamba/include/mamba/core/util_string.hpp
AntoinePrv: Similarly it's for the generic code. This function could use `std::size` but using it directly in the generic code would require some sort of ADL.

libmamba/include/mamba/core/util_string.hpp
AntoinePrv: I'll add some doc though.

Klaim (approved):
LGTM 👍🏽

PR #2040, opened 2022-10-24
Fix conda-forge required test conflict

AntoinePrv:
This test is failing (or will fail) since the conflict has disappeared.
This is a more specific pin.


jonashaag:
Do you know where this build problem is coming from? spdlog update it seems

AntoinePrv:
Yes, it's on all (recent) PRs.
Context https://github.com/mamba-org/mamba/issues/2044

PR #2047, opened 2022-10-26
Pin spdlog

jonashaag:
Temporarily work around broken builds with new spdlog build


jonashaag:
@AntoinePrv

AntoinePrv:
@jonashaag #2040 Was merged so this should pass if you rebase on `main`.

jonashaag:
Shall we wait for #2048 instead?

jonashaag:
@AntoinePrv i will leave the decision to use this or not to you since you are working on a proper fix I guess?

PR #2048, opened 2022-10-26
Direct dependency on fmt

### LINKED ISSUE #1694: Allow spdlog to be used with external fmt lib

### LINKED ISSUE #2044: spdlog and CMake
Dear mamba developers,

we are attempting to unvendor fmt from the conda-forge spdlog package. See here for context:

https://github.com/conda-forge/spdlog-feedstock/pull/50

Shortly after the merge of this PR, it was reported that building mamba against the updated spdlog package is now broken:

https://github.com/conda-forge/spdlog-feedstock/issues/52

I spent some time looking at the mamba build system and I am wondering why you have code like this:

```cmake
find_file(SPDLOG_TWEAK_H NAMES spdlog/tweakme.h)
if(SPDLOG_TWEAK_H)
    file(READ ${SPDLOG_TWEAK_H} SPDLOG_TWEAK_H_STRING)
    if("${SPDLOG_TWEAK_H_STRING}" MATCHES "\n *#define *SPDLOG_FMT_EXTERNAL *")
        find_package(fmt REQUIRED)
        target_link_libraries(bindings PRIVATE fmt::fmt)
    endif()
endif()
```

rather than just e.g.

```cmake
find_package(spdlog CONFIG)
target_link_libraries(mamba spdlog::spdlog)
```

Linking to the imported target ``spdlog::spdlog`` already takes care of setting up all the appropriate definitions (such as ``SPDLOG_FMT_EXTERNAL`` in this specific case).

I can open a PR if you want, but I first wanted to check with the developers about why things are done in this way in the current build system.

(NOTE: this may be tangentially related to #1694)

AntoinePrv:
Close #2044
Close #1694
Close #2047 (or revert #2047)
Close #2050


Klaim (approved):
LGTM 👍🏽 

I wonder however if forcing the header-only target of fmt doesnt mean the users of libmamba (not mamba or micromamba) will end up having linking issues if they use the "normal" version of fmt.

wolfv:
Awesome!

PR #2049, opened 2022-10-26
Add micromamba nightly releases

### LINKED ISSUE #1929: Change version name for CI builds of `micromamba`
It would be nice for the `micromamba` CI builds to have a custom version name instead of `0.25.1`.

Some suggestions for names:
- `nightly-yyyy-mm-dd`
- `nightly-yyyy-mm-dd-hh:mm`
- `nightly-sha-<commit-sha>`
- `nightly-<commit-run-id>` (the number in `https://github.com/mamba-org/mamba/actions/runs/<commit-run-id>`)

Other suggestions are welcome :)

pavelzw:
For each `main` commit where the `static-build.yml` workflow succeeds, the workflow will afterwards create a new tag on a separate repository (for example `mamba-org/micromamba-nightlies`) where the built artifacts get uploaded. This fixes the issue of static build artifacts being deleted after 90 days.

Since every release needs to point to a specific commit sha in this new repository, we need to pin each release to a specific commit. We have two options here:

1. For each run, create a new commit that modifies the readme in the nightlies repo (for example displaying the current SHA and commit details) and create a release on this commit. 
2. Create a tag every time on `HEAD` leading to hundreds of tags on `HEAD`.

Still to do:

- [ ] add counter to version number to show (for example make releases like this: `v0.26.0-<counter-increment>-<first-letters-of-sha>`)
- [ ] Add new version string to `micromamba --version` (this closes #1929)

Proof of concept can be seen in [`pavelzw/micromamba-releases`](https://github.com/pavelzw/micromamba-nightlies/releases) with corresponding workflow runs in [`pavelzw/mamba #7`](https://github.com/pavelzw/mamba/pull/7).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -14,6 +14,7 @@ concurrency:
 
 jobs:
   libmamba_static:
+    if: false
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
@@ -267,6 +268,7 @@ jobs:
           rm -rf $(micromamba info --json | jq -r '."env location"')/envs
 
   libmamba_static_win:
+    if: false
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
--- a/.github/workflows/static_build.yml
+++ b/.github/workflows/static_build.yml
@@ -178,6 +178,7 @@ jobs:
 
   create_nightly_release:
     name: "Create nightly release"
+    runs-on: ubuntu-latest
 #    depends-on: [conda-build, micromamba-full-static-win]
     steps:
       - run: echo "Hello, world!" >> artifact.txt
--- a/.github/workflows/static_build.yml
+++ b/.github/workflows/static_build.yml
@@ -195,7 +195,7 @@ jobs:
       - name: package-version-to-git-tag
         uses: pkgdeps/git-tag-action@v2
         with:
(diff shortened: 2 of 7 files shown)


pavelzw:
One easy-to-implement counter would be the actions run id, i.e., the thing that is in the url in `https://github.com/mamba-org/mamba/actions/runs/<run-id>`.

jonashaag:
Do we actually need a separate commit/tag per release? Is that a GH requirement?

I wonder if we should not include counter and version in the release name. We should optimize for ease of installation of a specific SHA (or date if we decide to do actual nightlies)

pavelzw:
> Do we actually need a separate commit/tag per release? Is that a GH requirement?

Yes, each release needs a separate tag (we can use the same commit with multiple tags, though).

<details>
<summary>Details</summary>

image


</details>

pavelzw:
> I wonder if we should not include counter and version in the release name.

The problem then is that it is not obvious which version is newer if comparing newer versions...

jonashaag:
Is it?

pavelzw:
Not completely, I just took a break from it 😅
The release on a custom channel part works but I still wanted to add the commit hash to `micromamba --version`.

wolfv:
Oh sorry

PR #2050, opened 2022-10-26
fix pytest-xprocess issue

wolfv:
Unfortunately, a dropped dependency on `py` forces us to install it explicitly: https://github.com/pytest-dev/pytest-xprocess/issues/110


jonashaag:
I prefer the pin as in #2048 and #2047

wolfv:
pytest dropped the dependency on `py` in the latest release and pytest-xprocess relied on it implicitly. but anyways, anything that fixes thsi si fine wtih me :)

PR #2052, opened 2022-10-26
Add Conda nightly test

jonashaag:
This adds a nightly test with latest Conda and creates an issue if the test suite fails.

This is a way to detect incompatibilities with the upcoming Conda release.

There is one weird problem with boa that I'm working around with. Boa seems to be ignored `--no-test`.

### DIFF
--- a/.github/workflows/bot_issue_template.md
+++ b/.github/workflows/bot_issue_template.md
@@ -1,0 +1,4 @@
+---
+title: Bot failure
+---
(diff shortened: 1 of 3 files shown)


jonashaag:
@AntoinePrv @wolfv

AntoinePrv (approved):
Not sure if this will create unreasonable amounts of issue if the CI is broken, but we can try it out.

PR #2058, opened 2022-10-27
Upload conda-bld artifacts for failed builds

pavelzw:


pavelzw:
Maybe decreasing the retention days would make sense for these large build folders... How about 6 days @wolfv or is this not enough?

pavelzw:
Then this is ready to merge/review 👍🏻

wolfv (approved):

PR #2064, opened 2022-10-28
[MER][T51|T52|T53|T54|T55] Tree error message

AntoinePrv:
Partial implementation of #1878.

### DIFF
--- a/libmamba/include/mamba/core/util_string.hpp
+++ b/libmamba/include/mamba/core/util_string.hpp
@@ -218,7 +218,17 @@ namespace mamba
                                       std::size_t threshold = 5,
                                       std::pair<std::size_t, std::size_t> show = { 2, 1 })
     {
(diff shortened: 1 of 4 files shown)


libmamba/include/mamba/core/util_string.hpp
```diff
@@ -218,7 +218,17 @@ namespace mamba
                                       std::size_t threshold = 5,
                                       std::pair<std::size_t, std::size_t> show = { 2, 1 })
     {
-        if ((last - first) <= threshold)
+        auto unsigned_cast = [](auto n)
```
Klaim, line 221: Interesting, you could move that in utilities. (no need to do it now)

Klaim (approved):
LGTM, I am not knowledgable of the domain-specific parts though.

SylvainCorlay:
Let's move forward and iterate on this!

PR #2076, opened 2022-11-01
fixes for ssl init and static build

wolfv:


Klaim (approved):
Looks good to me, not sure if the crash on CI is even related?

PR #2079, opened 2022-11-02
Fix export/set/unset environment variables for fish shell when activating a new env

JafarAbdi:
I was running into an issue where activating an environment isn't working for the fish shell and it was messing up the `PATH` environment variable 

```
PATH=/home/jafar/micromamba/bin:/home/jafar/micromamba/condabin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:set:-gx:CONDA_PREFIX:/home/jafar/micromamba:set:-gx:CONDA_SHLVL:1:set:-gx:CONDA_DEFAULT_ENV:base:set:-gx:CONDA_PROMPT_MODIFIER:(base)
```
as you can see from the variable above, it was being interpreted as appending to the list rather than a command, I fixed it by adding a semicolon after each command

The second commit renamed `MAMBA_PROMPT_MODIFIER` to `CONDA_PROMPT_MODIFIER`, grepping the repo I don't see other usages for `MAMBA_PROMPT_MODIFIER` and looking at the codebase `CONDA_PROMPT_MODIFIER` is the variable being set

Ubuntu version: `20.04`
Fish version `3.5.1`


jonashaag:
Thanks for the patch! I don't understand how this fixes the problem. Shouldn't `\n` suffice to terminate a fish command?

JafarAbdi:
> Shouldn't \n suffice to terminate a fish command?

That's what I thought, too, but it doesn't seem to be the case. I'll work on providing an easily reproducible example

wolfv:
I guess it wouldn't hurt either, right @jonashaag ?

jonashaag:
Yes but I wonder what's the underlying problem here. I'm worried that we're not actually fixing the problem.

JafarAbdi:
Hmm, I can't reproduce it using a fish shell directly, maybe it's related to how CLI11 is executing those commands in https://github.com/mamba-org/mamba/blob/main/libmamba/src/api/shell.cpp#L150

libmamba/src/core/activation.cpp
```diff
@@ -1204,17 +1204,17 @@ namespace mamba
 
         for (const std::string& uvar : env_transform.unset_vars)
         {
-            out << "set -e " << uvar << "\n";
+            out << "set -e " << uvar << ";\n";
```
JafarAbdi, line 1207: Should I remove the newline? It's redundant here!
```suggestion
            out << "set -e " << uvar << ";";
```

wolfv: It probably looks cleaner with a newline, no?

jonashaag:
Thanks for your contribution but I'm against merging this without understanding the problem, sorry!

jonashaag:
Cherry-picked changes in https://github.com/mamba-org/mamba/pull/2101

PR #2080, opened 2022-11-02
[MER] Add experimental flag for error messages

AntoinePrv:
Connects #1878 to the rest of mamba

Add a flag `experimental_sat_error_message` to `condarc`.

Right now it will look like this:
```
error    libmamba Could not solve for environment specs
================== Experimental satisfiability error messages ==================
You are seeing this because you set `experimental_sat_error_message: true`
Use the following issue to share feedback on this experimental feature
   https://github.com/mamba-org/mamba/issues/2078

============================ Legacy messages (old) =============================
  - nothing provides __cuda needed by pytorch-1.11.0-cuda102py310hdf4a2db_202

The environment can't be solved, aborting the operation
========================= Experimental messages (new) ==========================
Coming soon
================================================================================

critical libmamba UnsatisfiableError
```

I'm not sure about the `error` and `critical` message from the logger. In my head, error from the logger are bugs that should be reported, this is more of an expected output that we should handle gracefully. This is a more general discussion all over mamba though...


libmamba/include/mamba/core/solver.hpp
```diff
@@ -8,6 +8,7 @@
 #define MAMBA_CORE_SOLVER_HPP
 
 #include <string>
+#include <ostream>
```
Klaim, line 11: I think `<iosfwd>` would work here, to get only the declarations/forwardings - then you can add `<ostream>` in the cpp implementing the new functions using `std::ostream`

libmamba/src/core/solver.cpp
```diff
@@ -598,6 +602,36 @@ namespace mamba
         return problems.str();
     }
 
+    std::ostream& MSolver::explain_problems(std::ostream& out) const
+    {
+        bool experimental = Context::instance().experimental_sat_error_message;
+        if (experimental)
+        {
+            fmt::print(out, "{:=^80}\n", " Experimental satisfiability error messages ");
```
Klaim, line 610: Did you check there there will not be any issues of mixing spdlog's output with `fmt::print`? they use the same output?

AntoinePrv: I think they flush at different moments so the outputs are not intertwined.
Here the `out` is a parameter so this is managed outside this function.

libmamba/src/core/solver.cpp
```diff
@@ -598,6 +602,36 @@ namespace mamba
...
+            out << "Coming soon\n";
+            fmt::print(out, "{:=^80}\n", "");
+        }
+        return out;
+    }
+
+    std::string MSolver::explain_problems() const
+    {
+        std::stringstream ss;
+        explain_problems(ss);
```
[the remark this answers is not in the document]
AntoinePrv: Not sure I see how I could call `explain_problems(std::ostream&)` with `fmt::format`. Do you have a more detailed suggestion?

SylvainCorlay:
This apparently needs a rebase.

libmambapy/libmambapy/__init__.pyi
```diff
@@ -962,13 +962,15 @@ class Solver:
...
+    def explain_problems(self) -> str: ...
     def is_solved(self) -> bool: ...
+    def must_solve(self) -> None: ...
     def problems_to_str(self) -> str: ...
     def set_flags(self, arg0: typing.List[typing.Tuple[int, int]]) -> None: ...
     def set_postsolve_flags(
         self, arg0: typing.List[typing.Tuple[int, int]]
     ) -> None: ...
-    def solve(self) -> bool: ...
+    def try_solve(self) -> bool: ...
```
wolfv, line 973: this is going to break all downstream projects. Maybe add a deprecation warning and a shim that forwards to either of the two functions?

libmamba/include/mamba/core/solver.hpp
```diff
@@ -67,17 +68,21 @@ namespace mamba
...
-        std::string problems_to_str() const;
-        std::vector<std::string> all_problems() const;
-        std::vector<MSolverProblem> all_problems_structured() const;
-        std::string all_problems_to_str() const;
-
-        const std::vector<MatchSpec>& install_specs() const;
-        const std::vector<MatchSpec>& remove_specs() const;
-        const std::vector<MatchSpec>& neuter_specs() const;
-        const std::vector<MatchSpec>& pinned_specs() const;
+        [[nodiscard]] bool try_solve();
```
wolfv, line 71: how about we change the interface to return a `tl::expected` instead? Then we can keep a single function that we can test for solver problems?

libmamba/include/mamba/core/solver.hpp
```diff
@@ -68,20 +68,21 @@ namespace mamba
...
         void set_postsolve_flags(const std::vector<std::pair<int, int>>& flags);
-        [[nodiscard]] bool is_solved() const;
         [[nodiscard]] bool try_solve();
         void must_solve();
-        std::string problems_to_str() const;
-        std::vector<std::string> all_problems() const;
-        std::vector<MSolverProblem> all_problems_structured() const;
-        std::string all_problems_to_str() const;
+
+        [[nodiscard]] bool is_solved() const;
```
wolfv, line 74: I am not sure if the `nodiscard` really does something useful here besides adding noise to the header. I guess the only place where I think it makes a lot of sense is the `solve()` method (because we want to enforce users to check that it was actually solved.

Or am I misunderstanding `nodiscard`?

libmambapy/libmambapy/__init__.pyi
AntoinePrv: That's a good point! I did not do it at first because `Solver` was in `include/core`. Do we maintain API there as well?

libmamba/include/mamba/core/solver.hpp
AntoinePrv: I don't know. When I tried to use them I found they don't compose nicely. Here there would also not be any outcome, so does it make sense to have a `tl::expected<void, Error>`?

I also introduce `must_solve` that throws an exception because some code used `solve` without checking the outcome.
Since I would not know what to do in those cases, the exception seems like the natural first step.

libmamba/include/mamba/core/solver.hpp
AntoinePrv: I have the same understanding as you. On `solve` it is critical (as it happens, it is currently *not* checked everywhere).

It is however correct to use it here, because there is no point in calling `is_solved` without checking the result because this function has no side effect. Some best-practices suggest to use it liberally in those case, it's only more bugs you can catch (*e.g.* if the `const` function name is ambiguous). There's even a `clang-tidy` fix that would apply it everywhere it can.

libmambapy/libmambapy/__init__.pyi
AntoinePrv: Conda seems to not use it directly
https://cs.github.com/?scopeName=All+repos&scope=&q=org%3Aconda+.solve%28%29

wolfv: This function is 100% used in boa, the conda-forge bot and the conda-libmamba-solver: https://github.com/conda-incubator/conda-libmamba-solver/blob/96e907cfbde41704c3fb50e59995d33c2d10359d/conda_libmamba_solver/solver.py#L321-L322

libmamba/src/core/pool.cpp
```diff
@@ -21,119 +19,146 @@ extern "C"
...
-        return m_pool;
+        return pool();
     }
 
     std::vector<Id> MPool::select_solvables(Id matchspec, bool sorted) const
     {
         MQueue job, solvables;
         job.push(SOLVER_SOLVABLE_PROVIDES, matchspec);
-        selection_solvables(m_pool, job, solvables);
+        selection_solvables(const_cast<Pool*>(pool()), job, solvables);
```
AntoinePrv, line 120: The method was const (the `Pool` pointer was a `Pool* const`) but the libsolv function is not `const` (but libsolv does not seem fully const correct).

wolfv:
I think I am happy with the changes here! The only thing that I would've liked to look into more is the API changes on the `Solver` class, but we can also look at that in a follow up PR.

AntoinePrv:
> The only thing that I would've liked to look into more is the API changes on the Solver class, but we can also look at that in a follow up PR.

@wolfv sure, what would you like to look into? If it is the `try_solve`, I kept `solve` with a deprecation in Python, but removed it in C++. I can also do a deprecation if you'd like.

libmambapy/libmambapy/__init__.pyi
AntoinePrv: Added it back in Python with a deprecation.

libmamba/include/mamba/core/solver.hpp
```diff
@@ -86,6 +89,8 @@ namespace mamba
         bool force_reinstall = false;
 
     private:
+        static void delete_libsolve_solver(Solver* solver);
```
Klaim, line 92: Looks like this should be a free function in the cpp? The unique-ptr only need `void(Solver*)` type.

libmamba/src/api/install.cpp
```diff
@@ -567,8 +565,8 @@ namespace mamba
             fs::u8path pkgs_dirs(Context::instance().root_prefix / "pkgs");
             MultiPackageCache pkg_caches({ pkgs_dirs });
             prefix_data.add_packages(get_virtual_packages());
-            MRepo::create(
-                pool, prefix_data);  // Potentially re-alloc (moves in memory) Solvables in the pool
+            // Potentially re-alloc (moves in memory) Solvables in the pool
```
Klaim, line 568: I prefer that the comment be in the line it is commenting, mainly because future changes might insert lines here without seeing the comment only applies to that line.

libmamba/src/core/solver.cpp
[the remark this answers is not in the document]
Klaim: I was thinking about something like `return fmt::format(explain_problems(ss));` but that only replaces the return type. I suspect it's faster than the `.str()` but it's not worth changing at this point.

libmamba/include/mamba/core/pool.hpp
```diff
@@ -51,9 +38,24 @@ namespace mamba
...
-        std::pair<spdlog::logger*, std::string> m_debug_logger;
-        Pool* m_pool;
-        std::list<MRepo> m_repo_list;
+        struct MPoolData;
+
+        Pool* pool();
+        Pool const* pool() const;
+
+        /**
+         * Make MPool behave like a shared_ptr (with move and copy).
```
Klaim, line 47: I think the first part of this should be also in a documentation comment at the top of this type.

libmamba/include/mamba/core/pool.hpp
```diff
@@ -51,9 +38,24 @@ namespace mamba
...
+         * The pool is passed to the ``MSolver`` for its lifetime but the pool can legitimely
+         * be reused with different solvers (in fact it is in ``conda-forge``'s ``regro-bot``
+         * and ``boa``).
+         * An alternative considered was to make ``MPool`` a move-only type, moving it in and out
+         * of ``MSolver`` (effectively borrowing).
+         * It was decided to make ``MPool`` share ressources for the following reasons.
+         *    - Rvalue semantics would be unexpected in Python (and breaking ``conda-forge``);
+         *    - Facilitate (potential) future investigation of parallel solves.
+         */
+        std::shared_ptr<MPoolData> m_data;
```
Klaim, line 58: Didnt  we conclude that `MPool` would first have a move-only behavior (and implemented with `unique_ptr`), then later would be switched to shared when we will need it? You went directly with shared because you found cases where it was necessary already?

libmamba/include/mamba/core/solver.hpp
Klaim: I agree that `[[nodiscard]]` seems useful here, at least to catch cases where it's called but not checked.

libmamba/src/core/pool.cpp
```diff
@@ -21,119 +19,146 @@ extern "C"
...
+        pool()->debugmask |= SOLV_DEBUG_TO_STDERR;
         if (Context::instance().verbosity > 2)
         {
-            pool_setdebuglevel(m_pool, Context::instance().verbosity - 1);
+            pool_setdebuglevel(pool(), Context::instance().verbosity - 1);
             auto logger = spdlog::get("libsolv");
-            m_debug_logger.first = logger.get();
-            pool_setdebugcallback(m_pool, &libsolv_debug_callback, &m_debug_logger);
+            m_data->debug_logger.first = logger.get();
+            pool_setdebugcallback(pool(), &libsolv_debug_callback, &(m_data->debug_logger));
```
Klaim, line 97: I don't think the parenthesis around `m_data->debug_logger` are necessary?

libmamba/src/core/solver.cpp
```diff
@@ -147,53 +152,73 @@ namespace mamba
...
-        , m_solver(nullptr)
-        , m_pool(pool)
-    {
-        queue_init(&m_jobs);
-        pool_createwhatprovides(pool);
+        MSolverProblem make_solver_problem(
+            ::Solver* solver, SolverRuleinfo type, Id source_id, Id target_id, Id dep_id)
+        {
+            return {
+                /* type= */ type,
```
Klaim, line 189: When we have these comments for naming fields, I like to use the C++20 syntax so that we will just uncomment and it will work, for example here: `/* .type = */ type,` (add a dot)

Klaim (approved):
Ok as is, some nitpicking below.

libmamba/include/mamba/core/pool.hpp
AntoinePrv: No, for the reasons mentioned here we later decided to go that way.

PR #2085, opened 2022-11-03
Move to fmt::terminal_color and other output IO improvements

### LINKED ISSUE #1906: Unnecessary termcolor dependency (?)
We use [Termcolor](https://github.com/ikalnytskyi/termcolor) for getting a couple colors, but it looks like we could get them from [fmt::color](https://fmt.dev/latest/api.html#color-api), which we also use.

AntoinePrv:
Close #1906

- Reimplement termcolor::is_atty
- Remove termcolor for fmt
- Switch to fmt::terminal_color

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -31,16 +31,16 @@ repos:
     hooks:
         -   id: isort
             exclude: tests/data
--   repo: https://gitlab.com/pycqa/flake8
-    rev: 3.9.2
+-   repo: https://github.com/PyCQA/flake8
+    rev: 5.0.4
     hooks:
     -   id: flake8
         language_version: python3
         additional_dependencies:
-            - flake8-typing-imports==1.12.0
-            - flake8-builtins==1.5.3
-            - flake8-bugbear==22.1.11
-            - flake8-isort==4.1.1
+            - flake8-typing-imports==1.14.0
+            - flake8-builtins==2.0.1
+            - flake8-bugbear==22.10.27
+            - flake8-isort==5.0.0
 -   repo: https://github.com/pre-commit/mirrors-clang-format
     rev: v14.0.6
     hooks:
--- a/docs/source/installation.rst
+++ b/docs/source/installation.rst
@@ -92,7 +92,7 @@ Linux and macOS
 
 Download and unzip the executable (from the official conda-forge package):
 
-Ensure that basic utilities are installed. We need ``wget`` and ``tar`` with support for ``bzip2``.
+Ensure that basic utilities are installed. We need ``curl`` and ``tar`` with support for ``bzip2``.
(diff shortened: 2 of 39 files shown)


Klaim:
@AntoinePrv As the arguments are paths, I think it's better to let `fs::path` handle the string type internally (as it should be). It takes `std::wstring` and assumes it's the encoding of the system (whatever it is), keeps the data as the system needs it but also allows us to read the same data as UTF-8.

So maybe:
```
void set_autorun_registry_key(const fs::path& reg_path, const fs::path& value)
    auto out = Console::stream(); // not completely sure but I believe this requires utf-8 encoding
    fmt::print(
        out,
        "Setting cmd.exe AUTORUN to: {}\n",
         fmt::styled(value.string(), fmt::fg(fmt::terminal_color::green)) // value.string() is guaranteed to be utf-8
    );
}
```

or if you need to keep that function's interface:
```
void set_autorun_registry_key(const std::wstring& reg_path, const std::wstring& value)
    auto out = Console::stream(); // not completely sure but I believe this requires utf-8 encoding
    fmt::print(
        out,
        "Setting cmd.exe AUTORUN to: {}\n",
         fmt::styled(fs::path(value).string(), fmt::fg(fmt::terminal_color::green)) // `fs::path::string()` is guaranteed to return utf-8
    );
}
```
(is it normal that `reg_path` is not used here?)

AntoinePrv:
Here is what is happening:
- With `termcolor` ANSI color codes did not show up in Python's subprocess output, but with `fmt` they do, in fact they are never disabled (`micromamba list > file.txt` will now put escape codes in `file.txt`).
- The fact that the error shows up only on Windows is because it is the only place where we have a test for the content of the of colored string, but results could be duplicated on Windows.
- Before this PR, we were using `fmt` for progress bar (which should be disable if output is not a TTY) and some output of `micromamba run` on Windows.

The way forward would be to define a semantic color palette in the Context, which would be set to nothing if output is not a TTY (for which we already have a function).

AntoinePrv:
This check is failing on Windows.
https://github.com/mamba-org/mamba/blob/60413e955d37661f3715b4fd4b03f33816e206ac/libmamba/tests/test_progress_bar.cpp#L321

I believe that is it broken when checking it for the progress bar
https://github.com/mamba-org/mamba/blob/60413e955d37661f3715b4fd4b03f33816e206ac/libmamba/src/core/progress_bar_impl.cpp#L395
In this case `m_value` contains ANSI escape codes, as well as multi-byte character encoding, which are the reason for `m_value.size()` being so big. I see nothing in code indicating that this is taken into account for checking overflow.
If we disable colors (which is happening now with this PR during test since output is not a TTY) *and* we use ASCII bar (which is happening on Windows) then I can reproduce the failure.

I'd recommend removing this check with a comment, rather than forcing colors in CI,  since the latter only hides the problem.

wolfv (approved):

PR #2091, opened 2022-11-07
Fix libmamba CMake config file after dependency change

l2dy:
Direct dependency on fmt and spdlog was added in commit b8d7ec97a4324a53e2ed6ca8cff39a349b99c4bd.


wolfv:
Thanks!

AntoinePrv (approved):

PR #2092, opened 2022-11-07
Update pre-commit-config

AntoinePrv:


jonashaag:
Shall we also add Dependabot?

AntoinePrv:
That would be nice. Does dependabot work for pre-commit or conda environments?

jonashaag:
None of them 🙈 But it works for GitHub Actions workflows and maybe some other things in the repo, not sure.

jonashaag (approved):

Issue #2089, opened 2022-11-06
Do not show mamba ASCII art banner when `-q` if `-q` is passed


corneliusroemer:
It surprised me that you show the mamba ASCII art banner even when I ask for quiet output (`-q`).

Not quite sure what is quiet about the banner. I love mamba and have been advocating its use everywhere but this is a bit weird.

I know you have the `--no-banner`  flag, but that's an extra flag to add, and a lot of characters.

Quiet should mean quiet and not quiet plus verbose banner.

<img width="1223" alt="image" src="https://user-images.githubusercontent.com/25161793/200194187-04e8664b-40ca-4d74-af66-0b9602c2e65a.png">

And yes, I know one can set the `MAMBA_NO_BANNER=1` environment variable. But still, that doesn't justify showing verbose banner when someone asks for quiet output.

jonashaag:
Agreed! Should be simple to fix.

jonashaag:
@corneliusroemer would you be up for contributing a fix? Should be fairly easy, the relevant code for Mamba is here: https://github.com/mamba-org/mamba/blob/424a18e3bbc889e6f3de1461b9862caafcc89f46/mamba/mamba/mamba.py#L877

You could simply try the changes in your existing Mamba install:

```
vim $(python -c 'import mamba.mamba as m; print(m.__file__)')
```

corneliusroemer:
Here we go @jonashaag thanks for the pointer.

Tested on my local installation and it worked. Hope it passes more thorough tests but the build setup seems somewhat complicated.

https://github.com/mamba-org/mamba/pull/2097

PR #2097, opened 2022-11-07
fix: Don't print banner in quiet mode

corneliusroemer:
Resolves https://github.com/mamba-org/mamba/issues/2089

- fix: do not print banner in quiet mode
- typo(contributing.md): everyones -> everyone's


jonashaag (approved):

jonashaag:
Thanks!

Do we need a fix for Micromamba as well?

corneliusroemer:
I think Micromamba already does it the "right" way - just tested and it didn't show when I used `-q`

PR #2098, opened 2022-11-07
Docs: repoquery depends requires installed or channel (#2088)

### LINKED ISSUE #2088: mamba repoquery depends does not work without specifying a channel
https://mamba.readthedocs.io/en/latest/user_guide/mamba.html#repoquery gives a examples
`mamba repoquery depends xtensor` and `mamba repoquery depends -t xtensor`

When run these only give
```
Executing the query xtensor
No entries matching "xtensor" found
```
unless a channel is explicitly given as a argument, e.g.
`mamba repoquery depends --channel conda-forge -t xtensor`

`mamba repoquery search xtensor` works without a channel.

I think `depends` ought to work like `search`, as the docs suggest.

```
# mamba info

                  __    __    __    __
                 /  \  /  \  /  \  /  \
                /    \/    \/    \/    \
███████████████/  /██/  /██/  /██/  /████████████████████████
              /  / \   / \   / \   / \  \____
             /  /   \_/   \_/   \_/   \    o \__,
            / _/                       \_____/  `
            |/
        ███╗   ███╗ █████╗ ███╗   ███╗██████╗  █████╗
        ████╗ ████║██╔══██╗████╗ ████║██╔══██╗██╔══██╗
        ██╔████╔██║███████║██╔████╔██║██████╔╝███████║
        ██║╚██╔╝██║██╔══██║██║╚██╔╝██║██╔══██╗██╔══██║
        ██║ ╚═╝ ██║██║  ██║██║ ╚═╝ ██║██████╔╝██║  ██║
        ╚═╝     ╚═╝╚═╝  ╚═╝╚═╝     ╚═╝╚═════╝ ╚═╝  ╚═╝

        mamba (0.27.0) supported by @QuantStack

        GitHub:  https://github.com/mamba-org/mamba
        Twitter: https://twitter.com/QuantStack

█████████████████████████████████████████████████████████████


     active environment : base
    active env location : /mambaforge
            shell level : 1
       user config file : /root/.condarc
 populated config files : /mambaforge/.condarc
          conda version : 22.9.0
    conda-build version : not installed
         python version : 3.10.6.final.0
       virtual packages : __linux=5.15.0=0
                          __glibc=2.31=0
                          __unix=0=0
                          __archspec=1=x86_64
       base environment : /mambaforge  (writable)
      conda av data dir : /mambaforge/etc/conda
  conda av metadata url : None
           channel URLs : https://conda.anaconda.org/conda-forge/linux-64
          package cache : /mambaforge/pkgs
                          /root/.conda/pkgs
       envs directories : /mambaforge/envs
                          /root/.conda/envs
               platform : linux-64
             user-agent : conda/22.9.0 requests/2.28.1 CPython/3.10.6 Linux/5.15.0-52-generic ubuntu/20.04.1 glibc/2.31
                UID:GID : 0:0
             netrc file : None
           offline mode : False
```

I see this on ubuntu 20.04 installed from Mambaforge, and also in a fresh ubuntu docker environment.

samtygier:
The command `mamba repoquery depends` requires the package to already be installed in the environement. Otherwise the channel must be specified.

Closes #2088


hadim:
In my test, even if the channel is specified it does not work:

```bash
docker run -it --rm mambaorg/micromamba:1.0.0 bash -c "micromamba repoquery depends -c conda-forge xtensor"
```

It seems like the package specified must always be installed in the environment. If it's excepted then that should be reflected in the doc more explicitly no?

samtygier:
I am not sure if the difference between mamba and micromamba is expected, or should be considered a micromamba bug?

hadim:
My preference would be to stick to what mamba does already: not requiring the package to be installed. Maybe both behavior should be supported and one of them behind a boolean flag? I can see usecases for both.

Hind-M:
Thanks for submitting this PR!
`micromamba`'s behavior is indeed a bit different. I will open an issue for that.

JohanMabille (approved):

PR #2104, opened 2022-11-08
Enable list packages installed with pip in micromamba

### LINKED ISSUE #2059: `micromamba list` does not show packages installed with pip
Title says it all, couldn't find anything detailing whether this is a known limitation of micromamba at the present time, or if something is wrong with my environment. Cheers.

Hind-M:
Fix #2059


wolfv:
Awesome!

jonashaag:
Maybe naive question, why don't we just use `pip freeze` or something like that? Is the site-packages method guaranteed to work?

Hind-M:
> Maybe naive question, why don't we just use `pip freeze` or something like that? Is the site-packages method guaranteed to work?

That could be a way too I guess (but then we should be able to get the output of the launched command, is there another method allowing that since `reproc::run` doesn't?)... Using site-packages method was inspired from the way they do it in conda but less complicated. I haven't fully tested it yet because I was trying to activate a micromamba env within a conda env which doesn't seem to work... A bug or maybe I missed something?

Hind-M:
Hmm, actually I think that the way I did it won't do because I considered that `pip install pkg` passes necessarily through install command, which is only the case when we do install from a yaml file and `pip` is mentioned there... So I guess we should use a different method (maybe the `pip freeze` one if, as I said, there is a way to get the output)...
I kind of understand now the complicated way of conda...

zacharyburnett:
does this also cover `micromamba env export`?

Hind-M:
This PR is still a WIP, but when done the `env export` should be pretty straightforward I think.

zacharyburnett:
```
>       assert expected_link_packages.issubset(link_packages)
E       AssertionError: assert False
E        +  where False = <built-in method issubset of set object at 0x10a0cb300>({'openssl', 'pip', 'python', 'readline', 'setuptools', 'sqlite', ...})
E        +    where <built-in method issubset of set object at 0x10a0cb300> = {'python', 'python_abi'}.issubset
```
is this a MacOS issue?

Hind-M:
> ```
> >       assert expected_link_packages.issubset(link_packages)
> E       AssertionError: assert False
> E        +  where False = <built-in method issubset of set object at 0x10a0cb300>({'openssl', 'pip', 'python', 'readline', 'setuptools', 'sqlite', ...})
> E        +    where <built-in method issubset of set object at 0x10a0cb300> = {'python', 'python_abi'}.issubset
> ```
> 
> is this a MacOS issue?

Well it does happen only on the MacOS on CI, but I can't tell for sure if the error is related to this PR or not... (I'll have to investigate more...)

Hind-M:
cc @AntoinePrv 
Thanks :)

libmamba/src/core/prefix_data.cpp
```diff
@@ -144,4 +147,45 @@ namespace mamba
...
+        }
+
+        auto pkgs_info_list = split(strip(out), "\n");
+        for (auto& pkg_info_line : pkgs_info_list)
+        {
+            if (pkg_info_line.find("==") != std::string::npos)
+            {
+                auto pkg_info = split(strip(pkg_info_line), "==");
+                auto prec = PackageInfo(pkg_info[0], pkg_info[1], "pypi_0", "pypi");
+                m_package_records.insert({ prec.name, std::move(prec) });
```
AntoinePrv, line 186: Since conda and pip packages are not the same, I think we should store them separately, even use a different class than `PackageInfo` (maybe latter on).

Also I notice that `pip freeze` returns all packages, including the ones installed by mamba. One thing to be careful with is that mamba and pip package don't always have the same name, and a mamba package could contain multiple wheels...

AntoinePrv:
@Hind-M, playing around with pip, I found the following good ideas:

- `pip [...] --path [PrefixData]` could be a good idea to make sure we are using the right path (even if the wrong pip, see comment)
- `pip inspect` could be perfect here. It returns a json (therefore safe to parse) with plenty of useful information, in particular that `installed[pgk_idx].installer` (`pip` or `conda`) and `installed[pgk_idx].requested`

```
{
      "metadata": {
        "metadata_version": "2.1",
        "name": "args",
        "version": "0.1.0",
        "platform": [
          "any"
        ],
        "summary": "Command Arguments for Humans.",
        "home_page": "https://github.com/kennethreitz/args",
        "author": "Kenneth Reitz",
        "author_email": "me@kennethreitz.com",
        "license": "BSD",
        "classifier": [
          "Environment :: Web Environment",
          "Intended Audience :: Developers",
          "License :: OSI Approved :: BSD License",
          "Operating System :: OS Independent",
          "Programming Language :: Python",
          "Topic :: Internet :: WWW/HTTP :: Dynamic Content",
          "Topic :: Software Development :: Libraries :: Python Modules"
        ],
        "description": "\nargs\n~~~~\n\nThis simple module gives you an elegant interface for your command line argumemnts.\n\n\n\n"
      },
      "metadata_location": "/Users/antoine/.local/share/mamba/envs/mamba-dev/lib/python3.10/site-packages/args-0.1.0.dist-info",
      "direct_url": {
        "url": "file:///home/conda/feedstock_root/build_artifacts/args_1637347267685/work",
        "dir_info": {}
      },
      "installer": "conda",
      "requested": true
    },
```

RafailFridman:
I think this would be an amazing feature to have
Can't wait :)

agilebean:
any updates?

zacharyburnett:
I really appreciate y'all's work on this PR, and just wanted to say I still find this to be an important missing feature in my workflow; for now, I install `conda` in my workflow's environment solely to do `conda env export > environment.yml` so I can have the `pip:` section in the output.

Hind-M:
Superseded by #3565

PR #2115, opened 2022-11-15
Add some info to the doc

Hind-M:
Add some info to the documentation that I find may be helpful

### DIFF
--- a/docs/source/developer_zone/build_locally.rst
+++ b/docs/source/developer_zone/build_locally.rst
@@ -143,6 +143,9 @@ Alternatively you can use:
 
     make test
 
+.. note::
+    If you want to run specific or a subset of tests, you can use ``GTEST_FILTER`` environment variable or the ``--gtest_filter`` flag.
+
 Build ``libmambapy``
 ====================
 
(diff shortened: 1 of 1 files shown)


docs/source/developer_zone/build_locally.rst
```diff
@@ -323,8 +336,21 @@ or with ``libmamba``:
...
+
+.. code::
+
+    export TEST_MAMBA_EXE=build/micromamba/micromamba
+    export MAMBA_ROOT_PREFIX=YOUR_MICROMAMBA_ROOT_PREFIX
+
+Then, from the ``mamba`` root directory, you should be able to run the tests:
 
 .. code::
+    cd ..
```
AntoinePrv, line 349: Which folder did we start from? Should we simply remove this?

docs/source/developer_zone/build_locally.rst
```diff
@@ -323,8 +336,21 @@ or with ``libmamba``:
...
+    pytest micromamba/tests/
 
-    pytest ./micromamba/tests/
+Since running all the tests would take a great amount of time, you could choose to run only a specific test.
+To launch ``test_env`` for example, you can run:
```
AntoinePrv, line 353: We can also mention the `-k` option to filter by test name (across file, or for a subset of a file).

docs/source/developer_zone/build_locally.rst
Hind-M: We were in the build to run `make`, so I added this eventually so that it would be consistent with the previous steps and also added the  `Then, from the ``mamba`` root directory, ...` just before so that it's clearer...

AntoinePrv: Two folders are named `mamba`, the repo and the sub-folder.

I you care to change that, I think using `cmake -B build/ -D ...` & `cmake --build build/` above, then removing all `cd`/`mkdir` would be nicer.

Hind-M: Done!

docs/source/developer_zone/build_locally.rst
```diff
@@ -123,25 +117,23 @@ Tests
...
 
-    ./libmamba/tests/test_libmamba
+    ./build/libmamba/tests/test_libmamba
 
 Alternatively you can use:
 
 .. code::
 
-    make test
+    make test -C build/
```
AntoinePrv, line 136: Generator agnostic suggestion
```suggestion
    cmake --build build/ --target test
```

AntoinePrv (approved):

AntoinePrv:
Thanks for adding these!

I realize that there may be one pitfall to my suggestion: running `libmamba` tests form root directory may not work for some tests (a relative link to a test data file I believe). Happy to merge this as it is and fix the relative path issue in another PR.

Hind-M:
> I realize that there may be one pitfall to my suggestion: running `libmamba` tests form root directory may not work for some tests (a relative link to a test data file I believe)

I haven't found such a link, can you point it out if you already know what or where it is? 
Thanks!

Issue #1780, opened 2022-07-07
Use pre-commit.ci in this repo


jonashaag:
Example here: https://github.com/jonashaag/mamba/pull/3

I think it's great! Automatic formatting :)

wolfv:
Hmm, I really dislike these commits in the history though. That's the only downside

jonashaag:
You can always squash-merge, I think that's a good thing anyways.

PR #2116, opened 2022-11-15
Add PULL_REQUEST_TEMPLATE.md

Hind-M:
The reason of adding this is mainly to ask/put a reminder for running linters beforehand and avoid additional commits to fix possible linters failures.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,8 @@
+<!--
+Thank you for pull request.
+Below are a few things we ask you kindly to self-check before getting a review. Remove checks that are not relevant.
+-->
+Checklist
+* [ ] run linters using `pre-commit run --all`
+* [ ] note any issues closed by this PR with [closing keywords](https://help.github.com/articles/closing-issues-using-keywords)
+* [ ] put any other relevant information below


jonashaag:
Or maybe revisit https://github.com/mamba-org/mamba/issues/1780?

Hind-M:
Hum, I think that the automatic commits would make the history overloaded indeed. So if we decide to remove them with squash-merge manually, that would require more actions to be done, while running it locally won't necessarily (at least when the PR is created). But if you mean a squash-merge of all commits just before merging the PR into the main branch and always make the PR as a single commit (maybe automatically), maybe... But then for big PRs we lose the progress of the commits...
Or, we can set (maybe in a config file, or as a rule with git) an automatic run of the pre-commit when `git commit` is run locally, and if there is a failure, the commit pauses until it's fixed... (I don't know if that's possible though).

jonashaag:
> I don't know if that's possible though

That's exactly what this is doing: https://pre-commit.com/#3-install-the-git-hook-scripts

jonashaag:
Re squash commits, IMO in almost all cases it makes live easier: No need to manually clean the history before a merge. In the few cases where sub-PR history should be kept it does indeed take more work to get rid of the pre-commit commits, but in the other 90% of PRs you don't have to remind people to use pre-commit.

Change my mind ;)

PR #2118, opened 2022-11-15
Add dev mode command for micromamba shell init

Hind-M:
Following my experience in dev mode to activate an environment using a locally built micromamba, I added the working command in the displayed error message to hopefully make it clearer.

### DIFF
--- a/micromamba/src/activate.cpp
+++ b/micromamba/src/activate.cpp
@@ -28,14 +28,26 @@ set_activate_command(CLI::App* subcom)
         [&]()
         {
             std::string guessed_shell = guess_shell();
-            std::string shell_hook_command = "", shell_hook = "";
+            std::string shell_hook_command = "", shell_hook = "", shell_hook_dev_mode_cmd = "",
+                        shell_hook_dev_mode = "";
 
             if (guessed_shell == "powershell")
                 shell_hook_command
                     = "micromamba.exe shell hook -s powershell | Out-String | Invoke-Expression";
             else
+            {
                 shell_hook_command
                     = "eval \"$(micromamba shell hook --shell=" + guessed_shell + ")\"";
+
+                shell_hook_dev_mode_cmd = "eval \"$(./micromamba/micromamba shell hook --shell="
+                                          + guessed_shell + ")\"";
+
+                shell_hook_dev_mode = unindent((R"(
+
+                Or in dev mode, in your build directory:
+                    $ )" + shell_hook_dev_mode_cmd)
+                                                   .c_str());
+            }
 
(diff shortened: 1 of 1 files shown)


micromamba/src/activate.cpp
```diff
@@ -28,23 +28,35 @@ set_activate_command(CLI::App* subcom)
...
 
             if (guessed_shell == "powershell")
                 shell_hook_command
                     = "micromamba.exe shell hook -s powershell | Out-String | Invoke-Expression";
             else
+            {
                 shell_hook_command
                     = "eval \"$(micromamba shell hook --shell=" + guessed_shell + ")\"";
 
+                shell_hook_dev_mode_cmd = "eval \"$(./micromamba/micromamba shell hook --shell="
```
AntoinePrv, line 42: Won't this path change depending on user? It's not uncommon to put it under `build/`

micromamba/src/activate.cpp
```diff
@@ -28,23 +28,35 @@ set_activate_command(CLI::App* subcom)
...
+
             if (guessed_shell != "cmd.exe")
             {
                 shell_hook = unindent((R"(
 
                 To initialize the current )"
                                        + guessed_shell + R"( shell, run:
                     $ )" + shell_hook_command
-                                       + R"(
+                                       + shell_hook_dev_mode + R"(
```
AntoinePrv, line 59: This is going to be shown to all (non-dev) users? If so I would suggest a better approach would be to put that information in the docs.

micromamba/src/activate.cpp
AntoinePrv: In my experience, I also don't need to run the hook again. Since the hook already uses the `MAMBA_EXE` variable, I can set `export MAMBA_EXE="${PWD}/build/micromamba/micromamba`. Or perhaps I'm missing something.

Hind-M: > Won't this path change depending on user? It's not uncommon to put it under `build/`

If you are testing your built micromamba, I assumed you're in your build directory and so the path would be the same. But we could also make it `./build/micromamba/micromamba`, maybe that would be clearer.

micromamba/src/activate.cpp
Hind-M: > This is going to be shown to all (non-dev) users? 
Yes.

> If so I would suggest a better approach would be to put that information in the docs.
Well I thought about it but since the problem occurred while trying to activate an env with the built micromamba I had this message error (which wasn't solving the problem). So I think that having an error message with the right suggested command would really help because it's the first thing that I would try (follow the suggestion, before searching in the docs), but that doesn't mean not putting it there as well.

PR #2126, opened 2022-11-18
docs: consistently use `curl` for fetching files

wholtz:
Removes a conflicting referencing to using wget to retrieve micromamba


jonashaag (approved):

jonashaag:
Thanks! What do you think about adding a Docker section to the installation docs and linking the official Docker image (and Mambaforge) there? It’s not really „installing“ but serves the same purpose

wholtz:
Yes, that is a good idea. I'll add it to my todo list.

PR #2127, opened 2022-11-18
Bind SAT error messages to python

AntoinePrv:
- Bind ProblemsGraph and CompressProblemGraph
- Rename summary error message


wolfv:
Cool stuff! 

Do you know if this also exposes meaningful "structured" problems? E.g. for the conda-build use case, we need to know which package is missing (and find the "package not found [insert packge name]" error.

AntoinePrv:
@Wolf If you only want missing packages (ignoring conflicts), you should already be able to do filter `SOLVER_RULE_JOB_NOTHING_PROVIDES_DEP`, `SOLVER_RULE_JOB_UNKNOWN_PACKAGE` and `SOLVER_RULE_PKG_NOTHING_PROVIDES_DEP` from `Solver.all_problems_structured`.

https://github.com/mamba-org/mamba/blob/f32b42e3f23f585df28c59aa0e0c43b2333737eb/libmamba/src/core/satisfiability_error.cpp#L281

Could also double check that `expand_solvables` resolves to the empty list for said dependencies.

AntoinePrv:
This PR requires `networkx` to access the graph, which is not ideal. Perhaps a better option would be to return raw data (dicts and list) that can be passed to networkx).

libmambapy/src/main.cpp
```diff
@@ -54,6 +70,46 @@ deprecated(char const* message)
...
+        .def("build_strings_trunc",
+             &type::build_strings_trunc,
+             py::arg("sep") = "|",
+             py::arg("etc") = "...");
+    return pyclass;
+}
+
+template <typename Node, typename Edge>
+py::object
+make_networkx(mamba::DiGraph<Node, Edge> const& g)
```
AntoinePrv, line 98: This requires `networkx`. Perhaps we could return raw data (dicts and list) that can be forwarded (or not) by the user.

JohanMabille (approved):

Issue #2131, opened 2022-11-22
Micromamba remove with missing package gives "Nothing to do"


jonashaag:
```
$ micromamba create -n tmp1 'aimodelshare=0.0.146'
$ micromamba -n tmp1 remove aiomodelshare  # < Note the typo!
...
Nothing to do
...
```

Conda behaviour:

```
PackagesNotFoundError: The following packages are missing from the target environment:
  - xxx
```

Klaim:
> Conda behaviour:
> 
> ```
> PackagesNotFoundError: The following packages are missing from the target environment:
>   - xxx
> ```

Do you need the exact same output, or is it fine like in #2132 ?

jonashaag:
Nice, fine for me!

PR #2132, opened 2022-11-22
Report failure when packages to remove don't exist. (#2131)

Klaim:


Klaim:
I am not yet completely sure this is the right way to fix #2131 as this is not attempting to fix specifically `remove` but just the case where some internal data are inconsistent. I suspect explicitly checking the operation's result and/or checking if the package is present before attempting the operation (but that one might be expensive).

Here is how it looks for me with an environment where `fmt` is installed and we attempt to remove `ftm`:
```
╰─$ ./micromamba remove -n tmp1 ftm          

                                           __
      /_/

Transaction

  Prefix: /home/joel/micromamba/envs/tmp1

  Failure: packages to remove not found in the environment:

  - ftm


Transaction starting
Transaction finished
```

AntoinePrv (approved):

PR #2137, opened 2022-11-28
Fix micromamba shell completion when running 'shell hook' directly

### LINKED ISSUE #1748: micromamba shell completion doesn't work if you run 'shell hook' directly
To reproduce: (assuming Linux and bash)
```console
$ mkdir /tmp/aaaaa
$ cd /tmp/aaaaa
$ wget -qO- https://micro.mamba.pm/api/micromamba/linux-64/latest | tar -xvj bin/micromamba
$ eval "$( ./bin/micromamba shell hook -s posix )"
$ micromamba <TAB>
completer: command not found
```
Version:
```console
$ ./bin/micromamba --version
0.24.0
```

The reason for this bug is that _umamba_bash_completions() tries to run `$MAMBA_EXE completer "${COMP_WORDS[@]:1}"`, but the hook script never assigned $MAMBA_EXE. Therefore bash looks for a binary named `completer` (which doesn't exist) when it should've been `/tmp/aaaaa/bin/micromamba completer`. You can see it in the code below.

<details><summary>Full output of <code>./bin/micromamba shell hook -s posix</code></summary>

```bash
# Copyright (C) 2012 Anaconda, Inc
# SPDX-License-Identifier: BSD-3-Clause

__add_sys_prefix_to_path() {
    # In dev-mode MAMBA_EXE is python.exe and on Windows
    # it is in a different relative location to condabin.
    if [ -z "${MAMBA_ROOT_PREFIX}" ]; then
        return 0
    fi;

    if [ -n "${WINDIR+x}" ]; then
        PATH="${MAMBA_ROOT_PREFIX}/bin:${PATH}"
        PATH="${MAMBA_ROOT_PREFIX}/Scripts:${PATH}"
        PATH="${MAMBA_ROOT_PREFIX}/Library/bin:${PATH}"
        PATH="${MAMBA_ROOT_PREFIX}/Library/usr/bin:${PATH}"
        PATH="${MAMBA_ROOT_PREFIX}/Library/mingw-w64/bin:${PATH}"
        PATH="${MAMBA_ROOT_PREFIX}:${PATH}"
    else
        PATH="${MAMBA_ROOT_PREFIX}/bin:${PATH}"
    fi
    \export PATH
}


__conda_hashr() {
    if [ -n "${ZSH_VERSION:+x}" ]; then
        \rehash
    elif [ -n "${POSH_VERSION:+x}" ]; then
        :  # pass
    else
        \hash -r
    fi
}

__mamba_activate() {
    \local cmd="$1"
    shift
    \local ask_conda
    CONDA_INTERNAL_OLDPATH="${PATH}"
    __add_sys_prefix_to_path
    ask_conda="$(PS1="$PS1" "/tmp/aaaaa/bin/micromamba" shell --shell bash "$cmd" "$@")" || \return $?
    rc=$?
    PATH="${CONDA_INTERNAL_OLDPATH}"
    \eval "$ask_conda"
    if [ $rc != 0 ]; then
        \export PATH
    fi
    __conda_hashr
}

__mamba_reactivate() {
    \local ask_conda
    CONDA_INTERNAL_OLDPATH="${PATH}"
    __add_sys_prefix_to_path
    ask_conda="$(PS1="$PS1" "/tmp/aaaaa/bin/micromamba" shell --shell bash reactivate)" || \return $?
    PATH="${CONDA_INTERNAL_OLDPATH}"
    \eval "$ask_conda"
    __conda_hashr
}

micromamba() {
    if [ "$#" -lt 1 ]; then
        "/tmp/aaaaa/bin/micromamba"
    else
        \local cmd="$1"
        shift
        case "$cmd" in
            activate|deactivate)
                __mamba_activate "$cmd" "$@"
                ;;
            install|update|upgrade|remove|uninstall)
                CONDA_INTERNAL_OLDPATH="${PATH}"
                __add_sys_prefix_to_path
                "/tmp/aaaaa/bin/micromamba" "$cmd" "$@"
                \local t1=$?
                PATH="${CONDA_INTERNAL_OLDPATH}"
                if [ $t1 = 0 ]; then
                    __mamba_reactivate
                else
                    return $t1
                fi
                ;;
            *)
                CONDA_INTERNAL_OLDPATH="${PATH}"
                __add_sys_prefix_to_path
                "/tmp/aaaaa/bin/micromamba" "$cmd" "$@"
                \local t1=$?
                PATH="${CONDA_INTERNAL_OLDPATH}"
                return $t1
                ;;
        esac
    fi
}

if [ -z "${CONDA_SHLVL+x}" ]; then
    \export CONDA_SHLVL=0
    # In dev-mode MAMBA_EXE is python.exe and on Windows
    # it is in a different relative location to condabin.
    if [ -n "${_CE_CONDA+x}" ] && [ -n "${WINDIR+x}" ]; then
        PATH="${MAMBA_ROOT_PREFIX}/condabin:${PATH}"
    else
        PATH="${MAMBA_ROOT_PREFIX}/condabin:${PATH}"
    fi
    \export PATH

    # We're not allowing PS1 to be unbound. It must at least be set.
    # However, we're not exporting it, which can cause problems when starting a second shell
    # via a first shell (i.e. starting zsh from bash).
    if [ -z "${PS1+x}" ]; then
        PS1=
    fi
fi

if [ -n "${ZSH_VERSION:+x}" ]; then
  autoload -U +X compinit && compinit
  autoload -U +X bashcompinit && bashcompinit

  _umamba_zsh_completions()
  {
    COMPREPLY=($($MAMBA_EXE completer "${(@s: :)${(@s: :)COMP_LINE}:1}"))
  }

  complete -o default -F _umamba_zsh_completions micromamba
fi
if [ -n "${BASH_VERSION:+x}" ]; then
  _umamba_bash_completions()
  {
    COMPREPLY=($($MAMBA_EXE completer "${COMP_WORDS[@]:1}"))
  }
  complete -o default -F _umamba_bash_completions micromamba
fi

```

</details>

This is just my conjecture, but the problem seems to be that the hook script is created from two templates: [first](https://github.com/mamba-org/mamba/blob/master/libmamba/data/micromamba.sh) and [second](https://github.com/mamba-org/mamba/blob/master/libmamba/data/mamba_completion.posix). And only occurrences of `$MAMBA_EXE` in the first template were replaced by the program path.

I'm guessing this usually works for people who run "shell init" because it also writes `export MAMBA_EXE=...` directly to .bashrc. By the way, once you fix this bug, you might not even need to add that export anymore (at least for bash). But that's optional.

TomiBelan:
Fixes #1748.

Note: I didn't test this at all.

Caveat: With this patch, old users will most likely have to rerun "micromamba shell init" or "micromamba shell reinit" after doing a manual update. Otherwise the old script will stay cached in ~/micromamba/etc/profile.d/micromamba.sh and their bash/zsh completion will break. It would be good to document this. But at least users on 1.0.0+ can just use "micromamba self-update", which runs shell reinit automatically.

### DIFF
--- a/libmamba/data/mamba_completion.posix
+++ b/libmamba/data/mamba_completion.posix
@@ -10,7 +10,7 @@ if [ -n "${ZSH_VERSION:+x}" ]; then
 
   _umamba_zsh_completions()
   {
-    COMPREPLY=($($MAMBA_EXE completer "${(@s: :)${(@s: :)COMP_LINE}:1}"))
+    COMPREPLY=($(__mamba_exe completer "${(@s: :)${(@s: :)COMP_LINE}:1}"))
   }
 
   complete -o default -F _umamba_zsh_completions micromamba
@@ -18,7 +18,7 @@ fi
 if [ -n "${BASH_VERSION:+x}" ]; then
   _umamba_bash_completions()
   {
-    COMPREPLY=($($MAMBA_EXE completer "${COMP_WORDS[@]:1}"))
+    COMPREPLY=($(__mamba_exe completer "${COMP_WORDS[@]:1}"))
   }
   complete -o default -F _umamba_bash_completions micromamba
 fi


TomiBelan:
@wolfv, can you please review this?

wolfv (approved):

wolfv:
Thanks!

PR #2141, opened 2022-11-30
feat(fs): Replace 'last_write_time' by underlying 'utimesat' to support ACL's

coroa:
Co-authored-by: Bourdon Olivier <mail@olivier-bourdon.fr>


coroa:
@wolfv Do I need to ping someone for review? I am pretty confident this solves the `Permission denied` errors, people in multi-user installations with ACLs or setgid or in some container setups have been reporting.

The gist is that on linux only the owner is allowed to update the modification time of a file to a particular timestamp, while it is permitted to update the modification time to the current time (as long as one has write permission), but the `std::filesystem` layer does not provide an interface for that.

coroa:
Updated the PR description to make it easier to follow and hope to find someone for review!

wolfv:
Thanks, I looked through the PR and it looks good to me! Will ask @Klaim what he thinks, then happy to merge! :)

Do you know if this also works on macOS?

wolfv:
Judging by this issue on some other repo, it seems that the command was not available before macOS 10.12: https://github.com/bytecodealliance/rustix/issues/157

Klaim:
Thanks for the PR!

I just started reviewing this., other than informing myself of the details,
I'm trying to figure out if there isnt a way to do this without deferring to the toolchain's implementation of the standard's implementation to avoid future surprises or platforms not having `utimensat`. I'm not completely knowledgeable of the issue so bear with me ^^;

As the proposed function is an extension of the standard interface we try to keep, the overall idea of this PR seems reasonable to me.
At the moment I have a slight reticence to introducing `fs::now` because it's a name that could become problematic in the future - though we could also change it in a future version so I don't worry too much.

Could the `utimensat` implementation be restricted to non-windows but also non-apple platforms? Or are they also affected?

(I didn´t look at the ci issues yet)

coroa:
> I just started reviewing this., other than informing myself of the details, I'm trying to figure out if there isnt a way to do this without deferring to the toolchain's implementation of the standard's implementation to avoid future surprises or platforms not having `utimensat`. I'm not completely knowledgeable of the issue so bear with me ^^;

Great to see this is moving somewhere.

> As the proposed function is an extension of the standard interface we try to keep, the overall idea of this PR seems reasonable to me. At the moment I have a slight reticence to introducing `fs::now` because it's a name that could become problematic in the future - though we could also change it in a future version so I don't worry too much.

I am not married to the name, if you find a more fitting alternative i'm happy to oblige.

> Could the `utimensat` implementation be restricted to non-windows but also non-apple platforms? Or are they also affected?

There was no report about a multi-user issue on MacOS X, which given the nature of normal mac usage is not surprising, but i can actually try on my machine. Will report back.

On systems for which `utimensat` is not defined we could also defer back to utimes, but then we need an automatic build-system discovery step to differentiate between the two, i am wary of introducing this complexity but can put some research into it.

Klaim:
Thanks for the details @coroa 

I was looking at libcxx (from llvm) and libstdc++ (from gcc) and they actually do use `utimensat` by default when doing `std::filesystem::last_write_time(path, std::filesystem::file_time_type::now(), ...) ...)` (which is also what you proposed [for the windows case](https://github.com/mamba-org/mamba/pull/2141/files#diff-14842e873613261b88cfd612f98b39a7ded38892a8465daa346d2b3ee62891f4R1165-R1166).
When `utimensat` is not available (or not wanted) they switch to another implementation.

See: 
 - https://gcc.gnu.org/git/?p=gcc.git;a=blob;f=libstdc%2B%2B-v3/src/filesystem/ops.cc;hb=b97c33fbd28e74a206c81c96a9b0b9fa3c8545d6#l919
 - https://github.com/llvm/llvm-project/blob/main/libcxx/src/filesystem/operations.cpp#L1253 and https://github.com/llvm/llvm-project/blob/2e999b7dd1934a44d38c3a753460f1e5a217e9a5/libcxx/src/filesystem/filesystem_common.h#L532

Therefore, if my understanding is correct (I probably missed something), the issue is really that our calls to `fs::last_write_time(path, now)` should be rewritten to `fs::last_write_time(path, fs::time, std::filesystem::file_time_type::now())` systematically - or be replaced by your function with only the windows impl.

I cannot yet reproduce the initial issue, but if you do reproduce it, could you check if using only that implementation https://github.com/mamba-org/mamba/pull/2141/files#diff-14842e873613261b88cfd612f98b39a7ded38892a8465daa346d2b3ee62891f4R1165-R1166 for that new function actually fixes the issue?

Klaim:
Looks like the CI issues are unrelated to this PR?

wolfv:
yep, i think most of the ci issues are unrelated :)

coroa:
> Therefore, if my understanding is correct (I probably missed something), the issue is really that our calls to fs::last_write_time(path, now) should be rewritten to fs::last_write_time(path, std::filesystem::file_time_type::now()) systematically - or be replaced by your function with only the windows impl.

No, you are slightly off course. 

`fs::last_write_time(path, std::filesystem::file_time_type::now())` creates a timestamp for the current time, then (in the llvm libcxx impl's operations.cpp) converts this to a timespec and then calls `utimensat` to set file modification and access times to this timespec. Unfortunately, this is a privileged operation where you have to own the file (or be root or similar). Instead you want to pass in a TimeSpec with the special value `UTIME_NOW` or a `NULL` pointer. The latter is what my implementation does.

> If the tv_nsec field of one of the timespec structures has the special value UTIME_NOW, then the corresponding file timestamp is set to the current time.  If the tv_nsec field of one of the timespec structures has the special value UTIME_OMIT, then the corresponding file timestamp is left unchanged.  In both of these cases, the value of the corresponding tv_sec field is ignored.

(as described in the manpage [utimensat](https://man7.org/linux/man-pages/man2/utimensat.2.html) )

coroa:
I found that llvm's libcxx implementation uses the presence of the `UTIME_NOW` constant to decide whether use `utimensat`.

https://github.com/llvm/llvm-project/blob/2e999b7dd1934a44d38c3a753460f1e5a217e9a5/libcxx/src/filesystem/filesystem_common.h#L42-L44

I would propose to make the new code conditional on this constant being defined, then we fall back to the old implementation when `utimensat` does not exist and use the fixed code on newer systems?

Klaim:
> I would propose to make the new code conditional on this constant being defined, then we fall back to the old implementation when utimensat does not exist and use the fixed code on newer systems?

That makes sense to me indeed.

If that fixes also the macos build, then after fixing the ci I'm ok to merge this change 👍🏽 Thanks for the detailed explaination.

libmamba/include/mamba/core/mamba_fs.hpp
```diff
@@ -10,6 +10,12 @@
 #if !defined(_WIN32)
 #include <sys/stat.h>
 #include <fcntl.h>
+
+// We can use the presence of UTIME_OMIT to detect platforms that provide
+// utimensat.
+#if defined(UTIME_OMIT)
+#define USE_UTIMENSAT
```
coroa, line 17: I was wondering whether to prefix the constant, ie like `_MAMBA_USE_UTIMENSAT`?

coroa:
Ok, with this change it compiles and fixes the issue on ubuntu and MacOSX. Tests on MacOSX pass, on ubuntu i'm seeing some file locking errors, which i don't think are related the filesystem layer.

wolfv:
Thanks so much! I hope you had some wonderful holidays :)

wolfv (approved):

PR #2158, opened 2022-12-06
Expliclity point to libmamba test data independently of cwd

AntoinePrv:
This makes it possible to run tests from the project root directory, without changing into the test build directory.

### DIFF
--- a/libmamba/tests/test_data.hpp
+++ b/libmamba/tests/test_data.hpp
@@ -15,7 +15,7 @@ namespace mamba
 #ifndef MAMBA_TEST_DATA_DIR
 #error "MAMBA_TEST_DATA_DIR must be defined pointing to test data"
 #endif
-    static const fs::u8path test_data_dir = MAMBA_TEST_DATA_DIR;
+    inline static const fs::u8path test_data_dir = MAMBA_TEST_DATA_DIR;
 
 }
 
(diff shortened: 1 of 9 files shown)


libmamba/tests/test_data.hpp
```diff
@@ -0,0 +1,22 @@
...
+
+#include "mamba/core/fsutil.hpp"
+
+namespace mamba
+{
+
+#ifndef MAMBA_TEST_DATA_DIR
+#error "MAMBA_TEST_DATA_DIR must be defined pointing to test data"
+#endif
+    static const fs::u8path test_data_dir = MAMBA_TEST_DATA_DIR;
```
Klaim, line 18: To avoid ODR violation (or at least duplication in multiple object files) I think you can make this object `inline`, so that there is only one definition after link. Either that or mark it extern and put the definition in a cpp). But `inline` should be enough.

libmamba/tests/test_lockfile.cpp
```diff
@@ -20,8 +20,17 @@ extern "C"
...
 namespace mamba
 {
+
+#ifndef MAMBA_TEST_LOCK_EXE
+#error "MAMBA_TEST_LOCK_EXE must be defined pointing to testing_libmamba_lock"
+#endif
+    static const fs::u8path testing_libmamba_lock_exe = MAMBA_TEST_LOCK_EXE;
+
+
+    static const fs::u8path test_data_dir = MAMBA_TEST_DATA_DIR;
```
Klaim, line 33: Why does this need to be re-defined here if it's in `test_data.hpp`?

Klaim (commented):
Very good idea, just a few adjustments and LGTM

Klaim (approved):

Issue #1895, opened 2022-09-01
Having `micromamba` in PATH conflicts with `micromamba` shell function


jonashaag:
Investigate this https://github.com/mamba-org/mamba/blob/7a4f488568d71335f807b280719c2a03f39651ff/.github/workflows/test_micro.mamba.pm.yml#L42-L44

pavelzw:
This seems to be special for the GitHub Actions `ubuntu-latest` image. On a regular Ubuntu installation like the [Ubuntu docker container](https://hub.docker.com/_/ubuntu), `$HOME/.local/bin` is not in `$PATH`. 

Though the semantics of the `bin` folder usually implies that this folder *should* be in path imo. Maybe putting the `micromamba` executable in `$HOME/micromamba/bin/micromamba` would be better?

jonashaag:
IMO it doesn’t matter what’s the exact folder. Micromamba should work fine with the binary in the PATH, otherwise it’s a bug IMO.

AntoinePrv:
> This function would somehow need to be integrated into the micromamba executable... This seems like a kind of long journey.

The way I understand it is that it is not possible. When executed, `micromamba` is a sub-process and cannot change the user's env (`PATH`, `CFLAGS`...) during activation.

> Micromamba should work fine with the binary in the PATH, otherwise it’s a bug IMO.

I agree, I have `micromamba` in `PATH` and everything works fine. It is even necessary for scripts (*e.g.* using `micromamba run`) that do not have access to the function.

pavelzw:
> I agree, I have micromamba in PATH and everything works fine.

@AntoinePrv have you got an idea why [this](https://github.com/pavelzw/ci-test/actions/runs/3021562222) fails then (it belongs to [this](https://github.com/pavelzw/ci-test/blob/micromamba-path-broken/.github/workflows/ci.yml) CI file)? I thought it has something to do with the `micromamba` executable being in `PATH`...

pavelzw:
> It is even necessary for scripts (e.g. using micromamba run) that do not have access to the function.

Are you saying that `micromamba run` only works if it is in `PATH` and not if *only* [this](https://github.com/mamba-org/mamba/issues/1895#issuecomment-1234531093) function definition is sourced?

AntoinePrv:
[...]

PR #2162, opened 2022-12-11
Don't create a prefix which is missing conda-meta

maresb:
To illustrate the problem, run the following:

```bash
docker run --rm -it mambaorg/micromamba:1.1.0
unset MAMBA_ROOT_PREFIX
micromamba shell init
micromamba shell init
```

The first `shell init` command creates a root prefix `~/micromamba` which contains the single file `~/micromamba/etc/profile.d/micromamba.sh`. The second `shell init` validates the root prefix by checking for `~/micromamba/conda-meta`, which is missing, causing the following error:

```
critical libmamba Could not use default 'root_prefix': /home/mambauser/micromamba: Directory exists, is not empty and not a conda prefix.
```

Workaround with Bash:
```bash
docker run --rm -it mambaorg/micromamba:1.1.0
unset MAMBA_ROOT_PREFIX
micromamba shell init
mkdir -p ~/micromamba/conda-meta
micromamba shell init
```

This PR which I'm submitting is untested, but it should at least convey what I believe the fix should be.

### DIFF
--- a/.github/workflows/conda_nightly.yml
+++ b/.github/workflows/conda_nightly.yml
@@ -103,5 +103,5 @@ jobs:
         if: always()
         run: |
           # Do not cache temporary envs with 'cache-env: true'
-          find ~ -iname build_env -maxdepth 3 -type d -exec rm -rf {} \;
-          find ~ -iname conda-bld -maxdepth 3 -type d -exec rm -rf {} \;
+          rm -rf $(micromamba info --json | jq -r '."env location"')/envs
+          rm -rf $(micromamba info --json | jq -r '."env location"')/conda-build/conda-bld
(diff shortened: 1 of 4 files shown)


jonashaag:
Thanks! Do you have time to add a test?

jonashaag:
I think this is the same issue as https://github.com/mamba-org/mamba/issues/1895#issuecomment-1242808731 which I believed to be fixed but it’s not

maresb:
I was having some trouble getting pytest to run in my local environment. Thus I haven't yet tested the test. :smile:

maresb:
Thanks a lot for cleaning this up!!!

wolfv (approved):

PR #2173, opened 2022-12-16
Add the bzip2 package requirement

kozo2:
If the user is unaware of this, the install script will fail with:
```
100   889  100   889    0     0    800      0  0:00:01  0:00:01 --:--:--   800
tar (grandchild): bzip2: Cannot exec: No such file or directory
tar (grandchild): Error is not recoverable: exiting now
tar: Child died with signal 13
tar: Error is not recoverable: exiting now
bash: line 42: /home/kozo2/.local/bin/micromamba: No such file or directory
```

### DIFF
--- a/docs/source/installation.rst
+++ b/docs/source/installation.rst
@@ -73,6 +73,8 @@ For Linux, the default shell is ``bash``:
 
 .. code:: bash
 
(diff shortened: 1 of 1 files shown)


jonashaag:
Thanks a lot for the patch, we’re currently building a new install.sh that doesn’t need bzip2. Will give a pre release version next week so you can test if you’re curious :)

PR #2177, opened 2022-12-19
make tarballs look more similar to conda-package-handling

wolfv:
This fixes an bug where symlinks where not properly included in the archives and also adds a similar sorting function to what conda-package-handling has. Although one has to note that `conda-package-handling` currently uses the builtin `hash(...)` function which is more difficult to implement in C++.

We might want to switch to alphabetical sorting instead anyways (as discussed with @dholth)


wolfv:
Looks like we can even enable multi-threaded compression! :) https://github.com/libarchive/libarchive/pull/1567

dholth:
We stopped using libarchive for conda-package-handling 2.0, and conda-package-streaming has a rust extractor in an unmerged branch. Threaded compression is great.

wolfv:
yep, I am aware that you moved away from libarchive. Rust might be cool and we have some plans in that direction as well -- but these are some pretty lowhanging fruits with libarchive :)

libmamba/src/core/package_handling.cpp
```diff
@@ -142,24 +185,44 @@ namespace mamba
...
-                p = p.substr(1);
+                auto clean_path = dir_entry.path().lexically_relative("./");
+                files.push_back({ clean_path, zip_order(clean_path) });
             }
-            if (p[0] == '/')
+        }
+
+        std::sort(files.begin(),
+                  files.end(),
+                  [](const auto& a, const auto& b) { return a.second < b.second; });
```
JohanMabille, line 209: replacing this lambda with:
```cpp
[](const auto& a, const auto& b) { return a.second < b.second || (a.second == b.second && a.first < b.first); }
```
allows to return an `int` instead of `std::pair<int, std::string>` in the `order` and  `zip_order` function, and simplifies the files type.

JohanMabille: or even simpler:

```cpp
[](const auto& a, const auto& b) { return std::tie(a.second, a.first) < std::tie(b.second, b.first); }
```

Klaim: Is the comparison object even necessary? `std::pair` already have lexicographic comparison using both members by default.

wolfv: yes, exactly. I think what's confusing is that this is a pair of pairs and the latter pair is actually used with it's default comparison operator.

wolfv: Ah ok, I think I see what you want now :) will change it.

wolfv: lambda is gone :)

libmamba/src/core/package_handling.cpp
```diff
@@ -84,12 +84,38 @@ namespace mamba
...
+        return pair.second == prefix.std_path().end();
+    }
+
+    std::pair<int, std::string> order(const fs::u8path& path)
+    {
+        int is_info = path_has_prefix(path, "info");
+        return { !is_info, path.string() };
+    }
+
+    std::pair<int, std::string> zip_order(const fs::u8path& path)
```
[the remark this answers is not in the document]
wolfv: done

libmamba/src/core/package_handling.cpp
JohanMabille: If the order functions return `int` instead of `std::pair<int, string>` then we want to sort by the second member of the pair (the int)  first, and then the first (the path). Therefore the lambda is still required since the natural comparison of `std::pair` starts with the first member and then the second if the first ones are equal.

libmamba/src/core/package_handling.cpp
```diff
@@ -84,12 +84,38 @@ namespace mamba
...
+
+    int zip_order(const fs::u8path& path)
+    {
+        // sort info-...tar.zst file last in zip folder"
+        int init_order = starts_with(path.filename().string(), "info-");
+        // sort metadata.json first in zip folder
+        if (path.filename().string() == "metadata.json")
+            init_order = -1;
+        return init_order;
+    }
```
dholth, line 110: info- is now last so that it could be buffered in memory while the pkg- archive is written directly to the zip

libmamba/src/core/package_handling.cpp
```diff
@@ -127,10 +153,27 @@ namespace mamba
...
+            int res = archive_write_set_options(a, comp_level.c_str());
+            if (res != 0)
+            {
+                LOG_ERROR << "libarchive error (" << res << ") " << archive_error_string(a);
+            }
+
+            if (compression_threads > 2)
+            {
+                std::string comp_threads_level
+                    = std::string("zstd:threads=") + std::to_string(compression_threads);
```
dholth, line 170: great!

mamba_package/src/package.cpp
```diff
@@ -35,6 +36,10 @@ set_package_command(CLI::App* com)
         "-c,--compression-level",
         compression_level,
         "Compression level from 0-9 (tar.bz2, default is 9), and 1-22 (conda, default is 15)");
+    compress_subcom->add_option(
+        "--compression-threads",
+        compression_threads,
+        "Compression threads (only relevant for .conda packages, default is 1)");
```
dholth, line 42: I don't know if zstd does "one thread if 1, a single thread but doesn't use threads at all if 0" or -1 to automatically choose threading based on processors

micromamba/tests/test_package.py
```diff
@@ -0,0 +1,209 @@
...
+                with tarfile.open(fileobj=dcf, mode="r|") as z:
+                    assert_sorted(z.getnames())
+                    members = z.getmembers()
+                    for m in members:
+                        if f.name.startswith("info-"):
+                            assert m.name.startswith("info/")
+                        if not f.name.startswith("info-"):
+                            assert not m.name.startswith("info/")
+                        assert m.uid == 0
+                        assert m.gid == 0
```
dholth, line 209: our new comparison is https://github.com/conda/conda-package-handling/blob/main/src/conda_package_handling/validate.py#L59

tarballs can also store uname and gname are strings

to be bug-compatible, conda-package-handling 2 keeps the "regex that puts info/LICENSE into pkg-" which we should fix in a minor 2.x release. We may audit existing packages for info/ filenames to see if there was a good reason for that regex.

wolfv: I fixed that (also setting them to empty).

PR #2182, opened 2022-12-19
Add bug report issue template

jonashaag:
We have lots of bug reports that

- are duplicates
- are missing important info
- show that people haven't invested a lot of time into solving the problems on their own or were unable to find the right resources to do so

I propose to use an issue template to reduce the number of low quality reports.


AntoinePrv (commented):
Should we add a reminder to try with the latest version of mamba?

jonashaag:
Great idea, done

AntoinePrv (approved):

PR #2184, opened 2022-12-20
[CI] try to fix linux build

pavelzw:


wolfv:
Thanks for debugging this a bit! I think this might be caused by the latest `libsolv` release? Not sure.

More generally, we might be able to disable `GZIP` for the libsolv build (`-DENABLE_GZIP=OFF`) in the libsolv-feedstock as I don't think we use `gzip` (through libsolv at least).

wolfv:
I think this fixes it #2201

PR #2200, opened 2023-01-04
Faster conda decompress

wolfv:
Implement the `.conda` file extraction in a streaming fashion (instead of extracting to temporary directory first).

This makes it faster. Also disabling the `libarchive` `ARCHIVE_EXTRACT_SPARSE` feature makes it faster (on macOS) because the way it's implemented in `libarchive` results in many more calls to `write`: https://github.com/libarchive/libarchive/issues/1835

### DIFF
--- a/libmamba/include/mamba/core/package_handling.hpp
+++ b/libmamba/include/mamba/core/package_handling.hpp
@@ -34,7 +34,6 @@ namespace mamba
                         int compression_level);
 
     void extract_archive(const fs::u8path& file, const fs::u8path& destination);
-    void stream_extract_archive(struct archive* a, const fs::u8path& destination);
     void extract_conda(const fs::u8path& file,
                        const fs::u8path& dest_dir,
                        const std::vector<std::string>& parts = { "info", "pkg" });
--- a/libmamba/include/mamba/core/package_handling.hpp
+++ b/libmamba/include/mamba/core/package_handling.hpp
@@ -34,6 +34,7 @@ namespace mamba
                         int compression_level);
 
     void extract_archive(const fs::u8path& file, const fs::u8path& destination);
+    void stream_extract_archive(struct archive* a, const fs::u8path& destination);
     void extract_conda(const fs::u8path& file,
                        const fs::u8path& dest_dir,
                        const std::vector<std::string>& parts = { "info", "pkg" });
--- a/libmamba/src/core/package_handling.cpp
+++ b/libmamba/src/core/package_handling.cpp
@@ -374,7 +374,7 @@ namespace mamba
         flags |= ARCHIVE_EXTRACT_SECURE_NODOTDOT;
         flags |= ARCHIVE_EXTRACT_SECURE_SYMLINKS;
         flags |= ARCHIVE_EXTRACT_SECURE_NOABSOLUTEPATHS;
-        flags |= ARCHIVE_EXTRACT_SPARSE;
+        // flags |= ARCHIVE_EXTRACT_SPARSE;
         flags |= ARCHIVE_EXTRACT_UNLINK;
 
         a = archive_read_new();
@@ -488,7 +488,7 @@ namespace mamba
         flags |= ARCHIVE_EXTRACT_SECURE_NODOTDOT;
         flags |= ARCHIVE_EXTRACT_SECURE_SYMLINKS;
         flags |= ARCHIVE_EXTRACT_SECURE_NOABSOLUTEPATHS;
-        flags |= ARCHIVE_EXTRACT_SPARSE;
+        // flags |= ARCHIVE_EXTRACT_SPARSE;
         flags |= ARCHIVE_EXTRACT_UNLINK;
 
         ext = archive_write_disk_new();
--- a/libmamba/src/core/package_handling.cpp
+++ b/libmamba/src/core/package_handling.cpp
@@ -7,6 +7,7 @@
 
 #include <archive.h>
 #include <archive_entry.h>
+#include <iostream>
 
 #include <sstream>
 
@@ -451,38 +452,164 @@ namespace mamba
         fs::current_path(prev_path);
     }
 
+    namespace
+    {
+        constexpr size_t BUFFER_SIZE = 131072;
+
+        struct conda_extract_context
+        {
+            struct archive* source;
+            char buffer[BUFFER_SIZE];
+        };
+    }
+
+    void stream_extract_archive(archive* a, const fs::u8path& destination)
+    {
+        auto prev_path = fs::current_path();
+        if (!fs::exists(destination))
+        {
+            fs::create_directories(destination);
+        }
+        fs::current_path(destination);
+
+        struct archive* ext;
+        struct archive_entry* entry;
+        int flags;
+        int r;
+
+        /* Select which attributes we want to restore. */
+        flags = ARCHIVE_EXTRACT_TIME;
+        flags |= ARCHIVE_EXTRACT_PERM;
+        flags |= ARCHIVE_EXTRACT_SECURE_NODOTDOT;
+        flags |= ARCHIVE_EXTRACT_SECURE_SYMLINKS;
+        flags |= ARCHIVE_EXTRACT_SECURE_NOABSOLUTEPATHS;
+        flags |= ARCHIVE_EXTRACT_SPARSE;
+        flags |= ARCHIVE_EXTRACT_UNLINK;
+
+        ext = archive_write_disk_new();
+        archive_write_disk_set_options(ext, flags);
+        archive_write_disk_set_standard_lookup(ext);
+
+        for (;;)
+        {
+            r = archive_read_next_header(a, &entry);
+            if (r == ARCHIVE_EOF)
+            {
+                break;
+            }
+            if (r < ARCHIVE_OK)
+            {
+                throw std::runtime_error(archive_error_string(a));
+            }
+
+            r = archive_write_header(ext, entry);
+            if (r < ARCHIVE_OK)
+            {
+                throw std::runtime_error(archive_error_string(ext));
+            }
+            else if (archive_entry_size(entry) > 0)
+            {
+                std::cout << "Reading data into file: " << archive_entry_pathname(entry) << "\n";
+                
+                int fd = open(archive_entry_pathname(entry), O_WRONLY);
+                archive_read_data_into_fd(a, fd);
+                // r = copy_data(a, ext);
+                if (r < ARCHIVE_OK)
+                {
+                    const char* err_str = archive_error_string(ext);
+                    if (err_str == nullptr)
+                    {
+                        err_str = archive_error_string(a);
+                    }
+                    if (err_str != nullptr)
+                    {
+                        throw std::runtime_error(err_str);
+                    }
+                    throw std::runtime_error("Extraction: writing data was not successful.");
+                }
+            }
+            r = archive_write_finish_entry(ext);
+            if (r == ARCHIVE_WARN)
+            {
+                std::cout << "libarchive warning: " << archive_error_string(a);
+            }
+            else if (r < ARCHIVE_OK)
+            {
+                throw std::runtime_error(archive_error_string(ext));
+            }
+        }
+
+        archive_read_close(a);
+        archive_read_free(a);
+
+        archive_write_close(ext);
+        archive_write_free(ext);
+
+        fs::current_path(prev_path);
+    }
+
+
+    static la_ssize_t file_read(struct archive* a, void* client_data, const void** buff)
+    {
+        struct conda_extract_context* mine = static_cast<conda_extract_context*>(client_data);
+        *buff = mine->buffer;
+
+        auto read = archive_read_data(mine->source, mine->buffer, BUFFER_SIZE);
+        if (read < 0)
+        {
+            std::cout << archive_error_string(mine->source) << std::endl;
+            std::cout << "Error reading from archive" << std::endl;
+            return read;
+        }
+        // std::cout << "Read " << read << " bytes" << std::endl;
+        return read;
+    }
+
+    int archive_read_open_archive_entry(struct archive* a, struct archive* source)
+    {
+        struct conda_extract_context* mine;
+
+        archive_clear_error(a);
+        mine = (struct conda_extract_context*) calloc(1, sizeof(*mine));
+        mine->source = source;
+
+        archive_read_set_read_callback(a, file_read);
+        archive_read_set_callback_data(a, mine);
+        return archive_read_open1(a);
+    }
+
+
     void extract_conda(const fs::u8path& file,
                        const fs::u8path& dest_dir,
                        const std::vector<std::string>& parts)
     {
-        TemporaryDirectory tdir;
-        extract_archive(file, tdir);
+        // open outer zip archive
+        struct archive* a = archive_read_new();
+        assert(archive_read_support_format_zip(a) == ARCHIVE_OK);
 
-        auto fn = file.stem();
+        if (archive_read_open_filename(a, file.string().c_str(), BUFFER_SIZE) != ARCHIVE_OK) {
+            throw std::runtime_error(archive_error_string(a));
+        }
 
-        auto metadata_path = tdir.path() / "metadata.json";
-        if (fs::exists(metadata_path) && fs::file_size(metadata_path) != 0)
+        // read archive
+        struct archive_entry* entry;
+        while (archive_read_next_header(a, &entry) == ARCHIVE_OK)
         {
-            const auto condafile_path = tdir.path() / "metadata.json";
-            std::ifstream condafile_meta(condafile_path.std_path());
-            nlohmann::json j;
-            condafile_meta >> j;
-
-            if (j.find("conda_pkg_format_version") != j.end())
+            fs::u8path p(archive_entry_pathname(entry));
+            if (p.extension() == ".zst")
             {
-                if (j["conda_pkg_format_version"] != 2)
-                {
-                    throw std::runtime_error("Can only read conda version 2 files.");
-                }
+                // extract zstd file
+                struct archive* inner = archive_read_new();
+                archive_read_support_filter_zstd(inner);
+                archive_read_support_format_tar(inner);
+
+                archive_read_open_archive_entry(inner, a);
+                stream_extract_archive(inner, dest_dir);
             }
         }
-
-        for (auto& part : parts)
-        {
-            std::stringstream ss;
-            ss << part << "-" << fn.string() << ".tar.zst";
-            extract_archive(tdir.path() / ss.str(), dest_dir);
-        }
+        // close archive
+        archive_read_close(a);
+        archive_read_free(a);
     }
 
     static fs::u8path extract_dest_dir(const fs::u8path& file)
--- a/libmamba/src/core/package_handling.cpp
+++ b/libmamba/src/core/package_handling.cpp
@@ -8,6 +8,7 @@
 #include <archive.h>
 #include <archive_entry.h>
 #include <iostream>
+#include <zstd.h>
 
 #include <sstream>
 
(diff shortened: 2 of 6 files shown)


libmamba/include/mamba/core/package_handling.hpp
```diff
@@ -34,6 +34,7 @@ namespace mamba
                         int compression_level);
 
     void extract_archive(const fs::u8path& file, const fs::u8path& destination);
+    void stream_extract_archive(struct archive* a, const fs::u8path& destination);
```
JohanMabille, line 37: `struct` no required.

libmamba/src/core/package_handling.cpp
```diff
@@ -354,14 +356,53 @@ namespace mamba
         LOG_INFO << "Extracting " << file << " to " << destination;
         extraction_guard g(destination);
 
+        struct archive* a;
+        int r;
+
+        a = archive_read_new();
```
JohanMabille, line 362: `archive* a = archive_read_new();`

libmamba/src/core/package_handling.cpp
```diff
@@ -354,14 +356,53 @@ namespace mamba
...
+        archive_read_support_format_tar(a);
+        archive_read_support_format_zip(a);
+        archive_read_support_filter_all(a);
+
+        auto lock = LockFile(file);
+        r = archive_read_open_filename(a, file.string().c_str(), 10240);
+
+        if (r != ARCHIVE_OK)
```
Klaim, line 370: As this is C++ and not old version of C, I would prefer:
```
archive* a = archive_read_new();
...
const int r = archive_read_open_filename(...);
if (r !=ARCHIVE_OK)
```

libmamba/src/core/package_handling.cpp
```diff
@@ -354,14 +356,53 @@ namespace mamba
         LOG_INFO << "Extracting " << file << " to " << destination;
         extraction_guard g(destination);
 
+        struct archive* a;
+        int r;
```
JohanMabille, line 360: These variables can be initialized when defined (see below).

libmamba/src/core/package_handling.cpp
```diff
@@ -354,14 +356,53 @@ namespace mamba
...
+
+        a = archive_read_new();
+        archive_read_support_format_tar(a);
+        archive_read_support_format_zip(a);
+        archive_read_support_filter_all(a);
+
+        auto lock = LockFile(file);
+        r = archive_read_open_filename(a, file.string().c_str(), 10240);
```
JohanMabille, line 368: `int r = ...`

libmamba/src/core/package_handling.cpp
```diff
@@ -354,14 +356,53 @@ namespace mamba
...
+    {
+        struct conda_extract_context
+        {
+            conda_extract_context(struct archive* source)
+                : source(source)
+                , buffer(ZSTD_DStreamOutSize())
+            {
+            }
+
+            struct archive* source;
```
JohanMabille, line 392: no need for `struct` keyword.

libmamba/src/core/package_handling.cpp
```diff
@@ -354,14 +356,53 @@ namespace mamba
...
+        if (r != ARCHIVE_OK)
+        {
+            LOG_ERROR << "Error opening archive: " << archive_error_string(a);
+            throw std::runtime_error(file.string() + " : Could not open archive for reading.");
+        }
+
+        stream_extract_archive(a, destination);
+
+        archive_read_close(a);
+        archive_read_free(a);
```
Klaim, line 379: This definitely need to be RAII-ed/scoped as there is a potential exception just before.
Consider using  the `mamba::on_scope_exit()` for this.

libmamba/src/core/package_handling.cpp
```diff
@@ -354,14 +356,53 @@ namespace mamba
...
+    }
+
+    namespace
+    {
+        struct conda_extract_context
+        {
+            conda_extract_context(struct archive* source)
```
Klaim, line 386: `struct` is unnecessary here
Also make this constructor `explicit` otherwise you will be able to pass ints to functions taking this.

libmamba/src/core/package_handling.cpp
Klaim, line 392: `struct` is unnecessary here

libmamba/src/core/package_handling.cpp
```diff
@@ -354,14 +356,53 @@ namespace mamba
...
         if (!fs::exists(destination))
         {
             fs::create_directories(destination);
         }
         fs::current_path(destination);
 
-        struct archive* a;
         struct archive* ext;
         struct archive_entry* entry;
         int flags;
```
JohanMabille, line 408: Same remark here, when possible we should avoid variable definition without immediatly initializing them.

libmamba/src/core/package_handling.cpp
```diff
@@ -373,32 +414,22 @@ namespace mamba
...
-        a = archive_read_new();
-        archive_read_support_format_tar(a);
-        archive_read_support_format_zip(a);
-        archive_read_support_filter_all(a);
+        if (Context::instance().extract_sparse)
+        {
+            flags |= ARCHIVE_EXTRACT_SPARSE;
+        }
 
         ext = archive_write_disk_new();
```
JohanMabille, line 424: `archive* ext = ...`

libmamba/src/core/package_handling.cpp
```diff
@@ -373,32 +414,22 @@ namespace mamba
...
-
         for (;;)
         {
             if (is_sig_interrupted())
             {
-                break;
+                throw std::runtime_error("SIGINT received. Aborting extraction.");
             }
 
             r = archive_read_next_header(a, &entry);
```
JohanMabille, line 435: `int r = ...`

libmamba/src/core/package_handling.cpp
Klaim: BTW is it ok if `source` is `nullptr`? If not consider adding an assertion in the body.

libmamba/src/core/package_handling.cpp
```diff
@@ -354,14 +356,53 @@ namespace mamba
...
+    {
         auto prev_path = fs::current_path();
         if (!fs::exists(destination))
         {
             fs::create_directories(destination);
         }
         fs::current_path(destination);
 
-        struct archive* a;
         struct archive* ext;
```
Klaim, line 406: Same remarkes as `extract_archive()`

libmamba/src/core/package_handling.cpp
```diff
@@ -354,14 +356,53 @@ namespace mamba
...
+            {
+            }
+
+            struct archive* source;
+            std::vector<char> buffer;
+        };
+    }
+
+    void stream_extract_archive(archive* a, const fs::u8path& destination)
+    {
```
Klaim, line 398: consider adding an assertion here if `a` must not be `nullptr`.

libmamba/src/core/package_handling.cpp
```diff
@@ -443,46 +474,87 @@ namespace mamba
...
         archive_write_close(ext);
         archive_write_free(ext);
 
         fs::current_path(prev_path);
     }
 
+
+    static la_ssize_t file_read(struct archive* a, void* client_data, const void** buff)
+    {
+        struct conda_extract_context* mine = static_cast<conda_extract_context*>(client_data);
```
Klaim, line 487: `struct` unnecessary`

Are the arguments allowed to be nullptr? if not add an assertion

libmamba/src/core/package_handling.cpp
```diff
@@ -443,46 +474,87 @@ namespace mamba
...
+        auto read = archive_read_data(mine->source, mine->buffer.data(), mine->buffer.size());
+        if (read < 0)
+        {
+            throw std::runtime_error(
+                fmt::format("Error reading from archive: {}", archive_error_string(mine->source)));
+        }
+        return read;
+    }
+
+    int archive_read_open_archive_entry(struct archive* a, conda_extract_context* ctx)
```
Klaim, line 499: Same remarkes as `file_read`

libmamba/src/core/package_handling.cpp
```diff
@@ -443,46 +474,87 @@ namespace mamba
...
+
     void extract_conda(const fs::u8path& file,
                        const fs::u8path& dest_dir,
                        const std::vector<std::string>& parts)
     {
-        TemporaryDirectory tdir;
-        extract_archive(file, tdir);
+        // open outer zip archive
+        struct archive* a = archive_read_new();
+        struct archive_entry* entry;
```
Klaim, line 514: `struct` unnecessary and these needs RAII/scoping (`mamba::on_scope_exit`) as there are potential exceptions below.

libmamba/src/core/package_handling.cpp
```diff
@@ -443,46 +474,87 @@ namespace mamba
...
+    static la_ssize_t file_read(struct archive* a, void* client_data, const void** buff)
+    {
+        struct conda_extract_context* mine = static_cast<conda_extract_context*>(client_data);
+        *buff = mine->buffer.data();
+
+        auto read = archive_read_data(mine->source, mine->buffer.data(), mine->buffer.size());
+        if (read < 0)
+        {
+            throw std::runtime_error(
+                fmt::format("Error reading from archive: {}", archive_error_string(mine->source)));
```
JohanMabille, line 494: Potential resource leak if the caller does not catch the exception in order to close and free the archive.

libmamba/src/core/package_handling.cpp
```diff
@@ -443,46 +474,87 @@ namespace mamba
...
         {
-            const auto condafile_path = tdir.path() / "metadata.json";
-            std::ifstream condafile_meta(condafile_path.std_path());
-            nlohmann::json j;
-            condafile_meta >> j;
+            fs::u8path p(archive_entry_pathname(entry));
+            if (p.extension() == ".zst")
+            {
+                // extract zstd file
+                struct archive* inner = archive_read_new();
```
Klaim, line 532: `struct`
RAII if any of the following functions relies on any callback (believe they do).

JohanMabille (commented):
I think the `archive_read/write_new` and `archive_read/write_free` should be wrapped in a guard object to guarantee resources are freed whatever the execution path is (RAII).

libmamba/src/core/package_handling.cpp
```diff
@@ -443,46 +474,87 @@ namespace mamba
...
+
+                archive_read_close(inner);
+                archive_read_free(inner);
+            }
+            else if (p.extension() == ".json")
             {
-                if (j["conda_pkg_format_version"] != 2)
+                std::size_t json_size = archive_entry_size(entry);
+                std::string json;
+                json.resize(json_size);
```
Klaim, line 546: This is equivalent to `std::string json(json_size, char{});` which is shorter.
However I'm not sure if that's exactly what you want here. `\0` is the end of a `const char*` but not the end of `std::string` so it might be confusing using that string afterwards. I suppose it's not a big deal if the next function do overwrites the whole thing, but make sure that the last `\0` is not part of the `std::string` otherwise there might be surprised for code using that one.

Klaim (commented):
I did a first surface review pass, there is a lot of old-version C code in there that should be converted to C++ to avoid issues in particular with exception usage (or any kind of error handling that might skips the resource releases).

libmamba/src/core/package_handling.cpp
```diff
@@ -373,32 +414,22 @@ namespace mamba
...
-            LOG_ERROR << "Error opening archive: " << archive_error_string(a);
-            throw std::runtime_error(file.string() + " : Could not open archive for reading.");
-        }
-
         for (;;)
         {
             if (is_sig_interrupted())
             {
-                break;
+                throw std::runtime_error("SIGINT received. Aborting extraction.");
```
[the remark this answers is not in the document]
Klaim: We have [this](https://github.com/mamba-org/mamba/blob/main/libmamba/include/mamba/core/util_scope.hpp) potentially to quickly fix this kind of issues.

libmamba/src/core/package_handling.cpp
wolfv: I fixed this by taking the raii object as input now.

libmamba/src/core/package_handling.cpp
wolfv: 👍

wolfv:
I've added explicit raii wrapper classes for `libarchive` classes and used them in most signatures instead of the previous `archive*` (except for the callback where it is not possible).

I would love if someone has time to go over the error handling with either a macro or a function like this: https://github.com/DDoSolitary/LxRunOffline/blob/bdc6d7d77f886c6dcdab3b4c9c136557ca6694c4/src/lib/fs.cpp#L30-L42

libmamba/src/core/package_handling.cpp
```diff
@@ -57,7 +58,104 @@ namespace mamba
...
+        {
+            return libarchive_read_raii(archive_read_disk_new());
+        }
+
+        ~libarchive_read_raii()
+        {
+            archive_read_free(m_archive);
+        }
+
+        operator archive*()
```
Klaim, line 76: make it `explicit` otherwise conversion to int are possible

also need a const version

Personally I would avoid operators altogether and just provide an accessor, either `operator*` or `archive* raw()` (or similar name) or both.
Same remark for the other types.

libmamba/src/core/package_handling.cpp
```diff
@@ -57,7 +58,104 @@ namespace mamba
...
+        {
+            return m_archive;
+        }
+
+    private:
+        libarchive_read_raii(archive* a)
```
Klaim, line 82: `explicit`

libmamba/src/core/package_handling.cpp
```diff
@@ -57,7 +58,104 @@ namespace mamba
...
+                throw std::runtime_error("Could not create libarchive entry object");
+            }
+        }
+
+        ~libarchive_entry_raii()
+        {
+            archive_entry_free(m_entry);
+        }
+
+        operator archive_entry*()
```
Klaim, line 111: `explicit`

libmamba/src/core/package_handling.cpp
```diff
@@ -57,7 +58,104 @@ namespace mamba
...
+        {
+            return libarchive_write_raii(archive_write_disk_new());
+        }
+
+        ~libarchive_write_raii()
+        {
+            archive_write_free(m_archive);
+        }
+
+        operator archive*()
```
Klaim, line 138: `explicit`

libmamba/src/core/package_handling.cpp
```diff
@@ -57,7 +58,104 @@ namespace mamba
...
+        {
+            return m_archive;
+        }
+
+    private:
+        libarchive_write_raii(archive* a)
```
Klaim, line 144: `explicit`

libmamba/src/core/package_handling.cpp
```diff
@@ -57,7 +58,104 @@ namespace mamba
...
+    {
+    public:
+        libarchive_read_raii()
+            : libarchive_read_raii(archive_read_new()){};
+        static libarchive_read_raii read_disk()
+        {
+            return libarchive_read_raii(archive_read_disk_new());
+        }
+
+        ~libarchive_read_raii()
```
Klaim, line 71: While having a user-defined destructor deactivates the copy and move operators, could you add them explicitely `= delete;` so that it's clear that it's the intention? (until they are properly implemented later)
This is just for clarity.
Same remark for the other types.

libmamba/src/core/package_handling.cpp
```diff
@@ -57,7 +58,104 @@ namespace mamba
         const fs::u8path& m_file;
     };
 
-    static int copy_data(archive* ar, archive* aw)
+    class libarchive_read_raii
```
Klaim, line 61: I don't think having `raii` in the name is useful, nor is having the `lib` prefix. Suggestion: `scoped_archive_read` (same suggestion for the other names)

libmamba/src/core/package_handling.cpp
```diff
@@ -57,7 +58,104 @@ namespace mamba
...
+                throw std::runtime_error("Could not create libarchive write object");
+            }
+        }
+
+        archive* m_archive;
+    };
+
+    void stream_extract_archive(libarchive_read_raii& a, const fs::u8path& destination);
+
+    static int copy_data(libarchive_read_raii& ar, libarchive_write_raii& aw)
```
Klaim, line 158: I'm surprised that they are both modified?

libmamba/src/core/package_handling.cpp
Klaim: +1

libmamba/src/core/package_handling.cpp
```diff
@@ -354,51 +437,74 @@ namespace mamba
...
+            LOG_ERROR << "Error opening archive: " << archive_error_string(a);
+            throw std::runtime_error(file.string() + " : Could not open archive for reading.");
+        }
+
+        stream_extract_archive(a, destination);
+    }
+
+    namespace
+    {
+        struct conda_extract_context
```
Klaim, line 461: Consider making this type non-copyable explicitly (`= delete;`)

libmamba/src/core/package_handling.cpp
```diff
@@ -354,51 +437,74 @@ namespace mamba
...
         {
             fs::create_directories(destination);
         }
         fs::current_path(destination);
 
-        struct archive* a;
-        struct archive* ext;
-        struct archive_entry* entry;
-        int flags;
+        archive_entry* entry;
```
Klaim, line 483: Avoid uninitialized variables

libmamba/src/core/package_handling.cpp
wolfv: Using an explict accessor would modify a lot of functions and that's not the goal of this PR. Let's do that later if we want to.

libmamba/src/core/package_handling.cpp
wolfv: hmm not sure I like the amount of boilerplate code that adds...

libmamba/src/core/package_handling.cpp
wolfv: actually the `explicit` doesn't work for `archive*` operator because it can't be converted to `struct archive*` ...

libmamba/src/core/package_handling.cpp
wolfv: yes, e.g. the `a->error` is set on the archive if either read or write fails.

libmamba/src/core/package_handling.cpp
wolfv: done

libmamba/src/core/package_handling.cpp
wolfv: I need that pointer to retrieve the value from 

```
archive_read_next_header(a, &entry)
```

libmamba/src/core/package_handling.cpp
Klaim: ok for separate pr though I dont see how changing it here would change previously written functions?

> actually the explicit doesn't work for archive* operator because it can't be converted to struct archive*

I didnt understand what you mean here?

libmamba/src/core/package_handling.cpp
```diff
@@ -443,45 +547,102 @@ namespace mamba
...
+
+
     void extract_conda(const fs::u8path& file,
                        const fs::u8path& dest_dir,
                        const std::vector<std::string>& parts)
     {
-        TemporaryDirectory tdir;
-        extract_archive(file, tdir);
+        // open outer zip archive
+        archive_entry* entry;
```
Klaim, line 583: Move that line close the the `while` where it's used.

Klaim (approved):
LGTM except we really need that separate PR with various checks

PR #2201, opened 2023-01-04
Fixed static dependency order

JohanMabille:


wolfv:
noice!

wolfv (approved):

JohanMabille:
arg stil one failure ;(

wolfv:
that migth be an actual network issue. I would say let's merge and see if we see that issue again?

PR #2211, opened 2023-01-09
try new shebang style

wolfv:
cc @jaimergp pls let me know what you think :)

### DIFF
--- a/libmamba/include/mamba/core/link.hpp
+++ b/libmamba/include/mamba/core/link.hpp
@@ -23,6 +23,16 @@
 namespace mamba
 {
     std::string replace_long_shebang(const std::string& shebang);
+    std::string python_shebang(const std::string& python_exe);
+
+    static const std::regex shebang_regex(
+        "^(#!"                      // pretty much the whole match string
+        "(?:[ ]*)"                  // allow spaces between #! and beginning of the executable path
+        "(/(?:\\\\ |[^ \n\r\t])*)"  // the executable is the next text block without an escaped
+                                    // space or non-space whitespace character
+        "(.*))$");  // the rest of the line can contain option flags and end whole_shebang group
+
+    constexpr size_t MAX_SHEBANG_LENGTH = on_linux ? 127 : 512;
 
     struct python_entry_point_parsed
     {
--- a/libmamba/src/core/link.cpp
+++ b/libmamba/src/core/link.cpp
@@ -102,9 +102,11 @@ namespace mamba
                                   // or non-space whitespace character
         "(.*))$");                // end whole_shebang group
 
+    constexpr size_t MAX_SHEBANG_LENGTH = on_linux ? 127 : 512;
+
     std::string replace_long_shebang(const std::string& shebang)
     {
-        if (shebang.size() <= 127)
+        if (shebang.size() <= MAX_SHEBANG_LENGTH)
         {
             return shebang;
         }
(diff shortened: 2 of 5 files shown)


jaimergp (approved):
I am a bit lost in the `"` soup but I think that's it, yes :) LGTM.

wolfv:
@jaimergp do you know why this

https://github.com/conda/conda/blob/48f51e6c1d412270efbbdb1d9ff571087568b6ea/conda/core/portability.py#L233-L238

```
if len(whole_shebang) > MAX_SHEBANG_LENGTH or "\\ " in prefix:
```

Isn't `\\ ` going to be just fine (escaped whitespace?

libmamba/include/mamba/core/link.hpp
```diff
@@ -23,6 +23,16 @@
 namespace mamba
 {
     std::string replace_long_shebang(const std::string& shebang);
+    std::string python_shebang(const std::string& python_exe);
+
+    static const std::regex shebang_regex(
+        "^(#!"                      // pretty much the whole match string
+        "(?:[ ]*)"                  // allow spaces between #! and beginning of the executable path
+        "(/(?:\\\\ |[^ \n\r\t])*)"  // the executable is the next text block without an escaped
```
wolfv, line 31: previously we had a bug with no doubly escaped `\\`

wolfv:
tested and indeed, python entry points work better like this! :) Thanks for figuring this all out, @jaimergp :)

libmamba/include/mamba/core/link.hpp
```diff
@@ -23,6 +23,16 @@
...
+                                    // space or non-space whitespace character
+        "(.*))$");  // the rest of the line can contain option flags and end whole_shebang group
+
+    constexpr size_t MAX_SHEBANG_LENGTH = on_linux ? 127 : 512;
```
JohanMabille, line 35: This is an implementation detial and should live in the cpp file.

libmamba/include/mamba/core/link.hpp
```diff
@@ -23,6 +23,16 @@
 namespace mamba
 {
     std::string replace_long_shebang(const std::string& shebang);
+    std::string python_shebang(const std::string& python_exe);
```
JohanMabille, line 26: These functions look like implementation functions (there are not called anywhere else in the library and I'm not sure they will), they could be moved to the cpp file (and forward declared in the test). We can also do a pass on this during the refactoring of Mamba APIs.

libmamba/include/mamba/core/link.hpp
JohanMabille: We shoudl avoid global variables in headers and prefer functions to get them (to avoid a possible static initialization order fiasco).

libmamba/include/mamba/core/link.hpp
wolfv: that was like that before, but Windows complained in the test that it couldn't find the symbol :/

wolfv:
I basically only moved some stuff to the header to be able to test it. lmk if you know a good way to do that. tried `extern std::regex shebang_regex` but that didn't work :)

wolfv:
@jaimergp hmm, I have a remaining question -- do we need to enforce that conda / mamba are _not_ using the new shebang style (special casing `bin/conda` and `bin/mamba`) to make sure that $PS1 is properly passed through?

jaimergp:
> Isn't `\\ ` going to be just fine (escaped whitespace?

There's no such guarantee 😬 The kernel doesn't have a notion of escaped spaces, sadly. See https://github.com/conda/conda/issues/186.

wolfv:
Good to know! 

Do you think we should special case `bin/conda`?

jaimergp:
It wouldn't hurt! That said, it should only be a problem with initialized `conda`s. And the overlap of people needed an initialized `conda` and people installing it with (micro)mamba is really small, I'd say?

PR #2214, opened 2023-01-10
use local meta.yaml

wolfv:
try things differently


wolfv:
With this change we might have to sometimes manually sync the recipe back to this repository (or ideally, the other way around!).

jonashaag:
cc @pavelzw

JohanMabille:
This should not be very different form a regular process where you patch the recipe when you have fixes in the source repo and you don't want to release yet. The only difference is that we need to sync the dependencies in the local meta.yaml, but thats all, no more pathching.

Issue #2164, opened 2022-12-12
Feature request: soft version matching in micro.mamba.pm


maresb:
It would be nice to be able to use the link http://micro.mamba.pm/api/micromamba/linux-64/1.1 to download the latest v1.1.X.

I would open a PR to do this if the webservice were open-source :wink:

maresb:
We are also working on various installation script issues in https://github.com/eitsupi/mamba-devcontainer-features/pull/3. It would also be nice to achieve a unified `install.sh` which handles more edge cases, e.g. if the bzip2 package is missing.

jonashaag:
> if the bzip2 package is missing.

I'm working on something like this atm, will share soon

maresb:
@jonashaag, then you might want to take a quick look at the already-completed work [here](https://github.com/eitsupi/mamba-devcontainer-features/tree/questions/src/micromamba). The main body is `install.sh`, and the helper functions are `utils.sh`. Similarly, there are some really useful helper functions [here](https://github.com/devcontainers/features/blob/main/src/nix/utils.sh#L49-L65) which we haven't yet adapted for our scripts.

jonashaag:
Here's a version that does not require bzip2 and supports custom version:


```
#!/bin/bash

set -eu

ARCH=$(uname -m)
OS=$(uname)

if [[ "$OS" == "Linux" ]]; then
  PLATFORM="linux"
  if [[ "$ARCH" == "aarch64" ]]; then
    ARCH="aarch64"
  elif [[ $ARCH == "ppc64le" ]]; then
    ARCH="ppc64le"
  else
    ARCH="64"
  fi    
elif [[ "$OS" == "Darwin" ]]; then
  PLATFORM="osx"
  if [[ "$ARCH" == "arm64" ]]; then
    ARCH="arm64"
  else
    ARCH="64"
  fi
elif [[ "$OS" =~ "NT" ]]; then
  PLATFORM="win"
  ARCH="64"
else
  echo "Failed to detect your OS" >&2
  exit 1
fi

if [[ "${VERSION:-}" == "" ]]; then
  RELEASE_URL=https://github.com/mamba-org/boa-forge/releases/latest/download/micromamba-$PLATFORM-$ARCH
else
  RELEASE_URL=https://github.com/mamba-org/boa-forge/releases/download/micromamba-$VERSION/micromamba-$PLATFORM-$ARCH
fi

BIN_FOLDER=~/.local/bin
[...]

PR #2218, opened 2023-01-12
Update micromamba installation docs for Windows

Tiksagol:
No need to use 7zip as Windows 10 contains tar.exe (and the former might not be installed).

### DIFF
--- a/docs/source/installation.rst
+++ b/docs/source/installation.rst
@@ -158,8 +158,7 @@ Windows
 .. code:: powershell
 
   Invoke-Webrequest -URI https://micro.mamba.pm/api/micromamba/win-64/latest -OutFile micromamba.tar.bz2
-  C:\PROGRA~1\7-Zip\7z.exe x micromamba.tar.bz2 -aoa
-  C:\PROGRA~1\7-Zip\7z.exe x micromamba.tar -ttar -aoa -r Library\bin\micromamba.exe
+  tar xf micromamba.tar.bz2
 
   MOVE -Force Library\bin\micromamba.exe micromamba.exe
   .\micromamba.exe --help


wolfv:
@Tiksagol does it really? I haven't really seen that in the wild (except on CI systems that is indeed the case).

I think we're working more towards putting the non-compressed binary somewhere so that installation becomes even simpler.

jonashaag:
To that end, want to deploy this script? https://github.com/mamba-org/mamba/issues/2164#issuecomment-1359564081

Tiksagol:
> @Tiksagol does it really? I haven't really seen that in the wild (except on CI systems that is indeed the case).
> 
> I think we're working more towards putting the non-compressed binary somewhere so that installation becomes even simpler.

@wolfv - yes, works fine on Windows 10

jonashaag:
I think it ships bsdtar. You can see usage in the provision-with-micromamba code, which is compatible with bare bones Windows systems.

wolfv:
This article is from April 22 https://learn.microsoft.com/en-us/virtualization/community/team-blog/2017/20171219-tar-and-curl-come-to-windows

So I guess it's true! I'm fine with changing it to curl + tar in that case and keep the old one as reference for older systems.

Tiksagol:
> I think it ships bsdtar. You can see usage in the provision-with-micromamba code, which is compatible with bare bones Windows systems.

@jonashaag - right:

`C:\Users\me>tar --version
bsdtar 3.5.2 - libarchive 3.5.2 zlib/1.2.5.f-ipp bz2lib/1.0.6`

wolfv:
Hmmmmm, can you double check this for me? 

On a windows server installation I can access, teh windows shipped `tar` doesn't have `bzip2` support (see screenshot).

Screenshot 2023-01-22 at 01 13 07

Tiksagol:
> Hmmmmm, can you double check this for me?
> 
> On a windows server installation I can access, teh windows shipped `tar` doesn't have `bzip2` support (see screenshot).
> 
> Screenshot 2023-01-22 at 01 13 07

Seems like it has, as in my laptop it works and returned the same output (its end seems to be removed from your screenshot) - 
```
c:\Windows\System32>.\tar.exe --version
bsdtar 3.5.2 - libarchive 3.5.2 zlib/1.2.5.f-ipp **bz2lib/1.0.6**
```

PR #2228, opened 2023-01-17
Fix Windows static builds

jonashaag:


jonashaag:
FYI you can also use `if: false` to skip a job

wolfv:
idk why sed had these super weird problems but anyways, this hacky solution seems to be working OKish.

wolfv (approved):

PR #2229, opened 2023-01-17
add functionality to download lockfiles from internet

wolfv:


libmamba/src/api/install.cpp
```diff
@@ -656,12 +656,33 @@ namespace mamba
                 }
             }
 
+            static std::unique_ptr<TemporaryFile> tmp_lock_file;
```
JohanMabille, line 659: Why `static`? this makes the feature unusable in a library context since it will be destroyed at the program exit only.

JohanMabille (changes requested):

libmamba/src/api/install.cpp
wolfv: in the micromamba context, if it gets destroyed before, the file is gone. Let's look for a better solution :)

JohanMabille (approved):

Issue #2054, opened 2022-10-26
Implement `install --export`


jonashaag:
We'd like to automatically generate an `environment.yml` during an `install`, with the same format as exported by `env export`.

We want this so that we can create a fully solved `environment.yml` without actually installing by combining with `--dry-run`.

jonashaag:
Should it be `micromamba create --export` instead?

PR #2236, opened 2023-01-18
Implement `install --export`

jonashaag:
Very early version of https://github.com/mamba-org/mamba/issues/2054.

Mostly here for design feedback from @wolfv @Klaim @AntoinePrv and whoever else is willing to give feedback!

My initial idea was to retrieve the `PrefixData` to construct the environment, as is done in `env.cpp` for `env export`. But that doesn't work if you use dry-run (our primary use case) because it relies on the `conda-meta/*.json` files, which are only present after a non-dry-run installation.

Next idea is to return the `MTransaction` from `install_specs`, but I was incapable of returning it because the solver (and maybe other things?) are stack allocated and the pointers will be invalidated even if `MTransaction` is returned as a shared pointer. Thus, I've added an `printEnv` function to `MTransaction`.

There are a few unrelated cleanup changes to use `pool_id2solvable` where possible.

### DIFF
--- a/libmamba/src/core/transaction.cpp
+++ b/libmamba/src/core/transaction.cpp
@@ -37,6 +37,24 @@ namespace
 
 namespace mamba
 {
+    // todo dedup
+    std::string get_env_name(const fs::u8path& px)
+    {
+        const auto& ctx = Context::instance();
+        auto& ed = ctx.envs_dirs[0];
+        if (px == ctx.root_prefix)
+        {
+            return "base";
+        }
+        else if (mamba::starts_with(px.string(), ed.string()))
+        {
+            return fs::relative(px, ed).string();
+        }
+        else
+        {
+            return "";
+        }
+    }
     nlohmann::json solvable_to_json(Solvable* s)
     {
         return PackageInfo(s).json_record();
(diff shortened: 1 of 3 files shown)


jonashaag:
@wolfv @AntoinePrv @Klaim would be great to get some feedback here!

AntoinePrv:
For API `micromamba env export` directly prints the environment packages, but `micromamba install [...]` would print the installation details as well. What it the goal, to have it print only the new environment packages in the environment.yml format and skip the rest?

Would it make sense to use `install --export` without dry-run?

libmamba/src/core/transaction.cpp
```diff
@@ -1562,6 +1580,60 @@ namespace mamba
         t.print(out);
     }
 
+    void MTransaction::printEnv()
```
AntoinePrv, line 1583: Can we make this return an object (`std::string` or even `nl::json`) instead of printing (which we're working on keeping out of libmamba)?

libmamba/src/core/transaction.cpp
```diff
@@ -1562,6 +1580,60 @@ namespace mamba
...
+            /* todo need this?
+            Id real_repo_key = pool_str2id(pool, "solvable:real_repo_url", 1);
+            if (solvable_lookup_str(s, real_repo_key))
+            */
+
+            channels.insert(make_channel(s->repo->name).base_url());
+
+            auto name = pool_id2str(pool, s->name);
+            auto version = pool_id2str(pool, s->evr);
+            auto build_string = solvable_lookup_str(s, SOLVABLE_BUILDFLAVOR);
```
AntoinePrv, line 1624: Is there an auxiliary function to do this? Should there be one?

AntoinePrv (commented):
it would be nice if we could reuse some of the code from `env export`. Perhaps we need to move it form micromamba to libmamba

PR #2241, opened 2023-01-21
Install pip deps like conda

### LINKED ISSUE #1894: Problems with installing pip's `-r` dependencies
It seems like `micromamba` (and quite possibly `mamba` too) can't install `pip` dependencies specified in the `environment.yml` the same way as `conda` does.
Suppose we have following dependencies in `environment.yml`:
```yaml
dependencies:
  - pip
  - pip:
    - -r requirements.txt
    - -e deps/my_local_package
```
It is now impossible to install such dependencies with `pip` and therefore with `micromamba`.
And that is because `micromamba` creates its own temporary `requirements.txt` file in `/tmp` directory (https://github.com/mamba-org/mamba/blob/master/libmamba/src/api/install.cpp#L95-L100) and calls `pip` to install them.
For some reason `pip` resolves `-r` paths as relative to file that contains them, but `-e` paths as relative to current working directory.
So in this case it looks for `/tmp/requirements.txt` and `$PWD/deps/my_local_package` and unless whole project is thrown right into `/tmp` this installation will fail.

I found existing issue in `pip` repo that describes this behaviour: https://github.com/pypa/pip/issues/8765
And here are relevant code pieces from pip repo:
- [this one](https://github.com/pypa/pip/blob/main/src/pip/_internal/req/constructors.py#L76-L78) uses `os.path.normpath(os.path.abspath(path))` to create absolute path relative to CWD for `-e` paths
- [and this one](https://github.com/pypa/pip/blob/main/src/pip/_internal/req/req_file.py#L355-L357) - uses `os.path.join` to simply join dirname of `requirements.txt` file containing `-r` path.

I checked `conda` behavior in some of those cases and it looks like `conda` does the heavy lifting and does not let `pip` resolve those paths. At least in my tests when creating environment with `conda`, all paths were treated as relative to file they were placed in.

I also found issue https://github.com/mamba-org/mamba/pull/1160 from last year, but I am not sure if and how relevant it is to the problem with `-r` dependencies.

### Current behavior
`micromamba` can't install `pip` dependencies if they contain `-r requirements.txt` and whole project isn't directly in `/tmp`

### Expected behavior
At least somewhat similar to `conda`, so one can use `-r` dependency in `environment.yml` without issues

### Proposed solution
I think that creating this temporary `requirements.txt` file next to `environment.yml` file is a viable solution.
It shouldn't break other types of dependencies and `-r` dependencies were broken anyway so nothing bad should happen.

michalsieron:
Fixes #1894

### DIFF
--- a/libmamba/include/mamba/core/util.hpp
+++ b/libmamba/include/mamba/core/util.hpp
@@ -94,7 +94,9 @@ namespace mamba
     class TemporaryFile
     {
     public:
-        TemporaryFile(const std::string& prefix = "mambaf", const std::string& suffix = "", const std::string& dir = "");
+        TemporaryFile(const std::string& prefix = "mambaf",
+                      const std::string& suffix = "",
+                      const std::string& dir = "");
         ~TemporaryFile();
 
         TemporaryFile(const TemporaryFile&) = delete;
--- a/libmamba/include/mamba/core/util.hpp
+++ b/libmamba/include/mamba/core/util.hpp
@@ -96,7 +96,7 @@ namespace mamba
     public:
         TemporaryFile(const std::string& prefix = "mambaf",
                       const std::string& suffix = "",
-                      const std::string& dir = "");
+                      const std::optional<fs::u8path>& dir = std::nullopt);
         ~TemporaryFile();
 
         TemporaryFile(const TemporaryFile&) = delete;
--- a/libmamba/include/mamba/core/util.hpp
+++ b/libmamba/include/mamba/core/util.hpp
@@ -94,7 +94,7 @@ namespace mamba
     class TemporaryFile
     {
     public:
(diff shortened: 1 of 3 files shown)


jonashaag:
Thanks for the PR! Can you please add a test for this change?

michalsieron:
I think I fixed the formatting errors, however I am not sure where I should write those tests.
My changes are in libmamba and I can't find tests related to resolving and installing pip dependencies. Such test would need to start pip subprocess to test if it was successful in installing the dependencies specified in the temporary file.

jonashaag:
You could add a test in the micromamba Python tests, maybe here? https://github.com/mamba-org/mamba/blob/main/micromamba/tests/test_create.py Feel free to create additional test files. I just saw that there are `spec_file_{1,2}.txt` but they seem unused.

wolfv (approved):

PR #2257, opened 2023-02-01
[micromamba] Stop run command when given prefix does not exist

### LINKED ISSUE #2084: `micromamba run` with missing env still attempts to execute the command
```
$ micromamba run -n xxx python
warning  libmamba 'root_prefix' set with default value: /Users/j/micromamba
critical libmamba Cannot activate, prefix does not exist at: /Users/j/micromamba/envs/xxx
/var/folders/8v/x6lkdp6110d5w7z73g_jfm_00000gn/T/mambafjhLFdy8hm4: line 5: exec: python: not found
```

Env is missing so cannot be activated. Still the command is attempted to be executed as can be seen from the `exec` error message. This can lead to confusing behaviour if the command to be executed exists outside the environment (eg. system Python in this case). We should just stop after the critical error.

Hind-M:
Fix #2084 
The exception thrown during env activation is indeed not caught. Since `reproc` has the upper hand and doesn't catch it, I added a check at the beginning of `run_in_environment` function.

### DIFF
--- a/libmamba/src/core/run.cpp
+++ b/libmamba/src/core/run.cpp
@@ -254,7 +254,8 @@ namespace mamba
     {
         if (!fs::exists(Context::instance().target_prefix))
         {
-            LOG_CRITICAL << "The given prefix does not exist: " << Context::instance().target_prefix;
+            LOG_CRITICAL << "The given prefix does not exist: "
+                         << Context::instance().target_prefix;
             return 1;
         }
         std::vector<std::string> raw_command = command;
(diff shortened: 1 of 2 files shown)


jonashaag:
LGTM, can you please add a test?

Hind-M:
> LGTM, can you please add a test?

Done!

jonashaag (approved):
Thanks! Are the test failures related?

Hind-M:
> Are the test failures related?
I don't think so because I get the same failures in another PR, but not sure enough.
@wolf @klaim are the test failures related to something specific?

jonashaag:
Ah, it's because of `.tar.bz2` -> `.conda`

jonashaag:
See https://github.com/mamba-org/mamba/pull/2258

PR #2258, opened 2023-02-02
Fix .tar.gz URLs in tests

jonashaag:


wolfv:
nice, thanks!

jonashaag:
Hmm seems like that wasn't the right place

wolfv:
check out #2259

PR #2260, opened 2023-02-02
Fix `micromamba env export` to get channel name instead of full url

### LINKED ISSUE #2235: `micromamba env export` exports full channel URLs
### Troubleshooting docs

env export channels

### Latest version of Mamba

### Tried in Conda?

Not reproducible with Conda

### Describe your issue

```
$ mamba env export | head
name: mamba-dev
channels:
  - conda-forge
...

$ micromamba env export | head
name: mamba-dev
channels:
- https://conda.anaconda.org/conda-forge
...
```

_No response_

### Logs

_No response_

### environment.yml

_No response_

### ~/.condarc

_No response_

Hind-M:
Fix #2235
In case of `pkgs/main`, `defaults` is used to be consistent with conda.

### DIFF
--- a/micromamba/src/env.cpp
+++ b/micromamba/src/env.cpp
@@ -120,8 +120,15 @@ set_env_command(CLI::App* com)
                             dependencies << "=" << v.build_string;
                         dependencies << "\n";
                     }
-
-                    channels.insert(make_channel(v.url).base_url());
+                    const auto& chan_name = make_channel(v.url).name();
+                    if (chan_name == "pkgs/main")
+                    {
+                        channels.insert("defaults");
+                    }
+                    else
+                    {
+                        channels.insert(chan_name);
+                    }
                 }
 
                 for (const auto& c : channels)
--- a/micromamba/src/env.cpp
+++ b/micromamba/src/env.cpp
@@ -120,15 +120,8 @@ set_env_command(CLI::App* com)
                             dependencies << "=" << v.build_string;
                         dependencies << "\n";
                     }
-                    const auto& chan_name = make_channel(v.url).name();
-                    if (chan_name == "pkgs/main")
-                    {
-                        channels.insert("defaults");
-                    }
-                    else
-                    {
-                        channels.insert(chan_name);
-                    }
+
+                    channels.insert(make_channel(v.url).name());
                 }
 
                 for (const auto& c : channels)
--- a/micromamba/tests/test_env.py
+++ b/micromamba/tests/test_env.py
@@ -120,7 +120,7 @@ class TestEnv:
         create("", "-n", env_name, "-f", spec_file)
         ret = yaml.safe_load(run_env("export", "-n", env_name))
         assert ret["name"] == env_name
-        assert set(ret["channels"]) == {"https://conda.anaconda.org/conda-forge"}
+        assert set(ret["channels"]) == {"conda-forge"}
         assert "micromamba=0.24.0=0" in ret["dependencies"]
 
     def test_create(self):


jonashaag:
> In case of `pkgs/main`, `defaults` is used to be consistent with conda.

Hmm... is this a special case in the Conda source code? If so, could you post a link to the piece of code that does it? I'm curious to see how exactly it is implemented.

Hind-M:
Actually, I just run `conda env export` in a test environment which printed:
```
channels:
  - conda-forge
  - defaults
```
And looking at the Conda source code, I found this:
https://github.com/conda/conda/blob/fc92ea3353cce71df24fac31ce333ec9f8cf7ba7/tests/models/test_index_record.py#L30-L31
But didn't find the explicit implementation of it...

jonashaag:
I'd be interested if Conda always prints the name of the channel that has been used to install (like `--from-history`), or the name of the multichannel.

micromamba/src/env.cpp
```diff
@@ -120,8 +120,15 @@ set_env_command(CLI::App* com)
                             dependencies << "=" << v.build_string;
                         dependencies << "\n";
                     }
-
-                    channels.insert(make_channel(v.url).base_url());
+                    const auto& chan_name = make_channel(v.url).name();
+                    if (chan_name == "pkgs/main")
+                    {
+                        channels.insert("defaults");
+                    }
```
AntoinePrv, line 127: It looks like using either `pkgs/main` or `defaults` works so I would be conservative and say we do not need to change? I'm no expert here. What do you think?

In the refactorign reflection, this should go in `Channel`, eventually through a new method name like `cannonical_name` so that we don't manage edge cases everywhere.

micromamba/tests/test_env.py
```diff
@@ -120,7 +120,7 @@ def test_env_export(self):
         create("", "-n", env_name, "-f", spec_file)
         ret = yaml.safe_load(run_env("export", "-n", env_name))
         assert ret["name"] == env_name
-        assert set(ret["channels"]) == {"https://conda.anaconda.org/conda-forge"}
+        assert set(ret["channels"]) == {"conda-forge"}
```
AntoinePrv, line 123: If we do add that the `defaults` naming, we should also add a test for it, so that we do not loose it through a refactoring.

micromamba/src/env.cpp
Hind-M: Yes we could leave `pkgs/main` then (less prone to bugs and avoid edge cases). And yes, I agree that we should use something like `canonical_name` like Conda does IIRC.

micromamba/tests/test_env.py
Hind-M: Indeed.

AntoinePrv (approved):

jonashaag:
> I'd be interested if Conda always prints the name of the channel that has been used to install (like `--from-history`), or the name of the multichannel.

Did anyone check this?

Hind-M:
> > I'd be interested if Conda always prints the name of the channel that has been used to install (like `--from-history`), or the name of the multichannel.
> 
> Did anyone check this?

Nope. I've looked a bit into their code base but it wasn't obvious to spot and have the full picture. I would probably spend more time on it later if no one else checks it or has already some prior background to narrow the search.

jonashaag:
OK! Should we wait until we have a definite answer? I guess solving this issue isn't urgent, and if we change behavior twice that might be confusing to users.

Hind-M:
Yes, that seems reasonable!

jonashaag:
I had a look:

```
$ cat ~/.condarc
channels:
- defaults
- conda-forge

custom_multichannels:
  my_multi:
  - conda-forge

$ mamba create -n t1 conda-forge::zlib
$ mamba create -n t2 defaults::zlib  # => does not work
$ mamba create -n t2 pkgs/main::zlib
$ mamba create -n t3 my_multi::zlib

$ conda env export -n t1
name: t1
channels:
  - my_multi
  - defaults
  - conda-forge
...

$ conda env export -n t2
name: t2
channels:
  - defaults
  - conda-forge
  # => NOTE: No my_multi
...

$ conda env export -n t3
name: t3
channels:
  - my_multi
  - defaults
  - conda-forge
...

$ micromamba env export -p /opt/conda/envs/t1
name:
channels:
- https://conda.anaconda.org/conda-forge
- https://repo.anaconda.com/pkgs/main
# => NOTE: No my_multi, but mamba env export has it
...

$ micromamba env export -p /opt/conda/envs/t2
name:
channels:
- https://repo.anaconda.com/pkgs/main
# => NOTE: No conda-forge and my_multi, but mamba env export have it
...

$ micromamba env export -p /opt/conda/envs/t3
name:
channels:
- https://conda.anaconda.org/conda-forge
- https://repo.anaconda.com/pkgs/main
# => NOTE: No my_multi, but mamba env export has it
...
```

Now with `--from-history`:

```
$ conda env export -n t1 --from-history
name: t1
channels:
  - defaults
  - conda-forge
dependencies:
  - conda-forge::zlib
prefix: /opt/conda/envs/t1

$ conda env export -n t2 --from-history
name: t2
channels:
  - defaults
  - conda-forge
dependencies:
  - pkgs/main::zlib
prefix: /opt/conda/envs/t2

$ conda env export -n t3 --from-history
name: t3
channels:
  - defaults
  - conda-forge
dependencies:
  - my_multi::zlib
prefix: /opt/conda/envs/t3

$ micromamba env export -p /opt/conda/envs/t1 --from-history
name:
channels:
- https://conda.anaconda.org/conda-forge
# => NOTE: No defaults, but mamba env export has it
dependencies:
- conda-forge::zlib

$ micromamba env export -p /opt/conda/envs/t2 --from-history
name:
channels:
- https://repo.anaconda.com/pkgs/main
# => NOTE: No conda-forge, but mamba env export has it
dependencies:
- pkgs/main::zlib

$ micromamba env export -p /opt/conda/envs/t3 --from-history
name:
channels:
- https://conda.anaconda.org/conda-forge
# => NOTE: No defaults, but mamba env export has it
dependencies:
- my_multi::zlib
```

IIUC we should always print the *multichannel name* in `channels:`, and not special-case `defaults`.

jonashaag:
Interestingly, the order of channels is different with Micromamba. Is that a bug? cc @wolfv

jonashaag (approved):
Agreed!

PR #2266, opened 2023-02-06
docs: add micromamba docker image

wholtz:
In the micromamba installation docs, add that it can be run from a docker image


wholtz:
As discussed in https://github.com/mamba-org/mamba/pull/2126#issuecomment-1319630218

wolfv (approved):

maresb:
I think there is potential here for confusion. The first interpretation which comes to my mind when reading

> can be used to run `micromamba` without installing it

is that the Docker image could be used as an installation alternative, namely that the Docker image could be used to install packages on the host system.

Perhaps it would be better to say something like

> provides `micromamba` pre-installed on various base images.

PR #2272, opened 2023-02-06
Fix #2271

### LINKED ISSUE #2271: `micromamba shell -n myenv` broken in 1.3.0
### Troubleshooting docs

/

### Latest version of Mamba

### Tried in Conda?

Not applicable

### Describe your issue

```
⋊> ~ micromamba shell -n base
critical libmamba The given prefix does not exist: "base"
⋊> ~ micromamba run -n base ls
...works fine...
```

Curious why this isn't covered by a test

_No response_

### Logs

_No response_

### environment.yml

_No response_

### ~/.condarc

_No response_

jonashaag:
Attempt to fix https://github.com/mamba-org/mamba/issues/2271 (introduced in #2257, cc @Hind-M)

- Adds a test
- Makes exit codes identical to `run` and `conda run`
- Doesn't actually fix the bug, because I wasn't sure where — @Hind-M any insights?

### DIFF
--- a/libmamba/src/core/run.cpp
+++ b/libmamba/src/core/run.cpp
@@ -391,7 +391,7 @@ namespace mamba
             if (ec)
             {
                 std::cerr << ec.message() << '\n';
-                return 1;
+                return 127;
             }
 
 #ifndef _WIN32
--- a/libmamba/src/core/run.cpp
+++ b/libmamba/src/core/run.cpp
@@ -391,7 +391,7 @@ namespace mamba
             if (ec)
             {
                 std::cerr << ec.message() << '\n';
-                return 127;
+                return 1;
             }
 
 #ifndef _WIN32
--- a/libmamba/src/core/activation.cpp
+++ b/libmamba/src/core/activation.cpp
@@ -721,7 +721,9 @@ namespace mamba
             }
         }
 
-        if (Context::instance().auto_activate_base)
+        // if we are in a `mamba shell -n <env>` we don't want to activate base
+        auto has_prefix = env::get("CONDA_PREFIX");
+        if (Context::instance().auto_activate_base && !has_prefix.has_value())
         {
             builder << "micromamba activate base\n";
         }
--- a/micromamba/src/shell.cpp
+++ b/micromamba/src/shell.cpp
@@ -84,13 +84,14 @@ set_shell_command(CLI::App* subcom)
             auto& action = config.at("shell_action").compute().value<std::string>();
             auto& shell = config.at("shell_type").compute().value<std::string>();
             auto& stack = config.at("shell_stack").compute().value<bool>();
+
             if (action.empty())
             {
                 if (prefix.empty() || prefix == "base")
-                    Context::instance().target_prefix = env::home_directory() / "micromamba";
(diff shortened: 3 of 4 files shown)


libmamba/src/core/run.cpp
```diff
@@ -391,7 +391,7 @@ namespace mamba
             if (ec)
             {
                 std::cerr << ec.message() << '\n';
-                return 1;
```
jonashaag, line 394: Not sure about this.

Hind-M:
Oh I see, that's because `micromamba shell` doesn't go through `configuration` where the `target_prefix` is set:
https://github.com/mamba-org/mamba/blob/main/libmamba/src/api/configuration.cpp#L447-L465
So adding something like this piece of code could do the trick:

```
if (prefix.empty() || prefix == "base") // could be only base depending of the behavior that we want
    Context::instance().target_prefix = env::home_directory() / "micromamba";
else
    Context::instance().target_prefix = env::home_directory() / "micromamba/envs/" / prefix;
```
as a replacement of https://github.com/mamba-org/mamba/blob/main/micromamba/src/shell.cpp#L89

Hind-M:
I am not sure though if we should do it this way in case using the shell isn't meant to manipulate/handle the prefix... If so, maybe checking that the `target_prefix` got in `run_in_environment` is a path before checking its existence...
But I suspect that there was already a bug in shell since nothing was checked nor tested?...

Hind-M:
Yes, I meant hidden bugs or special cases since the prefix is not checked.

wolfv:
Planning to do a bugfix release today. Do you think we can finish this one?

Hind-M:
The snippet I posted seems to work on my side, but I don't think I have the rights to push here.

jonashaag:
We are hardcoding the root prefix here though...

Hind-M:
> Also, does the name `base` have special meaning in Micromamba?

Well It seems to be the name referring to the root prefix as in the following files:

https://github.com/mamba-org/mamba/blob/main/libmamba/src/api/shell.cpp#L95-L112
https://github.com/mamba-org/mamba/blob/main/libmamba/src/api/shell.cpp#L137-L144
https://github.com/mamba-org/mamba/blob/main/libmamba/src/api/configuration.cpp#L450-L457

but maybe it shouldn't...

Hind-M:
> We are hardcoding the root prefix here though...

Yes, but I am actually wondering why we don't have access to the root prefix in the configuration when using the shell this way...

libmamba/src/core/run.cpp
wolfv: sorry, I reverted this change. I didn't get the motivation. Is that what conda does?

jonashaag:
Why do we even need a special case for setting up the prefix here? I’m not familiar with the prefix code but this looks weird

wolfv:
It's because shell is also used for shell init etc where the normal way of finding root prefix/target prefix might not work yet

jonashaag:
Ok! The test I added still fails for some reason

wolfv:
I think the file is not executable

wolfv:
hmm there was something else going on on linux with the file being busy or so. I think I fixed that in a clean way. Let's get this merged and release! :)

PR #2276, opened 2023-02-09
fix rename or remove

wolfv:
This fixes the usage of the confusing `ec` boolean values :)

It would be much nicer if the `error_code` had a `has_error()` function to make it easier to understand what we're testing.

Maybe @JohanMabille or @Klaim can give this a review to make sure it's correct this time around.

### DIFF
--- a/libmamba/include/mamba/core/fsutil.hpp
+++ b/libmamba/include/mamba/core/fsutil.hpp
@@ -136,6 +136,10 @@ namespace mamba
     namespace mamba_fs
     {
         // Like std::rename, but works across file systems by moving the file instead
+        // if the rename fails.
+        // if both rename and move fail, the error code is set to the error code of the
+        // copy_file operation and the `to` file is removed. You will have to handle removal
+        // of the `from` file yourself.
         inline void rename_or_move(const fs::u8path& from,
                                    const fs::u8path& to,
                                    std::error_code& ec)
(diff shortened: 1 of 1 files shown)


JohanMabille (approved):

JohanMabille:
I agree that it's a bit too close to the C error code mechanism.

libmamba/include/mamba/core/fsutil.hpp
```diff
@@ -145,9 +149,10 @@ namespace mamba
             {
                 ec.clear();
                 fs::copy_file(from, to, ec);
-                if (!ec)
+                if (ec)
                 {
-                    fs::remove(from, ec);
+                    std::error_code ec2;
+                    fs::remove(to, ec2);
```
AntoinePrv, line 155: Not so familiar with error code, but shouldn't we not checking the result here?

wolfv: we don't really care about the result because at this point the operation has already failed.

Klaim: Adding `[[maybe_unused]]` here would clarify that and avoid warnings.

Klaim:
> It would be much nicer if the `error_code` had a `has_error()` function to make it easier to understand what we're testing.

I remember people in Boost and C++ committee agreeing with you (and adding some other issues with `boost/std::error_code`) and planning for an alternative which would have a better interface, but it's note in the standard (yet) and I dont know where the effort ended up. But yeah it's among the known and agreed issues.

PR #2277, opened 2023-02-09
Add channel specific job with new str

AntoinePrv:
Currently, using a spec like `conda-forge::python` will segfault in `MSolver::all_problems_structured` because channel specific jobs are added as a queue qith SOLVER_JOB_ONE_OF. In `all_prblem_structured`, the queue `id` is popping back as a `dep` but isn't.
There is not easy way to check for it explicitly, and no way to kepp track of the the fact that it was an explicit channel job.

The fix proposed here is rather to change the implementation of `add_channel_specific_jobs` to dynamically add a spec for the channel specific spec.

### DIFF
--- a/libmamba/src/core/solver.cpp
+++ b/libmamba/src/core/solver.cpp
@@ -271,8 +271,14 @@ namespace mamba
             LOG_ERROR << "Selected channel specific (or force-reinstall) job, but "
                          "package is not available from channel. Solve job will fail.";
         }
-        Id d = pool_queuetowhatprovides(pool, selected_pkgs);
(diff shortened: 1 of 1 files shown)


AntoinePrv:
@wolfv Regarding the `conda-forge::python` syntax, this is mamba specific?
I wonder if an `python@conda-forge` would be better to work with URL channels (a separate reflection)

wolfv:
No, it is not mamba specific. The `channel:namespace:packagename` syntax comes from conda https://github.com/conda/conda/blob/2e3f909cba109cdd7642387722223aa5cb79fd5c/conda/models/match_spec.py#L88-L92

jonashaag:
fyi @thomasblauthqc

wolfv (approved):

Issue #1242, opened 2021-10-26
recursive dependency list


craigim:
I am trying to get tensorflow and its dependencies onto an air-gapped network. To do so, I need to get a list of the dependencies with versions, get the wheel files from conda-forge, and then put them onto portable media.

If I run `mamba repoquery depends tensorflow` I get only the top-level dependencies.

| Name      |           Version | Build      |        Channel |
| ----------- | ------------- | ------------- | -------------- |
| tensorflow       |    2.6.0   | mkl_py38h874c8e8_0 |
| _tflow_select     |   2.3.0 |  mkl |
| python    |            3.8.12  | h12debd9_0 |
| tensorboard     |      2.6.0   | py_1 |
| tensorflow-estimator | 2.6.0   | pyh7b7c402_0 |
| tensorflow-base    |   2.6.0   | mkl_py38h3d85931_0 |

 If I add `-t` I get the full tree, which I pipe to a file, open the file in a text editor, and then erase all of the lines with "already visited" and all of the tree lines to get the list. (I have to do this seldom enough that I haven't yet hacked together a script to do this for me).

What I'd like would be to have a `--recursive` flag that would give me the list of _all_ dependencies as I get with `-t`, but, formatted the same way as if I had no flags into a table. Looking at verbose output, repoquery is doing the recursive search anyway, it just isn't exposing the data.

craigim:
If this were a pure python project, I'd be game to give it a try, but I have never written a line of C++ and I don't have any idea where to even start.

PR #2283, opened 2023-02-12
repoquery: Implemented recursive dependency printout. Closes: #1242

timostrunk:
https://github.com/mamba-org/mamba/issues/1242 describes a way to get recursive dependency information from repoquery depends.

I implemented it as specified in the issue:

    mamba repoquery depends cmake --recursive 

behaves exactly like 

    mamba repoquery depends cmake --tree

but lists the packages in a table. --tree overrides --recursive. Both mamba and micromamba have been updated. Please let me know if I should write additional tests and where.

### DIFF
--- a/docs/source/user_guide/mamba.rst
+++ b/docs/source/user_guide/mamba.rst
@@ -68,10 +68,14 @@ Here are some examples:
     # you can also specify more constraints on this search query
     $ mamba repoquery search "xtensor>=0.18"
 
-    # will show you a list of the dependencies of xtensor.
+    # will show you a list of the direct dependencies of xtensor.
     $ mamba repoquery depends xtensor
 
-With the ``-t,--tree`` flag, you can get the same information in a tree.
+    # will show you a list of the dependencies (including dependencies of dependencies).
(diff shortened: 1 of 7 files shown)


timostrunk:
Thanks for starting the tests @wolfv  Sorry for the broken pre-commit. Should pass now.

timostrunk:
I regenerated the stubs and reformatted them according to the pre-commit.

jonashaag (approved):
Thanks a lot!

Do we want to mention the word "transitive" (vs. "recursive") or do you think that's too technical?

Do we have any tests for `repoquery` that should get an addition?

timostrunk:
I think transitive fits much better than recursive. Should I adapt the pull request and change everything from recursive to transitive?

jonashaag:
Although, "transitive" might be a less well known term. I'll let you make the decision.

timostrunk:
True, I added the explanation that --recursive means transitive to the helpstring and the docs.

jonashaag (approved):
Awesome!

Issue #2250, opened 2023-01-26
std::bad_alloc if package download has HTTP 401 response


jonashaag:
### Troubleshooting docs

- [X] My problem is not solved in the Troubleshooting docs

### Search tried in issue tracker

-

### Latest version of Mamba

- [X] My problem is not solved with the latest version

### Tried in Conda?

Didn't try

### Describe your issue

```
libmamba Failed to download package from https://conda.mydomain.com/mychannel/linux-64/mypackage.tar.bz2 (status 401)
  @#�NnU�(�NnU0-hdcd2b5c_0.con`� extraction failed
  error    libmamba Error when extracting package: std::bad_alloc
```

jonashaag:
I don't remember how to reproduce this anymore but I assume it is fixed

pbsds:
i now get this with `mamba create --yes --name foobar conda-forge::python==3.10.13 conda-forge::poetry==1.7.1 conda-forge::just==1.24.0 conda-forge::fd-find==8.5.3 conda-forge::jq==1.7.1 conda-forge::yq==3.2.2 conda-forge::bat==0.24.0 dnachun::gum`, but only on `Ubuntu 22.04.4 LTS`. More modern distros are fine

PR #2287, opened 2023-02-13
Fix segmentation fault in case of an invalid package name

timostrunk:
Issue #2250 says that trying to install a not existing package from a URL leads to a bad alloc. I cannot reproduce this. I can however produce a segfault with a non existing package:

    micromamba install  https://www.heise.de/not_existing_package.tar.bz2
    Segmentation fault (core dumped)

The error is not that the package does not exist in this case, but rather that the version information from the package is being parsed, even though it is not present (the error is the package name not_existing_package.tar.bz2).

For example, this crashes correctly:

    micromamba install  https://www.heise.de/linux-64/fbBss-gpu-1.7.3-py3.9_h28a55e0_0_cuda11.3.tar.bz2
    https://www.heise.de/linux-64                       37.0kB @ 121.3kB/s 404 failed  0.3s
    https://www.heise.de/noarch                         37.0kB @ 120.4kB/s 404 failed  0.3s
    critical libmamba Multiple errors occured:
     Multi-download failed. Reason: Transfer finalized, status: 404 [https://www.heise.de/noarch/repodata.json] 36966 bytes
     Subdir https://www.heise.de/noarch not loaded!

In this PR, I removed an unnecessary try except block and instead verified the existence of the version information in the file string. In this case, mamba now errors out without a segfault:

    ./build/micromamba/micromamba install  https://www.heise.de/not_existing_package.tar.bz2
    error    libmamba dist_str not_existing_package did not split into a correct version info.
    critical libmamba Invalid package filename

- [ ] @jonashaag Please verify, if this covers your issue #2250.

### DIFF
--- a/libmamba/src/core/match_spec.cpp
+++ b/libmamba/src/core/match_spec.cpp
@@ -17,14 +17,14 @@ namespace mamba
 {
     std::vector<std::string> parse_legacy_dist(std::string dist_str)
     {
-        try
+        dist_str = strip_package_extension(dist_str).string();
+        auto split_str = rsplit(dist_str, "-", 2);
+        if (split_str.size() != 3)
         {
-            dist_str = strip_package_extension(dist_str).string();
+            LOG_ERROR << "dist_str " << dist_str << " did not split into a correct version info.";
+            throw std::runtime_error("Invalid package filename");
         }
-        catch (...)
-        {
-        }
-        return rsplit(dist_str, "-", 2);
+        return split_str;
     }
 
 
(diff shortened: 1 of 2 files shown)


jonashaag:
Wow @timostrunk thanks for the spree of PRs! If you need any help working on anything, please let me know. We can definitely use more (wo)manpower :)

Will review this shortly. Could you add a test for this? I guess the simplest would be to add something in the Micromamba `install` tests.

timostrunk:
Sure, I will add a test in the evening.

- [ ] Add test for failing install

timostrunk:
I added a test to check that the correct exception is raised.

jonashaag (approved):
Perfect!

PR #2289, opened 2023-02-14
Isolate solv::ObjQueue

AntoinePrv:
Part of #2302.

Here the Queue is isolated in a safe RAII object.
It was made to use std conventions rather than libsolv ones.

This is private API so does not have any consequence.

On the immediate side, it caught some memory leaks in `MSolver`!

### DIFF
--- a/libmamba/CMakeLists.txt
+++ b/libmamba/CMakeLists.txt
@@ -123,6 +123,8 @@ endforeach()
 set(LIBMAMBA_SOURCES
     longpath.manifest
     ${LIBMAMBA_SOURCE_DIR}/version.cpp
+    # C++ wrapping of libsolv
+    ${LIBMAMBA_INCLUDE_DIR}/mamba/solv-cpp/queue.hpp
     # Core API (low-level)
     ${LIBMAMBA_SOURCE_DIR}/core/singletons.cpp
     ${LIBMAMBA_SOURCE_DIR}/core/activation.cpp
@@ -212,7 +214,6 @@ set(LIBMAMBA_HEADERS
     ${LIBMAMBA_INCLUDE_DIR}/mamba/core/progress_bar.hpp
     ${LIBMAMBA_INCLUDE_DIR}/mamba/core/pinning.hpp
     ${LIBMAMBA_INCLUDE_DIR}/mamba/core/query.hpp
-    ${LIBMAMBA_INCLUDE_DIR}/mamba/core/queue.hpp
     ${LIBMAMBA_INCLUDE_DIR}/mamba/core/repo.hpp
     ${LIBMAMBA_INCLUDE_DIR}/mamba/core/run.hpp
     ${LIBMAMBA_INCLUDE_DIR}/mamba/core/shell_init.hpp
--- a/libmamba/CMakeLists.txt
+++ b/libmamba/CMakeLists.txt
@@ -124,7 +124,8 @@ set(LIBMAMBA_SOURCES
     longpath.manifest
     ${LIBMAMBA_SOURCE_DIR}/version.cpp
     # C++ wrapping of libsolv
-    ${LIBMAMBA_INCLUDE_DIR}/mamba/solv-cpp/queue.hpp
+    ${LIBMAMBA_SOURCE_DIR}/solv-cpp/queue.hpp
+    ${LIBMAMBA_SOURCE_DIR}/solv-cpp/queue.cpp
     # Core API (low-level)
     ${LIBMAMBA_SOURCE_DIR}/core/singletons.cpp
     ${LIBMAMBA_SOURCE_DIR}/core/activation.cpp
--- a/libmamba/CMakeLists.txt
+++ b/libmamba/CMakeLists.txt
@@ -124,7 +124,6 @@ set(LIBMAMBA_SOURCES
     longpath.manifest
     ${LIBMAMBA_SOURCE_DIR}/version.cpp
     # C++ wrapping of libsolv
-    ${LIBMAMBA_SOURCE_DIR}/solv-cpp/queue.hpp
     ${LIBMAMBA_SOURCE_DIR}/solv-cpp/queue.cpp
     # Core API (low-level)
     ${LIBMAMBA_SOURCE_DIR}/core/singletons.cpp
@@ -186,7 +185,7 @@ foreach(script ${SHELL_SCRIPTS})
     list(APPEND LIBMAMBA_SOURCES shell_scripts/${script}.cpp)
 endforeach()
 
-set(LIBMAMBA_HEADERS
+set(LIBMAMBA_PUBLIC_HEADERS
     ${LIBMAMBA_INCLUDE_DIR}/mamba/version.hpp
     # Core API (low-level)
     ${LIBMAMBA_INCLUDE_DIR}/mamba/core/activation.hpp
@@ -254,6 +253,10 @@ set(LIBMAMBA_HEADERS
     ${LIBMAMBA_INCLUDE_DIR}/mamba/api/update.hpp
 )
 
+set(LIBMAMBA_PRIVATE_HEADERS
+    ${LIBMAMBA_SOURCE_DIR}/solv-cpp/queue.hpp
+)
+
 # Targets and link
 # ================
 
@@ -270,7 +273,12 @@ macro(libmamba_create_target target_name linkage deps_linkage output_name)
 
     # Output
     # ======
-    add_library(${target_name} ${linkage_upper} ${LIBMAMBA_SOURCES} ${LIBMAMBA_HEADERS})
+    add_library(
+        ${target_name} ${linkage_upper}
+        ${LIBMAMBA_PUBLIC_HEADERS}
+        ${LIBMAMBA_PRIVATE_HEADERS}
+        ${LIBMAMBA_SOURCES}
+    )
 
     mamba_target_add_compile_warnings(
         ${target_name}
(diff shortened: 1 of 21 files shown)


Hind-M (approved):
Nice!
A slight nitpicking, new files should have 2023 instead of 2019 or 2022 in the license section for history consistency.

libmamba/include/mamba/core/solver.hpp
```diff
@@ -101,7 +114,7 @@ namespace mamba
         // Order of m_pool and m_solver is critical since m_pool must outlive m_solver.
         MPool m_pool;
         std::unique_ptr<::Solver, void (*)(::Solver*)> m_solver;
-        Queue m_jobs;
+        std::unique_ptr<solv::ObjQueue> m_jobs;
```
wolfv, line 117: Is there a reason why this should be a unique_ptr now?

libmamba/src/core/query.cpp
```diff
@@ -275,29 +270,26 @@ namespace mamba
...
         int depth = tree ? -1 : 1;
 
-        auto find_latest = [&](MQueue& solvables) -> Solvable*
+        auto find_latest_in_non_empty = [&](solv::ObjQueue& solvables) -> Solvable*
         {
-            Solvable* latest = pool_id2solvable(m_pool.get(), solvables[0]);
-            for (int i = 1; i < solvables.count(); ++i)
+            Solvable* latest = pool_id2solvable(m_pool.get(), solvables.front());
+            for (Id const solv : solvables)
             {
```
wolfv, line 291: the previous loop started at 1 - is this still correct?

libmamba/src/core/solver.cpp
```diff
@@ -585,28 +592,27 @@ namespace mamba
...
-        queue_init(&problem_rules);
-        Id count = solver_problem_count(m_solver.get());
+        solv::ObjQueue problem_rules;
+        Id const count = solver_problem_count(m_solver.get());
         for (Id i = 1; i <= count; ++i)
         {
-            solver_findallproblemrules(m_solver.get(), i, &problem_rules);
-            for (Id j = 0; j < problem_rules.count; ++j)
+            solver_findallproblemrules(m_solver.get(), i, problem_rules.get());
+            for (Id const r : problem_rules)
```
wolfv, line 600: is `Id const <var>` better vs `const Id <var>`?

wolfv: I just saw that it was like this before :) so that's fine

libmamba/src/core/transaction.cpp
```diff
@@ -1449,17 +1439,16 @@ namespace mamba
         };
 
         int mode = SOLVER_TRANSACTION_SHOW_OBSOLETES | SOLVER_TRANSACTION_OBSOLETE_IS_UPGRADE;
-        transaction_classify(m_transaction, mode, &classes);
-        Id cls;
-        for (int i = 0; i < classes.count; i += 4)
+        transaction_classify(m_transaction, mode, classes.get());
+        auto const cls_end = classes.end();
+        for (auto cls_iter = classes.begin(); cls_iter < cls_end; cls_iter += 4)
```
wolfv, line 1444: is it really better to use iterators here? I feel like it would be safer to use the index instead of going through an iterator (potentially we could do bounds checking with the indedx function, and it's more straightforward).

libmamba/tests/repodata_json_cache/test_7.state.json
```diff
@@ -11,5 +11,7 @@
         "value": true
     },
     "mod": "Fri, 11 Feb 2022 13:52:44 GMT",
+    "mtime_ns": -4762331256009361200,
```
wolfv, line 14: maybe remove this? We should add this to the .gitignore files nad figure out why the `mtime_ns` is negative on linux.

wolfv:
Looks very good to me! 

I was wondering about the header location (`solv-cpp/queue.hpp` vs something like `mamba/internal/solv-cpp/queue.hpp`). But I guess it doesn't matter much since we're not installing that header, right?

And the rest was just some tiny comments.

libmamba/include/mamba/core/solver.hpp
AntoinePrv: Yes, we want to isolate `libsolv` from the public API so we cannot know the `solv::ObjQueue` type.

AntoinePrv: Yes, we want to isolate `libsolv` from the public API so we cannot know the `solv::ObjQueue` type.

libmamba/src/core/query.cpp
AntoinePrv: Yes, this is like a `min` function where the running `min` is initialized to the first element.
You don't need to compare it again with itself but it does not change either and make for a simpler use of range-based for loop.

libmamba/src/core/transaction.cpp
AntoinePrv: Changed back to index and `ObjQueue::at`.

AntoinePrv:
@wolfv Yes these headers are private (hence the `std::unique_ptr`).
On the longer run we could make it its own implementation library, that is in CMake but never packaged or installed.
That would further ensure that we are keeping a strict separation of abstractions.

libmamba/src/solv-cpp/queue.hpp
```diff
@@ -0,0 +1,146 @@
...
+        auto rend() const -> const_reverse_iterator;
+        auto crend() const -> const_reverse_iterator;
+        auto data() -> pointer;
+        auto data() const -> const_pointer;
+
+        template <template <typename, typename...> class C>
+        auto as() -> C<value_type>;
+
+        auto get() -> ::Queue*;
+        auto get() const -> ::Queue const*;
```
Klaim, line 90: I feel like the naming here could be misleading when reading the code using this queue type.
Consider alternatives like:
 - `native()`
 - `impl()`
 - `raw()`
 To convey that we get something lower level, not just "the content".
 (not a big issue)

libmamba/src/solv-cpp/queue.hpp
```diff
@@ -0,0 +1,146 @@
...
+
+    private:
+        friend void swap(ObjQueue& a, ObjQueue& b) noexcept;
+
+        ::Queue m_queue = {};
```
Klaim, line 95: Side note that stylistically I would prefer default initialization to not have an assignation. I understand that it's valid in C++17, but it looks redundant and noisy to me when reading the code.
Weak "meh" here, to be clear.

libmamba/src/solv-cpp/queue.hpp
```diff
@@ -0,0 +1,146 @@
...
+
+        ObjQueue(std::nullptr_t);
```
Klaim, line 97: First time I see this kind of hack/technique XD
Consider making this one `explicit` to avoid unintional conversions inside the cpp

libmamba/src/solv-cpp/queue.cpp
```diff
@@ -0,0 +1,288 @@
...
+#include <sstream>
+
+#include <solv/queue.h>
+
+#include "solv-cpp/queue.hpp"
+
+namespace mamba::solv
+{
+    ObjQueue::ObjQueue(std::nullptr_t)
+        : m_queue{}
```
Klaim, line 20: This is redundant with the member initialization, you can remove that line.

libmamba/src/solv-cpp/queue.cpp
```diff
@@ -0,0 +1,288 @@
...
+
+    auto ObjQueue::empty() const -> bool
+    {
+        return size() == 0;
+    }
+
+    auto ObjQueue::erase(const_iterator pos) -> iterator
+    {
+        auto const offset = offset_of(pos);
+        assert(offset <= std::numeric_limits<int>::max());
```
Klaim, line 106: Why `int` and not `std::ptrdiff_t`?
Same question in `reserve()`

libmamba/src/solv-cpp/queue.cpp
```diff
@@ -0,0 +1,288 @@
...
+    }
+
+    namespace
+    {
+        void ensure_valid_pos(ObjQueue::size_type size, ObjQueue::size_type pos)
+        {
+            if (pos >= size)
+            {
+                std::stringstream ss = {};
+                ss << "Index " << pos << " is greater that the number of elements (" << size << ')';
```
Klaim, line 164: Use fmt here (maybe with a comment to use `std::format` once C++20 is available)

libmamba/CMakeLists.txt
```diff
@@ -252,6 +253,10 @@ set(LIBMAMBA_HEADERS
     ${LIBMAMBA_INCLUDE_DIR}/mamba/api/update.hpp
 )
 
+set(LIBMAMBA_PRIVATE_HEADERS
+    ${LIBMAMBA_SOURCE_DIR}/solv-cpp/queue.hpp
```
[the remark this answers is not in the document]
AntoinePrv: They are not really private, they are installed, and needed by public headers.

libmamba/include/mamba/core/package_info.hpp
```diff
@@ -7,19 +7,15 @@
...
 #include <string>
 #include <vector>
 #include <set>
 
-#include "nlohmann/json.hpp"
+#include <nlohmann/json.hpp>
+#include <solv/pool.h>
+#include <solv/poolid.h>
+#include <solv/repo.h>
+#include <solv/solvable.h>
```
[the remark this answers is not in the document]
AntoinePrv: Because the `extern "C"` is in libsolv and so was never really needed.

libmamba/src/solv-cpp/queue.cpp
AntoinePrv: The idea is to keep these wrappers dependency free so I think we should avoid using `fmt` until there is some strong need for it.

libmamba/src/solv-cpp/queue.cpp
AntoinePrv: Because `int` is the type used by the `libsolv` functions `queue_delete`, `queue_prealloc`...

AntoinePrv:
Bad branch manipulation.

Issue #2189, opened 2022-12-26
Unable to not show banner when using mamba


harelwa:
Hi,

I would like to disable `banner` printing to console for all mamba cli subcommands ( `list`, `config`, `search` etc. ).

Inferring from related PRs, I've configured `show_banner` to false, in `~/.condarc` as follows - 

```bash
> $ cat ~/.condarc

show_channel_urls: true
auto_activate_base: true
show_banner: false
```

I'm using mamba version `1.1.0` - 

```bash
> $ mamba list mamba

# packages in environment at /Users/harelos/mambaforge:
#
# Name                    Version                   Build  Channel
libmamba                  1.1.0                h8b63968_2    conda-forge
libmambapy                1.1.0           py310hb15139c_2    conda-forge
mamba                     1.1.0           py310h6bde348_2    conda-forge
```

I found `show banner` reading in `libmamba`, thus, e.g. when I run `mamba create` no banner is shown.

However when I run `mamba search` or `mamba info` the banner is shown:

```bash
> $ mamba info

                  __    __    __    __
                 /  \  /  \  /  \  /  \
                /    \/    \/    \/    \
███████████████/  /██/  /██/  /██/  /████████████████████████
              /  / \   / \   / \   / \  \____
             /  /   \_/   \_/   \_/   \    o \__,
            / _/                       \_____/  `
            |/
        ███╗   ███╗ █████╗ ███╗   ███╗██████╗  █████╗
        ████╗ ████║██╔══██╗████╗ ████║██╔══██╗██╔══██╗
        ██╔████╔██║███████║██╔████╔██║██████╔╝███████║
        ██║╚██╔╝██║██╔══██║██║╚██╔╝██║██╔══██╗██╔══██║
        ██║ ╚═╝ ██║██║  ██║██║ ╚═╝ ██║██████╔╝██║  ██║
        ╚═╝     ╚═╝╚═╝  ╚═╝╚═╝     ╚═╝╚═════╝ ╚═╝  ╚═╝

        mamba (1.1.0) supported by @QuantStack

        GitHub:  https://github.com/mamba-org/mamba
        Twitter: https://twitter.com/QuantStack

█████████████████████████████████████████████████████████████


     active environment : base
    active env location : /Users/harelos/mambaforge
            shell level : 1
       user config file : /Users/harelos/.condarc
 populated config files : /Users/harelos/mambaforge/.condarc
                          /Users/harelos/.condarc
          conda version : 4.14.0
[...]

PR #2299, opened 2023-02-16
Fix #1973: Micromamba banner should go to stderr

### LINKED ISSUE #1973: Micromamba banner should go to stderr
Currently it goes to stdout.

@wolfv what's your opinion on removing the banner from Mamba and Micromamba entirely?

jonashaag:


AntoinePrv:
@jonashaag what problem is this addressing? I think we should keep `stderr` for errors...

jonashaag:
No, in the Unix philosophy stderr is for everything that’s not output that you want to process further (in a pipe). So logging, errors, progress, etc. The banner causes problems when people feed the output of Mamba into programs.

libmamba/include/mamba/core/output.hpp
```diff
@@ -12,6 +12,7 @@
 #include <string>
 #include <string_view>
 #include <vector>
+#include <iostream>
```
Klaim, line 15: I think this should be `iosfwd` as it is in headers and we want to avoid bring unnecessary heavy headers.

Hind-M:
Hmm, I think that having `show_banner` configurable addresses the problem (even if it seems to be not used everywhere, it should be fixed, cf. #2189 for example).
If we don't want the banner at all or if the flag `show_banner` is always set to `false`, we should just remove the banner completely...

Klaim (approved):
Seems ok to me.

Issue #2226, opened 2023-01-17
Explicit channel specifiers causes mamba to drop unsolvable packages 


shughes-uk:
### Troubleshooting docs

- [X] My problem is not solved in the Troubleshooting docs

### Search tried in issue tracker

Tried!

### Latest version of Mamba

- [X] My problem is not solved with the latest version

### Tried in Conda?

Not reproducible with Conda

### Describe your issue

When Installing packages that should not solve, but using an explicit channel specifier `<channel>::<somepackage>`, mamba will warn that the solve will fail but then proceeds to install the solvable packages and ignore the error.

```
micromamba create -r ./tmp -p /bad  conda-forge::python==3.11.0 conda-forge::poetry==1.0.0
```

This results in poetry not being installed but no error code and just an easily missed error log, without the channel specifier you get an appropriate error message and the command fails.

```
micromamba install -c conda-forge python==3.11.0 poetry==1.0.0
```

Conda does not have this issue.

### mamba info / micromamba info

```shell
__
          __  ______ ___  ____ _____ ___  / /_  ____ _
         / / / / __ `__ \/ __ `/ __ `__ \/ __ \/ __ `/
        / /_/ / / / / / / /_/ / / / / / / /_/ / /_/ /
       / .___/_/ /_/ /_/\__,_/_/ /_/ /_/_.___/\__,_/
      /_/


            environment : /Users/samanthahughes/miniconda3/envs/coiled (active)
           env location : /Users/samanthahughes/miniconda3/envs/coiled
      user config files : /Users/samanthahughes/.mambarc
 populated config files : /Users/samanthahughes/.condarc
       libmamba version : 1.2.0
     micromamba version : 1.2.0
           curl version : libcurl/7.86.0 SecureTransport (OpenSSL/1.1.1s) zlib/1.2.13 libssh2/1.10.0 nghttp2/1.47.0
     libarchive version : libarchive 3.6.2 zlib/1.2.13 bz2lib/1.0.8 libzstd/1.5.2
       virtual packages : __unix=0=0
                          __osx=13.1=0
                          __archspec=1=arm64
               channels : 
       base environment : /Users/samanthahughes/micromamba
               platform : osx-arm64
```

### Logs

```shell
info     libmamba Parsing MatchSpec conda-forge::python==3.11.0
info     libmamba Parsing MatchSpec conda-forge::poetry==1.0.0
[...]

Issue #2041, opened 2022-10-24
Subdir ignored in `channel/subdir` specifier


jonashaag:
```
conda create -n xxx 'conda-forge/noarch::git'
...
PackagesNotFoundError: The following packages are not available from current channels:

  - conda-forge/noarch::git
...
```

```
mamba create -n xxx 'conda-forge/noarch::git'
...
  added / updated specs:
    - conda-forge/osx-arm64::git
...
```

This also applies when installing from `environment.yml` files.

jonashaag:
In 1.5.1:

```
micromamba create -n xxx conda-forge/osx-64::git
conda-forge/osx-arm64                                       Using cache
conda-forge/noarch                                          Using cache
critical libmamba The package "conda-forge/osx-64::git" is not available for the specified platform
```

jonashaag:
> * Was it different before?

Yes, it ignored the constraint, see initial post.

> * Does it work with `micromamba create -n xxx conda-forge/osx-64::git -c conda-forge/osx-64`?

No.


> * I think the following should not throw but `return false;`

Not sure, this was changed in #2300 for good reasons I think

jonashaag:
Related issue: https://github.com/mamba-org/mamba/issues/2431

delsner:
I could take a look at this. But before I get started, I would like to understand what's the precise problem at the moment and whether we would like to fix it for `1.x`.

## Status Quo

**1.x** 

```
❯ build/1.x/micromamba/micromamba create -n xxx conda-forge/osx-64::git -c conda-forge/osx-64
conda-forge/osx-64                                          Using cache
conda-forge/osx-arm64                                       Using cache
conda-forge/noarch                                          Using cache
critical libmamba The package "conda-forge/osx-64::git" is not available for the specified platform
```
returns that the package is not available, even though it is available in subdir `osx-64`. 

**main**

```
❯ build/main/micromamba/micromamba create -n xxx conda-forge/osx-64::git -c conda-forge/osx-64
conda-forge/osx-64                                          Using cache
conda-forge/noarch                                          Using cache
conda-forge/osx-arm64                                       Using cache
conda-forge/noarch                                          Using cache
conda-forge/osx-arm64                                       Using cache

Transaction

  Prefix: /Users/delsner/opt/micromamba/envs/xxx

  Updating specs:

   - conda-forge/osx-64::git
```

seems to work fine (?).

## Possible TODOs

- [ ] Fix channel handling in `1.x` to match `main`
[...]

PR #2300, opened 2023-02-16
Respect subdir in match spec

ThomasBlauthQC:
This PR addresses #2041 and #2226. 

If a package can't be found in the specified subdir, the output now looks like this.
```
micromamba create -n xxx 'conda-forge/noarch::git'
...
critical libmamba Selected channel specific (or force-reinstall) job, but package is not available from channel.
...
```

I'm not too proficient in C++, so any feedback is highly appreciated! 🙂

cc @jonashaag

### DIFF
--- a/libmamba/src/core/pool.cpp
+++ b/libmamba/src/core/pool.cpp
@@ -221,8 +221,8 @@ namespace mamba
                         }
                     }
                 }
-
             );
+            
             solv::StringId const repr_id = pool.add_string(repr);
             ::Id const offset = pool_queuetowhatprovides(pool.raw(), selected_pkgs.raw());
             // FRAGILE This get deleted when calling ``pool_createwhatprovides`` so care
--- a/libmamba/src/core/pool.cpp
+++ b/libmamba/src/core/pool.cpp
@@ -160,6 +160,25 @@ namespace mamba
             return false;
         }
 
+        bool
+        subdir_match(std::string candidate_url, std::string needle_spec)
+        {
+            std::string needle_channel = split(needle_spec, ":", 1)[0];
+            if (!contains(needle_channel, "/")){
+                return true;
+            }
+            std::string needle_subdir = rsplit(needle_channel, "/", 1)[1];
+
+            std::string candidate_subdir = rsplit(candidate_url, "/", 1)[1];
+
+            if (candidate_subdir == needle_subdir){
+                return true;
+            }
+            throw std::runtime_error(
+                fmt::format("The package \"{}\" is not available for the specified platform", needle_spec
+            ));
+        }
+
         /**
          * Add function to handle matchspec while parsing is done by libsolv.
          */
@@ -197,9 +216,12 @@ namespace mamba
                     auto const url = std::string(repo.url());
                     if (channel_match(channel_context, channel_context.make_channel(url), c))
                     {
-                        selected_pkgs.push_back(s.id());
+                        if (subdir_match(url, ms.spec)){
+                            selected_pkgs.push_back(s.id());
+                        }
                     }
                 }
+
             );
             solv::StringId const repr_id = pool.add_string(repr);
             ::Id const offset = pool_queuetowhatprovides(pool.raw(), selected_pkgs.raw());
--- a/libmamba/src/core/pool.cpp
+++ b/libmamba/src/core/pool.cpp
@@ -173,9 +173,9 @@ namespace mamba
             }
             std::string needle_subdir = rsplit(needle_channel, "/", 1)[1];
 
-            std::string candidate_subdir = rsplit(candidate_url, "/", 1)[1];
+            std::string candidate_repo_subdir = rsplit(candidate_repo_url, "/", 1)[1];
 
-            if (candidate_subdir == needle_subdir)
+            if (candidate_repo_subdir == needle_subdir)
             {
                 return true;
             }
--- a/libmamba/src/core/pool.cpp
+++ b/libmamba/src/core/pool.cpp
@@ -160,22 +160,28 @@ namespace mamba
             return false;
         }
 
-        bool
-        subdir_match(std::string candidate_url, std::string needle_spec)
+        bool subdir_match(std::string candidate_repo_url, std::string needle_spec)
         {
+            // Example candidate_repo_url: https://.../conda-forge/linux-64
+            // example needle_spec: conda-forge/osx-64::xtensor
+
             std::string needle_channel = split(needle_spec, ":", 1)[0];
-            if (!contains(needle_channel, "/")){
+            if (!contains(needle_channel, "/"))
+            {
+                // Subdir not specified, so any subdir is fine
                 return true;
             }
             std::string needle_subdir = rsplit(needle_channel, "/", 1)[1];
 
             std::string candidate_subdir = rsplit(candidate_url, "/", 1)[1];
 
-            if (candidate_subdir == needle_subdir){
+            if (candidate_subdir == needle_subdir)
+            {
                 return true;
             }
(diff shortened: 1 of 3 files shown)


libmamba/src/core/solver.cpp
```diff
@@ -227,6 +227,15 @@ namespace mamba
         if (!chan)
             return false;
 
+        // Check the subdirectory.
+        // This assumes, that explicitly specified subdirectories
+        // (e.g. conda-forge/linux-64::mamba) result in the platform member of needle
+        // only containing one entry. When the subdirectory is omitted, the paltfom
+        // member instead contains implicit <platform> and noarch entries.
```
jonashaag, line 234: Typo in "platform", also I find this a little easier to parse:

```suggestion
        // This assumes that explicitly specified subdirectories
        // (e.g. conda-forge/linux-64::mamba) result in needle.platforms.size() == 1 (e.g. needle.platforms() == {"linux-64"}). If the subdirectory is omitted from the spec, needle.platforms() is {"noarch", "<platform>"}.
```

jonashaag (commented):
So many tests 😍😍😍

libmamba/src/core/solver.cpp
ThomasBlauthQC: I also like this! Not sure how useful the comment is in general, or if that is "common knowledge". I just wanted to make the `== 1` comparison come a little less out of the blue.

libmamba/tests/test_satisfiability_error.cpp
```diff
@@ -634,4 +638,100 @@ namespace mamba
                                              create_scip,
                                              create_jupyterlab,
                                              create_double_python));
+
+    template <typename PkgRange>
+    auto create_mrepo_with_repo_meta(PkgRange const& packages,
+                                     std::string channel_name,
+                                     std::string subdir)
+    {
+        auto const tmp_dir = dir_guard(fs::temp_directory_path() / "mamba/tests"
```
AntoinePrv, line 647: `test_satisfiability`  is not meant for testing this. Would you mind movinf these new tests to a `test_solver.cpp`, possibly moving test utility code to a common header.

AntoinePrv (changes requested):
Thanks for the PR

libmamba/tests/test_satisfiability_error.cpp
ThomasBlauthQC: Sure, that makes a lot of sense!

jonashaag:
@ThomasBlauthQC do you plan on following up on this?

ThomasBlauthQC:
@AntoinePrv, @jonashaag, I'm very sorry for the delay; this fell through the cracks way too many times! I adapted the pr, the string parsing feels very hacky though... Let me know what you think!

libmamba/src/core/pool.cpp
```diff
@@ -160,6 +160,25 @@ namespace mamba
             return false;
         }
 
+        bool
+        subdir_match(std::string candidate_url, std::string needle_spec)
+        {
```
jonashaag, line 165: Would be nice to have an example for `candidate_url` and  `needle_spec` in a comment here.

jonashaag, line 165:
```suggestion
        {
            // Example candidate_url: https://.../micromamba-1.4.7-0.tar.gz2
            // example needle_spec: conda-forge/osx-64
```

libmamba/src/core/pool.cpp
```diff
@@ -160,6 +160,25 @@ namespace mamba
...
+            std::string needle_channel = split(needle_spec, ":", 1)[0];
+            if (!contains(needle_channel, "/")){
```
jonashaag, line 167:
```suggestion
            if (!contains(needle_channel, "/")){
                // Subdir not specified, so any subdir is fine
```

libmamba/src/core/pool.cpp
```diff
@@ -160,6 +160,25 @@ namespace mamba
...
+                return true;
+            }
+            std::string needle_subdir = rsplit(needle_channel, "/", 1)[1];
+
+            std::string candidate_subdir = rsplit(candidate_url, "/", 1)[1];
+
+            if (candidate_subdir == needle_subdir){
+                return true;
+            }
+            throw std::runtime_error(
```
jonashaag, line 177: Not sure if this is the right way to handle this case here

jonashaag:
Thanks for updating @ThomasBlauthQC!

libmamba/src/core/pool.cpp
ThomasBlauthQC: I'm not sure myself, open for any suggestions!

jonashaag (approved):

AntoinePrv:
> I adapted the pr, the string parsing feels very hacky though... Let me know what you think!

@ThomasBlauthQC thanks for working on this, however I am currently doing significant rework of the repo/pool/solver/transcaction and I would prefer a different solution that will only be possible later on.

Is this feature blocking for you? Or can it wait a couple more months?

jonashaag:
Would be nice to have this earlier than in a couple of months.

jonashaag:
@AntoinePrv if it's fine with you I'd like to merge this next week.

libmamba/src/core/pool.cpp
```diff
@@ -160,6 +160,31 @@ namespace mamba
...
+            // Example candidate_repo_url: https://.../conda-forge/linux-64
+            // example needle_spec: conda-forge/osx-64::xtensor
+
+            std::string needle_channel = split(needle_spec, ":", 1)[0];
+            if (!contains(needle_channel, "/"))
+            {
+                // Subdir not specified, so any subdir is fine
+                return true;
+            }
+            std::string needle_subdir = rsplit(needle_channel, "/", 1)[1];
```
jaimergp, line 174: @jonashaag, concerning https://github.com/mamba-org/mamba/issues/2431,  I think this is the problematic part. It assumes that if the channel name contains a slash it's because it is specifying a subdir, but `pkgs/main` is a valid channel name (whether we like it or not 😬). So I'd say that we add some logic after this to check whether `needle_subdir` is indeed a known subdir, and ignore it otherwise.

micromamba/tests/test_create.py
```diff
@@ -613,6 +613,17 @@ def test_spec_with_channel(tmp_home, tmp_root_prefix, tmp_path):
             assert link["version"].startswith("0.22.")
 
 
+def test_spec_with_channel_and_subdir():
```
jaimergp, line 616: We should extend this test to make sure `pkgs/main::package_name` works correctly.

libmamba/src/core/pool.cpp
jaimergp: The logic implemented in `conda` can be found at https://github.com/conda/conda/blob/a9b1838f059501528e6f48c35972fd5f379dec0a/conda/models/match_spec.py#L188-L199

jonashaag: Sounds correct. That's really unfortunate :/

jonashaag: Are there any known channels with `/` other than the official Anaconda ones? I wonder if we could get away with hardcoding them if the subdir lookup thing is too complicated to implement.

jonashaag: @jaimergp I did some investigation. This part of the code is never run, I think because this PR collided with some other refactorings that were done. So while incorrect, it's not the cause of the bug.

jonashaag: TIL `KNOWN_PLATFORMS` in `channel.cpp`

PR #2301, opened 2023-02-16
docs: Clarify "so" meaning

jrwrigh:
At least I assume that what this is supposed to mean. If not, correct clarification is needed.


jonashaag:
No, here „or so“ is probably a Germanism and means „or similar“. However in this context I don’t know why it says zip file because a tarball is always as tarball, not a zip file and nothing else either

PR #2303, opened 2023-02-17
Fix some warnings

AntoinePrv:


libmamba/include/mamba/core/satisfiability_error.hpp
```diff
@@ -93,7 +93,7 @@ namespace mamba
         };
         struct PackageNode : PackageInfo
         {
-            std::optional<SolverRuleinfo> problem_type;
+            std::optional<SolverRuleinfo> problem_type = {};
```
Klaim, line 96: Was there a warning about this line?

Klaim (approved):
LGTM 👍🏽

libmamba/include/mamba/core/satisfiability_error.hpp
AntoinePrv: I think there was a warning when using the constructor with `PackageNode(PackageInfo())` leaving it the `problem_type` so I made a default which was the intended behavior.

PR #2306, opened 2023-02-18
Add conda.rc Options for Existing Remote Settings

### LINKED ISSUE #1871: Controlling mamba fetch retries
Does micromamba respect the conda.rc options `remote_max_retries` and `remote_read_timeout_secs`?

When we run `mamba install` on 1000 EC2 `c5.xlarge` , installing 234 packages on each instance, we consistently see at least one fail with

 ```
critical libmamba Download error (28) Timeout was reached [https://repo.anaconda.com/pkgs/main/noarch/pexpect-4.8.0-pyhd3eb1b0_3.conda]
 Operation too slow. Less than 30 bytes/sec transferred the last 60 seconds
```

We need to be able to adjust mamba to be more tolerant of situations like this and retry slightly harder. 1/1000 does not seem like a lot but we regularly run this many thousands of times a day so we have multiple failures to this per day!

srilman:
Closes #1871. Makes the following conda.rc options configurable inside of libmamba. Note that all of these settings already exist in libmamba's Context, and are already configurable inside of mamba.
- `remote_connect_timeout_secs`
- `remote_backoff_factor`
- `remote_max_retries`

### DIFF
--- a/libmamba/src/api/configuration.cpp
+++ b/libmamba/src/api/configuration.cpp
@@ -1243,23 +1243,28 @@ namespace mamba
                         the value is the url of the proxy server, optionally with username and password
                         in the form of scheme://username:password@hostname.)")));
 
-        insert(Configurable("remote_connect_timeout_secs", &ctx.connect_timeout_secs)
-                    .group("Network")
-                    .set_rc_configurable()
-                    .set_env_var_names()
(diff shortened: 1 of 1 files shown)


srilman:
@wolfv Who should I request to get a review on this?

JohanMabille (approved):

PR #2310, opened 2023-02-20
Extend issue template

jonashaag:
- Clarify "reproducible with Conda"
- Add stronger warning that Mambaforge is the only supported installation method

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.yml
+++ b/.github/ISSUE_TEMPLATE/bug.yml
@@ -9,6 +9,19 @@ body:
       options:
         - label: "My problem is not solved in the Troubleshooting docs"
           required: true
+  - type: dropdown
+    id: mamba_installation_method
+    attributes:
+      label: How did you install Mamba?
+      description: |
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/bug.yml
```diff
@@ -9,6 +9,19 @@ body:
...
         - label: "My problem is not solved in the Troubleshooting docs"
           required: true
+  - type: dropdown
+    id: mamba_installation_method
+    attributes:
+      label: How did you install Mamba?
+      description: |
+        Note that Mamba installation methods other than [Mambaforge](https://mamba.readthedocs.io/en/latest/installation.html#fresh-install) are **NOT SUPPORTED**.
+
+        If this issue is about Micromamba, please select "Other".
```
AntoinePrv, line 19: Perhaps, I would rather say something like, *if you use mamba* you need to install through mambaforge.

.github/ISSUE_TEMPLATE/bug.yml
```diff
@@ -9,6 +9,19 @@ body:
...
+      options:
+        - "Mambaforge"
+        - "Other (please describe)"
```
AntoinePrv, line 22: I would add an option for micromamba

AntoinePrv (approved):

PR #2315, opened 2023-02-21
Add checked numeric cast

AntoinePrv:
This PR adds a checked numeric cast to make checked conversion whenever we currently have a warning or a silencing `static_cast` without checks.

### DIFF
--- a/libmamba/include/mamba/core/util_cast.hpp
+++ b/libmamba/include/mamba/core/util_cast.hpp
@@ -1,0 +1,117 @@
+// Copyright (c) 2023, QuantStack and Mamba Contributors
+//
+// Distributed under the terms of the BSD 3-Clause License.
+//
+// The full license is in the file LICENSE, distributed with this software.
+
+#ifndef MAMBA_CORE_UTIL_CAST_HPP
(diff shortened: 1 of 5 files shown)


libmamba/include/mamba/core/util_cast.hpp
```diff
@@ -0,0 +1,117 @@
...
+                {
+                    throw detail::make_overflow_error<To>(val);
+                }
+            }
+
+            return static_cast<To>(val);
+        }
+        else
+        {
+            using float_type = std::common_type_t<From, To>;
```
wolfv, line 85: Do we do a lot of casting between float types? Seems to me we should rather get rid of it if possible.

The integer ones are obviously necessary since libsolv uses 32 bit integers in quite some places.

wolfv (approved):

libmamba/include/mamba/core/util_cast.hpp
AntoinePrv: I think I've seen a couple casting from integer to float but perhaps they need not be exact.

Klaim (approved):

PR #2314, opened 2023-02-21
DependencyInfo is redundant

AntoinePrv:
- Rename `MatchSpec::build` > `MatchSpec::build_string` for consistency and ease of use in the SAT error messages.
- Use Compare for equality in `vector_set` as done in `std::set`
- Remove redundant `DependencyInfo`

PR #2318, opened 2023-02-22
Improve Python conflicts

AntoinePrv:
A heuristic to better handle the (almost) cyclic conflict due to the reversal of the dependencies bewteen `python` and `python_abi`.
The heuristic looks for this pattern and tries to simplify it.

The heuristics required removing nodes/conflicts, which led to most of the changes in this PR.

- [x] Based on #2314


libmamba/include/mamba/core/satisfiability_error.hpp
```diff
@@ -70,31 +77,14 @@ namespace mamba
         };
         struct PackageNode : PackageInfo
```
JohanMabille, line 78: Why keeping empty types? If you find the name more expressive, you can alias the type (although it won't prevent to pass the real type to functions accepting the alias).

libmamba/include/mamba/core/satisfiability_error.hpp
```diff
@@ -70,31 +77,14 @@ namespace mamba
...
         {
-            std::optional<SolverRuleinfo> problem_type = {};
-
-            PackageNode(const PackageNode&) = default;
-            PackageNode(PackageNode&&) noexcept = default;
-            PackageNode& operator=(const PackageNode&) = default;
-            PackageNode& operator=(PackageNode&&) noexcept = default;
         };
         struct UnresolvedDependencyNode : MatchSpec
```
JohanMabille, line 81: Same here.

libmamba/include/mamba/core/satisfiability_error.hpp
```diff
@@ -70,31 +77,14 @@ namespace mamba
...
         {
-            SolverRuleinfo problem_type;
-
-            UnresolvedDependencyNode(const UnresolvedDependencyNode&) = default;
-            UnresolvedDependencyNode(UnresolvedDependencyNode&&) noexcept = default;
-            UnresolvedDependencyNode& operator=(const UnresolvedDependencyNode&) = default;
-            UnresolvedDependencyNode& operator=(UnresolvedDependencyNode&&) noexcept = default;
         };
         struct ConstraintNode : MatchSpec
```
JohanMabille, line 84: Same here.

libmamba/include/mamba/core/satisfiability_error.hpp
AntoinePrv: Strong typing. They are put in a `std::variant` so they cannot be the same type.

JohanMabille: Oh, indeed.

JohanMabille (approved):

PR #2249, opened 2023-01-25
Fix #2248

jonashaag:
Closes #2248

PR #2324, opened 2023-02-26
Fix `mamba init --no-user`

xylar:
This merge makes changes similar to
https://github.com/conda/conda/pull/11949
to fix `mamba init` so that the `--no-user` flag actually has the intended effect of not changing a user's `.bashrc` or equivalent. Without this change, `--no-user` has no effect as was previously the case in conda, see:
https://github.com/conda/conda/issues/11948

### DIFF
--- a/mamba/mamba/mamba_shell_init.py
+++ b/mamba/mamba/mamba_shell_init.py
@@ -81,12 +81,7 @@ def shell_init(args):
         return initialize_dev(shell)
 
     else:
-        for_user = args.user
-        if not (args.install and args.user and args.system):
-            for_user = True
-        # no_user removed in Conda 23.1.0, see https://github.com/mamba-org/mamba/issues/2248
-        if hasattr(args, "no_user") and args.no_user:
-            for_user = False
+        for_user = args.user and not args.system
 
         anaconda_prompt = on_win and args.anaconda_prompt
         exit_code = initialize(
--- a/mamba/mamba/mamba_shell_init.py
+++ b/mamba/mamba/mamba_shell_init.py
@@ -81,7 +81,13 @@ def shell_init(args):
         return initialize_dev(shell)
 
     else:
-        for_user = args.user and not args.system
+        if hasattr(args, "no_user"):
+            # this seems to be conda < 23.1.0
+            # the `no_user` flag, if set, gets erroneously set to False
+            for_user = (args.no_user != False and args.user != False and
+                        not args.system)
+        else:
+            for_user = args.user and not args.system
 
         anaconda_prompt = on_win and args.anaconda_prompt
         exit_code = initialize(
--- a/mamba/mamba/mamba_shell_init.py
+++ b/mamba/mamba/mamba_shell_init.py
@@ -83,9 +83,14 @@ def shell_init(args):
     else:
         if hasattr(args, "no_user"):
             # this seems to be conda < 23.1.0
-            # the `no_user` flag, if set, gets erroneously set to False
-            for_user = (args.no_user != False and args.user != False and
-                        not args.system)
+            # Here, we use the old behavior even though the --no-user flag
+            # doesn't behave as desired, see:
+            # https://github.com/conda/conda/issues/11948
+            for_user = args.user
+            if not (args.install and args.user and args.system):
+                for_user = True
+            if args.no_user:
+                for_user = False
         else:
             for_user = args.user and not args.system
 


xylar:
I tested this locally by making a fresh install of mambaforge, then doing:
```
python -m pip install -e mamba
mamba init --no-user
mamba init --user
mamba init
```
As expected, `mamba init` and `mamba init --user` both lead to a modified `.bashrc`, whereas `mamba init --no-user` does not modify my `.bashrc`.

This behavior is needed in situations where we do not wish to have mambaforge activate automatically in a user's shell (e.g. because we have multiple conda or mambaforge installations available on an HPC system and shell scripts that call python scripts may revert back to the wrong conda base environment).

A workaround has been to copy a user's `.bashrc` (etc.) to a temporary file, then copy it back after calling `mamba init` but this shouldn't be necessary if `--no-user` functioned properly.

jonashaag:
Will this break backwards compat with older Conda versions?

xylar:
@jonashaag, I can test that but I don't believe so beyond the fact that the `--no-user` option didn't work for any version prior to the latest release.

xylar:
@jonashaag, yes, unfortunately with `conda<23`, this appears to have the undesirable effect of making the behavior of `mamba init` equivalent to `mamba init --no-user`, whereas with `conda>=23`, `mamba init` behaves like `mamba init --user` (as it is intended to).

I don't see an obvious way to fix this for all conda versions.  I will give it some thought but would appreciate suggestions.  It is broken as written now so I don't think leaving it as it is would be a very desirable option, especially because I put the work into fixing `conda` only to discover that `mamba` is bypassing my fix.

jonashaag:
Thanks for checking! I think just adding a Conda version check will be fine:

```
if conda version is before removal of no_user:
  use old code (before #2249)
else:
  use your new impl
```

xylar:
@jonashaag, do you have code somewhere for detecting the conda version?  I didn't want to add any new dependencies for comparing versions but if you already have such a dependency,

For now, I'm using the presence of `args.no_user` as the way to detect this.

mamba/mamba/mamba_shell_init.py
```diff
@@ -81,12 +81,13 @@ def shell_init(args):
...
-        if not (args.install and args.user and args.system):
-            for_user = True
-        # no_user removed in Conda 23.1.0, see https://github.com/mamba-org/mamba/issues/2248
-        if hasattr(args, "no_user") and args.no_user:
-            for_user = False
+        if hasattr(args, "no_user"):
+            # this seems to be conda < 23.1.0
+            # the `no_user` flag, if set, gets erroneously set to False
+            for_user = (args.no_user != False and args.user != False and
+                        not args.system)
```
xylar, line 88: I think I'll replace this with the old code even though it is broken for `--no-user`.  The fix is just so non-intuitive that I think it's maybe better if things are broken for older conda as they used to be.

jonashaag: Agreed! Let’s just roll back to the old (broken) code since our „fix“ in #2249 didn’t  actually fix anything

jonashaag:
> @jonashaag, do you have code somewhere for detecting the conda version? I didn't want to add any new dependencies for comparing versions but if you already have such a dependency,
> 
> For now, I'm using the presence of `args.no_user` as the way to detect this.

This is fine IMO, actually it’s better than checking for version: checking for availability of a feature. Version numbers are just proxies for that.

mamba/mamba/mamba_shell_init.py
```diff
@@ -81,12 +81,18 @@ def shell_init(args):
...
+        if hasattr(args, "no_user"):
+            # this seems to be conda < 23.1.0
+            # Here, we use the old behavior even though the --no-user flag
+            # doesn't behave as desired, see:
+            # https://github.com/conda/conda/issues/11948
+            for_user = args.user
+            if not (args.install and args.user and args.system):
+                for_user = True
+            if args.no_user:
+                for_user = False
```
xylar, line 93: Okay, this is the block of code before #2249.  So we'll behave as before if `args.no_user` is present but things will work properly (including the `--no-user` flag for newer versions where the argument is no longer there.

xylar:
Thanks @jonashaag.  I think we're on the same page.  Let me know what you think at this point.

jonashaag (approved):
Awesome thanks so much!

PR #2329, opened 2023-02-28
Add risc-v support for micromamba

### LINKED ISSUE #2308: RISC-V support and micromamba bootstrap build
Hi! I want to build a `micromamba` for my RISC-V SBC, but to setup a dev env for micromamba, a micromamba needs to be installed first as said in the documentation. This gets me stuck. 

So now I have two questions:
* Do we have a plan to add RISC-V support? I think a minimum support for managing pure-python (or any architecture-independent) packages should be possible and easy.
* Can we somehow bootstrap building `micromamba` without an existing `micromamba`?

dtcxzyw:
+ Add platform definitions for linux-riscv32/64. Please see also https://github.com/conda/conda/pull/12319/.
+ ~~Link with ```zstd_static``` when building dynamically linked ```libmamba``` if  ```zstd_shared``` is not available.~~

Tested with Ubuntu 22.04.1 LTS (GNU/Linux 5.19.0-1012-generic riscv64, GCC 11.3.0). Only statically linked ```micromamba``` was built and tested. 

closes #2308

### DIFF
--- a/libmamba/CMakeLists.txt
+++ b/libmamba/CMakeLists.txt
@@ -448,9 +448,10 @@ macro(libmamba_create_target target_name linkage deps_linkage output_name)
             ${LIBSOLV_LIBRARIES}
             ${LIBSOLVEXT_LIBRARIES}
             ${LibArchive_LIBRARIES}
-            $<IF:$<TARGET_EXISTS:zstd::libzstd_shared>,zstd::libzstd_shared,zstd::libzstd_static>
+            zstd::libzstd_shared
             ${CURL_LIBRARIES}
             ${OPENSSL_LIBRARIES}
+            zstd::libzstd_shared
             BZip2::BZip2
             yaml-cpp
             reproc++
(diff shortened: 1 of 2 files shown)


libmamba/CMakeLists.txt
```diff
@@ -448,10 +448,9 @@ macro(libmamba_create_target target_name linkage deps_linkage output_name)
             ${LIBSOLV_LIBRARIES}
             ${LIBSOLVEXT_LIBRARIES}
             ${LibArchive_LIBRARIES}
-            zstd::libzstd_shared
+            $<IF:$<TARGET_EXISTS:zstd::libzstd_shared>,zstd::libzstd_shared,zstd::libzstd_static>
```
JohanMabille, line 451: MIxing static and shared libraries may lead to undefined behavior because of possibly different runtimes (especially on Windows), in which case the dynamic library could be missing?

dtcxzyw: ```zstd``` is built statically on risc-v platforms because ```VCPKG_LIBRARY_LINKAGE``` is ```static``` in triplets ```riscv64-linux```. I will try to build a statically-linked ```mamba``` and roll back this change.

dtcxzyw:
I am confused about CI failures. It seems that ```libz``` cannot be found and linked correctly. Should I update this pull request branch by rebasing?

PR #2330, opened 2023-02-28
Fix `repoquery` issues

### LINKED ISSUE #2320: [Micromamba] `repoquery depends` and `repoquery whoneeds` only works with installed packages
With `mamba`, these sub-commands work if the package is installed in the environment, or if a channel is specified.
cf. #2088 and #2098 for more details.

**TODO**

- Update `micromamba` doc
- Add more info when running these commands (specify working conditions) - do the same for `mamba`.
- Make it work when the channel is specified like `mamba`? (to investigate)

Hind-M:
Fix #2320 items

### DIFF
--- a/micromamba/src/repoquery.cpp
+++ b/micromamba/src/repoquery.cpp
@@ -69,8 +69,12 @@ set_common_search(CLI::App* subcom, bool is_repoquery)
     static std::vector<std::string> specs;
     subcom->add_option("specs", specs, "Specs to search")->required();
 
-    static int local = 0;
-    subcom->add_flag("--local,!--remote", local, "Use installed data or remote repositories");
+    static int local = -1;
+    subcom->add_option(
+        "--use-local",
+        local,
+        "Use installed data (--use-local=1, default for `depends` and `whoneeds`) or remote repositories (--use-local=0, default for `search`)"
+    );
 
     auto& platform = config.at("platform");
     subcom->add_option("--platform", platform.get_cli_config<std::string>(), platform.description());
@@ -80,19 +84,21 @@ set_common_search(CLI::App* subcom, bool is_repoquery)
         {
             auto qtype = str_to_qtype(query_type);
             QueryResultFormat format = QueryResultFormat::kTABLE;
+            bool use_local = true;
             switch (qtype)
             {
                 case QueryType::kSEARCH:
                     format = QueryResultFormat::kTABLE;
-                    local = (local == 0) ? false : local > 0;
+                    use_local = (local == -1) ? false : (local > 0);  // use remote repodata by
+                                                                      // default for `search`
                     break;
                 case QueryType::kDEPENDS:
                     format = QueryResultFormat::kTABLE;
-                    local = (local == 0) ? true : local > 0;
+                    use_local = (local == -1) ? true : (local > 0);
                     break;
                 case QueryType::kWHONEEDS:
                     format = QueryResultFormat::kTABLE;
-                    local = (local == 0) ? true : local > 0;
+                    use_local = (local == -1) ? true : (local > 0);
                     break;
             }
             if (qtype == QueryType::kDEPENDS && recursive)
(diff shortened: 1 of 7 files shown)


micromamba/src/repoquery.cpp
```diff
@@ -80,19 +84,21 @@ set_common_search(CLI::App* subcom, bool is_repoquery)
         {
             auto qtype = str_to_qtype(query_type);
             QueryResultFormat format = QueryResultFormat::kTABLE;
+            bool use_local = true;
             switch (qtype)
             {
                 case QueryType::kSEARCH:
                     format = QueryResultFormat::kTABLE;
-                    local = (local == 0) ? false : local > 0;
+                    use_local = (local == -1) ? false : (local > 0);  // use remote repodata by
```
JohanMabille, line 92: can be simplified in `use_local = local > 0;`

micromamba/src/repoquery.cpp
```diff
@@ -80,19 +84,21 @@ set_common_search(CLI::App* subcom, bool is_repoquery)
...
+                                                                      // default for `search`
                     break;
                 case QueryType::kDEPENDS:
                     format = QueryResultFormat::kTABLE;
-                    local = (local == 0) ? true : local > 0;
+                    use_local = (local == -1) ? true : (local > 0);
```
JohanMabille, line 97: Can be simplified in `use_local = (locall == -1) || (local > 0);`

micromamba/src/repoquery.cpp
```diff
@@ -80,19 +84,21 @@ set_common_search(CLI::App* subcom, bool is_repoquery)
...
                     break;
                 case QueryType::kWHONEEDS:
                     format = QueryResultFormat::kTABLE;
-                    local = (local == 0) ? true : local > 0;
+                    use_local = (local == -1) ? true : (local > 0);
```
JohanMabille, line 101: Same here

micromamba/src/repoquery.cpp
```diff
@@ -110,10 +116,17 @@ set_common_search(CLI::App* subcom, bool is_repoquery)
...
+            {
+                format = QueryResultFormat::kJSON;
+            }
 
-            repoquery(qtype, format, local, specs[0]);
+            auto& channels = config.at("channels").compute().value<std::vector<std::string>>();
+            if (!channels.empty())
+            {
+                use_local = false;
+            }
```
JohanMabille, line 128: This can be simpified in `use_local = use_local && channels.empty()` or even `use_local = channels.empty()`

JohanMabille (commented):
After the review, I was wondering: should we switch to `use_local` automatically when the channels are empty, and warn the users? This will simplify the logic (no need for an additional option in the command line for instance).

Or do we want to give the user the ability to use the local packags only even if a channel(s) is (are) provided?

Hind-M:
> After the review, I was wondering: should we switch to `use_local` automatically when the channels are empty, and warn the users? This will simplify the logic (no need for an additional option in the command line for instance).
> 
> Or do we want to give the user the ability to use the local packags only even if a channel(s) is (are) provided?

Hmm I think that there is a case where the channels can be configured using a rc file or other config sources, and the user may want to choose to use local repoquery for installed packages or remote (`--channel` option is not given by the user either way but in case of `--use-local=0`, the configured channels would be used).

JohanMabille (approved):

jaimergp:
The extra message is breaking the JSON output in my dev environment :(

```
{
    "query": {
        "query": "conda >23.1.0",
        "type": "search"
    },
    "result": {
        "msg": "No entries matching \"conda-forge::conda>23.1.0\" found",
        "pkgs": [],
        "status": "OK"
    }
}conda-forge::conda>23.1.0 may not be installed. Try specifying a channel with '-c,--channel' option
```

Parsing this blindly gives `json.decoder.JSONDecodeError: Extra data: line 11 column 2 (char 223)`

JohanMabille:
I think I found the culprit. I'll ping you in the incoming PR.

PR #2339, opened 2023-03-01
Clean util_string

AntoinePrv:
- [x] Clean up the `util_string` header
- [x] Add `util_string.cpp`
- [x] Decouple `util.hpp` from `util_string.hpp` (lots of header changes related to this)
- [x] ~~Add `strip_if` function (needed for following PR)~~ Done in #2344
- [x] ~~Add `split_once` functions (needed for following PR)~~ Done in #2345
- [ ] ~~Add `split_for_each` as a replacement of split?~~

### DIFF
--- a/libmamba/include/mamba/core/util_string.hpp
+++ b/libmamba/include/mamba/core/util_string.hpp
@@ -21,108 +21,282 @@
 
 namespace mamba
 {
-    inline const char* check_char(const char* ptr)
-    {
-        return ptr ? ptr : "";
-    }
+    /**
+     * Return the string if the pointer is not null, otherwise a pointer to an empty string.
+     */
+    const char* raw_str_or_empty(const char* ptr);
 
-    constexpr const char* WHITESPACES(" \r\n\t\f\v");
-    constexpr const wchar_t* WHITESPACES_WSTR(L" \r\n\t\f\v");
+    char to_lower(char c);
+    wchar_t to_lower(wchar_t c);
+    std::string to_lower(std::string_view str);
+    std::wstring to_lower(std::wstring_view str);
+    template <typename Char>
+    std::basic_string<Char> to_lower(std::basic_string<Char>&& str);
+    template <typename Char, typename Iter>
+    void to_lower_tranform(std::basic_string_view<Char> str, Iter out);
 
-    bool starts_with(const std::string_view& str, const std::string_view& prefix);
-    bool ends_with(const std::string_view& str, const std::string_view& suffix);
-    bool contains(const std::string_view& str, const std::string_view& sub_str);
+    char to_upper(char c);
+    wchar_t to_upper(wchar_t c);
+    std::string to_upper(std::string_view str);
+    std::wstring to_upper(std::wstring_view str);
+    template <typename Char>
+    std::basic_string<Char> to_upper(std::basic_string<Char>&& str);
+    template <typename Char, typename Iter>
+    void to_upper_tranform(std::basic_string_view<Char> str, Iter out);
 
-    // TODO: add concepts here, or at least some contraints
-    template <typename T, typename AssociativeContainer>
-    auto contains(const AssociativeContainer& values, const T& value_to_find)
-        -> decltype(values.find(value_to_find) != values.end())  // this should make invalid usage
-                                                                 // SFINAE
-    {
-        return values.find(value_to_find) != values.end();
-    }
+    bool starts_with(std::string_view str, std::string_view prefix);
 
-    bool any_starts_with(const std::vector<std::string_view>& str, const std::string_view& prefix);
+    bool ends_with(std::string_view str, std::string_view suffix);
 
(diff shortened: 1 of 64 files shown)


AntoinePrv:
Can be merged anytime to avoid conflicts, the rest can be merged in another PR.

libmamba/src/core/run.cpp
```diff
@@ -290,7 +292,7 @@ namespace mamba
         }
 
         LOG_DEBUG << "Currently running processes: " << get_all_running_processes_info();
-        LOG_DEBUG << "Remaining args to run as command: " << join(" ", command);
+        fmt::print(LOG_DEBUG, "Remaining args to run as command: {}", fmt::join(command, " "));
```
Hind-M, line 295: This is more a general question, but are we using `fmt` now everywhere for logging?
(btw, the `LOG_DEBUG` in the line above could be replaced as well)

Hind-M (approved):
This PR is so long :) I expect the build/tests to complain if there is any problem, otherwise, I guess we will find out soon!

libmamba/src/core/run.cpp
AntoinePrv: I think we can use `fmt` wherever we need it, here it avoids allocating the result of `join`.

libmamba/src/core/util_string.cpp
```diff
@@ -0,0 +1,408 @@
...
+        return static_cast<wchar_t>(std::tolower(static_cast<wchar_t>(c)));
+    }
+
+    namespace
+    {
+        template <typename Char>
+        std::basic_string<Char> to_lower_impl(std::basic_string_view<Char> str)
+        {
+            auto out = std::basic_string<Char>();
+            to_lower_tranform<Char>(str, std::back_inserter(out));
```
JohanMabille, line 36: Typo. Also it could operate on an already allocated string.

libmamba/include/mamba/core/util_string.hpp
```diff
@@ -21,108 +21,282 @@
...
+    const char* raw_str_or_empty(const char* ptr);
+
+    char to_lower(char c);
+    wchar_t to_lower(wchar_t c);
+    std::string to_lower(std::string_view str);
+    std::wstring to_lower(std::wstring_view str);
+    template <typename Char>
+    std::basic_string<Char> to_lower(std::basic_string<Char>&& str);
+    template <typename Char, typename Iter>
+    void to_lower_tranform(std::basic_string_view<Char> str, Iter out);
```
JohanMabille, line 36: Typo. Also it would be more C++ish to have a full iterator interface.
And design question: Do we want to keep a template function for `to_lower` and `to_lower_transform` in the API? We could also have it as an implementation detail and a restricted API on `string` and `wstring` like we have for `string_view` and `wstring_view`. Or we could have a global template for handling these 4 use cases.

libmamba/include/mamba/core/util_string.hpp
```diff
@@ -21,108 +21,282 @@
...
+
+    char to_upper(char c);
+    wchar_t to_upper(wchar_t c);
+    std::string to_upper(std::string_view str);
+    std::wstring to_upper(std::wstring_view str);
+    template <typename Char>
+    std::basic_string<Char> to_upper(std::basic_string<Char>&& str);
+    template <typename Char, typename Iter>
+    void to_upper_tranform(std::basic_string_view<Char> str, Iter out);
```
JohanMabille, line 45: Same remark as for `to_lower_transform`

libmamba/include/mamba/core/util_string.hpp
```diff
@@ -21,108 +21,282 @@
...
+    }
+
+    template <typename StrRange>
+    bool any_starts_with(const StrRange& strs, std::wstring_view prefix)
+    {
+        return any_starts_with<StrRange, decltype(prefix)::value_type>(strs, prefix);
+    }
+
+    extern template bool any_starts_with(const std::vector<std::string>&, std::string_view);
+    extern template bool any_starts_with(const std::vector<std::string_view>&, std::string_view);
```
JohanMabille, line 259: extern template are meant to be used in cpp files, not hpp.

libmamba/include/mamba/core/util_string.hpp
```diff
@@ -21,108 +21,282 @@
...
-    std::vector<std::string>
-    rsplit(const std::string_view& input, const std::string_view& sep, std::size_t max_split = SIZE_MAX);
+    template <typename StrRange>
+    bool starts_with_any(std::wstring_view str, const StrRange& prefix)
+    {
+        return starts_with_any<StrRange, wchar_t>(str, prefix);
+    }
+
+    extern template bool starts_with_any(std::string_view, const std::vector<std::string>&);
+    extern template bool starts_with_any(std::string_view, const std::vector<std::string_view>&);
```
JohanMabille, line 287: Same remark here

PR #2340, opened 2023-03-02
Isolate PackageInfo from libsolv

AntoinePrv:
- The PackageInfo becomes (almost) a data type, independant from libsolv.
- The `Solvable*` ->` PackageInfo` is done in the `MPool`.
- `MTransaction` needs to capture the `MPool` to use that functions (but anyways it was using a `Pool*` already, at least it's explicit).

The rest of the changes are to cope with these changes throughout the code base.


wolfv:
nice!

libmambapy/src/main.cpp
```diff
@@ -210,7 +210,14 @@ PYBIND11_MODULE(bindings, m)
...
 
     py::class_<MTransaction>(m, "Transaction")
-        .def(py::init<MSolver&, MultiPackageCache&>())
+        .def(py::init<>(
+            [](MSolver& solver, MultiPackageCache& mpc)
+            {
+                deprecated("Use Transaction(Pool, Solver, MultiPackageCache) instead");
+                return std::make_unique<MTransaction>(solver.pool(), solver, mpc);
+            }
+        ))
```
AntoinePrv, line 219: Keeping deprecated API in Python

libmamba/src/core/transaction.cpp
```diff
@@ -670,17 +674,18 @@ namespace mamba
             solver.install_specs()
         );
 
-        if (m_transaction_context.relink_noarch && pool->installed != nullptr)
+        ::Pool* const pool_ptr = m_pool;
```
Hind-M, line 677: What's the reason for having another pool here `pool_ptr`?

libmamba/src/core/transaction.cpp
```diff
@@ -1468,7 +1473,7 @@ namespace mamba
                 }
                 else
                 {
-                    if (!need_pkg_download(s, m_multi_cache))
+                    if (!need_pkg_download(mk_pkginfo(m_pool, s), m_multi_cache))
```
Hind-M, line 1476: `mk_pkginfo` returns a `PackageInfo` right? It replaces a solvable here?

libmamba/include/mamba/core/solver.hpp
```diff
@@ -95,7 +95,7 @@ namespace mamba
         [[nodiscard]] const std::vector<MatchSpec>& neuter_specs() const;
         [[nodiscard]] const std::vector<MatchSpec>& pinned_specs() const;
 
-
+        operator const Solver*() const;
```
Klaim, line 98: I would prefer the conversions to be explicit, maybe better provided by a function like `raw()` that we used somewhere else?

libmamba/include/mamba/core/transaction.hpp
```diff
@@ -126,7 +122,7 @@ namespace mamba
             const std::vector<MatchSpec>& specs_to_install,
             MultiPackageCache& caches
         );
-        MTransaction(MSolver& solver, MultiPackageCache& caches);
+        MTransaction(MPool& pool, MSolver& solver, MultiPackageCache& caches);
```
Klaim, line 125: Because the member `m_pool` is a value and not a reference, it is not clear what is going on here. In the constructor implementation we pass `m_pool(pool)`,  what does this mean? Is it a copy?

libmamba/src/core/pool.cpp
```diff
@@ -150,13 +159,146 @@ namespace mamba
         return id;
     }
 
-    std::optional<PackageInfo> MPool::id2pkginfo(Id id)
+    namespace
+    {
+        auto make_package_info(::Solvable* s) -> PackageInfo
```
Klaim, line 164: Looks like this should be a reference as we access `*s` directly without checking. I understand this function is just moved but we can do that change too while we're at it.

libmamba/src/core/transaction.cpp
AntoinePrv: it's not a copy, `m_pool` is a `MPool` and `pool_ptr` a `::Pool*`, due to `MPool::operator ::Pool*`.
A static cast would be more explicit...

libmamba/src/core/transaction.cpp
AntoinePrv: Yes, the previous line was calling an implicit conversion from `Solvable*` to `PackageInfo`.

libmamba/include/mamba/core/solver.hpp
AntoinePrv: Not sure why this appears in the diff but this is current behavior. The isolation of libsolv into `mamba/solv-cpp` will surely get rid of these conversion, but this is not the scope of this PR.

libmamba/include/mamba/core/transaction.hpp
AntoinePrv: It's a shared reference. We had the discussion last autumn to make `MPool` interally based on an `std::shared_ptr`.
This will also change as we refactor the pool.

Hind-M (approved):

libmamba/include/mamba/core/solver.hpp
JohanMabille: Indeed these implicit conversions are everywhere regarding libsolv objects, and will be removed progressively while they are isolated from libmamba.

libmamba/include/mamba/core/transaction.hpp
Klaim: Ah ok, then why not pass it by value? that would make it clear that whatever the model, there is ownership being acquired. There is no condition to that so no reason to pass a reference.

JohanMabille: I missed your answer before I merged, I can make the change in a mechanical PR.

PR #2341, opened 2023-03-02
Remove unused libarchive header in fetch

### LINKED ISSUE #2337: Libarchive leftovers removal?
Libarchive headers are included in:

- https://github.com/mamba-org/mamba/blob/main/libmamba/include/mamba/core/fetch.hpp#L12
(apparently not used?)
- https://github.com/mamba-org/mamba/blob/main/libmamba/src/api/info.cpp#L19
used in
https://github.com/mamba-org/mamba/blob/main/libmamba/src/api/info.cpp#L152
(to get libarchive version (won't be needed in case of complete removal))
- https://github.com/mamba-org/mamba/blob/main/libmamba/src/core/package_handling.cpp
(where it's mainly used)

Hind-M:
Fix #2337 
It seems that `libarchive` usage is quite contained in `package_handling.cpp`, so I guess that no further actions are needed.
cc @JohanMabille

### DIFF
--- a/libmamba/include/mamba/core/fetch.hpp
+++ b/libmamba/include/mamba/core/fetch.hpp
@@ -9,7 +9,6 @@
 
 extern "C"
 {
-#include <archive.h>
 #include <curl/curl.h>
 }
 


JohanMabille:
It is also included in `info.cpp` and we don't seem to use it (but I might have missed it).

Hind-M:
It is used to get libarchive version in:
https://github.com/mamba-org/mamba/blob/main/libmamba/src/api/info.cpp#L152
(cf. #2337)

JohanMabille (approved):

PR #2342, opened 2023-03-02
Only full shared or full static build

JohanMabille:
This PR simplifies the cmake and removes an hybrid build that might be problematic because of mixing different runtimes.
- Removed the BUILD_STATIC_DEPS option
- The BUILD_STATIC option is now equivalent to the old BUILD_STATIC_DEPS
- Target linkage is now the same as the build type (so we do not mix static and shared anymore)
- In case we build both shared and static libraries, binaries (micromamba and mamba-package) will link with the shared libraries by default. To change that, the options `MICROMAMBA_LINKAGE` and `MAMBA_PACKAGE_LINKAGE` can be set to `STATIC`


jonashaag:
fyi @pavelzw

wolfv (approved):

wolfv:
Looks good to me!

libmamba/CMakeLists.txt
```diff
@@ -48,19 +47,15 @@ message(STATUS "libmamba binary version: v${LIBMAMBA_BINARY_VERSION}")
 
 option(BUILD_TESTS "Build libmamba C++ tests" OFF)
 option(BUILD_SHARED "Build shared libmamba library" OFF)
-option(BUILD_STATIC "Build static libmamba library" OFF)
-option(BUILD_STATIC_DEPS "Build static libmamba library with static linkage to its dependencies" OFF)
+option(BUILD_STATIC "Build static libmamba library with static linkage to its dependencies" OFF)
```
AntoinePrv, line 50: Should we take this opportunity to introduce an prefix on options: `MAMBA_BUILD_STATIC` *etc.* ?

libmamba/CMakeLists.txt
```diff
@@ -505,7 +496,7 @@ macro(libmamba_create_target target_name linkage deps_linkage output_name)
         set_target_properties(
             ${target_name}
             PROPERTIES
-            PUBLIC_HEADER "${LIBMAMBA_PUBLIC_HEADERS}"
+            #PUBLIC_HEADER "${LIBMAMBA_PUBLIC_HEADERS}"
```
AntoinePrv, line 499: Typo or should we remove?

libmamba/CMakeLists.txt
```diff
@@ -516,7 +507,7 @@ macro(libmamba_create_target target_name linkage deps_linkage output_name)
...
```
AntoinePrv, line 510: Same

libmamba/CMakeLists.txt
```diff
@@ -552,23 +543,13 @@ if (BUILD_STATIC)
...
-        libmamba_create_target(libmamba-full-static STATIC STATIC libmamba_full_static)
+        libmamba_create_target(libmamba-static STATIC libmamba_static)
     else ()
-        libmamba_create_target(libmamba-full-static STATIC STATIC libmamba_full_static)
+        libmamba_create_target(libmamba-static STATIC libmamba)
     endif ()
 endif ()
 
-if (NOT (BUILD_SHARED OR BUILD_STATIC OR BUILD_STATIC_DEPS))
+if (NOT (BUILD_SHARED OR BUILD_STATIC))
```
AntoinePrv, line 552: Should there be s single option, where true means static, false means shared?

libmamba/CMakeLists.txt
JohanMabille: I don't know, `MAMBA_BUILD_MICROMAMBA` sounds a bit weird. And there is already a lot to type to build micromamba for instance

libmamba/CMakeLists.txt
wolfv: yeah I think we should default to `SHARED`

libmamba/CMakeLists.txt
JohanMabille: I want to keep it as a reminder that we should use this instead of a manual copy of the headers. The reason for not doing it is because we have subfolder under the include directory which is not handled correctly by cmake (it might be fixed in a recent version though).

libmamba/CMakeLists.txt
JohanMabille: I think `BUILD_SHARED` and `BUILD_STATIC` are more expressive. And you may want to build both at the same time.

AntoinePrv: Ha right, it's too target, not one that transmute

libmamba/CMakeLists.txt
AntoinePrv: `MAMBA_BUILD_MICRO`?

libmamba/CMakeLists.txt
JohanMabille: > yeah I think we should default to `SHARED`

In that case if you want to build the static version only, you need to pass two options ( `BUILD_SHARED=OFF`  and `BUILD_STATIC=ON`) instead of a single one

libmamba/CMakeLists.txt
JohanMabille: still weird. And I don't think we need this prefix as long as this is not used as a framework in another repo building with cmake.

libmamba/CMakeLists.txt
JohanMabille: I can add options like:
- BUILD_ALL_SHARED
- BUILD_ALL_STATIC

The only question is should we build tests with these targets? (sometimes you just want to build micromamba for reproducing an issue or debugging, and you don't want tot build the tests).

PR #2344, opened 2023-03-03
Add strip_if functions

AntoinePrv:
- [x] Based on  #2339

### DIFF
--- a/libmamba/include/mamba/core/util_string.hpp
+++ b/libmamba/include/mamba/core/util_string.hpp
@@ -10,11 +10,12 @@
 #include <algorithm>
 #include <cstring>
 #include <iomanip>
-#include <ostream>
+#include <iterator>
 #include <sstream>
 #include <string>
 #include <string_view>
 #include <type_traits>
+#include <utility>
 #include <vector>
 
 #include "mamba/core/util_compare.hpp"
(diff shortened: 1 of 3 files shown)


libmamba/include/mamba/core/util_string.hpp
```diff
@@ -286,7 +398,7 @@ namespace mamba
         }
     }
 
-    // TODO(C++20) Take an input range and return a range
+    // TODO(C++20) This is a range
```
Hind-M, line 401: Are we willing to change something here later with C++20? If yes, maybe specify what exactly as a comment, otherwise, remove the comment.

libmamba/src/core/util_string.cpp
```diff
@@ -7,15 +7,117 @@
...
 namespace mamba
 {
-    /***************************************************
-     *  Implementation of to_lower to_upper functions  *
-     ***************************************************/
+    /****************************************
+     *  Implementation of cctype functions  *
+     ****************************************/
+
+    char is_control(char c)
```
Hind-M, line 22: Aren't all these functions supposed to return a boolean instead?

AntoinePrv: yes, good catch !

libmamba/src/core/util_string.cpp
```diff
@@ -7,15 +7,117 @@
...
+     ****************************************/
+
+    bool is_control(char c)
+    {
+        return std::iscntrl(static_cast<unsigned char>(c)) != 0;
+    }
+
+    bool is_control(wchar_t c)
+    {
+        return std::iswcntrl(static_cast<wint_t>(c)) != 0;
```
Hind-M, line 29: According to the [documentation](https://en.cppreference.com/w/cpp/string/wide/iswcntrl), I don't think that we need the `static_cast` to call this function. It also makes it inconsistent with the cast in the equivalent function above which uses `char` and that we cast to a type different from the one specified in the definition of the function, for the sake of a defined behavior (same applies for the other similar functions below).

Hind-M (approved):

libmamba/src/core/util_string.cpp
```diff
@@ -27,6 +129,10 @@ namespace mamba
         return static_cast<wchar_t>(std::tolower(static_cast<wchar_t>(c)));
```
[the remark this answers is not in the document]
JohanMabille: You can even remove the static_cast

PR #2345, opened 2023-03-03
Add split_once functions

AntoinePrv:
- [ ] Based on #2344


AntoinePrv:
Not worth it since these functions cannot differentiate between `split_once("hello$", '$')` and `split_once("hello", '$')`.

PR #2347, opened 2023-03-03
Add strip_parts functions

AntoinePrv:
Like `strip` but lets you keep what was stripped.

- [x] ~~Based on #2345~~
- [x] Add `strip_if_parts`

### DIFF
--- a/libmamba/src/core/util_string.cpp
+++ b/libmamba/src/core/util_string.cpp
@@ -253,77 +253,58 @@ namespace mamba
      *  Implementation of strip functions  *
      ***************************************/
 
-    namespace
-    {
-        // string_view has a different overload for ``find(char)`` and ``find(string_view)``
-        // so we want to leverage that.
-        template <typename Char, typename CharOrStrView>
-        std::basic_string_view<Char>
-        lstrip_impl(std::basic_string_view<Char> input, CharOrStrView chars)
-        {
-            std::size_t const start = input.find_first_not_of(chars);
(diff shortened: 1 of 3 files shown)


libmamba/src/core/util_string.cpp
```diff
@@ -554,5 +635,10 @@ namespace mamba
         {
             return 1;
         }
+        std::size_t size(const wchar_t c)
+        {
+            return 1;
```
Hind-M, line 640: Hmm is this supposed to return the size of `wchar_t`? If so, it's wrong. Otherwise, I would say that we should use `sizeof(wchar_t)` instead of returning a hardcoded value.
(I also realized that there is a `return 1;` for `char` case in the same file in `master`, to be changed to `sizeof(char)`).

AntoinePrv: It's supposed to return the size of a string. Since some functions work with both a string or a character, we use this generic overload to view a characters as a 1-character strings.

For instance in `concatenate` we can concatenate `string` or `char`. We compute the final size by calling this functions on all inputs.

Hind-M: Hmm ok, yeah so maybe rename it `str_size` or `str_len` or something similar?

Hind-M (approved):

JohanMabille (changes requested):

JohanMabille (approved):

PR #2352, opened 2023-03-06
add Repo.set_channel()

jaimergp:
I am [trying](https://github.com/conda/conda-libmamba-solver/pull/65) to use `libmambapy` with already downloaded repodatas from `conda` side. I just want `libmamba` to load a given JSON for a given channel.

I've tried setting `context.offline = True` and other tricks so `SubdirData` finds the downloaded JSON just fine, but there's some disagreement in how caching works right now, and actually at that point we already know that there's no download needed, and I have the direct path anyway.

As a result, I am trying to load it use `libmamba.Repo` directly, but the Python-exposed constructors differ from the ones `SubdirData` is using.

- `SubdirData` uses the [constructor that relies on the `RepoMetadata` helper](https://github.com/mamba-org/mamba/blob/c4a7bb86d708fa1467adb9e7a4254e74ba763e7f/libmamba/src/core/repo.cpp#L39-L53). But this is not exposed in the Python bindings.
- We can do _almost_ the same by using this [other constructor](https://github.com/mamba-org/mamba/blob/c4a7bb86d708fa1467adb9e7a4254e74ba763e7f/libmamba/src/core/repo.cpp#L55-L61) (only strings), but this one doesn't set `p_channel`. (Check what happens after `read_file(index)`.

I am assuming this is the reason why I can't use `api.Repo` directly with channel containing `MatchSpec` like `conda-forge::python`: `libsolv` will say "unsupported request" for these!

So if I can set a channel separately, then things _should_ work. This PR adds a couple bindings in `libmambapy`, hoping that's it :D


jaimergp:
This seems to fix the issue I was seeing in https://github.com/conda/conda-libmamba-solver/pull/65!

wolfv:
Looks good to me!

JohanMabille:
Hey, sorry for the late review. Would you consider binding the constructor of Repo used by SubdirData instead of adding a setter for the Channel? I think it makes more sense to fully initiliaze the object with the constructor rather than through a consrtuctor call and a setter later. Also there's a chance that we modify the API of the constructor taking only strings since it does not initialize all the data members (or even that we completely remove it).

JohanMabille:
Hey, sorry for the late review. Would you consider binding the constructor used by SubdirData instead of adding a setter for the channel? I think it makes more sense to fully initialize an object within its consturctor rather than partially initiliaze it in the constructor and then call a setter later. Also, there is a chance that we change the API of the constructor accepting strings only (or even that we remove it), since it does not fully initialize the repo object.

jaimergp:
Makes sense! I'll give it a try, thanks!

AntoinePrv:
@jaimergp I'm working on changes that will make possible to use a `Pool` and `Solver` directly from a `repodata.json` without interacting with the rest of Mamba (downloads *etc*). Would that be of interest?

AntoinePrv:
What is you timeline? We're about to to a Mamba release so we could try to get this one in. Otherwise the larger refactoring will be in another release where we will also break libmamba API (maybe a couple months).

jaimergp:
Ideally we can get this one in now 🙏

JohanMabille:
Closing this one in favor of #2398 which is definitely simpler and safer.

PR #2353, opened 2023-03-06
Fixed repoquery output of mamba when query format is JSON

JohanMabille:
@jaimergp this should fix the issue you discribe in https://github.com/mamba-org/mamba/pull/2330


Hind-M (approved):

jonashaag:
IMO we should start properly separating stdout and stderr, then this wouldn't be a problem since you output JSON to stdout and errors to stderr...

jaimergp:
Thanks!

PR #2354, opened 2023-03-06
Removed libmamba warnings

JohanMabille:


libmamba/src/core/output.cpp
```diff
@@ -122,7 +122,11 @@ namespace mamba
                 m_padding = std::vector<int>(n_col, 1);
             }
 
-            std::size_t total_length = std::accumulate(cell_sizes.begin(), cell_sizes.end(), 0);
+            std::size_t total_length = std::accumulate(
+                cell_sizes.begin(),
+                cell_sizes.end(),
+                static_cast<std::size_t>(0)
```
AntoinePrv, line 128: Nit: also using `0u` is shorter to read.

libmamba/src/core/output.cpp
```diff
@@ -146,7 +150,7 @@ namespace mamba
                             out,
                             "{: >{}}",
                             fmt::styled(row[j].s, row[j].style),
-                            cell_sizes[j] + m_padding[j]
+                            cell_sizes[j] + static_cast<std::size_t>(m_padding[j])
```
AntoinePrv, line 153: Should `m_padding` `valut_type` be unsigned?

libmamba/src/core/output.cpp
```diff
@@ -221,13 +226,14 @@ namespace mamba
...
-            int block_width = padding + data_max_width;
+            std::size_t block_width = padding + data_max_width;
 
             auto sorted_data = data;
             std::sort(sorted_data.begin(), sorted_data.end(), string_comparison);
             for (const auto& d : sorted_data)
             {
-                int p = block_width - d.size();
+                // p is positive by construction of block_width
+                size_t p = block_width - d.size();
```
AntoinePrv, line 236:
```suggestion
                std::size_t p = block_width - d.size();
```

libmamba/src/core/satisfiability_error.cpp
```diff
@@ -1144,13 +1144,13 @@ namespace mamba
...
             auto [out, _] = visit_node(m_pbs.root_node(), path.begin());
-            path.resize(out - path.begin());
+            path.resize(static_cast<std::size_t>(out - path.begin()));
             return path;
         }
 
         auto TreeDFS::node_uninstallable(node_id id) -> Status
         {
-            auto installables_contains = [&](auto&& id) { return leaf_installables.contains(id); };
+            auto installables_contains = [&](auto&& lid) { return leaf_installables.contains(lid); };
```
AntoinePrv, line 1153: `lid` feels weird here. Perhaps using `the_id`?

AntoinePrv (commented):
Thanks for doing this cleanup! My general comments, although not necessarily within the scope of this PR, are:
- Perhaps some types are wrong in the first place: unsigned types for padding, sizes, timeout... If a negative value has a specific meaning, I prefer to see a `std::optional`, if multiple negative values have a special meaning, a `std::variant` with an enum...
- I think we should add `const`, even to local vars when we have the opportunity.
- More related to the changes here is the generalized use of `static_cast`. My idea when introducing `safe_num_cast` was that it would be the default, and that `static_cast` would be used only when one can prove that the conversion is safe, mostly around signed and unsigned conversion in the std lib for instance, or leaving a comment why the cast is safe. @Klaim @Hind-M  any opinion on this?

libmamba/include/mamba/core/util.hpp
```diff
@@ -215,7 +215,7 @@ namespace mamba
         // Returns 0 if `is_locked() == false`.
         std::size_t count_lock_owners() const
         {
-            return impl.has_value() ? impl.value().use_count() : 0;
+            return std::size_t(impl.has_value() ? impl.value().use_count() : 0);
```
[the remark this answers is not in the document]
JohanMabille: I prefer to keep the explicit instantiation, this way it is easier to see that we have potentially the wrong type during a later refactoring

libmamba/src/core/fetch.cpp
```diff
@@ -247,7 +247,7 @@ namespace mamba
 
     std::size_t DownloadTarget::get_default_retry_timeout()
     {
-        return Context::instance().retry_timeout;
+        return static_cast<std::size_t>(Context::instance().retry_timeout);
```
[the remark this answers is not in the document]
JohanMabille: We could do that in a further refactoring, I want to limit the data member type changes for now to keep this mechanical.

libmamba/src/core/fetch.cpp
```diff
@@ -615,30 +618,33 @@ namespace mamba
 
         if (!total_to_download && target->expected_size())
         {
-            target->m_progress_bar.update_current(now_downloaded);
+            target->m_progress_bar.update_current(static_cast<std::size_t>(now_downloaded));
```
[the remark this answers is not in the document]
JohanMabille: It was not checked before, I want to keep this mechanical

libmamba/src/core/output.cpp
JohanMabille: 0u means unsigned, but not necessarily unsigned long long.

libmamba/src/core/output.cpp
JohanMabille: Yes, but this i sa data member

libmamba/src/core/satisfiability_error.cpp
JohanMabille: lid stands for `local id`, this is king of a convention like `m_` and `p_` for data member.

libmamba/include/mamba/core/util.hpp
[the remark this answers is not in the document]
AntoinePrv: What about the following, that way we still get an error in case of narrowing during a future refactoring.
```suggestion
            return std::size_t{impl.has_value() ? impl.value().use_count() : 0};
```

libmamba/src/core/output.cpp
AntoinePrv: But the implicit conversion is well defined then

AntoinePrv (approved):

Hind-M:
Yeah I think that if we want to keep this mechanical to get rid of the warnings and refactor later, we can go like this. But I agree that we should recheck that all types are relevant and change them accordingly. I'm wondering about the best/easiest way to do it, remove the warnings and go through the whole code base again to check everything (or almost) or take these warnings as an opportunity to have pointers to where to look...
Anyway it's merged now ^^

JohanMabille:
> I'm wondering about the best/easiest way to do it, remove the warnings and go through the whole code base again to check everything (or almost) or take these warnings as an opportunity to have pointers to where to look..

For most of them, we will make a pass when refactoring the code, therefore we will "naturally" wonder if the types are correct (because a `static_cast` is often a bad smell for me, except when interfacing with external libraries, especially C libraries)

PR #2185, opened 2022-12-20
Micromamba server

wolfv:
So we can resolve more than 1 environment in a row

TODO:

- [x] kick out virtual packages from Pool / kick out previous "installed" repo

PR #2362, opened 2023-03-08
Revert "Micromamba server (#2185)"

JohanMabille:
This reverts commit 7866468c516c004998756cb74434c01d61fbea5e.

Just to test if the responsible of the failure is #2185 (test failure after fixing the compilation error) or the PR that was merged before (and most of the builds were canceled because #2185 was merged right after).


JohanMabille:
Ok so this proves that the responsible for the failure is not #2360. LEt's close this.

PR #2370, opened 2023-03-10
Improved Sanitizers options and usage

Klaim:


Klaim:
As observed locally, enabling ASAN will block execution of both libmamba C++ tests and micromamba tests.

My investigation lead me to believe this issue is present at least since v1.0.0 but probably before. I can confirm however that it was not present when I first introduced the sanitizer options for v0.23.0 so something changed in-between that leads to ASAN (or maybe libsolv?) blocking completely. Debugging the libmamba C++ tests leads me to code stopping in ASAN allocation code. Searching about that do lead to some bugs related to ASAN on GCC having deadlocks on allocation when libc is not ABI-compatible or something like that, but none of the issues have been fixed because the reports were not reliable or updated enough.

PR #2373, opened 2023-03-13
Add parsing of Conda version

AntoinePrv:
This makes it possible to parse Conda versions, the goal is to eventually use it inside libsolv.
Currenlty however, this is not used anywhere.

Improvement:
- [ ] By the specs (or tests), a version is an arbitrary sized vector of arbitrary sized vector. However we knkow that in most case those would contain only a few (less thatn 4) elements. We can optimize this case with a small vector or an allocator (this may be important when we use it in libsolv?)
- [ ] Comparison operations iterate multiple times over input for simplicity. Perhaps this could all be improved by doing some C++20-like three-way comparison (but need to re-write three-way comparison for ``std::string``).


libmamba/include/mamba/specs/version.hpp
```diff
@@ -0,0 +1,202 @@
...
+    class VersionPartAtom
+    {
+    public:
+
+        VersionPartAtom() noexcept = default;
+        VersionPartAtom(std::size_t numeral) noexcept;
+        VersionPartAtom(std::size_t numeral, std::string_view litteral);
+        // The use of a template is only meant to prevent ambiguous conversions
+        template <typename Char>
+        VersionPartAtom(std::size_t numeral, std::basic_string<Char>&& litteral);
```
Klaim, line 36: Do we actually need to handle anything else than `char` and assume UTF-8?

libmamba/include/mamba/specs/version.hpp
```diff
@@ -0,0 +1,202 @@
...
+     * A sequence of VersionPart meant to represent all parts of a version.
+     *
+     * CommonVersion are composed of an aribtrary postive number parts, such as major, minor.
+     * They are typically separated by dots, for instance the three parts in 3.0post1dev.4 are
+     * {{3, ""}}, {{0, "post"}, {1, "dev"}}, and {{4, ""}}.
+     *
+     * @see  Version::parse for how this is computed from strings.
+     * @todo Use a small vector of expected size 4 if performance ar not good enough.
+     */
+    using CommonVersion = std::vector<VersionPart>;
```
Klaim, line 81: So ok we need arbitrary number parts (assuming that's a requirement here) but having vectors of vectors feels like it could be simplified/flatten in some ways.
Would it be possible to use one vector of VersionPartAtom with a way to know the begin and end of each version part?

libmamba/src/specs/version.cpp
```diff
@@ -0,0 +1,562 @@
...
+#include "mamba/core/error_handling.hpp"
+#include "mamba/core/util_cast.hpp"
+#include "mamba/core/util_string.hpp"
+#include "mamba/specs/version.hpp"
+
+namespace mamba::specs
+{
+    namespace
+    {
+        // TODO(C++20) use operator<=>
```
Klaim, line 23: Note: most tools detecting TODOs will detect `TODO:` (with the colon) specifically.

libmamba/src/specs/version.cpp
```diff
@@ -0,0 +1,562 @@
...
+    {
+        return fmt::format("{}", *this);
+    }
+
+    namespace
+    {
+        /**
+         * Compare two range where some trailing elements can be considered as empty.
+         *
+         * If ``0`` is considered "empty" then all the ranges ``[1, 2] and ``[1, 2, 0]``,
```
Klaim, line 216: What is this double anti-quote syntax?

libmamba/src/specs/version.cpp
```diff
@@ -0,0 +1,562 @@
...
+                Version::part_delim,
+                Version::part_delim_alt,
+                Version::part_delim_special,
+            };
+            static constexpr auto delims = std::string_view{ delims_buf.data(), delims_buf.size() };
+
+            CommonVersion parts = {};
+            auto tail = str;
+            std::size_t tail_delim_pos = 0;
+            for (;;)
```
Klaim, line 488: I have a slight preference for `while(true)` but ok.

libmamba/include/mamba/specs/version.hpp
```diff
@@ -0,0 +1,202 @@
...
+#include <utility>
+#include <vector>
+
+#include <fmt/format.h>
+
+namespace mamba::specs
+{
+
+    /**
+     * A succession of a number and a  lowercase litteral.
```
wolfv, line 21: extra space, and `literal` is without double `t` :)

libmamba/include/mamba/specs/version.hpp
Klaim: That would however complicate the tests...

Klaim (approved):
I made some comments but globally it looks fine to me.

wolfv (approved):
Impressive - very nice :)

libmamba/include/mamba/specs/version.hpp
AntoinePrv: Would complicate much the very custom comparisons as well. It would help to have some `std::span` and `std::ranges` for sure!

Or would an allocator aware design make this better?

libmamba/src/specs/version.cpp
AntoinePrv: Same as `0` in makrdown but also works with reStructured (Sphinx).

libmamba/src/specs/version.cpp
AntoinePrv: Hopefully they won't stay here for too long ;)

libmamba/src/specs/version.cpp
Klaim: Ah, I thought we were using Markdown over Doxygen syntax

PR #2375, opened 2023-03-14
Arrange header sorting

### LINKED ISSUE #2368: Change clang-format rules
Because of #2290 and `includes` sorting, it may happen quite often that local/mamba headers are included before third party libraries or even c++ standard library headers which is not recommended and prone to bugs.
cc @AntoinePrv

AntoinePrv:
Close #2368


Hind-M (approved):
Nice! Thanks!

PR #2381, opened 2023-03-15
is this file being used

AntoinePrv:


Hind-M:
Hmm, it seems that it's not used indeed... I guess that this should be replaced with a cpp test instead.
Opening an issue to not forget.

PR #2385, opened 2023-03-16
add option to relocate prefix.

DerThorsten:
## Problem
Let's say we have two machines:

Alice with a homedir: `/home/alice`
Bob: `/home/bob`
Bob wants a new env at `/home/bob/myenv`
but we cannot create conda/mamba envs on bobs machine directly.
So we need to create the env at Alice's machine, zip it and send it to Bob.

So alide tries

* `micromamba create -p /home/bob/myenv`
But this does not work because creating `/home/bob/myenv` on Alice's machine is impossible.

## Solution

add `relocate-prefix` option to the create command.

`micromamba create  -p /home/alice/myenv  --relocate-prefix /home/bob/myenv` will create all files at `/home/alice/myenv` but they are relocated to `/home/bob/myenv`
st. we can send this to Bob, and on Bob's machine, the env would behave like an env created on his machine.


libmamba/include/mamba/core/transaction_context.hpp
```diff
@@ -39,17 +39,30 @@ namespace mamba
...
         bool has_python;
         fs::u8path target_prefix;
         fs::u8path site_packages_path;
         fs::u8path python_path;
         std::string python_version;
         std::string old_python_version;
         std::string short_python_version;
+
+        fs::u8path relocate_prefix;
+        bool relocate = false;
```
wolfv, line 64: I think we can just always store two `u8path`. One that is the place where files are installed to, and one that is used to exchange prefix information from binarys. Most of the time they will just hbe the same.

wolfv (changes requested):

libmamba/include/mamba/core/transaction_context.hpp
DerThorsten: agreed

JohanMabille (approved):

PR #2398, opened 2023-03-23
Handle `url` via ChannelBuilder in Repo constructor

jaimergp:
Alternative to https://github.com/mamba-org/mamba/pull/2352.

This seems like an easier fix and works for my intended use case. No new APIs, just a fix where we do use that `url` parameter. But maybe this is the wrong way to do it?

### DIFF
--- a/libmamba/src/core/repo.cpp
+++ b/libmamba/src/core/repo.cpp
@@ -13,13 +13,13 @@ extern "C"  // Incomplete header
 #include <solv/repo_conda.h>
 }
 
+#include "mamba/core/channel_builder.hpp"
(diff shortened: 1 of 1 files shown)


jaimergp:
Confirmed this approach works too. @JohanMabille @AntoinePrv @wolfv - up to you to choose which one is better, but this one looks simpler.

JohanMabille (approved):

JohanMabille:
@jaimergp sorry I forgot this one before releasing, I can tag a patch release once this one is in.

jaimergp:
That'd be awesome, thank you @JohanMabille!

Issue #2302, opened 2023-02-16
Solving Isolation


AntoinePrv:
This is a meta issue to plan the isolation of libsolv and the definition of a new solving interface.

### A -`solv-cpp` Safe and useful wrapping of libsolv
The goal is to build self contained C++ wrappers around libsolv that makes it safer and consolidate verbose patterns into utility functions.
Things such as:
- RAII and value semantic
- Iterators
- Visitors (to replace macros)

But not:
- Mamba specific logic
- Context
- Any `#include` from other mamba headers
- Wrap all `libsolv` functions for the sake of wrapping
- Hide the original `libsolv` object (because only selected functions would be wrapped).

All this wrapping will be private to Mamba. Current `MSolver`, `MPool`... will make use of these new wrappers but keep their interface. This will help getting some code outside of the current implementation to keep only the Mamba logic.

### B- Specs ownership
Currently, version comparison, matchspecs, even repodata.json parsing is done in libsolv.
This leads to duplication (e.g. `mamba::MatchSpecs`), and difficulty to evolve (let alone lack of unit testing).

Right now, some functions such as `MSolver::add_channel_specific_job`, implement custom libsolv logic when this could be handled (and unit tested) through proper abstraction (MatchSpecs, Version...).

The goal is to implement and test (with conda test cases) these abstractions and use them in `libsolv` through function callbacks (need to send PR on libsolv to make it possible).

The repodata would also be read by  `Mamba` and the packages added explicitly into libsolv.

### C - New solving interface
Once all abstractions are in place, libsolv is only used as a SAT solver. We can design the new solving interface that we would like.
There would be a generic interface (or trait). All solvers, starting with libsolv and other in the future, would implement such interface, hiding all their dependencies. This would make it possible to switch solver without any code change, possibly even at runtime (which shows that things are properly abstracted).

_To be completed for more details on the new interface..._

PR #2401, opened 2023-03-23
Wrap ::Pool and ::Repo

AntoinePrv:
For  #2302

- Add ObjPool
- Use ObjPool in MPool
- Add skeleton ObjRepoView
- Add solv Repo io

If https://github.com/openSUSE/libsolv/pull/526 is merged and release, we can use `std::string_view` for setting solvable attributes.

### DIFF
--- a/libmamba/src/solv-cpp/ids.hpp
+++ b/libmamba/src/solv-cpp/ids.hpp
@@ -17,6 +17,7 @@ namespace mamba::solv
     using SolvableId = ::Id;
 
     using RelationFlag = int;
+    using DistType = int;
 }
 
 #endif
--- a/libmamba/src/solv-cpp/ids.hpp
+++ b/libmamba/src/solv-cpp/ids.hpp
@@ -1,0 +1,22 @@
+// Copyright (c) 2023, QuantStack and Mamba Contributors
+//
+// Distributed under the terms of the BSD 3-Clause License.
+//
+// The full license is in the file LICENSE, distributed with this software.
+
+#ifndef MAMBA_SOLV_IDS_HPP
+#define MAMBA_SOLV_IDS_HPP
+
+#include <solv/pooltypes.h>
+
+namespace mamba::solv
+{
+    using StringId = ::Id;
+    using DependencyId = ::Id;
+    using RepoId = ::Id;
+    using SolvableId = ::Id;
+
+    using RelationFlag = int;
+}
+
+#endif
--- a/libmamba/src/solv-cpp/pool.cpp
+++ b/libmamba/src/solv-cpp/pool.cpp
@@ -242,18 +242,18 @@ namespace mamba::solv
 
     auto ObjPool::get_solvable(SolvableId id) const -> std::optional<ObjSolvableViewConst>
     {
-        if (const ::Solvable* s = ::pool_id2solvable(raw(), id); s != nullptr)
+        if (const ::Solvable* s = ::pool_id2solvable(raw(), id))
         {
-            return ObjSolvableViewConst{ s };
+            return ObjSolvableViewConst{ *s };
         }
         return std::nullopt;
     }
 
     auto ObjPool::get_solvable(SolvableId id) -> std::optional<ObjSolvableView>
     {
-        if (::Solvable* s = ::pool_id2solvable(raw(), id); s != nullptr)
+        if (::Solvable* s = ::pool_id2solvable(raw(), id))
         {
-            return ObjSolvableView{ s };
+            return ObjSolvableView{ *s };
         }
         return std::nullopt;
     }
--- a/libmamba/src/solv-cpp/pool.cpp
+++ b/libmamba/src/solv-cpp/pool.cpp
@@ -198,13 +198,21 @@ namespace mamba::solv
         return false;
     }
 
-    auto ObjPool::get_solvable(SolvableId id) const -> ObjSolvableViewConst
+    auto ObjPool::get_solvable(SolvableId id) const -> std::optional<ObjSolvableViewConst>
     {
-        return ObjSolvableViewConst{ ::pool_id2solvable(raw(), id) };
+        if (const ::Solvable* s = ::pool_id2solvable(raw(), id); s != nullptr)
+        {
+            return ObjSolvableViewConst{ s };
+        }
+        return std::nullopt;
     }
 
-    auto ObjPool::get_solvable(SolvableId id) -> ObjSolvableView
+    auto ObjPool::get_solvable(SolvableId id) -> std::optional<ObjSolvableView>
     {
-        return ObjSolvableView{ ::pool_id2solvable(raw(), id) };
+        if (::Solvable* s = ::pool_id2solvable(raw(), id); s != nullptr)
+        {
+            return ObjSolvableView{ s };
+        }
+        return std::nullopt;
     }
 }
--- a/libmamba/src/solv-cpp/pool.cpp
+++ b/libmamba/src/solv-cpp/pool.cpp
@@ -42,7 +42,7 @@ namespace mamba::solv
     auto ObjPool::find_string(std::string_view str) const -> std::optional<StringId>
     {
         assert(str.size() <= std::numeric_limits<unsigned int>::max());
-        ::Id const id = pool_strn2id(
+        ::Id const id = ::pool_strn2id(
             const_cast<::Pool*>(raw()),  // Safe because we do not create
             str.data(),
             static_cast<unsigned int>(str.size()),
@@ -55,7 +55,7 @@ namespace mamba::solv
     {
         assert(str.size() <= std::numeric_limits<unsigned int>::max());
         // Note: libsolv cannot report failure to allocate
-        ::Id const id = pool_strn2id(
+        ::Id const id = ::pool_strn2id(
             raw(),
             str.data(),
             static_cast<unsigned int>(str.size()),
@@ -68,13 +68,13 @@ namespace mamba::solv
     auto ObjPool::get_string(StringId id) const -> std::string_view
     {
         assert(!ISRELDEP(id));
-        return pool_id2str(raw(), id);
+        return ::pool_id2str(raw(), id);
     }
 
     auto ObjPool::find_dependency(StringId name_id, RelationFlag flag, StringId version_id) const
         -> std::optional<DependencyId>
     {
-        ::Id const id = pool_rel2id(
+        ::Id const id = ::pool_rel2id(
             const_cast<::Pool*>(raw()),  // Safe because we do not create
             name_id,
             version_id,
@@ -88,7 +88,7 @@ namespace mamba::solv
         -> DependencyId
     {
         // Note: libsolv cannot report failure to allocate
-        ::Id const id = pool_rel2id(
+        ::Id const id = ::pool_rel2id(
             raw(),
             name_id,
             version_id,
@@ -103,25 +103,25 @@ namespace mamba::solv
     auto ObjPool::get_dependency_name(DependencyId id) const -> std::string_view
     {
         assert(ISRELDEP(id));
-        return pool_id2str(raw(), id);
+        return ::pool_id2str(raw(), id);
     }
 
     auto ObjPool::get_dependency_version(DependencyId id) const -> std::string_view
     {
         assert(ISRELDEP(id));
-        return pool_id2evr(raw(), id);
+        return ::pool_id2evr(raw(), id);
     }
 
     auto ObjPool::get_dependency_relation(DependencyId id) const -> std::string_view
     {
         assert(ISRELDEP(id));
-        return pool_id2rel(raw(), id);
+        return ::pool_id2rel(raw(), id);
     }
 
     auto ObjPool::dependency_to_string(DependencyId id) const -> std::string
     {
         assert(ISRELDEP(id));
-        return pool_dep2str(
+        return ::pool_dep2str(
             // Not const in because function may allocate in pool tmp alloc space
             const_cast<::Pool*>(raw()),
             id
@@ -143,14 +143,14 @@ namespace mamba::solv
 
     void ObjPool::create_whatprovides()
     {
-        pool_createwhatprovides(raw());
+        ::pool_createwhatprovides(raw());
     }
 
     auto ObjPool::add_repo(std::string_view name) -> RepoId
     {
         const auto* repo_ptr = repo_create(
             raw(),
-            pool_id2str(raw(), add_string(name))  // Shenanigan to make it string_view compatible
+            ::pool_id2str(raw(), add_string(name))  // Shenanigan to make it string_view compatible
         );
         // No function exists to create a repo id
         assert(raw()->repos[raw()->nrepos - 1] == repo_ptr);
@@ -159,17 +159,34 @@ namespace mamba::solv
 
     auto ObjPool::get_repo(RepoId id) -> ObjRepoView
     {
-        return ObjRepoView{ pool_id2repo(raw(), id) };
+        return ObjRepoView{ ::pool_id2repo(raw(), id) };
     }
 
     auto ObjPool::get_repo(RepoId id) const -> ObjRepoViewConst
     {
         // Safe because we make the Repo deep const
-        return ObjRepoViewConst{ pool_id2repo(const_cast<::Pool*>(raw()), id) };
+        return ObjRepoViewConst{ ::pool_id2repo(const_cast<::Pool*>(raw()), id) };
+    }
+
+    auto ObjPool::n_repos() const -> std::size_t
+    {
+        // Id 0 is special
+        assert(raw()->nrepos >= 1);
+        return static_cast<std::size_t>(raw()->nrepos - 1);
     }
 
     void ObjPool::remove_repo(RepoId id, bool reuse_ids)
     {
         repo_free(get_repo(id).raw(), static_cast<int>(reuse_ids));
+    }
+
+    auto ObjPool::get_solvable(SolvableId id) const -> ObjSolvableViewConst
+    {
+        return ObjSolvableViewConst{ ::pool_id2solvable(raw(), id) };
+    }
+
+    auto ObjPool::get_solvable(SolvableId id) -> ObjSolvableView
+    {
+        return ObjSolvableView{ ::pool_id2solvable(raw(), id) };
     }
 }
--- a/libmamba/src/solv-cpp/pool.cpp
+++ b/libmamba/src/solv-cpp/pool.cpp
@@ -1,0 +1,147 @@
+// Copyright (c) 2023, QuantStack and Mamba Contributors
+//
+// Distributed under the terms of the BSD 3-Clause License.
+//
+// The full license is in the file LICENSE, distributed with this software.
+
+#include "solv-cpp/pool.hpp"
+
+#include <cassert>
+#include <limits>
+
+#include <solv/pool.h>
+#include <solv/poolid.h>
+#include <solv/pooltypes.h>
+#include <solv/selection.h>
+
+namespace mamba::solv
+{
+    void ObjPool::PoolDeleter::operator()(::Pool* ptr)
+    {
+        ::pool_free(ptr);
+    }
+
+    ObjPool::ObjPool()
+        : m_pool(::pool_create())
+    {
+    }
+
+    ObjPool::~ObjPool() = default;
+
+    auto ObjPool::raw() -> ::Pool*
+    {
+        return m_pool.get();
+    }
+
+    auto ObjPool::raw() const -> const ::Pool*
+    {
+        return m_pool.get();
+    }
+
+    auto ObjPool::find_string(std::string_view str) const -> std::optional<StringId>
+    {
+        assert(str.size() <= std::numeric_limits<unsigned int>::max());
+        ::Id const id = pool_strn2id(
+            const_cast<::Pool*>(raw()),  // Safe because we do not create
+            str.data(),
+            static_cast<unsigned int>(str.size()),
+            /* .create= */ 0
+        );
+        return (id == 0) ? std::nullopt : std::optional(id);
+    }
+
+    auto ObjPool::add_string(std::string_view str) -> StringId
+    {
+        assert(str.size() <= std::numeric_limits<unsigned int>::max());
+        // Note: libsolv cannot report failure to allocate
+        ::Id const id = pool_strn2id(
+            raw(),
+            str.data(),
+            static_cast<unsigned int>(str.size()),
+            /* .create= */ 1
+        );
+        assert(id != 0);
+        return id;
+    }
+
+    auto ObjPool::get_string(StringId id) const -> std::string_view
+    {
+        assert(!ISRELDEP(id));
+        return pool_id2str(raw(), id);
+    }
+
+    auto ObjPool::find_dependency(StringId name_id, RelationFlag flag, StringId version_id) const
+        -> std::optional<DependencyId>
+    {
+        ::Id const id = pool_rel2id(
+            const_cast<::Pool*>(raw()),  // Safe because we do not create
+            name_id,
+            version_id,
+            flag,
+            /* .create= */ 0
+        );
+        return (id == 0) ? std::nullopt : std::optional(DependencyId{ id });
+    }
+
+    auto ObjPool::add_dependency(StringId name_id, RelationFlag flag, StringId version_id)
+        -> DependencyId
+    {
+        // Note: libsolv cannot report failure to allocate
+        ::Id const id = pool_rel2id(
+            raw(),
+            name_id,
+            version_id,
+            flag,
+            /* .create= */ 1
+        );
+        assert(id != 0);
+        assert(ISRELDEP(id));
+        return id;
+    }
+
+    auto ObjPool::get_dependency_name(DependencyId id) const -> std::string_view
+    {
+        assert(ISRELDEP(id));
+        return pool_id2str(raw(), id);
+    }
+
+    auto ObjPool::get_dependency_version(DependencyId id) const -> std::string_view
+    {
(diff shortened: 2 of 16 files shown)


libmamba/src/solv-cpp/pool.cpp
```diff
@@ -0,0 +1,254 @@
...
+
+    auto ObjPool::add_string(std::string_view str) -> StringId
+    {
+        assert(str.size() <= std::numeric_limits<unsigned int>::max());
+        // Note: libsolv cannot report failure to allocate
+        ::Id const id = ::pool_strn2id(
+            raw(),
+            str.data(),
+            static_cast<unsigned int>(str.size()),
+            /* .create= */ 1
```
Klaim, line 74: Consider using a named constant to clarify these calls parameters? An `enum class`?

libmamba/src/solv-cpp/ids.hpp
```diff
@@ -0,0 +1,23 @@
...
+#include <solv/pooltypes.h>
+
+namespace mamba::solv
+{
+    using StringId = ::Id;
+    using DependencyId = ::Id;
+    using RepoId = ::Id;
+    using SolvableId = ::Id;
+
+    using RelationFlag = int;
```
Klaim, line 19: Comments documenting what these types are supposed to represent and which values are valid would be helpful.

libmamba/src/solv-cpp/pool.cpp
```diff
@@ -0,0 +1,255 @@
...
+        return ::pool_id2evr(raw(), id);
+    }
+
+    auto ObjPool::get_dependency_relation(DependencyId id) const -> std::string_view
+    {
+        assert(ISRELDEP(id));
+        return ::pool_id2rel(raw(), id);
+    }
+
+    auto ObjPool::dependency_to_string(DependencyId id) const -> std::string
```
Klaim, line 133: Why does this functoin returns `std::string` while the other functions returns `std::string_view`?

libmamba/src/solv-cpp/solvable.hpp
```diff
@@ -0,0 +1,226 @@
...
+#include <string>
+#include <string_view>
+#include <utility>
+
+#include <solv/pooltypes.h>
+
+#include "solv-cpp/ids.hpp"
+#include "solv-cpp/queue.hpp"
+
+typedef struct s_Solvable Solvable;
```
Klaim, line 19: `using Solvable = s_Solvable`

libmamba/src/solv-cpp/pool.cpp
```diff
@@ -0,0 +1,255 @@
...
+        return ::pool_dep2str(
+            // Not const in because function may allocate in pool tmp alloc space
+            const_cast<::Pool*>(raw()),
+            id
+        );
+    }
+
+    auto ObjPool::select_solvables(const ObjQueue& job) const -> ObjQueue
+    {
+        ObjQueue solvables = {};
```
Klaim, line 145: Shouldnt all this be part of a constructor of `ObjQueue`?

libmamba/src/solv-cpp/pool.cpp
```diff
@@ -0,0 +1,255 @@
...
+        if (!has_repo(id))
+        {
+            return std::nullopt;
+        }
+        return { ObjRepoView{ ::pool_id2repo(raw(), id) } };
+    }
+
+    auto ObjPool::get_repo(RepoId id) const -> std::optional<ObjRepoViewConst>
+    {
+        if (!has_repo(id))
```
Klaim, line 188: Consider implementing this function as `return const_cast<ObjPool*>(this)->get_repo();` ?

libmamba/src/solv-cpp/pool.cpp
```diff
@@ -0,0 +1,255 @@
...
+        {
+            return std::nullopt;
+        }
+        // Safe because we make the Repo deep const
+        return { ObjRepoViewConst{ ::pool_id2repo(const_cast<::Pool*>(raw()), id) } };
+    }
+
+    auto ObjPool::n_repos() const -> std::size_t
```
Klaim, line 196: Consider renaming this to `repo_count()` or `count_repos()` if it's usage is strictly for our C++ codebase.

libmamba/src/solv-cpp/pool.cpp
```diff
@@ -0,0 +1,255 @@
...
+    {
+        // Id 0 is special
+        assert(raw()->urepos >= 0);
+        return static_cast<std::size_t>(raw()->urepos);
+    }
+
+    auto ObjPool::remove_repo(RepoId id, bool reuse_ids) -> bool
```
Klaim, line 203: Consider using an `enum class : int` (or just `enum : int`) for `reuse_ids` to clarify the whole function.

libmamba/src/solv-cpp/pool.cpp
```diff
@@ -0,0 +1,255 @@
...
+        {
+            ::repo_free(get_repo(id).value().raw(), static_cast<int>(reuse_ids));
+            return true;
+        }
+        return false;
+    }
+
+    auto ObjPool::installed_repo() const -> std::optional<ObjRepoViewConst>
+    {
+        if (const auto* const installed_ptr = raw()->installed; installed_ptr != nullptr)
```
Klaim, line 215: Note that the second part of this if (after `;`) is not necessary, but ok if you want to make it explicit/clear.

libmamba/src/solv-cpp/pool.hpp
```diff
@@ -0,0 +1,214 @@
...
+    void ObjPool::for_each_solvable(UnaryFunc func)
+    {
+        // Safe optional unchecked because we iterate over available values
+        return for_each_solvable_id([this, func](SolvableId id) { func(get_solvable(id).value()); });
+    }
+
+    template <typename Func>
+    void ObjPool::set_debug_callback(Func callback)
+    {
+        m_user_debug_callback.reset(new Func(std::move(callback)));
```
Klaim, line 201: Did you consider `m_user_debug_callback` as `std::function<...>` and passing a pointer to it in `pool_set_debug_callback` instead?
It looks like `m_user_debug_callback` is passed as data to `pool_set_debug_callback` and we have the `debug_callback` reinterpreting the data anyway, so we dont have to store the real callback in a complicated way. Am I missing something?

libmamba/src/solv-cpp/pool.hpp
```diff
@@ -0,0 +1,214 @@
...
+ *  Implementation of ObjPool  *
+ *******************************/
+
+#include <solv/pool.h>
+
+namespace mamba::solv
+{
+
+    template <typename UnaryFunc>
+    void ObjPool::for_each_repo_id(UnaryFunc func) const
```
Klaim, line 116: Consider passing `UnaryFunc&&` here.

libmamba/src/solv-cpp/pool.hpp
```diff
@@ -0,0 +1,214 @@
...
+    {
+        const ::Pool* const pool = raw();
+        const ::Repo* repo = nullptr;
+        RepoId repo_id = 0;
+        FOR_REPOS(repo_id, repo)
+        {
+            func(repo_id);
```
Klaim, line 123: `func` can potentially throw. Is it ok here?
If not, consider using `safe_invoke` to wrap the call and gather the errors (see https://github.com/mamba-org/mamba/blob/main/libmamba/include/mamba/core/invoke.hpp) - which lacks documentation, sorry!

libmamba/src/solv-cpp/pool.hpp
```diff
@@ -0,0 +1,214 @@
...
+
+    template <typename UnaryFunc>
+    void ObjPool::for_each_whatprovides_id(DependencyId dep, UnaryFunc func) const
+    {
+        auto* const pool = const_cast<::Pool*>(raw());
+        SolvableId id = 0;
+        ::Id offset = 0;  // Not really an Id
+        FOR_PROVIDES(id, offset, dep)
+        {
+            func(id);
```
Klaim, line 149: Same remark than for `for_each_repo_id`.

libmamba/src/solv-cpp/repo.hpp
```diff
@@ -0,0 +1,149 @@
...
+#include <solv/repo.h>
+
+namespace mamba::solv
+{
+    /****************************************
+     *  Implementation of ObjRepoViewConst  *
+     ****************************************/
+
+    template <typename UnaryFunc>
+    void ObjRepoViewConst::for_each_solvable_id(UnaryFunc func) const
```
Klaim, line 113: Same safe invocation remark for these functions.

libmamba/src/solv-cpp/repo.cpp
```diff
@@ -0,0 +1,278 @@
...
+        return m_repo;
+    }
+
+    auto ObjRepoViewConst::id() const -> RepoId
+    {
+        // No function exists to get a repo id
+        const ::Pool* pool = raw()->pool;
+        const ::Repo* repo = nullptr;
+        int repoid = 0;
+        FOR_REPOS(repoid, repo)
```
Klaim, line 42: Could we cache the id after having done the search once? If the id doesnt change for the same value that might avoid having to loop through the ids each time we call this function that looks like a cheap getter from outside...

libmamba/src/solv-cpp/repo.cpp
```diff
@@ -0,0 +1,278 @@
...
+    }
+
+    void ObjRepoViewConst::write(std::filesystem::path solv_file) const
+    {
+        auto* fp = std::fopen(solv_file.string().c_str(), "wb");
+        if (!fp)
+        {
+            throw std::system_error(errno, std::generic_category());
+        }
+        const auto write_res = ::repo_write(const_cast<::Repo*>(raw()), fp);
```
Klaim, line 98: If `repo_write` can call callbacks, then we should raii `fp`?

libmamba/src/solv-cpp/repo.cpp
```diff
@@ -0,0 +1,278 @@
...
+        const auto close_res = std::fclose(fp);
+        if (write_res != 0)
+        {
+            // TODO(C++20) fmt::format
```
Klaim, line 102: Why not just use `fmt` here?

libmamba/src/solv-cpp/repo.cpp
```diff
@@ -0,0 +1,278 @@
...
+            ss << "Unable to write file '" << solv_file << '\'';
+            throw std::runtime_error(ss.str());
+        }
+    }
+
+    /***********************************
+     *  Implementation of ObjRepoView  *
+     ***********************************/
+
+    ObjRepoView::ObjRepoView(::Repo* repo) noexcept
```
Klaim, line 120: BTW is `nullptr` valid here?

libmamba/src/solv-cpp/repo.cpp
```diff
@@ -0,0 +1,278 @@
...
+    }
+
+    void ObjRepoView::read(std::filesystem::path solv_file) const
+    {
+        auto* fp = std::fopen(solv_file.string().c_str(), "rb");
+        if (!fp)
+        {
+            throw std::system_error(errno, std::generic_category());
+        }
+        const auto read_res = ::repo_add_solv(raw(), fp, 0);
```
Klaim, line 143: Same raii remark than for `write`

libmamba/src/solv-cpp/repo.cpp
```diff
@@ -0,0 +1,278 @@
...
+    }
+
+    namespace
+    {
+        auto ptr_to_strview(const char* ptr) -> std::string_view
+        {
+            static constexpr std::string_view null = "<NULL>";
+            if ((ptr == nullptr) || (ptr == null))
+            {
+                return "";
```
Klaim, line 210: Is this equivalent to `return {}`? Should it be?

libmamba/src/solv-cpp/solvable.hpp
```diff
@@ -0,0 +1,226 @@
...
+        auto raw() const -> ::Solvable*;
+
+        void set_name(std::string_view str) const;
+        void set_version(std::string_view str) const;
+
+        /** The following attributes need a call to @ref ObjRepoView::internalize. */
+        void set_build_number(std::size_t n) const;
+        void set_build_string(std::string_view bld) const;
+        void set_file_name(raw_str_view fn) const;
+        void set_file_name(const std::string& fn) const;
```
Klaim, line 77: Why not use `std::string_view` for all the const setters?
Also, why are these setters const?

libmamba/src/solv-cpp/solvable.hpp
```diff
@@ -0,0 +1,226 @@
...
+        void add_provides(const Range& deps) const;
+
+        /**
+         * Set all constraints.
+         *
+         * Setting this attribute requires a call to @ref ObjRepoView::internalize before it
+         * can be used.
+         **/
+        void set_constraints(const ObjQueue& q) const;
+        /**
```
Klaim, line 111: Add a space between functions, including their comments/documentation.

Klaim (commented):
I didn't check the tests yet.

Could you add comments/documentation for each of the types so that we have an overview of what it is about, maybe a link to the equivalent in libsolv if that makes sense, and document the functions which are not trivial to understand (like you did in `solvable.hpp`) like what does `whatproviedes` mean etc.

libmamba/src/solv-cpp/pool.cpp
AntoinePrv: This is calling a libsolv function that takes an `int` as a parameter (to represent a bool). I don't see how an enum would apply.

libmamba/src/solv-cpp/pool.cpp
AntoinePrv: It returns a temporary, vs previous function a view on an inner "attribute".

libmamba/src/solv-cpp/pool.cpp
AntoinePrv: This is not as easy here, because the return type is not the same. I believe the optional manipulation would be as verbose.

libmamba/src/solv-cpp/pool.hpp
AntoinePrv: I did, but passing the pointer to an attribute does not play well with move semantics.
I could try something with `std::function::target`, but I am also not sure it is heap allocated.

libmamba/src/solv-cpp/pool.hpp
AntoinePrv: Right, I never know which on is best since it done by value in the std lib.

libmamba/src/solv-cpp/pool.hpp
AntoinePrv: Throwing is not a problem here.

libmamba/src/solv-cpp/repo.cpp
AntoinePrv: I thought the same, and that we could even pass it a construction time when we have it available.

The list of repo is likely to be small (less than 10) and this function should not be called much so I did not bother to (for the sake of keeping a simple wrapper around a pointer).

libmamba/src/solv-cpp/repo.cpp
AntoinePrv: Hmm, I'd say not. What should it do, throw? Same remark for Solvables.

libmamba/src/solv-cpp/repo.cpp
AntoinePrv: I think it is not exactly the same in term of `.data` and. I have no preference between the two.

libmamba/src/solv-cpp/solvable.hpp
AntoinePrv: Unfortunately we cannot here because we need a `\0` terminated string...

libmamba/src/solv-cpp/pool.hpp
```diff
@@ -0,0 +1,193 @@
...
+     */
+    class ObjPool
+    {
+    public:
+
+        ObjPool();
+        ~ObjPool();
+
+        auto raw() -> ::Pool*;
+        auto raw() const -> const ::Pool*;
```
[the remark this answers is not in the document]
AntoinePrv: I'd say so, because the point is not to wrap all libsolv functions, only the most used / unsafe ones.

libmamba/src/solv-cpp/pool.cpp
```diff
@@ -0,0 +1,254 @@
...
+            /* .create= */ 0
+        );
+        return (id == 0) ? std::nullopt : std::optional(id);
+    }
+
+    auto ObjPool::add_string(std::string_view str) -> StringId
+    {
+        assert(str.size() <= std::numeric_limits<unsigned int>::max());
+        // Note: libsolv cannot report failure to allocate
+        ::Id const id = ::pool_strn2id(
```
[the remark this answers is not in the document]
AntoinePrv: Haha I found the way to trick clang-format.

libmamba/src/solv-cpp/pool.hpp
```diff
@@ -0,0 +1,193 @@
...
+        template <typename UnaryFunc>
+        void for_each_whatprovides(DependencyId dep, UnaryFunc func);
+
+        auto select_solvables(const ObjQueue& job) const -> ObjQueue;
+
+        auto add_repo(std::string_view name) -> std::pair<RepoId, ObjRepoView>;
+        auto has_repo(RepoId id) const -> bool;
+        auto get_repo(RepoId id) -> std::optional<ObjRepoView>;
+        auto get_repo(RepoId id) const -> std::optional<ObjRepoViewConst>;
+        auto n_repos() const -> std::size_t;
```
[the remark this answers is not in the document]
AntoinePrv: @Klaim suggested `repo_count`

libmamba/src/solv-cpp/pool.cpp
Klaim: It's more for making clear for the reader what is the intend. Here:
```
enum class SomeRelevantName : bool { do_not_create, create };

pool_strn2id( ... , SomeRelevantName.create); // make it clear what is the meaning of the value, instead of 1/0 or true/false.
```

libmamba/src/solv-cpp/pool.hpp
Klaim: Ok moving would imply re-setting the callback on move, so it's either that or having this allocation.
Let's keep it like that.

libmamba/src/solv-cpp/pool.hpp
Klaim: Yeah depends on a lot of things. Here I would have used forward reference as we directly invoke the function, we dont store it, we have no intention to copy it etc. But a copy is fine too, it's not a big deal.

libmamba/src/solv-cpp/pool.hpp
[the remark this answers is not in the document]
JohanMabille: ha even better.

libmamba/src/solv-cpp/repo.hpp
```diff
@@ -0,0 +1,149 @@
...
+         * @param solv_file A standard path with system encoding.
+         */
+        void write(std::filesystem::path solv_file) const;
+
+    private:
+
+        const ::Repo* m_repo = nullptr;
+    };
+
+    class ObjRepoView : public ObjRepoViewConst
```
JohanMabille, line 59: I forgot to add a comment here. I'm not in favor of mixing public inheritance and value semantic. A solution could be to have:
- `ObjRepoViewConstBase` and `ObjRepoViewBase` with constructors, destructors and assignment operators in the protected section.
- ObjRepoViewConst and ObjRepoView that prvately inherits from the previous bases, and "import" the method in thir public section.
 
This would avoid mixing / assigning objects of different types.

AntoinePrv: Is this value semantic, not reference/view semantic? A copy still points to the same data, same as in `std::span`.

The two types here serve to represent the inner `const` of an `std::span`.
We do want a `ObjRepoView` to convert to an `ObjRepoViewConst`, so that one can call`f(ObjRepoViewConst)` with a  `ObjRepoView`.

There are no invalid operations in the current design, so IMHO using base types etc would be amount to reproducing what is already here.

JohanMabille: > Is this value semantic, not reference/view semantic?

As soon as the object is manipulated with its real type, and can be copied and/or moved, it is value semantic; the opposite is entity semantic, where you manipulate objects via pointers or reference to a common base type (and ideally you never want to know the real type after the instantiation).

If you want to authorize passing `ObjRepoView` to functions accepting `ObjReoiViewConst` then indeed the current design works, but keep in mind that this authorizes the following statements:

```cpp
ObjRepoView v(...);
ObjRepoViewConst cv = v; // (potentially partial copy, here it works because only the base class has data members)

ObjRepoView* v = new ObjRepoView(....);
void free_view(ObjRepoViewConst* cv)
{
    delete cv; // The destructor of ObjRepoView is not called since it is not virtual (and it should not be).
}
```

I agree this is not problematic here because there is only one data member in the base class, the destructor of `ObjRepoView` is trivial and the hierarchy has two classes only. It's just potentially dangerous regarding future changes, and a kind of weird to read when you are used to a clean distinction between those semantics.

Here is a detailed article that states this a bit differently (it is more about virtuality and entity semantics), epecially the second question about the destructor.

But again, if we really want to be able to bind `view` objects to `const view` references, then I'm not sure we can do it differently (with conversion constructors, we could bind to const references only).

AntoinePrv:
If this https://github.com/openSUSE/libsolv/pull/526 is merged, we can use it with `std::string_view` for setting solvable attributes.

libmamba/src/solv-cpp/repo.cpp
```diff
@@ -83,34 +83,74 @@ namespace mamba::solv
...
+
+        FilePtr::FilePtr(std::FILE* ptr, std::string name)
+            : m_ptr{ ptr }
+            , m_name{ name }
+        {
+        }
+
+        auto FilePtr::open(std::filesystem::path path, const char* mode) -> FilePtr
+        {
+            std::string name = path;
```
Klaim, line 118: This doesnt work cross-platform as it's translated back to whatever is the default encoding, resulting in giberish if on Windows with utf-8 non-ascii characters in the path.
To fix it the simplest way is to use `mamba::fs::path` instead of `std::filesystem::path`

libmamba/src/solv-cpp/repo.cpp
```diff
@@ -83,34 +83,74 @@ namespace mamba::solv
...
-            std::string_view name = {};
+        public:
 
-            ~FilePtr() noexcept(false)
+            /**
+             * Open a file with C API.
+             *
+             * @param path must hae filesystem default encoding.
+             */
+            static auto open(std::filesystem::path path, const char* mode) -> FilePtr;
```
Klaim, line 95: Why not use a constructor and make this object testable instead (with `is_open() const` and `explicit operator bool() const`? Basically like in the standard. (thats also whatś in powerloader if my memory is correct)

libmamba/src/solv-cpp/repo.cpp
```diff
@@ -83,34 +83,74 @@ namespace mamba::solv
...
             {
-                const auto close_res = std::fclose(ptr);  // This flush too
-                if (close_res != 0)
-                {
-                    // TODO(C++20) fmt::format
-                    auto ss = std::stringstream();
-                    ss << "Unable to close file " << name;
-                    throw std::runtime_error(ss.str());
-                }
+                throw std::system_error(errno, std::generic_category());
```
Klaim, line 122: Why not throw filesystem error? I can see it was the previous code , but that makes little sense (I think it was like this previously because we ddidnt depend on the standard library filesystem)

Also, if this function throw, then it should be called in an opening constructor?

libmamba/src/solv-cpp/repo.cpp
```diff
@@ -83,34 +83,74 @@ namespace mamba::solv
...
             }
-        };
+            return { ptr, std::move(name) };
+        }
+
+        FilePtr::~FilePtr() noexcept(false)
```
Klaim, line 127: Allowing throwing destructor is definitely a recipe for failure and make some debugging harder.
Instead of `noexcept(false)`, I would recommend logging the error as error (if you want to provide a special error call funciton for this almost-separate-lib, do that) and just finish the destructor - no throw.

libmamba/src/solv-cpp/repo.cpp
```diff
@@ -83,34 +83,74 @@ namespace mamba::solv
 
     namespace
     {
-        struct FilePtr
+        class FilePtr
```
Klaim, line 86: At this point you can rename it, it's not just a ptr.

libmamba/src/solv-cpp/repo.cpp
AntoinePrv: > Why not throw filesystem error?

That would be better yes, but I this is the only way I found to make it work with the C error code...

libmamba/src/solv-cpp/repo.cpp
Klaim: You can use `mamba::fs::path::std_path()` to get a `std::filesystem::path`

libmamba/src/solv-cpp/repo.cpp
AntoinePrv: I would prefer throwing than logging. Since this is only used as a local variable in a function, I would argue this is OKish

Klaim: 😬

AntoinePrv: @Klaim Otherwise, I would `if(m_ptr == nullptr) std::terminate()` in `~FilePtr` to force the user to call a throwing `FilePtr::close`, a la `std::tread`.

PR #2113, opened 2022-11-15
Initial repodata.zst support

jonashaag:
@corneliusroemer :)

It works but we need to find a way of identifying channels that have a `.zst` variant.

PR #2404, opened 2023-03-23
Improvement: Support ZST in mamba and enable ZST by default

johnhany97:
## Before this PR

Only micromamba benefited from ZST support added in https://github.com/mamba-org/mamba/pull/2113

In addition, it was only enabled behind a flag, where users are required to add `repodata_use_zst: true` to their .condarc file.

## After this PR

mamba benefits from ZST support as well, and users no longer need to manually enable the flag. From now on, zst repodatas will be used where they're available.


mamba/mamba/utils.py
```diff
@@ -274,6 +277,7 @@ def get_base_url(url, name=None):
     api_ctx.retry_backoff = context.remote_backoff_factor
     api_ctx.add_pip_as_python_dependency = context.add_pip_as_python_dependency
     api_ctx.use_only_tar_bz2 = context.use_only_tar_bz2
+    api_ctx.repodata_use_zst = True
```
johnhany97, line 280: I hard-coded all around to true, and I should obviously pass it along from the context.

What's your preference on the path forwards? Should i make `repodata_use_zst` configurable using an env var (a la `MAMBA_REPODATA_USE_ZST`), read that in here and set it accordingly?

johnhany97:
I built this locally and verified that I now hit the .zst repodatas:

image

jonashaag:
It seems like zstd support is pretty stable in Micromamba, do we just want to remove the `repodata_use_zst` option? @wolfv @AntoinePrv

johnhany97:
I'm happy to repurpose this PR to do so if y'all are onboard 👍

JohanMabille:
> It seems like zstd support is pretty stable in Micromamba, do we just want to remove the `repodata_use_zst` option? @wolfv @AntoinePrv

I agree, we should use zstd everywhere and remove the option

jonashaag:
OK let's do it!

jonashaag:
Should fix this first https://github.com/mamba-org/mamba/issues/2418

corneliusroemer:
Thanks for tackling this @johnhany97 @jonashaag 

#2418 should be fixed now

johnhany97:
I've opened a separate PR to discuss the removal of the `repodata_use_zst` flag all together -> https://github.com/mamba-org/mamba/pull/2525

This PR for now assumes this flag is here to stay, and thus wires it up accordingly, if we end up merging the other PR first, then I can remove these flags in this PR.

johnhany97:
Windows CI tasks seem to be failing with:
```
CMake Error: The following variables are used in this project, but they are set to NOTFOUND.
Please set them or make sure they are set and tested correctly in the CMake files:
D:/a/mamba/mamba/libmamba/WINREG_INCLUDE_DIR
   used as include directory in directory D:/a/mamba/mamba/libmamba
```
which I believe is not relevant to this PR, but i'm not sure how to fix.

AntoinePrv:
The Windows failure is unrelated and a fix is proposed in https://github.com/mamba-org/mamba/pull/2526.

mamba/mamba/utils.py
```diff
@@ -102,10 +102,23 @@ def fixup_channel_spec(spec):
...
+                        "channel": channel,
+                        "needs_finalising": needs_finalising,
+                    },
+                )
             )
-            dlist.add(sd)
+
+    for (sd, info) in index:
+        if info["needs_finalising"]:
+            sd.finalize_checks()
```
jonashaag, line 120: I'm not too familiar with this part of the code base, can you explain what is going on here? Are there any alternatives to having this logic here in Python?

johnhany97: This is effectively just me mimicking the code that's in [`channel_loader`](https://github.com/mamba-org/mamba/blob/15f6b6e6062ce33a4562ef7003b90c42adfd329a/libmamba/src/api/channel_loader.cpp#L111-L119) which atm is duplicated between mamba and micromamba.

Let me know if you have an idea in mind that I could test out to make this nicer.

johnhany97: More concretely though, what's happening in the zst-enabled codepaths is that we do a HEAD request first to check if the zst variant exists, and if it does, we add another target that's the actual GET request for the zst variant.

jonashaag (approved):

johnhany97:
Looks like this is good to merge :D

AntoinePrv:
@johnhany97 @jonashaag this PR was reverted in #2637. I t was likely merged out of date with `main` and was the likely reason for the CI failing in `main`.

jonashaag:
Thanks for letting me know and sorry for the trouble. I should have rebased before merging because the PR was already pretty old.

Issue #2157, opened 2022-12-06
micromamba in powershell Invoke-Expression "warning" despite shell init


JannisZeller:
Hello, I ran into the following issue: I use micromamba under Windows 11 with Powershell 7.3. I did
```console
\...\micromamba.exe shell init -s powershell -p C:\mamba\base\
```
where `-p` is where I want (and have) my base environment. This resulted in the following modification to my Powershell-profile (which I know is excuted):
```console
#region mamba initialize
# !! Contents within this block are managed by 'mamba shell init' !!
$Env:MAMBA_ROOT_PREFIX = "C:\mamba\base\"
$Env:MAMBA_EXE = "C:\mamba\micromamba.exe"
(& $Env:MAMBA_EXE 'shell' 'hook' -s 'powershell' -p $Env:MAMBA_ROOT_PREFIX) | Out-String | Invoke-Expression
#endregion
```
This looks good to me. Now calling e. g. `micromamba install numpy` works as expected but with the following flaw: After `Transaction finished` is printed, there is some additional output:
![issue](https://user-images.githubusercontent.com/91051790/205934439-2b62598e-e5bc-4437-8a6b-8c34f33fb781.png)
Despite micromamba working correctly as far as I can tell, this is pretty tedious and finishing basically every install with an error message is painful. Using 
```console
C:\mamba\micromamba.exe install numpy
```
works without the additional error message. Of the "mamba-related directories" I only have `C:\mamba\base\Scripts\` in my PATH.
I have no problems in git bash. Could this be an issue with the newest Powershell version and the way `micomamba.exe shell init` works? I have little to no Powershell knowledge so please excuse me if this is senseless.

I would be grateful for any help. Keep up the good work!

moqmar:
That's coming from the "reactive environment" part in https://github.com/mamba-org/mamba/blob/main/libmamba/data/Mamba.psm1#L120.

The issue always occurs if "(micro)mamba install ..." *fails*, as nothing is returned then.

A workaround is wrapping this line in `if ($activateCommand -Ne "") { ... }`, which for microconda can be achieved by modifying the corresponding line in the powershell profile like this:

```
((& $Env:MAMBA_EXE 'shell' 'hook' -s 'powershell' -p $Env:MAMBA_ROOT_PREFIX) | Out-String).Replace('Invoke-Expression -Command $activateCommand', 'if ($activateCommand -Ne "") { Invoke-Expression -Command $activateCommand }') | Invoke-Expression
```

[...]

PR #2405, opened 2023-03-23
Work around bug in `Invoke-Mamba`

dhirschfeld:
A work-around for https://github.com/mamba-org/mamba/issues/2157#issuecomment-1481265749

---

I'm not sure this is the correct fix as I'm not sure exactly what `micromamba shell reactivate` is supposed to do. From my testing it gives no output - maybe that's another bug?

There certainly seems to be a bug in the function here, as a command such as:
```
micromamba install -n bin bat
```
...will result in
```
micromamba shell reactivate --shell powershell install -n bin bat
```
being called as the `$Args` are passed directly to the command:
https://github.com/mamba-org/mamba/blob/bc7fa860c982c796f3c35acaa011304128bb4e62/libmamba/data/Mamba.psm1#L118

The command:
```
micromamba shell reactivate --shell powershell install -n bin bat
```
will fail with exit code 114 and error message:
```
--prefix: At Most 1 required but received 2
```

As a result of the command failing `$activateCommand` will be an empty string. Since the exit code of the command isn't checked (which is a [PowerShell wart](https://github.com/mamba-org/mamba/issues/2157#issuecomment-1481204381)), the script continues on to line 120:
https://github.com/mamba-org/mamba/blob/bc7fa860c982c796f3c35acaa011304128bb4e62/libmamba/data/Mamba.psm1#L120
...which results in the error:
```
Invoke-Expression:
Line |
 120 |                      Invoke-Expression -Command $activateCommand;
     |                                                 ~~~~~~~~~~~~~~~~
     | Cannot bind argument to parameter 'Command' because it is an empty string.
```


dhirschfeld:
Resolved by #2499

PR #2409, opened 2023-03-24
No Storing Channel* and MRepo* in Solvables

AntoinePrv:
- For #2302
- Based on #2399

This removes
- the capture to `mamba::Channel*` in `mamba::MRepo`
- the capture of `mamba::MRepo*` in `::Repo::appdata`
- break dependency between `mamba::MRepo` and `mamba::Channel`


libmamba/src/core/repo.cpp
```diff
@@ -101,11 +94,23 @@ namespace mamba
 
     void MRepo::init(MPool& pool)
     {
-        m_repo->appdata = this;
         m_real_repo_key = pool_str2id(pool, "solvable:real_repo_url", 1);
+        m_mrepo_key = pool_str2id(pool, "solvable:mrepo_url", 1);
```
JohanMabille, line 98: What is the difference between `repo` and `real_repo` ?

libmamba/include/mamba/core/transaction.hpp
```diff
@@ -165,6 +165,8 @@ namespace mamba
 
         History::UserRequest m_history_entry;
         Transaction* m_transaction;
+        ::Id m_real_repo_key = 0;  // Must match that in ``MRepo``
+        ::Id m_mrepo_key = 0;      // Must match that in ``MRepo``
```
JohanMabille, line 169: I guess this leaking of libsolv details in the transaction class is temporary and will be replaced with PackageInfo list on the long run? In the meantime, they (and those in MRepo) should call the same initialization function to ensure their "synchronization"

libmamba/src/core/repo.cpp
AntoinePrv: I'd like to know too! This is a mechanical operation that stores the value of `MRepo::url` for retrival.
I tried to merge `real_repo_url` and `mrepo_url` but it was significantly more tricky than the scope of this PR.

libmamba/include/mamba/core/transaction.hpp
AntoinePrv: This constraint was already here, the PR just made it more obvious.
At the libsolv level (which will be private), #2401 will introduce clean getter/setter in the bindings so that we don't need to match the keys manually.
At the user level, this will be invisible. The user will only deal in PackageInfo style of data.

libmamba/src/core/repo.cpp
AntoinePrv: There is also `m_metadata.url`/"mamba:url" in the mix

JohanMabille (approved):

PR #2416, opened 2023-03-28
Refactor test_create, test_proxy, and test_env for test isolation

AntoinePrv:
Big PR, sorry. Basically it uses the new fixture to create temporary root_prefix and home directory inside a /tmp folder.

- Deactivate PyCQA line-length warnings
- Proper os.environ restore in pytest
- Add --mamba-pkgs-dir for cache reuse in Pytest
- Set pytest tmp retention policy
- Allow Path in test helpers
- Refactor test_create for isolation
- Improve pytest fixtures
- Refactor test_env.py for test isolation
- Refactor test_proxy for isolating tests

### DIFF
--- a/micromamba/tests/test_env.py
+++ b/micromamba/tests/test_env.py
@@ -1,152 +1,103 @@
 import os
-import platform
 import shutil
 from pathlib import Path
 
 import pytest
 import yaml
 
-from .helpers import *
+from . import helpers
+
+__this_dir__ = Path(__file__).parent.resolve()
 
 
-class TestEnv:
+def test_env_list(tmp_home, tmp_root_prefix, tmp_empty_env):
+    env_json = helpers.run_env("list", "--json")
 
-    current_root_prefix = os.environ["MAMBA_ROOT_PREFIX"]
-    current_prefix = os.environ["CONDA_PREFIX"]
-    cache = os.path.join(current_root_prefix, "pkgs")
+    assert "envs" in env_json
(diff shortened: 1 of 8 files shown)


micromamba/tests/test_env.py
```diff
@@ -1,152 +1,103 @@
...
+
+    all_lines = res.splitlines()
+    for line in all_lines:
+        if "*" in line:
+            active_env_l = line
+    assert str(tmp_prefix) in active_env_l
+
+
+def test_register_new_env(tmp_home, tmp_root_prefix):
+    helpers.create("-n", "env2", "--json", no_dry_run=True)
```
Hind-M, line 47: Do we remove the created test environments after the tests are done? It was handled by the `teardown_class` before... (This applies wherever `create` is used)

AntoinePrv: This is handled by the `tmp_root_prefix` fixture that sets a temporary root prefix for all environments.
It is created under temporary storage, independent for each test, and safely cleaned up afterwards.

Hind-M: Hmm yes, but it's not the case for the explicitly created envs `env2`, `env3`, etc.  Is it?

Hind-M: Oh ok I see, you mean that deleting the `root_prefix` deletes implicitely everything... ok!

Hind-M (approved):

Issue #2153, opened 2022-12-04
.\micromamba.exe --help has no output


eliannelavoie:
Following the instructions here: https://mamba.readthedocs.io/en/latest/installation.html#windows

I've ran those commands:
```
Invoke-Webrequest -URI https://micro.mamba.pm/api/micromamba/win-64/latest -OutFile micromamba.tar.bz2
C:\PROGRA~1\7-Zip\7z.exe x micromamba.tar.bz2 -aoa
C:\PROGRA~1\7-Zip\7z.exe x micromamba.tar -ttar -aoa -r Library\bin\micromamba.exe

MOVE -Force Library\bin\micromamba.exe micromamba.exe
.\micromamba.exe --help
```

The download and extraction worked fine, but the executable itself doesn't seem to work.
![image](https://user-images.githubusercontent.com/13744678/205522035-3cb097e8-56a0-4e4c-95c6-16b884aeced0.png)

No commands seem to work.

eliannelavoie:
I did find relevant errors in the Event Viewer, under "Windows Logs > Application" that mention a "Problem signature".

```
Log Name:      Application
Source:        Windows Error Reporting
Date:          2022-12-04 6:36:19 PM
Event ID:      1001
Task Category: None
Level:         Information
Keywords:      Classic
User:          N/A
Computer:      ncc-2000
Description:
Fault bucket 2104183035507384322, type 5
Event Name: BEX64
Response: Not available
Cab Id: 0

Problem signature:
P1: micromamba.exe
P2: 0.0.0.0
P3: 6385f292
P4: micromamba.exe
P5: 0.0.0.0
P6: 6385f292
P7: 00000000005d7559
P8: c0000409
P9: 0000000000000007
P10: 

Attached files:
\\?\C:\ProgramData\Microsoft\Windows\WER\Temp\WERC3B4.tmp.mdmp
\\?\C:\ProgramData\Microsoft\Windows\WER\Temp\WERC3E4.tmp.WERInternalMetadata.xml
\\?\C:\ProgramData\Microsoft\Windows\WER\Temp\WERC3F5.tmp.xml
\\?\C:\ProgramData\Microsoft\Windows\WER\Temp\WERC417.tmp.csv
\\?\C:\ProgramData\Microsoft\Windows\WER\Temp\WERC437.tmp.txt

These files may be available here:
[...]

PR #2417, opened 2023-03-28
Windows: Fixed environment variables not read as unicode.

Klaim:
The code acquiring environment variables on Windows was doing so using the non-unicode API which lead to some environment variables containing unicode path (or other unicode values) leading to random crashes.
This version uses a simpler API call which is unicode-enabled and should fix https://github.com/mamba-org/mamba/issues/2153

### DIFF
--- a/libmamba/include/mamba/core/util_os.hpp
+++ b/libmamba/include/mamba/core/util_os.hpp
@@ -13,8 +13,8 @@
 #include "mamba/core/fsutil.hpp"
 
 #ifdef _WIN32
-#   define WIN32_LEAN_AND_MEAN
-#   include <Windows.h>
+#define WIN32_LEAN_AND_MEAN
+#include <Windows.h>
 #endif
 
 namespace mamba
--- a/libmamba/include/mamba/core/util_os.hpp
+++ b/libmamba/include/mamba/core/util_os.hpp
@@ -12,13 +12,13 @@
 
 #include "mamba/core/fsutil.hpp"
 
-#ifdef _WIN32
-#define WIN32_LEAN_AND_MEAN
-#include <Windows.h>
-#endif
-
 namespace mamba
 {
+#ifdef _WIN32
+    // Intention is to avoid including `Windows.h`, while still using the basic Windows API types.
+    using DWORD = unsigned long;
+#endif
+
     bool is_admin();
     fs::u8path get_self_exe_path();
 
--- a/libmamba/include/mamba/core/util_os.hpp
+++ b/libmamba/include/mamba/core/util_os.hpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2019, QuantStack and Mamba Contributors
+// Copyright (c) 2019-2023, QuantStack and Mamba Contributors
 //
 // Distributed under the terms of the BSD 3-Clause License.
 //
--- a/libmamba/include/mamba/core/util_os.hpp
+++ b/libmamba/include/mamba/core/util_os.hpp
@@ -12,6 +12,11 @@
 
 #include "mamba/core/fsutil.hpp"
 
+#ifdef _WIN32
+#   define WIN32_LEAN_AND_MEAN
+#   include <Windows.h>
+#endif
+
 namespace mamba
 {
     bool is_admin();
--- a/libmamba/include/mamba/core/util_os.hpp
+++ b/libmamba/include/mamba/core/util_os.hpp
@@ -40,9 +40,10 @@ namespace mamba
     void reset_console();
 
 #ifdef _WIN32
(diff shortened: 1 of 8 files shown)


Klaim:
For info I'm trying to rewrite the solution with the safe functions and to handle the setenv because otherwse tests cannot pass.

libmamba/src/core/environment.cpp
```diff
@@ -18,52 +21,86 @@ namespace mamba
...
+                                                              // functions are not thread-safe, this
+                                                              // is to prevent related issues.
+
+            const auto on_failed = [&](errno_t error_code)
+            {
+                LOG_ERROR << fmt::format(
+                    "Failed to acquire environment variable '{}' : errcode = {}",
+                    key,
+                    error_code
+                );
```
AntoinePrv, line 37: Nit: One can also do (with the right header)
```suggestion
                fmt::format_to(LOG_ERROR,
                    "Failed to acquire environment variable '{}' : errcode = {}",
                    key,
                    error_code
                );
```

AntoinePrv (commented):
Thanks :tada: 

Can we take this opportunity to add some libmamba tests? Since we have `env::get` and `env::set` we can at least test them together.

Testing `env::get` alone would require setting an env var from outside the test exe. Not perfect, but maybe necessary?

libmamba/src/core/environment.cpp
Klaim: BTW I know this is possible but I dont know if that changes something performance-wise for fmt? Do you know?

Klaim:
I will also add a basic test for the unicode functions.

libmamba/src/core/environment.cpp
Klaim: Actually I tried that syntax and it doenst seem to compile, probably lacks some compatibility layer to add. I'll keep it as is for now.

Klaim:
@AntoinePrv I added basic tests using `mamba::env::get` to get environment variables that are always available on windows, probably will also fail if the user running the test is using a unicode path (we migth want to do that on the CI too...).
If you want to add another test for linux be my guest, I cant check right now such code as I'm stuck on windows until next week.

AntoinePrv (approved):
Thanks!

libmamba/include/mamba/core/util_os.hpp
```diff
@@ -12,6 +12,11 @@
 
 #include "mamba/core/fsutil.hpp"
 
+#ifdef _WIN32
+#define WIN32_LEAN_AND_MEAN
+#include <Windows.h>
+#endif
```
JohanMabille, line 18: Is it be possible to avoid including <Windows.h> in a "public" header? I'm a bit ocncerned with all the Windows headers this pulls in.

libmamba/tests/src/core/test_util_os.cpp
```diff
@@ -0,0 +1,34 @@
+﻿// Copyright (c) 2019, QuantStack and Mamba Contributors
```
Hind-M, line 1: Same here.
And btw, should we put this test file in `tests/src/util` folder instead of `core`? But I guess `test_system_env` could be kept in core?

AntoinePrv: I would move the tests file at the same time as we are able to move the header under similar `mamba/util` `mamba::util` scope.

libmamba/include/mamba/core/util_os.hpp
Klaim: No, I had to add it because of the usage of `DWORD`. Note that the reason it was not visible before is because there is no test before this PR including this header alone. This means that the other headers already include `windows.h`.
Note that `WIN32_LEAN_AND_MEAN` should drastically reduce the issue, but the only way to remove that would be to not use `DWORD` OR to use our definition of `DWORD`, which is a possibility, like they did in Boost.WinAPI for example.

libmamba/tests/src/core/test_util_os.cpp
Klaim: Yeah I tried to match where the current header is, not all utils are moved in `mamba/util/` at the moment.

libmamba/include/mamba/core/util_os.hpp
Klaim: Used an alias instead.

libmamba/tests/src/core/test_system_env.cpp
```diff
@@ -0,0 +1,68 @@
+﻿// Copyright (c) 2019, QuantStack and Mamba Contributors
```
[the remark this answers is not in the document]
Klaim: should be fixed now

Hind-M (approved):

Issue #2356, opened 2023-03-07
Isolate libcurl


JohanMabille:
libcurl should be isolated in a wrapper and not exposed to the rest of mamba. We should reuse code from powerloader for that (such as `https://github.com/mamba-org/powerloader/blob/main/include/powerloader/curl.hpp` and  `https://github.com/mamba-org/powerloader/blob/main/include/powerloader/url.hpp`.

The wrapper itself should be an implementation detail of libmamba and not exposed in public APIs.

PR #2421, opened 2023-03-29
Wrap more libcurl calls

Hind-M:
Part of #2356 
This PR includes:
- Moving `curl_initialized` from context to `DownloadTarget` class member
- Moving `hide_secrets` function from `url.{hpp, cpp}` to `util.{hpp, cpp}`
- Adding a getter for `m_errorbuffer`
- Defining `set_curl_handle` for more curl calls wrapping

### DIFF
--- a/libmamba/src/core/curl.cpp
+++ b/libmamba/src/core/curl.cpp
@@ -4,12 +4,108 @@
 //
 // The full license is in the file LICENSE, distributed with this software.
 
+// TODO remove all these includes later?
 #include <spdlog/spdlog.h>
+
+#include "mamba/core/mamba_fs.hpp"  // for fs::exists
+#include "mamba/core/util.hpp"      // for hide_secrets
 
 #include "curl.hpp"
 
 namespace mamba
 {
+    namespace curl
+    {
+        void set_curl_handle(
+            CURL* handle,
+            const std::string& url,
+            const bool set_low_speed_opt,
+            const long& connect_timeout_secs,
+            const bool ssl_no_revoke,
+            const std::optional<std::string> proxy,
+            const std::string& ssl_verify
+        )
+        {
+            curl_easy_setopt(handle, CURLOPT_URL, url.c_str());
+            curl_easy_setopt(handle, CURLOPT_NETRC, CURL_NETRC_OPTIONAL);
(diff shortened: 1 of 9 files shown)


libmamba/src/core/fetch.cpp
```diff
@@ -159,12 +95,12 @@ namespace mamba
     {
         auto& ctx = Context::instance();
 
-        if (!ctx.curl_initialized)
+        if (!m_curl_initialized)
```
JohanMabille, line 98: I think there is a risk that we do many times the same thing here since `ctx.curl_initialized` is a global, shared variable, while `m_curl_initialized` is a data member specific to a DownloadTarget data member.

This might not be a big deal here since we just set another variable in the context, but it might become later if we add stuff in this function.

What is the motivation for removing this variable from the context? Would it make sense to have it as part of a "DownloaderOption" subobject in the Context? (that could be done in another PR, but the current variable would remain in the context in this PR).

libmamba/src/core/curl.cpp
```diff
@@ -4,12 +4,108 @@
...
+#include "mamba/core/mamba_fs.hpp"  // for fs::exists
+#include "mamba/core/util.hpp"      // for hide_secrets
+
 #include "curl.hpp"
 
 namespace mamba
 {
+    namespace curl
+    {
+        void set_curl_handle(
```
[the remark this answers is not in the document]
Hind-M: Totally agreed, I left it out of the `CURLHandle` class for now since it's used in `DownloadTarget::resource_exists()` and the given `handle` is not the one of the `CURLHandle` class.

libmamba/src/core/fetch.cpp
Hind-M: Well I actually found out that it's not used anywhere but here in the `fetch`... so no reason to keep it in the context! Do you see any, or think it would be used somewhere else?

JohanMabille: I think it should live with the ssl_verify variable and maybe move to CURLSetup on the long run. This should be set once and for good. Keeping it in the context is ok for now.

JohanMabille (approved):

PR #2432, opened 2023-03-31
Context structuring

Hind-M:
This is a first pass on `Context` structuring.
I'm holding back any additional changes on this PR since it already involves many files (even if the changes are quite straightforward)...
If this is the way to go, work can be resumed on the `Context` entirely in another PR.


libmamba/include/mamba/core/context.hpp
```diff
@@ -118,11 +118,59 @@ namespace mamba
...
+            std::string user_agent{ "mamba/" LIBMAMBA_VERSION_STRING };
+
+            int connect_timeout_secs{ 10 };
+            // int read_timeout_secs { 60 };
+            int retry_timeout{ 2 };  // seconds
+            int retry_backoff{ 3 };  // retry_timeout * retry_backoff
+            int max_retries{ 3 };    // max number of retries
+        };
+
+        struct OutputInfo  // Configurable, TODO change name of struct?
```
Hind-M, line 140: Should we use names for structs which would reflect configurability like `InternalDesignInfo` and `ConfigSrcInfo`?

JohanMabille: I don't think so, Configurability is something that can evolve in the time, and it will be hard to keep things synchronized. Besides, once the configuration has been load, we don't really care where the values come from.

JohanMabille (approved):

libmamba/include/mamba/core/context.hpp
```diff
@@ -118,11 +118,58 @@ namespace mamba
...
+            log_level logging_level{ log_level::warn };
+
+            bool json{ false };
+            bool quiet{ false };
+
+            std::string log_pattern{ "%^%-9!l%-8n%$ %v" };
+            std::size_t log_backtrace{ 0 };
+        };
+
+        struct DesignInfo
```
Klaim, line 151: Did you meant "graphics" here? Design isn´t necessarilly about graphics.

Hind-M:
> While I generally approve, I feel like the data being grouped are not "info" but "config(uration)". It's the configuration driving the library/program, it's not just various information or status. If I am correct, consider renaming? Hopefully it's not too hard to do using refactoring tools.

Yeah there is a lot of it which is configuration, I used `Info` because I was planning to separate configurable and internal/non configurable parameters but at the end we decided to leave it like this (mixing both in the same struct and grouping only by purpose/use instead).  If we rename `Info` to `Config` I am afraid that it could be confusing since everything is not configurable... But at the same time, using `Info` could be not that relevant as you pointed out. So there is a choice to make here...

Klaim:
Hmmm well maybe it's not too important and keeping it vague might be better for now... I'll let you decide.

Now just "design" seems not clear to me, so maybe that needs to change - otherwise I'm approving this PR 👍🏽

libmamba/include/mamba/core/context.hpp
Klaim: Or maybe "display"?

Hind-M: Graphics or Display then? :)

Klaim: Both are fine to me, also this will change once we separate the logic from the "view" (which is another possible name XD)
I'll let you chose. My initial comment is because "design" can mean "concept" but the intention, if I understand correctly, is more like "the visible thing we will show", so any word more like "view", "graphcs" et.c works for me.

PR #2434, opened 2023-03-31
CI test should not need static build

AntoinePrv:
Static builds often fail for packaging reasons (or even downloading reasons).
This should not prevent tests from running
They provide feedback for dev working on unrelated features.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -52,7 +52,6 @@ jobs:
           rm -rf ~/micromamba*/envs/*/envs
 
   libmamba_cpp_tests:
-    needs: [libmamba_static]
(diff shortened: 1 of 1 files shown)


JohanMabille:
This dependency ensures the static library and the dynamic one can live together, without having to rebiuld the static library. I think we should fix the static build instead.

AntoinePrv:
@JohanMabille That's only the case for Windows. Otherwise there are no artifacts being shared for Unix.

Then I think we should have a job that builds both static and shared together and also does a CMake install+ `find_package` sanity check.

JohanMabille:
I agree that option would be better

PR #2436, opened 2023-03-31
Migrated to doctest

JohanMabille:


AntoinePrv (approved):
Thanks for doing this! I'm looking forward to nicer testing!
I mostly checked the test I wrote and everything looks good.

PR #2440, opened 2023-04-04
Remove const ref to string_view in codebase

Hind-M:


Hind-M:
This makes me wondering about the `const`:
https://github.com/mamba-org/mamba/blob/main/libmamba/src/core/util_os.cpp#L551
Should we remove it everywhere or leave it?

AntoinePrv (approved):

AntoinePrv:
This `const` is typically not added, but is correct.

For view types, one needs to think like pointers.
- `std::string_view` is like `char const*` -> the inner data is const but we can make the pointer point to something else. 
- `const std::string_view` is like `char const* const` -> the inner data is const, and we cannot rebind the pointer.

Issue #2418, opened 2023-03-28
Invalid cross-device link error with `repodata_use_zst`


hadim:
### Troubleshooting docs

- [X] My problem is not solved in the Troubleshooting docs

### Search tried in issue tracker

Invalid cross-device link 

### Latest version of Mamba

- [X] My problem is not solved with the latest version

### Tried in Conda?

Didn't try

### Describe your issue

When trying to install a package while `xxxxx=true`, it fails with the below error:

```
critical libmamba Multiple errors occured:
    Could not move repodata file from '/tmp/mambafylQyHBbIXP' to '/home/hadim/local/micromamba/pkgs/cache/d924db24.json': Invalid cross-device link
```

As a side note, `/tmp` and `/home/hadim` are on different partitions on my system.

See also https://github.com/mamba-org/mamba/issues/2239

### mamba info / micromamba info

```shell
environment : base (active)
           env location : /home/hadim/local/micromamba
      user config files : /home/hadim/.mambarc
 populated config files : /home/hadim/.mambarc
                          /home/hadim/.condarc
       libmamba version : 1.4.1
     micromamba version : 1.4.1
           curl version : libcurl/7.88.1 OpenSSL/3.1.0 zlib/1.2.13 zstd/1.5.2 libssh2/1.10.0 nghttp2/1.52.0
     libarchive version : libarchive 3.6.2 zlib/1.2.13 bz2lib/1.0.8 libzstd/1.5.2
       virtual packages : __unix=0=0
                          __linux=5.19.0=0
                          __glibc=2.36=0
                          __archspec=1=x86_64
                          __cuda=12.0=0
               channels : https://conda.anaconda.org/t/*****/conda-forge/linux-64
                          https://conda.anaconda.org/t/*****/conda-forge/noarch
       base environment : /home/hadim/local/micromamba
               platform : linux-64
```

hadim:
What's weird is that it sometimes works and sometime does not. Could that be because not all the `fs::rename` has been replaced?

See https://github.com/search?q=repo%3Amamba-org%2Fmamba%20%22fs%3A%3Arename%22&type=code

jonashaag:
Is there anything else that looks suspicious in the logs?

I wonder if there is a way for `fs::rename` to throw an exception instead of setting `ec`.

hadim:
It did happen again during a `micromamba install`. Full trace:

```
debug    libmamba Loading configuration
trace    libmamba Compute configurable 'create_base'
[...]

Issue #2239, opened 2023-01-20
repodata_use_zst causes Invalid cross-device link error


nathanweeks:
### Troubleshooting docs

- [X] My problem is not solved in the Troubleshooting docs

### Search tried in issue tracker

repodata_use_zst

### Latest version of Mamba

- [X] My problem is not solved with the latest version

### Tried in Conda?

Not applicable

### Describe your issue

On a host where /tmp and $HOME are on different partitions, `micromamba create -n <env>...` results in an `Invalid cross-device link error`.

### mamba info / micromamba info

```shell
environment : None (not found)
           env location : -
      user config files : /home/user/.mambarc
 populated config files : /home/user/.condarc
       libmamba version : 1.2.0
     micromamba version : 1.2.0
           curl version : libcurl/7.87.0 OpenSSL/1.1.1s zlib/1.2.13 libssh2/1.10.0 nghttp2/1.47.0
     libarchive version : libarchive 3.6.2 zlib/1.2.13 bz2lib/1.0.8 libzstd/1.5.2
       virtual packages : __unix=0=0
                          __linux=3.10.0=0
                          __glibc=2.17=0
                          __archspec=1=x86_64
               channels : https://conda.anaconda.org/conda-forge/linux-64
                          https://conda.anaconda.org/conda-forge/noarch
                          https://conda.anaconda.org/bioconda/linux-64
                          https://conda.anaconda.org/bioconda/noarch
                          https://repo.anaconda.com/pkgs/main/linux-64
                          https://repo.anaconda.com/pkgs/main/noarch
                          https://repo.anaconda.com/pkgs/r/linux-64
                          https://repo.anaconda.com/pkgs/r/noarch
       base environment : /home/user/micromamba
               platform : linux-64
```

### Logs

```shell
$ micromamba create -v -n samtools samtools

                                           __
          __  ______ ___  ____ _____ ___  / /_  ____ _
         / / / / __ `__ \/ __ `/ __ `__ \/ __ \/ __ `/
        / /_/ / / / / / / /_/ / / / / / / /_/ / /_/ /
       / .___/_/ /_/ /_/\__,_/_/ /_/ /_/_.___/\__,_/
      /_/

warning  libmamba 'root_prefix' set with default value: /home/user/micromamba
info     libmamba You have not set the 'root_prefix' environment variable.
    To permanently modify the root prefix location, either:
    - set the 'MAMBA_ROOT_PREFIX' environment variable
[...]

PR #2447, opened 2023-04-07
Avoid using /tmp

johnhany97:
Attempts to fix https://github.com/mamba-org/mamba/issues/2418

I realised that we already attempt to rename, and if that failed, we attempt to copy then delete the file. However, this seems to still cause `Invalid cross-device link error`.

My proposal is to just avoid storing the temporary file in `/tmp`, and instead use the same cache dir we already use for writing the .json files.

### DIFF
--- a/libmamba/src/core/subdirdata.cpp
+++ b/libmamba/src/core/subdirdata.cpp
@@ -905,7 +905,9 @@ namespace mamba
     void MSubdirData::create_target()
     {
         auto& ctx = Context::instance();
-        m_temp_file = std::make_unique<TemporaryFile>();
+        fs::u8path writable_cache_dir = create_cache_dir(m_writable_pkgs_dir);
+        auto lock = LockFile(writable_cache_dir);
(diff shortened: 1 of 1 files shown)


johnhany97:
Given the nondeterministic nature of the issue at hand, i'm likely to try and get a patched version of mamba with this change and run it for a bit on our internal development infra to see if the issue still pops up or not. I won't be able to do so for at least a week though.

jonashaag:
Not sure if this fixes anything, see discussion here: https://github.com/mamba-org/mamba/issues/2239#issuecomment-1400016398

johnhany97:
I guess the alternative then would be to read the file and write it again ourselves to the new location? Happy to repurpose the code to do so.

johnhany97:
I guess the alternative then would be to read the file and write it again ourselves to the new location? Happy to repurpose the code to do so.

johnhany97:
I guess the alternative then would be to read the file and write it again ourselves to the new location? Happy to repurpose the code to do so.

wolfv (approved):

PR #2453, opened 2023-04-12
Use libsolv wrappers in MPool and MRepo

AntoinePrv:
- For #2302 
- Based on  #2401

### DIFF
--- a/libmamba/include/mamba/core/repo.hpp
+++ b/libmamba/include/mamba/core/repo.hpp
@@ -13,6 +13,7 @@
 #include <utility>
 #include <vector>
 
+#include <nlohmann/json_fwd.hpp>
 #include <solv/pooltypes.h>
 
 #include "mamba_fs.hpp"
@@ -35,17 +36,17 @@ namespace mamba
      */
     struct RepoMetadata
     {
-        std::string url;
+        std::string url = {};
+        std::string etag = {};
+        std::string mod = {};
         bool pip_added = false;
-        std::string etag;
-        std::string mod;
     };
 
-    inline bool operator==(const RepoMetadata& lhs, const RepoMetadata& rhs)
-    {
-        return lhs.url == rhs.url && lhs.pip_added == rhs.pip_added && lhs.etag == rhs.etag
-               && lhs.mod == rhs.mod;
-    }
+    auto operator==(const RepoMetadata& lhs, const RepoMetadata& rhs) -> bool;
+    auto operator!=(const RepoMetadata& lhs, const RepoMetadata& rhs) -> bool;
+
+    void to_json(nlohmann::json& j, const RepoMetadata& m);
+    void from_json(const nlohmann::json& j, RepoMetadata& p);
 
     /**
      * A wrapper class of libsolv Repo.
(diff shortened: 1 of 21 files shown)


libmamba/include/mamba/core/repo.hpp
```diff
@@ -56,91 +62,52 @@ namespace mamba
...
-         * Static constructor.
-         * @param pool ``libsolv`` pool wrapper
-         * @param name Name
-         * @param uris Matchspecs pointing to unique resources (URL or files)
-         */
-        static MRepo&
-        create(MPool& pool, const std::string& name, const std::vector<PackageInfo>& uris);
 
-    private:
+        struct [[deprecated]] PyExtraPkgInfo
```
Klaim, line 80: I'm not sure if `[[deprecated]]` is useful here as we'll change the major version?

libmamba/src/core/query.cpp
```diff
@@ -411,6 +411,16 @@ namespace mamba
         return table(out, { "Name", "Version", "Build", "Channel" });
     }
 
+    namespace
+    {
+        /** Remove potential subdir from channel name (not url!). */
+        auto cut_subdir(std::string_view str) -> std::string
```
Klaim, line 417: I think we have similar code somewhere else in mamba, Iĺl try to find it.

libmamba/src/core/repo.cpp
```diff
@@ -16,145 +21,136 @@ extern "C"  // Incomplete header
...
 #include "mamba/core/pool.hpp"
 #include "mamba/core/prefix_data.hpp"
 #include "mamba/core/repo.hpp"
-#include "mamba/core/util_string.hpp"
+#include "solv-cpp/pool.hpp"
+#include "solv-cpp/repo.hpp"
 
 
-#define MAMBA_TOOL_VERSION "1.1"
+#define MAMBA_TOOL_VERSION "1.3"
```
Klaim, line 34: What is this version? libmamba is version 1.4.x at this point

libmamba/src/solv-cpp/repo.cpp
```diff
@@ -229,6 +235,27 @@ namespace mamba::solv
...
+            {
+                ss << ", error was: " << str;
+            }
+            throw std::runtime_error(ss.str());
+        }
+    }
+
+    void ObjRepoView::legacy_read_conda_repodata(const fs::u8path& repodata_file, int flags) const
+    {
+        auto file = CFile::open(repodata_file, "rb");
```
Klaim, line 248: Shouldnt `file` be tested here?

Klaim (approved):
Globally LGTM 👍🏽

libmamba/src/solv-cpp/repo.cpp
AntoinePrv: This is done in the `CFile::open` function.

libmamba/src/core/repo.cpp
AntoinePrv: This is the version of the binary serialization of libsolv data. We slap our own version into it to update when we changes the internal libsolv data.

Probably not the best way to do this, because you need to deserialize it to be able to read it. We should think of something better...

libmamba/include/mamba/core/repo.hpp
AntoinePrv: It's mostly for internal use, we'll know to remove it in the major release and it is to indicate that it is not needed internally anymore.

AntoinePrv:
Merging because this is a blocker and many reviewers are unavailable.

PR #2454, opened 2023-04-12
Remove dead code / attribute

AntoinePrv:
The keys `"solvable:extra_xxx"` are not set in any place, and are never used in libsolv.
Therefore that code is dead and `PackageInfo::extra_metadata` is unused.


wolfv:
The idea of these was that we could bit-for-bit reconstruct the entry from repodata.json and compare the signature. Not all entries of repodata.json are necessarily parsed into libsolv. I think Adrien used to have a patched `libsolv` version where these might have been set (https://github.com/openSUSE/libsolv/pull/469)

JohanMabille (approved):

PR #2455, opened 2023-04-12
No singleton: ChannelContext, ChannelBuilder and channel cache

Klaim:
Both `ChannelBuilder` and channel cache are now fusioned into `ChannelContext` which is now responsible for storing and providing `Channel` instances, and is not a singleton anymore.


Klaim:
Cancelling the build because this is definitely not ready yet.
I'm having trouble making sens of the usage of `make_channel(...)` as it seems to often be used not really to create channel but to obtain related names/values. As I'm trying to not break the meaning of the code, for now I'm avoiding blindly adding parameters where it might not really make sense.
So this is a work-in-progress.

AntoinePrv (approved):
Well done :tada: I could not really follow everywhere, but I have not seen anything shocking.
Let's merge this before it starts conflicting with everything.

PR #2459, opened 2023-04-13
Wapped curl multi handle

JohanMabille:


libmamba/src/core/curl.cpp
```diff
@@ -307,4 +308,180 @@ namespace mamba
...
+    bool operator!=(const CURLHandle& lhs, const CURLHandle& rhs)
+    {
+        return !(lhs == rhs);
+    }
+
+    /*****************
+     * CURLReference *
+     *****************/
+
+    CURLReference::CURLReference(CURL* handle)
```
Klaim, line 330: Note that in other wrappers we have been using `.raw()`

libmamba/src/core/curl.cpp
```diff
@@ -307,4 +308,180 @@ namespace mamba
...
+        : p_handle(rhs.p_handle)
+        , m_max_parallel_downloads(rhs.m_max_parallel_downloads)
+    {
+        rhs.p_handle = nullptr;
+        rhs.m_max_parallel_downloads = 0u;
+    }
+
+    CURLMultiHandle& CURLMultiHandle::operator=(CURLMultiHandle&& rhs)
+    {
+        std::swap(p_handle, rhs.p_handle);
```
Klaim, line 409: Shouldnt we set `rhs` in a non-owning state? Just swapping here will make it own what we have, it might not be immediately destroyed.

libmamba/src/core/curl.cpp
```diff
@@ -307,4 +308,180 @@ namespace mamba
...
+     * CURLMultiHandle *
+     *******************/
+
+    CURLMultiHandle::CURLMultiHandle(std::size_t max_parallel_downloads)
+        : p_handle(curl_multi_init())
+        , m_max_parallel_downloads(max_parallel_downloads)
+    {
+        if (p_handle == nullptr)
+        {
+            throw curl_error("Could not initialize CURL multi handle");
```
Klaim, line 380: Can't we use `curl_multi_strerror` to get the error message here?

libmamba/src/core/curl.cpp
```diff
@@ -307,4 +308,180 @@ namespace mamba
...
+            int numfds;
+            code = curl_multi_wait(p_handle, NULL, 0, static_cast<int>(curl_timeout), &numfds);
+            if (code != CURLM_OK)
+            {
+                throw std::runtime_error(curl_multi_strerror(code));
+            }
+
+            if (!numfds)
+            {
+                std::this_thread::sleep_for(std::chrono::milliseconds(100));
```
Klaim, line 482: I dont get what this sleep is for? The duration is arbitrary?

JohanMabille: Oh good catch, it should be the timeout computed, I will fix this.

libmamba/src/core/curl.cpp
JohanMabille: It's ok to have rhs freeing the curl handle in its destructor later.

libmamba/src/core/curl.cpp
JohanMabille: indeed, good catch

libmamba/src/core/curl.cpp
Klaim: Oups, this should be pointing at the `unwrap` line,  sorry

libmamba/src/core/curl.cpp
JohanMabille: I just ported the code that is currently in fetch, I think it was a copy paste from [there](https://curl.se/libcurl/c/curl_multi_wait.html)

Klaim: Yeah but this code is in a loop, while this function is not. If `CURLMultiHandle::wait` is called in a loop, shouldnt that last loop-sleep be in that loop too, and not in `CURLMultiHandle::wait`?

libmamba/src/core/curl.cpp
JohanMabille: Yes, the object it is swapped with is in a valid state as long as it has been constructed.

libmamba/src/core/curl.cpp
JohanMabille: Not sure to get you, the sleep is tighly coupled to the `curl_multi_wait` function since it depends on the numdfs returned. `wait` is meant to be called in a loop (otherwise you just wait for nothing), this `sleep` is just an implementation detail to distinguish timeouts from not having any filedescriptor left.

Klaim (approved):
LGTM!

PR #2467, opened 2023-04-17
Add topological sort

AntoinePrv:
- Fix `is_reachable` signature
- Make `depth_first_search` a free function to mirror networkx, rename it `dfs_`
- Add depth first search on all nodes
- Add preorder and postorder dfs aliases
- Improve tests
- Add topological sort

### DIFF
--- a/libmamba/include/mamba/util/graph.hpp
+++ b/libmamba/include/mamba/util/graph.hpp
@@ -8,6 +8,7 @@
 #define MAMBA_UTIL_GRAPH_HPP
 
 #include <algorithm>
+#include <cassert>
 #include <functional>
 #include <iterator>
 #include <map>
@@ -539,6 +540,9 @@ namespace mamba::util
             const typename Graph::adjacency_list& adjacency
         )
         {
+            assert(status.size() == graph.successors().size());
+            assert(adjacency.size() == graph.successors().size());
+            assert(start <= status.size());
             status[start] = Visited::ongoing;
             visitor.start_node(start, graph);
             for (auto child : adjacency[start])
--- a/libmamba/include/mamba/util/graph.hpp
+++ b/libmamba/include/mamba/util/graph.hpp
@@ -542,7 +542,7 @@ namespace mamba::util
         {
             assert(status.size() == graph.successors().size());
             assert(adjacency.size() == graph.successors().size());
-            assert(start <= status.size());
+            assert(start < status.size());
             status[start] = Visited::ongoing;
             visitor.start_node(start, graph);
             for (auto child : adjacency[start])
--- a/libmamba/include/mamba/util/graph.hpp
+++ b/libmamba/include/mamba/util/graph.hpp
@@ -66,11 +66,6 @@ namespace mamba::util
         template <typename UnaryFunc>
         UnaryFunc for_each_root_id_from(node_id source, UnaryFunc func) const;
 
-        // TODO C++20 better to return a range since this search cannot be interupted from the
-        // visitor
-        template <class V>
-        void depth_first_search(V& visitor, node_id start = node_id(0), bool reverse = false) const;
-
     protected:
 
         using derived_t = Derived;
@@ -89,25 +84,8 @@ namespace mamba::util
 
     private:
 
-        enum class visited
-        {
-            no,
-            ongoing,
-            yes
-        };
-
-        using visited_list = std::vector<visited>;
-
         template <class V>
         node_id add_node_impl(V&& value);
-
-        template <class V>
-        void depth_first_search_impl(
-            V& visitor,
-            node_id node,
-            visited_list& status,
-            const adjacency_list& successors
-        ) const;
 
         // Source of truth for exsising nodes
         node_map m_node_map;
@@ -118,17 +96,22 @@ namespace mamba::util
         std::size_t m_number_of_edges = 0;
     };
 
-    template <typename Graph>
-    auto
-    is_reachable(const Graph& graph, typename Graph::node_id source, typename Graph::node_id target)
-        -> bool;
+    // TODO C++20 better to return a range since this search cannot be interupted from the
+    // visitor
+    template <typename Graph, typename Visitor>
+    void
+    dfs_raw(const Graph& graph, Visitor&& visitor, typename Graph::node_id start, bool reverse = false);
 
-    template <class G>
-    class default_visitor
+    // TODO C++20 rather than providing an empty visitor, use a concept to detect the presence
+    // of member function.
+    // @warning Inheriting publicly from this class risks calling into the empty overloaded
+    // function.
+    template <typename Graph>
+    class EmptyVisitor
     {
     public:
 
-        using graph_t = G;
+        using graph_t = Graph;
         using node_id = typename graph_t::node_id;
 
         void start_node(node_id, const graph_t&)
@@ -154,6 +137,11 @@ namespace mamba::util
         {
         }
     };
+
+    template <typename Graph>
+    auto
+    is_reachable(const Graph& graph, typename Graph::node_id source, typename Graph::node_id target)
+        -> bool;
 
     template <typename Node, typename Edge = void>
     class DiGraph : private DiGraphBase<Node, DiGraph<Node, Edge>>
@@ -188,8 +176,6 @@ namespace mamba::util
         using Base::for_each_node_id;
         using Base::for_each_root_id;
         using Base::for_each_root_id_from;
-
-        using Base::depth_first_search;
 
         using Base::add_node;
         bool add_edge(node_id from, node_id to, const edge_t& data);
@@ -441,7 +427,7 @@ namespace mamba::util
     template <typename UnaryFunc>
     UnaryFunc DiGraphBase<N, G>::for_each_leaf_id_from(node_id source, UnaryFunc func) const
     {
-        struct LeafVisitor : default_visitor<derived_t>
+        struct LeafVisitor : EmptyVisitor<derived_t>
         {
             UnaryFunc& m_func;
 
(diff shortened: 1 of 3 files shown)


libmamba/tests/src/util/test_graph.cpp
```diff
@@ -339,11 +375,139 @@ TEST_SUITE("graph")
     {
         DiGraph<int> g;
         test_visitor<DiGraph<int>> vis;
-        g.depth_first_search(vis);
+        dfs_raw(g, vis, 0);
```
Hind-M, line 378: Should we convert `0` to `node_id`? I haven't looked into the details of it to check if the conversion is necessary, just saying because it was done in the declaration of previous used function:
`void depth_first_search(V& visitor, node_id start = node_id(0), bool reverse = false) const;`

libmamba/tests/src/util/test_graph.cpp
```diff
@@ -339,11 +375,139 @@ TEST_SUITE("graph")
...
+        }
+
+        SUBCASE("dfs_preorder on all nodes")
+        {
+            REQUIRE(g.has_node(n0));
+            REQUIRE(g.has_node(n1));
+            REQUIRE(g.has_node(n2));
+            dfs_preorder_nodes_for_each_id(g, [&nodes](node_id n) { nodes.push_back(n); });
+            CHECK(is_node_id_permutation(g, nodes.cbegin(), nodes.cend()));
+            CHECK_EQ(nodes, std::vector<node_id>{ n0, n1, n2 });
```
Hind-M, line 444: IIUC shouldn't `n2` come first (before `n1`)? Since `n1` is `n2`'s child?

libmamba/include/mamba/util/graph.hpp
```diff
@@ -443,127 +456,241 @@ namespace mamba::util
...
-        {
-            UnaryFunc& m_func;
-
-            LeafVisitor(UnaryFunc& func)
-                : m_func{ func }
-            {
-            }
-
-            void start_node(node_id n, const derived_t& g)
+        dfs_preorder_nodes_for_each_id(
```
Hind-M, line 459: I'm not sure about this, but shouldn't the `Leaf ` be replaced with postorder instead? (or maybe the trick was using reverse in the `Root`?).
I think that root/leaf keywords give more insight about this, maybe add a comment if we plan to remove them?

libmamba/tests/src/util/test_graph.cpp
AntoinePrv: No that's desired outcome. The dfs on all node starts with n0, finds n1, and stops. Then it starts again with the next unvisited node, n2, and finishes as n1 is already visited.

libmamba/include/mamba/util/graph.hpp
AntoinePrv: > I'm not sure about this, but shouldn't the Leaf be replaced with postorder instead?

It makes no difference, as they are no more nodes to visit after this one and the algorithm backtracks.

> or maybe the trick was using reverse in the Root?

Exactly

>  I think that root/leaf keywords give more insight about this, maybe add a comment if we plan to remove them?

I'm not sure what you mean.

libmamba/tests/src/util/test_graph.cpp
Hind-M: Even in preorder? Ok!

libmamba/include/mamba/util/graph.hpp
```diff
@@ -118,19 +96,48 @@ namespace mamba::util
...
+    // TODO should let user implement reverse with a reverse view when available
+    template <typename Graph, typename Visitor>
+    void
+    dfs_raw(const Graph& graph, Visitor&& visitor, typename Graph::node_id start, bool reverse = false);
+
+    template <typename Graph, typename Visitor>
+    void dfs_raw(const Graph& graph, Visitor&& visitor, bool reverse = false);
+
+    template <typename Graph, typename UnaryFunc>
+    void dfs_preorder_nodes_for_each_id(
```
JohanMabille, line 110: NItpicking: depth_first_search was more expressive than `dfs`. I nunderstand you changed it to avoid very long names for these overloads, can we just omit the `for_each_id` part? So we would have `depth_first_search_preorder_nodes`.

Another (maybe more elegant) solution could be to have all the DFS related stuff into a `depth_first_search` namespace, and remove the `dfs` prefix.

libmamba/tests/src/util/test_graph.cpp
AntoinePrv: It does not make a difference (`node_id` is an alias for `std::size_t`) but it's more explicit. I'll add it.

libmamba/include/mamba/util/graph.hpp
```diff
@@ -157,6 +164,14 @@ namespace mamba::util
         }
     };
 
+    template <typename Graph>
+    auto
+    is_reachable(const Graph& graph, typename Graph::node_id source, typename Graph::node_id target)
+        -> bool;
+
+    template <typename Graph, typename UnaryFunc>
+    void topological_sort_for_each_node_id(const Graph& graph, UnaryFunc&& func);
```
JohanMabille, line 173: Same remark regarding the name, do we really need the `for_each_node_id` part of the name?

libmamba/include/mamba/util/graph.hpp
```diff
@@ -443,127 +456,241 @@ namespace mamba::util
...
+                }
+            },
+            source,
+            /* reverse= */ true
+        );
+        return func;
+    }
+
+    template <typename N, typename G>
+    auto DiGraphBase<N, G>::number_of_node_id() const noexcept -> node_id
```
JohanMabille, line 493: Naming: `successor_count`?

libmamba/include/mamba/util/graph.hpp
Hind-M: > > I think that root/leaf keywords give more insight about this, maybe add a comment if we plan to remove them?
> 
> I'm not sure what you mean.

`LeafVisitor` and `RootVisitor` don't exist anymore, but the naming was maybe helping to understand the purpose of the code? So I suggested adding a comment for that (to explain what it does).

libmamba/include/mamba/util/graph.hpp
AntoinePrv: I changed to `dfs_` only to mirror `networkx`.

Namespace-wise, this is not even in a graph namespace, only in `mamba::util`. We could improve that as we make a proper utility lib.

libmamba/include/mamba/util/graph.hpp
AntoinePrv: No this is meant to be the number of ids generated, for being able to use an `std::vector` instead of a map.

libmamba/tests/src/util/test_graph.cpp
AntoinePrv: Yes, a better way to think of it it's that is is like two calls to `dfs` (as dfs does not explore everything on the first run).

libmamba/include/mamba/util/graph.hpp
AntoinePrv: > LeafVisitor and RootVisitor don't exist anymore,

Ha yes, the visitor pattern was more verbose so I switch to the preorder function.

> So I suggested adding a comment for that (to explain what it does).

Got it.

libmamba/include/mamba/util/graph.hpp
AntoinePrv: The idea with all the `for_each` function I write (here and in other places) is to not force a return type like `std::vector` because the user may not need it.

A better API would be in C++20 when, instead of having a visitor pattern, we will return a range that can properly iterate through the desired value without allocating and copying data. At that point we would drop the `for_each`

JohanMabille: ok, let's keep the `dfs` prefix for now then.

libmamba/include/mamba/util/graph.hpp
JohanMabille: Got it

Hind-M (approved):

libmamba/include/mamba/util/graph.hpp
```diff
@@ -443,127 +456,250 @@ namespace mamba::util
...
+        template <typename Graph, typename Visitor>
+        void dfs_raw_impl(
+            const Graph& graph,
+            Visitor&& visitor,
+            typename Graph::node_id start,
+            std::vector<Visited>& status,
+            const typename Graph::adjacency_list& adjacency
+        )
+        {
+            status[start] = Visited::ongoing;
```
Klaim, line 542: Is it safe to do this instead of using `.at` here?

Klaim: If it's required that `start` is in boundaries of `status` size, maybe just add an assertion?

Klaim (approved):
LGTM

libmamba/include/mamba/util/graph.hpp
Klaim: I think that's a contract of the functions and the asserts are enough checks yeah.

libmamba/include/mamba/util/graph.hpp
```diff
@@ -539,6 +540,9 @@ namespace mamba::util
             const typename Graph::adjacency_list& adjacency
         )
         {
+            assert(status.size() == graph.successors().size());
+            assert(adjacency.size() == graph.successors().size());
+            assert(start <= status.size());
```
Hind-M, line 545: Shouldn't it be a strict comparison here?

AntoinePrv: Yes, thanks!

PR #2469, opened 2023-04-17
Remove unused CURLSetup

Hind-M:
It seems that this is not used anywhere, was it supposed to be at some point?

### DIFF
--- a/libmamba/src/core/singletons.cpp
+++ b/libmamba/src/core/singletons.cpp
@@ -4,11 +4,6 @@
 #include <mutex>
 #include <regex>
 
-extern "C"
-{
-#include <curl/urlapi.h>
-}
-
 #include "mamba/api/configuration.hpp"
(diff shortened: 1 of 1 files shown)


libmamba/src/core/singletons.cpp
```diff
@@ -35,61 +30,6 @@ namespace mamba
...
-            }
-        }
-
-        ~CURLSetup()
-        {
-            curl_global_cleanup();
-        }
-    };
-
-    static CURLSetup curl_setup;
```
wolfv, line 91: This line creates a static global curl setup which initialized the curl runtime so it is used :)

wolfv (changes requested):

libmamba/src/core/singletons.cpp
Hind-M: Ohh ok! Thanks!

PR #2470, opened 2023-04-18
Add doctest printer for pair and vector

AntoinePrv:
Include these files for doctest to properly print the value.
They can be used and improved gradually.

### DIFF
--- a/libmamba/tests/src/doctest-printer/pair.hpp
+++ b/libmamba/tests/src/doctest-printer/pair.hpp
@@ -1,0 +1,22 @@
+// Copyright (c) 2023, QuantStack and Mamba Contributors
+//
+// Distributed under the terms of the BSD 3-Clause License.
+//
+// The full license is in the file LICENSE, distributed with this software.
+
+#include <utility>
+
+#include <doctest/doctest.h>
+#include <fmt/format.h>
+
+namespace doctest
+{
+    template <typename T, typename U>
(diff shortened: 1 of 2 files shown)


libmamba/tests/src/doctest-printer/pair.hpp
```diff
@@ -0,0 +1,22 @@
...
+#include <fmt/format.h>
+
+namespace doctest
+{
+    template <typename T, typename U>
+    struct StringMaker<std::pair<T, U>>
+    {
+        static auto convert(const std::pair<T, U>& value) -> String
+        {
+            return { fmt::format("std::pair{{{}, {}}}", value.first, value.second).c_str() };
```
Hind-M, line 19: Why do we convert to a c-string in the middle, to have it null-terminated?
A more general question, do you think there will be a lot of these functions so that it could be more readable to put them in separate files, or maybe the opposite? Because otherwise we could also put everything in one file.

AntoinePrv: Well, doctest does not use `std::string` anywhere, not even as conversion...
They have a `String` that can be built from a `const char*`, so I pass it the `c_str()` 

I think more would come yes. Also it is to avoid including all of `<map>` `<set>`, `<tuple>`, `"graph.hpp"` in every test file.

JohanMabille (approved):

PR #2474, opened 2023-04-19
Improve micromamba transaction message

ruben-arts:
Before:
```bash
Transaction finished

To activate this environment, use:

    $ [micro]mamba activate <environment>

Or to execute a single command in this environment, use:

    $ [micro]mamba run -n <environment> mycommand
```
Examples of the result now:
```bash
$ time ./build/micromamba/micromamba create -c conda-forge --yes -n test python

# results in

Transaction finished

To activate this environment, use:

    micromamba activate test

Or to execute a single command in this environment, use:

    micromamba run -n test mycommand
    

$ time ./micromamba create -c conda-forge --yes -p ~/testing python

# results in, note the -p or -n

Transaction finished

To activate this environment, use:

   micromamba activate /home/rarts/testing

Or to execute a single command in this environment, use:

    micromamba run -p /home/rarts/testing mycommand
```

Notable improvements:
- If is_micromamba is true micromamba will be used as the executable
- Use no `$` to make the line double click + copy + paste.
- Use `-p` when using prefix, use `-n` for a named env.

### DIFF
--- a/libmamba/src/core/transaction.cpp
+++ b/libmamba/src/core/transaction.cpp
@@ -1101,8 +1101,10 @@ namespace mamba
                               << "To activate this environment, use:\n\n"
                               << "    " << executable << " activate " << environment << "\n\n"
                               << "Or to execute a single command in this environment, use:\n\n"
-                              << "    " << executable << " run "
-                              // Use -n or -p depending on if the env_name is a full prefix or just a name.
+                              << "    " << executable
+                              << " run "
+                              // Use -n or -p depending on if the env_name is a full prefix or just
+                              // a name.
                               << (environment == ctx.target_prefix ? "-p " : "-n ") << environment
                               << " mycommand\n";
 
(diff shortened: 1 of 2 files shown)


libmamba/src/core/transaction.cpp
```diff
@@ -1086,15 +1086,28 @@ namespace mamba
...
         {
-            // TODO: add executable name
-            // TODO: add environment name/prefix
             LOG_INFO << "Waiting for pyc compilation to finish";
             m_transaction_context.wait_for_pyc_compilation();
+
+            // Get the name of the executable used directly from the command.
+            std::istringstream iss(ctx.command_params.current_command);
+            std::string executable;
+            std::getline(iss, executable, ' ');
```
AntoinePrv, line 1095: Could this be a method of the command?

wolfv:
Unfortunately the `executable` won't work for `activate` because the activation is actually implemented on the shell level. So it has to be `micromamba activate ...`.

AntoinePrv:
So maybe you can switch with `ctx.is_micromamba`?

ruben-arts:
> Unfortunately the `executable` won't work for `activate` because the activation is actually implemented on the shell level. So it has to be `micromamba activate ...`.

Should we keep the run with the executable or keep the also move that to micromamba?

wolfv:
I think the executable path can be long and weird so maybe just micromamba is better

Hind-M (approved):
Thanks!

PR #2477, opened 2023-04-20
Add mamba version to `mamba info` output

### LINKED ISSUE #2458: `mamba info` seems to not include the Mamba version
In some/all? of the bug reports we see `mamba info` output that doesn’t include the mamba version. It’s very important for diagnosis though

Hind-M:
Fix #2458

### DIFF
--- a/mamba/mamba/mamba.py
+++ b/mamba/mamba/mamba.py
@@ -766,7 +766,7 @@ def do_call(args, parser):
             else:
 
                 def format_param(nm, val):
-                    return "%23s : %s" % (nm, val)  # To match conda's format
+                    return f"{nm:>23} : {val}"  # To match conda's format
 
                 print(format_param("mamba version", mamba.__version__))
                 print(output.lstrip("\n"))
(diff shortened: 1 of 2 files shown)


mamba/mamba/mamba.py
```diff
@@ -747,7 +747,31 @@ def do_call(args, parser):
...
+
+                from conda.common.serialize import json_dump
+
+                json_obj = loads(output)
+                json_obj["mamba_version"] = mamba.__version__
+                print(json_dump(json_obj))  # To match conda's format
+            else:
+
+                def format_param(nm, val):
+                    return "%23s : %s" % (nm, val)  # To match conda's format
```
AntoinePrv, line 769: Should we convert to an fstring instead?

JohanMabille (approved):

mamba/mamba/mamba.py
Hind-M: Done!

Klaim (approved):

PR #2485, opened 2023-04-24
fix: parsing of empty track_features

AntoinePrv:
Fix on #2478

### DIFF
--- a/libmamba/src/core/package_info.cpp
+++ b/libmamba/src/core/package_info.cpp
@@ -114,7 +114,11 @@ namespace mamba
         assign_or(j, "sha256", sha256, ""s);
         if (j.contains("track_features"))
         {
-            track_features = split(j["track_features"].get<std::string_view>(), ",");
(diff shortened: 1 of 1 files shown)


Hind-M (approved):
Should we add a check in `split` function to throw in case the input is empty?
Something similar to this?
https://github.com/mamba-org/mamba/blob/main/libmamba/src/core/util_string.cpp#L478-L481

AntoinePrv:
I think it's a well defined operation, in Python you have `"".split(",") == [""]`.

JohanMabille (approved):

Hind-M:
Oh so the purpose of this PR was more avoiding to call `split` when the input is empty than a fix right?

AntoinePrv:
In the case of `track_feature`, the default behaviour of split (i.e. the Python behaviour) is not what we want.
This fixes the invocation of split, not split itself (which need not be fixed).

PR #2494, opened 2023-04-26
Support future deprecated API for `Context`

Hind-M:


JohanMabille:
I think you need to update the stubs too.

Hind-M:
@AntoinePrv I was wondering if the `deprecated` function works?
https://github.com/mamba-org/mamba/blob/main/libmambapy/src/main.cpp#L65-L70
I tested locally (in the context of this PR) but it doesn't seem to. Did I miss something?

AntoinePrv:
@Hind-M they should work properly, but maybe what you are seeing is that Python does not show them by default.
The reason for that is that they would otherwise show up for `conda` users, which we do not want.

At least that the distinction I understand between `DeprecationWarning` and `FutureWarning` ([doc](https://docs.python.org/3/library/warnings.html#warning-categories)).

You can try for youself:

```python
import warnings

warnings.warn("Le deprecation warning", DeprecationWarning, stacklevel=2)
warnings.warn("Le future warning", FutureWarning, stacklevel=2)
```

With `python [file.py]` vs `python -Walways::DeprecationWarning [file.py]`.

Note that test frameworks, such as Pytest do enable warnings.

Hind-M:
Ok I see, so I should remove enabling them by default in my last commit. Thanks!

JohanMabille (approved):

PR #2497, opened 2023-04-26
Forward NETRC environment variable to curl, if exported

### LINKED ISSUE #2496: (micro-)mamba does not support NETRC export
conda supports exporting NETRC to redirect the used netrc file, which usually resides at $HOME/.netrc (on linux) for authentication, while mamba does not.

This is used e.g., if you want to use a private conda repository, which requires authentication in docker builds. If you would use
https://user:secret@myrepo/ in your environment file, it would be hard to mask this in the installed environment.
mamba supports this already, but only in the $HOME/.netrc file.

conda reads this in conda/cli/main_info.py and uses get_netrc_auth of python-requests to do actual requests. Full disclosure: I have never tried, if this actually works in conda, but I think it should. To use this in mamba, I think we just have to use the correct curl option: CURLOPT_NETRC_FILE

The fix in mamba is simple and I will send an untested draft PR soon (and test it on Friday).

timostrunk:
If NETRC is exported, we use it in configure_curl_handle to pass it on.

This resolves #2496

- [ ] This is still untested. I will test it on Friday.

### DIFF
--- a/libmamba/src/core/curl.cpp
+++ b/libmamba/src/core/curl.cpp
@@ -7,8 +7,9 @@
 // TODO remove all these includes later?
 #include <spdlog/spdlog.h>
 
-#include "mamba/core/mamba_fs.hpp"  // for fs::exists
-#include "mamba/core/util.hpp"      // for hide_secrets
+#include "mamba/core/environment.hpp"  // for NETRC env var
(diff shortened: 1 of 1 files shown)


jonashaag (approved):
Thanks! I think this is fine. Note thought that condarc also has its own way or specifying authentication with URLs and environment variables.

PR #2512, opened 2023-05-03
add bearer token authentication

wolfv:

PR #2520, opened 2023-05-07
Use only vcpkg for static windows build

pavelzw:
I suggest we switch to vcpkg completely for the windows static build process for micromamba. This makes the build pipeline for windows more maintainable since the dependencies in vcpkg are actively maintained not only by mamba-org. Also, fewer weird errors occur when building outside of a conda environment with MSVC.

This also allows us to use `windows-latest` instead of `windows-2019`.

The build time for static win-64 sinks from [40min](https://github.com/mamba-org/mamba/actions/runs/4891770895/jobs/8732705840) down to 7min or [2min](https://github.com/mamba-org/mamba/actions/runs/4912518885/jobs/8771623891?pr=2520) with sccache (as long as the vcpkg dependencies are cached which is now easier outside of a conda env).


jonashaag:
Thanks a lot for working on this!

I wonder if there is a way to share the build instructions with the feedstock. Maybe we can use the feedstock `bld.bat`?

JohanMabille:
So here is my suggestion: let's merge https://github.com/mamba-org/mamba/pull/2526 to fix the CI, then merge https://github.com/mamba-org/mamba/pull/2512 and other stuff when ready so that we can quickly push out a release.

Then we merge this one and setup the CI so that it is possible to test the feedstock before releasing.

jonashaag (approved):

pavelzw:
No worries 👍🏻

PR #2525, opened 2023-05-14
Improvement: Remove repodata_use_zst flag

johnhany97:
## Before this PR

Users would have to add `repodata_use_zst: true` in their .condarc file to leverage the much lighter .zst variant of repodatas, should they exist.

## After this PR

We remove the flag and make it default to always use zst where possible for downloading repodatas


johnhany97:
Okay great! I've merged this into the other PR because I realised they need to go in together because the mamba codepaths would fail otherwise. Closing this one to continue iterating on my other PR.

PR #2527, opened 2023-05-15
[Internal] Use subsub commands for `micromamba shell`

AntoinePrv:
The code was previously not using sub command, which creates a mix of invlaid CLI arguments.
This is taking some load off #2475.

### DIFF
--- a/libmamba/data/Mamba.psm1
+++ b/libmamba/data/Mamba.psm1
@@ -58,7 +58,7 @@ function Exit-MambaEnvironment {
         if ($deactivateCommand.Trim().Length -eq 0) {
             return;
         }
-        Write-Verbose "[micromamba shell deactivate -s powershell]`n$deactivateCommand";
+        Write-Verbose "[micromamba shell deactivate --shell powershell]`n$deactivateCommand";
         Invoke-Expression -Command $deactivateCommand;
     }
     process {}
--- a/micromamba/src/shell.cpp
+++ b/micromamba/src/shell.cpp
@@ -204,6 +204,65 @@ namespace
             }
         );
     }
+
+    template <typename Arr>
+    void set_shell_launch_command(CLI::App* subcmd, const Arr& all_subsubcmds)
+    {
+        // The initial parser had the subcmdmand as an action so both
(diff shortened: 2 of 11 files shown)


micromamba/src/shell.cpp
```diff
@@ -8,127 +8,337 @@
...
+
+        auto& prefix = config.insert(
+            Configurable("shell_prefix", std::string(""))
+                .group("cli")
+                .description("The root prefix to configure (for init and hook), and the prefix "
+                             "to activate for activate, either by name or by path"),
+            true
+        );
+        subcmd->add_option(
+            "prefix,-p,--prefix,-n,--name",
```
Hind-M, line 45: Oh so at the end, we keep handling `--name` and `--prefix` with the same option. As long as it works and isn't confusing...

libmamba/data/Mamba.psm1
```diff
@@ -58,7 +58,7 @@ function Exit-MambaEnvironment {
         if ($deactivateCommand.Trim().Length -eq 0) {
             return;
         }
-        Write-Verbose "[micromamba shell deactivate -s powershell]`n$deactivateCommand";
+        Write-Verbose "[micromamba shell deactivate --shell powershell]`n$deactivateCommand";
```
Hind-M, line 61: Is this change aiming to make the command more explicit?
If so, should we change it everywhere?

micromamba/src/shell.cpp
AntoinePrv: The goal is to split them, but this PR was already enough work to introduce the necessary structure to be able to split them. This will be done in a follow-up.

libmamba/data/Mamba.psm1
AntoinePrv: Yhea, it's cosmetic. At first I started looking at them because I wanted to remove usages of  micromamba shell --shell <shell> <cmd>` in favor of  `micromamba shell <cmd> --shell <shell>`

Hind-M (approved):

Klaim (approved):

PR #2528, opened 2023-05-15
Fix CURLHandle::get_info on 32bit platform

e8035669:
I am building mamba and micromamba on armv7l device, which is 32bit platform. Both of them will run into segmentation fault.
I found the problem is using wrong type with `libcurl`. when calling `get_info(CURLINFO_SPEED_DOWNLOAD_T)` the type should be `curl_off_t`, but the actual type of `curl_off_t` can be `long` or `long long` depends on 64bit or 32bit platform.
So I change `CURLHandle::get_info<std::size_t>` make it use `curl_off_t` type, I think it will match one of `long` or `long long` template function.
This patch can work on my armv7 device, but I haven't test on my PC.

Edit: I tested on my PC, this patch does not affect 64bit computer and all works great.


wolfv:
Should we remove the `long` option and instead only use `curl_off_t`?

Hind-M (commented):
Thanks for this PR!
I think that you can rather replace `long` with `curl_off_t` in `get_integer_info` to have it more generic and leave `CURLHandle::get_info<std::size_t>` as it is.
In this line:
https://github.com/mamba-org/mamba/blob/main/libmamba/src/core/curl.cpp#L294

Hind-M:
> Should we remove the `long` option and instead only use `curl_off_t`?

Ideally we are aiming to keep curl specific code only in internal curl files (and remove everything explicit about it in `fetch.cpp`). So using `curl_off_t` instead of `long` in `get_integer_info` should be enough to fix the issue.

e8035669:
> Oh or did you mean in here? https://github.com/mamba-org/mamba/blob/main/libmamba/src/core/curl.cpp#L286 Then yes! :)

I think we can't just change `long` to `curl_off_t`, because `get_info(CURLINFO_RESPONSE_CODE)` in `fetch.cpp` required a `long` type value, according to [libcurl document](https://curl.se/libcurl/c/CURLINFO_RESPONSE_CODE.html)

so I think we need both `long` and `long long` version template specialization to fit their needs.

e8035669:
> > Should we remove the `long` option and instead only use `curl_off_t`?
> 
> Ideally we are aiming to keep curl specific code only in internal curl files (and remove everything explicit about it in `fetch.cpp`). So using `curl_off_t` instead of `long` in `get_integer_info` should be enough to fix the issue.

Yeah, I agree to remove everything explicit about it in fetch.cpp, so I only change curl.cpp and keep fetch.cpp as it is.

wolfv:
thanks @e8035669 I was also just checking precisely that.

e8035669:
No matter `curl_off_t` is `long` or `long long`, they will cast to `std::size_t`, even if `std::size_t` has only 32 bits length on 32 bit platform. (overflow on 32bit platform potentially)

Hind-M (approved):

Hind-M:
@JohanMabille Should we remove `get_integer_info` completely since it's only used now in `get_info` with `int`? (and so it's explicit that it's converted using `get_info<long>`)

JohanMabille:
I think we should open a PR upstream to fix this [page of the doc](https://curl.se/libcurl/c/curl_easy_getinfo.html) which is misleading: 

"Request internal information from the curl session with this function. The third argument MUST be a pointer to a long, a pointer to a char *, a pointer to a struct curl_slist * or a pointer to a double (as this documentation describes further down)"

The third argument should also accept pointer to long long.

@Hind-M yeah we should probably clean a bit the code.

PR #2532, opened 2023-05-17
Common CMake presets

AntoinePrv:
Some presets that can be used to configure.
Right now this is mostly developer oriented, so let me know if that works for you.
BTW, cutom presets can go in `CMakeUserPresets.jons`, reusing the presets given here.


Hind-M:
> Right now this is mostly developer oriented, so let me know if that works for you. 
Are we aiming to use it for CI, for the project in general to be used by default, or just as a reference to be shared here?

AntoinePrv:
I think right now we can use it between us. For CI we rebuild every time with [libmamba,libmamba+test,libmamba+micromamba,libmamba+mamba] so that's not super nice to set up.

I think we could move to build everything once, then reuse the build in different test (mamba, micomamba...)

JohanMabille (approved):

PR #2534, opened 2023-05-17
[Libcurl] Cleaning and comments

Hind-M:


libmamba/src/core/curl.cpp
```diff
@@ -281,21 +281,30 @@ namespace mamba
...
-    tl::expected<I, CURLcode> CURLHandle::get_integer_info(CURLINFO option)
+    template <>
+    tl::expected<std::size_t, CURLcode> CURLHandle::get_info(CURLINFO option)
     {
-        auto res = get_info<long>(option);
+        auto res = get_info<curl_off_t>(option);
         if (res)
         {
-            return static_cast<I>(res.value());
+            return static_cast<std::size_t>(res.value());
```
wolfv, line 307: I maybe should've asked earlier but isn't casting to `size_t` a mistake since before it was a signed integer?

Hind-M: Because it's used to set `speed` and `m_downloaded_size` which are eventually converted to `std::size_t` in `set_speed`, `update_progress` and `set_total` of the `ProgressBar`...
These notions with `m_retry_wait_seconds` are also intuitively representing something that should be unsigned... I wonder why it's not handled in `libcurl` that way btw...

wolfv: Maybe they use negative values for errors.

JohanMabille: If that's the case, we should check the return value and return the appropriate `tl::unexpected`.

JohanMabille (approved):

PR #2539, opened 2023-05-23
fix different behavior between --version and -V options

alaniwi:
Fix a little bug where `mamba --version` was displaying the mamba version but `mamba -V` was not (despite the two options being described as equivalent in the `mamba -h` output).

For completeness, I've tested it as follows:
 - install miniforge (without mamba)
 - activate base env
 - `conda install libmambapy`
 - `python setup.py install`
 - try `mamba -V` and `mamba --version` (and a couple of unrelated options)

Thanks to @jaimergp for identifying which line of code needed to be fixed.


AntoinePrv (approved):

Hind-M:
Thanks!

PR #2541, opened 2023-05-23
No Singleton: `Configuration`

Klaim:


libmamba/include/mamba/api/configuration.hpp
```diff
@@ -149,10 +137,13 @@ namespace mamba
             virtual YAML::Node yaml_value() const = 0;
             virtual void dump_json(nlohmann::json& node, const std::string& name) const = 0;
 
+            bool is_config_loading() const;
```
AntoinePrv, line 140: What does that do?

AntoinePrv (commented):
Thanks! Just a few questions

libmamba/include/mamba/api/configuration.hpp
```diff
@@ -149,10 +137,13 @@ namespace mamba
...
+
             std::string m_name;
             std::string m_group = "Default";
             std::string m_description = "No description provided";
             std::string m_long_description = "";
+            Configuration* m_config = nullptr;
```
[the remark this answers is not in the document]
Klaim: See:
- https://github.com/mamba-org/mamba/pull/2541/files#diff-ec708ed47303bbb952a95a186b9463393bdddf3762aa7fbade9f5c21114cc632L662
- https://github.com/mamba-org/mamba/pull/2541/files#diff-ec708ed47303bbb952a95a186b9463393bdddf3762aa7fbade9f5c21114cc632L793

My understanding of the original code is that Configurables in a Configuration can be manipulated before `Configuration::load()` is done and in that case `Configurable::value()` can throw an exception if called before `Configuration::load()`. This is not ideal but I didnt want to change the logic and Configurable was accessing Configuration through the singleton accessor, so removing it lead to Configurable needing to be able to check that the Configuration is or not in the loading phase. The only way I found was adding the pointer to the Configuration once the Configurable is stored in the Configuration (because it could be created before that point).
When we rewrite Configuration itself to be as discussed last live meeting (real fields, no magically casting dynamic fields), this issue can be completely removed by having 2 types of configurations: one that is being set and one "locked".

libmamba/include/mamba/api/configuration.hpp
Klaim: See my last comment.

AntoinePrv (approved):

JohanMabille (approved):

libmamba/include/mamba/api/install.hpp
```diff
@@ -54,9 +56,9 @@ namespace mamba
 
         void create_empty_target(const fs::u8path& prefix);
 
-        void file_specs_hook(std::vector<std::string>& file_specs);
+        void file_specs_hook(std::vector<std::string>& file_specs, Configuration& config);
```
JohanMabille, line 59: Nitpicking: config should be the first argument to be consistent with other functions in this file.

Klaim: Ah yes nice catch, will fix

PR #2495, opened 2023-04-26
Switch to setup-micromamba

pavelzw:

PR #2545, opened 2023-05-25
Move GHA to setup-micromamba

AntoinePrv:


.github/workflows/main.yml
```diff
@@ -45,10 +48,9 @@ jobs:
       - name: build cache statistics
         run: sccache --show-stats
       - name: Cleanup
-        shell: bash
         if: always()
         run: |
-          # Do not cache temporary envs with 'cache-env: true'
+          # Do not cache temporary envs with 'cache-environment: true'
```
jonashaag, line 53: I think we don't need this step anymore. In setup-micromamba we changed the point in time when the environment is cached to right after environment creation, not after the rest of the CI.

pavelzw:
Thanks for fixing what I [gave up on](https://github.com/mamba-org/mamba/pull/2495)!

.github/workflows/main.yml
```diff
@@ -588,19 +602,20 @@ jobs:
...
       - name: create build environment
-        uses: mamba-org/provision-with-micromamba@main
+        uses: mamba-org/setup-micromamba@v1
         with:
           environment-file: ./micromamba/environment-dev.yml
           environment-name: build_env
-          extra-specs: menuinst
-
+          create-args: menuinst
+          init-shell: bash cmd.exe powershell
```
pavelzw, line 610: Setting `init-shell: none` could also be an option. This would then definitely not interfere with the shell initialization tests...

We would then need to activate the micromamba environment by hand or call `micromamba run -n ...`

AntoinePrv: I think when `init-shell` was not used, the shell was not finding `micromamba` executable

pavelzw: Yes, this is because the micromamba executable is not on PATH. You could add it to PATH manually beforehand:

```yml
      - name: Add micromamba to GITHUB_PATH
        run: echo "${HOME}/micromamba-bin" >> "$GITHUB_PATH"
```

jonashaag: Maybe I lobby for https://github.com/mamba-org/mamba/issues/2577

.github/workflows/main.yml
```diff
@@ -59,40 +53,38 @@ jobs:
...
         with:
           variant: sccache
           key: ${{ github.job }}-${{ matrix.os }}
           restore-keys: |
             libmamba_static-${{ matrix.os }}
-      - name: build and run libmamba C++ tests
+      - name: Build libmamba
         shell: bash -l {0}
         run: |
           micromamba activate build_env
```
jonashaag, line 71:
```suggestion
```

.github/workflows/main.yml
```diff
@@ -59,40 +53,38 @@ jobs:
...
+            -D BUILD_SHARED=ON \
+            -D BUILD_LIBMAMBA_TESTS=ON \
+            -D CMAKE_CXX_COMPILER_LAUNCHER=sccache \
+            -D CMAKE_C_COMPILER_LAUNCHER=sccache \
+            -G Ninja
+          cmake --build build/ --parallel --target test_libmamba testing_libmamba_lock
+      - name: Run libmamba tests
+        shell: bash -l {0}
+        run: |
+          micromamba activate build_env
```
jonashaag, line 83:
```suggestion
```

.github/workflows/main.yml
```diff
@@ -118,16 +110,15 @@ jobs:
         shell: bash -l {0}
         run: |
           micromamba activate build_env
```
jonashaag, line 112:
```suggestion
```

.github/workflows/main.yml
```diff
@@ -143,15 +134,10 @@ jobs:
       - name: micromamba python based tests
         shell: bash -l -eo pipefail {0}
         run: |
-          export TEST_MAMBA_EXE=$(pwd)/build/micromamba/micromamba
           micromamba activate build_env
```
jonashaag, line 137:
```suggestion
```

.github/workflows/main.yml
```diff
@@ -185,23 +171,23 @@ jobs:
         shell: bash -l {0}
         run: |
           micromamba activate build_env
```
jonashaag, line 173:
```suggestion
```

.github/workflows/main.yml
```diff
@@ -185,23 +171,23 @@ jobs:
...
+            -D BUILD_MAMBA_PACKAGE=ON \
+            -D CMAKE_CXX_COMPILER_LAUNCHER=sccache \
+            -D CMAKE_C_COMPILER_LAUNCHER=sccache \
+            -G Ninja
+          cmake --build build/ --parallel
+          cmake --install build/
       - name: install libmambapy
         shell: bash -l {0}
         run: |
           micromamba activate build_env
```
jonashaag, line 189:
```suggestion
```

.github/workflows/main.yml
```diff
@@ -213,23 +199,26 @@ jobs:
...
```
jonashaag, line 201:
```suggestion
```

.github/workflows/main.yml
```diff
@@ -213,23 +199,26 @@ jobs:
...
-          pip install ./mamba[test] --no-deps
+          python -m pip install --no-deps --no-build-isolation ./mamba[test]
       - name: run mamba tests suite
         shell: bash -l {0}
         run: |
           micromamba activate build_env
```
jonashaag, line 206:
```suggestion
```

jonashaag:
LGTM, I think all of the `micromamba activate build_env` can go, will try that in a new PR

JohanMabille (approved):

.github/workflows/main.yml
JohanMabille: > provision-with-micromamba did the shell initialization thing for every shell that was on the system. setup-micromamba allows you to specify the shell and also gives you the option to do the initialization for no shell at all. I think this would be the option that makes the most sense for shell init tests.

I agree this would be better, but I'm also afraid by potential issues this could lead to and the time it would take to fix them. Let's merge this one since it passes the, and and tackle this in a dedicated PR.

pavelzw (approved):

PR #2556, opened 2023-05-31
Fix stubgens

AntoinePrv:


AntoinePrv:
Looking into the details of https://github.com/mamba-org/mamba/pull/2558, this seems to be a partial fix

JohanMabille (approved):

PR #2557, opened 2023-05-31
Add Python-like set operations to flat_set

AntoinePrv:
- Add flat-set doctest printer
- Add Python-like set operations to flat_set

### DIFF
--- a/libmamba/include/mamba/util/flat_set.hpp
+++ b/libmamba/include/mamba/util/flat_set.hpp
@@ -15,6 +15,14 @@
 namespace mamba::util
 {
 
+
+    struct sorted_unique_t
+    {
+        explicit sorted_unique_t() = default;
+    };
+
+    inline constexpr sorted_unique_t sorted_unique{};
+
     /**
      * A sorted vector behaving like a set.
      *
@@ -61,6 +69,14 @@ namespace mamba::util
             key_compare compare = key_compare(),
             const allocator_type& alloc = Allocator()
         );
+        template <typename InputIterator>
+        flat_set(
+            sorted_unique_t,
+            InputIterator first,
+            InputIterator last,
+            key_compare compare = key_compare(),
+            const allocator_type& alloc = Allocator()
+        );
         flat_set(const flat_set&) = default;
         flat_set(flat_set&&) = default;
         explicit flat_set(std::vector<Key, Allocator>&& other, key_compare compare = key_compare());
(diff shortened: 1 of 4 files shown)


Klaim:
I feel like it's not a good idea to wrap generic algorithms into member functions for convenience because
- as soon as we switch to C++20 we will get ranges, making these functions obsolete or equivalent;
- usually these algorithms are defined as free functions and/or have different names in the c++ world (also see: https://www.boost.org/doc/libs/1_82_0/doc/html/boost/container/flat_set.html and the version that was voted in C++23: https://en.cppreference.com/w/cpp/header/flat_set )

I'm not completely against adding these functions but could you make them free instead of member at least?

AntoinePrv:
The algorithms exists as member functions because they can leverage the specific internals of the class for better algorithmic complexity. Just like we don't use `std::find` on a map.

IMHO C++ `<algorithm>` header is confusing. There are "range manipulation" algorithms (filter, transform...) that are in fact generic and the C++20 range library does amazingly well. Then there are more complex algorithms (e.g. non-linear), and these cannot be decoupled from data structure. I think set operations fall in the second category.

libmamba/include/mamba/util/flat_set.hpp
```diff
@@ -30,6 +30,7 @@ namespace mamba::util
     public:
 
         using Base = std::vector<Key, Allocator>;
+        using Self = flat_set<Key, Compare, Allocator>;
```
JohanMabille, line 33: bitpicking: I think we can simply use `flat_set` in the declaration and implementation, the template parameters will be deduced.

JohanMabille (approved):

libmamba/include/mamba/util/flat_set.hpp
```diff
@@ -307,17 +409,151 @@ namespace mamba::util
...
+         *
+         * Edited from https://en.cppreference.com/w/cpp/algorithm/set_intersection
+         * Distributed under the terms of the Copyright/CC-BY-SA License.
+         * The full license can be found at the address
+         * https://en.cppreference.com/w/Cppreference:Copyright/CC-BY-SA
+         */
+        template <class InputIt1, class InputIt2, class Compare>
+        auto
+        set_disjoint(InputIt1 first1, InputIt1 last1, InputIt2 first2, InputIt2 last2, Compare comp)
+            -> bool
```
JohanMabille, line 430: :smiling_face_with_tear:

PR #2571, opened 2023-06-08
Fix shadow warning

Hind-M:


JohanMabille (approved):

Klaim:
This will break https://github.com/mamba-org/mamba/pull/2541 which I believe also fixes the same shadowing problem (by using the config passed as parameter).

PR #2572, opened 2023-06-08
Try new vcpkg

AntoinePrv:


JohanMabille (approved):

AntoinePrv:
@jonashaag I'm no VCPKG expert, so I went for a greedy approach.
For the line you point too, I'm also wondering if the cache is working as intended (need to investigate).

PR #2584, opened 2023-06-12
Create Solver solution

AntoinePrv:
- [x] Based on #2602

The new `Solution` class represents the outcome of a Solver solve, which is now help inside a `::Transaction*`.
This PR introduce slowly start rolling it out in `mamba::MTransaction`.
The `MSolver` flags are also refactored into a `struct`.
Other transitory changes are added.


libmamba/include/mamba/core/solver.hpp
```diff
@@ -60,6 +60,16 @@ namespace mamba
     {
     public:
 
+        struct Flags
```
Klaim, line 63: For future maintainance and optimization, it might be best to make this expose only functions to read or write the values, as the implementation could change (to bits for example, in particular if we add more values later). Changing the implementation would then be easy without changing the interface.
An alternative, IFF flags are always boolean, would be to use an enum as key and have 1 functoin for reading, 1 functoin for storing.

libmamba/include/mamba/core/solution.hpp
```diff
@@ -0,0 +1,202 @@
...
+
+        Solution() = default;
+        Solution(const action_list& actions);
+        Solution(action_list&& actions);
+
+        [[nodiscard]] auto empty() const -> bool;
+        [[nodiscard]] auto size() const -> size_type;
+
+        [[nodiscard]] auto actions() const -> const action_list&;
+        [[nodiscard]] auto actions() -> action_list&;
```
Klaim, line 61: Why allow modifying the action list?

libmamba/include/mamba/core/solver.hpp
```diff
@@ -25,9 +25,9 @@
 
 #include "match_spec.hpp"
 
-#define MAMBA_NO_DEPS 0b0001
-#define MAMBA_ONLY_DEPS 0b0010
-#define MAMBA_FORCE_REINSTALL 0b0100
+#define PY_MAMBA_NO_DEPS 0b0001
```
Klaim, line 28: I believe these can be replaced by an enum or at least constexpr int.

libmamba/src/core/solution.cpp
```diff
@@ -0,0 +1,40 @@
...
+//
+// Distributed under the terms of the BSD 3-Clause License.
+//
+// The full license is in the file LICENSE, distributed with this software.
+
+#include "mamba/core/solution.hpp"
+
+namespace mamba
+{
+    Solution::Solution(const action_list& actions)
```
Klaim, line 11: Note that if you have only one constructor that takes by value and move the copy in, it's equivalent to the 3 constructors here, as this code cannot be inlined.

libmamba/src/core/solver.cpp
```diff
@@ -241,29 +241,59 @@ namespace mamba
...
+    }
+
+    auto MSolver::libsolv_flags() -> std::vector<std::pair<int, int>>&
+    {
+        return m_libsolv_flags;
+    }
+
+    void MSolver::apply_libsolv_flags()
     {
         // TODO use new API
```
Klaim, line 295: Is this comment still relevant?

libmamba/include/mamba/core/solver.hpp
```diff
@@ -73,12 +83,20 @@ namespace mamba
...
         void add_pin(const std::string& pin);
         void add_pins(const std::vector<std::string>& pins);
-        void set_flags(const std::vector<std::pair<int, int>>& flags);
-        void set_postsolve_flags(const std::vector<std::pair<int, int>>& flags);
+
+        [[deprecated]] void py_set_postsolve_flags(const std::vector<std::pair<int, int>>& flags);
+
+        void set_flags(const Flags& flags);
+        [[nodiscard]] auto flags() const -> const Flags&;
+        [[nodiscard]] auto flags() -> Flags&;
```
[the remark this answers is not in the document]
AntoinePrv: > Why allow writing? There is already a set_flags function.

Maybe the `set_flags` function should go then. Say you want to change only one flag, you have to first copy all flags, change the one you want, and copy them back. Not great as the number of flags grows.

>  Also, why not only returning by value? auto flags() const -> Flags?

Why return by value? If the struct grows we don't want to copy all fields just to maybe get only one.

libmamba/include/mamba/core/solver.hpp
```diff
@@ -73,12 +83,20 @@ namespace mamba
...
+        void set_libsolv_flags(const std::vector<std::pair<int, int>>& flags);
+        [[nodiscard]] auto libsolv_flags() const -> const std::vector<std::pair<int, int>>&;
+        [[nodiscard]] auto libsolv_flags() -> std::vector<std::pair<int, int>>&;
```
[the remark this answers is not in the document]
AntoinePrv: Again, maybe `set_libsolv_flags` should go. What if you want to push_back a single flag at the end?

libmamba/src/core/transaction.cpp
```diff
@@ -197,9 +192,9 @@ namespace mamba
         );
         trans().order(pool);
 
-        if (solver.no_deps || solver.only_deps)
+        if (!solver.flags().keep_dependencies || !solver.flags().keep_specs)
```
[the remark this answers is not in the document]
AntoinePrv: I agree for calling the function once, but not sure about `auto` vs `auto const&`.

Note this is a transitory interface. In the final interface, I would imagine the flags are passed at contruction time and not modified.

libmamba/include/mamba/core/solver.hpp
AntoinePrv: > An alternative, IFF flags are always boolean, would be to use an enum as key and have 1 functoin for reading, 1 functoin for storing.

That would be nice, but I don't want to limit the future of the flags to boolean only.

> For future maintainance and optimization, it might be best to make this expose only functions to read or write the values, as the implementation could change (to bits for example, in particular if we add more values later). Changing the implementation would then be easy without changing the interface.

I would agree in general, but in this case I think there is value in just having a plain, dumb, data only class.
I don't the complexity of storing flags as bits  would ever make sense for such high-level flags.

libmamba/include/mamba/core/solution.hpp
AntoinePrv: In general I would say, there is no invariant the this violates, so why not make it writable too.

In particular, there is currently some logic that needs to appends more entries to the actions, such as adding Python noarch packages for reinstall.
Hopefully I can change mamba behavior for this to be computed before creating this class, but this would be outside this PR.

libmamba/include/mamba/core/solver.hpp
AntoinePrv: These are kept for 1.x compatibility but will be removed.

libmamba/src/core/solver.cpp
AntoinePrv: Probably. The solver flags API will be reworked with the new Solver API.

libmamba/src/core/solution.cpp
AntoinePrv: Potentially one extra move, no?

libmamba/src/core/transaction.cpp
```diff
@@ -197,9 +192,10 @@ namespace mamba
         );
         trans().order(pool);
 
-        if (solver.no_deps || solver.only_deps)
+        const auto& solver_flags = solver.flags();
+        if (!solver_flags.keep_dependencies || !solver_flags.keep_specs)
         {
-            m_filter_type = solver.only_deps ? FilterType::keep_only : FilterType::ignore;
+            m_filter_type = !(solver.flags().keep_specs) ? FilterType::keep_only : FilterType::ignore;
```
Klaim, line 198: Should be `solver_flags.keep_specs`?

libmamba/src/core/transaction.cpp
```diff
@@ -210,7 +206,7 @@ namespace mamba
             }
         }
 
-        if (solver.only_deps)
+        if (!solver.flags().keep_specs)
```
Klaim, line 209: `solver_flags.keep_specs`?

Klaim (approved):

libmamba/src/core/solver.cpp
```diff
@@ -28,11 +28,11 @@
 namespace mamba
 {
     MSolver::MSolver(MPool pool, const std::vector<std::pair<int, int>> flags)
```
JohanMabille, line 30: - No need for `const` for a variable passed by value (and this actually can hits performance, see below).
- Not sure why this compiles since the constructor is declare as:
   ` MSolver(MPool, std::vector<std::pair<int, int>>)`

libmamba/src/core/solver.cpp
```diff
@@ -28,11 +28,11 @@
...
-        : m_flags(std::move(flags))
-        , m_is_solved(false)
+        : m_libsolv_flags(std::move(flags))
```
JohanMabille, line 31: This results in a copy because `flags` is `const` and `std::vector` does not provide a constructor accepting a const rvalue reference.

JohanMabille (changes requested):

libmamba/include/mamba/core/solution.hpp
```diff
@@ -0,0 +1,218 @@
...
+namespace mamba
+{
+    /********************************
+     *  Implementation of Solution  *
+     ********************************/
+
+    namespace detail
+    {
+        template <typename Action>
+        auto to_remove_ptr(Action& action)
```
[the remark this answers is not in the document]
AntoinePrv: Maybe this is not the idiomatic way of doing this, but this was to deduce the constness of Action.

PR #2591, opened 2023-06-13
Add boolean expression tree

AntoinePrv:
This is a flat boolean expression tree with Infix parser, needed to implement `VersionSpec` boolean grammar such as `(>=1.2,<1.3)|(==1.7)`.


JohanMabille (changes requested):

libmamba/include/mamba/util/flat_bool_expr_tree.hpp
```diff
@@ -0,0 +1,738 @@
...
+#include <utility>
+#include <variant>
+#include <vector>
+
+namespace mamba::util
+{
+    /**
+     * In array binary tree.
+     *
+     * A binary tree, each node is either a leaf, or a node with exactly two children.
```
Klaim, line 20: It's not clear to me if this is supposed to be a container like `flat_set` in boost (or the standard one but it's 2 vectors) or more like a logical state thing?

libmamba/include/mamba/util/flat_bool_expr_tree.hpp
```diff
@@ -0,0 +1,738 @@
...
+
+        [[nodiscard]] auto size() const -> size_type;
+        [[nodiscard]] auto empty() const -> bool;
+
+        void clear();
+        void reserve(size_type size);
+
+        auto add_leaf(const leaf_type& leaf) -> idx_type;
+        auto add_leaf(leaf_type&& leaf) -> idx_type;
+        auto add_branch(const branch_type& branch, idx_type left_child, idx_type right_child)
```
Klaim, line 53: Side note C++23: `auto add_branch(this auto&& self, ...)` would have been useful here haha

libmamba/include/mamba/util/flat_bool_expr_tree.hpp
```diff
@@ -0,0 +1,738 @@
...
+    /**
+     * A parser for postfix expressions.
+     *
+     * The parser creates an expression tree and validate that the expression being pushed
+     * is a valid postfix expression.
+     * For example, for the expression ``a + b * c`` on might push ``a b c * +``
+     * or ``b c * a +``.
+     */
+    template <typename Variable, typename Operator>
+    class PostfixParser
```
Klaim, line 89: So this header is not "just" for flat_binary_tree, is it nammed aptly for everything it contains?

libmamba/include/mamba/util/flat_bool_expr_tree.hpp
```diff
@@ -0,0 +1,738 @@
...
+// The full license is in the file LICENSE, distributed with this software.
+
+#ifndef MAMBA_UTIL_FLAT_EXPR_TREE_HPP
+#define MAMBA_UTIL_FLAT_EXPR_TREE_HPP
+
+#include <functional>
+#include <utility>
+#include <variant>
+#include <vector>
+
```
Klaim, line 14: Consider making the types constexpr as much as possibly allowed by the C++ version we are using (in C++23 they look like they could be completely usable at compile-time)

libmamba/include/mamba/util/flat_bool_expr_tree.hpp
```diff
@@ -0,0 +1,738 @@
...
+        using node_type = std::variant<branch_node, leaf_node>;
+        using node_list = std::vector<node_type>;
+        using size_type = typename node_list::size_type;
+        using idx_type = size_type;
+
+        [[nodiscard]] auto size() const -> size_type;
+        [[nodiscard]] auto empty() const -> bool;
+
+        void clear();
+        void reserve(size_type size);
```
Klaim, line 49: Consider documenting at least the pre and post conditions of the important functions in these types. For example, does reserve have some guarantees? Does add_leaf necessarilly allocates (I´d say yes if the node_list is not customizable) etc. And invariants.

I know we lack time so just the ones that are not that obvious would be helpful for future maintenance.

libmamba/include/mamba/util/flat_bool_expr_tree.hpp
```diff
@@ -0,0 +1,738 @@
...
+        using tree_type = flat_binary_tree<operator_type, variable_type>;
+
+        void push_variable(const variable_type& var);
+        void push_variable(variable_type&& var);
+        void push_operator(const operator_type& op);
+        void push_operator(operator_type&& op);
+        void finalize();
+
+        [[nodiscard]] auto tree() const& -> const tree_type&;
+        [[nodiscard]] auto tree() && -> tree_type&&;
```
Klaim, line 104: Im not completely sure if it's ok to keep that same name here, as it could lead to unexpectedly extract the tree from theparser and making it unusable silently. Did you consider renaming this function something like `extract_tree()` to clarify?

libmamba/include/mamba/util/flat_bool_expr_tree.hpp
```diff
@@ -0,0 +1,738 @@
...
+        void push_right_parenthesis();
+        void finalize();
+
+        [[nodiscard]] auto tree() const& -> const tree_type&;
+        [[nodiscard]] auto tree() && -> tree_type&&;
+
+    private:
+
+        using postfix_parser_type = PostfixParser<variable_type, operator_type>;
+        struct LeftParenthesis
```
Klaim, line 160: does the style forces you multiple lines here? to me that would be better as a one-liner
but if the style is like that dont bother changing that.

libmamba/include/mamba/util/flat_bool_expr_tree.hpp
```diff
@@ -0,0 +1,738 @@
...
+        auto stack_top_is_op_with_greater_precedence_than(const operator_type&) const -> bool;
+
+        template <typename V>
+        void push_variable_impl(V&& var);
+        template <typename O>
+        void push_operator_impl(O&& op);
+    };
+
+    // TODO C++20 this is std::identity
+    struct identity
```
Klaim, line 187: I thought we had something like that already somewhere... maybe move that in one of the utility headers?

libmamba/include/mamba/util/flat_bool_expr_tree.hpp
```diff
@@ -0,0 +1,738 @@
...
+    {
+        template <typename T>
+        constexpr auto operator()(T&& t) const noexcept -> T&&;
+    };
+
+    enum struct BoolOperator
```
Klaim, line 193: Why `struct` and not `class`? It doesnt seem to change a thing right?

libmamba/include/mamba/util/flat_bool_expr_tree.hpp
```diff
@@ -0,0 +1,738 @@
...
+    {
+    public:
+
+        using self_type = flat_bool_expr_tree<Variable>;
+        using operator_type = BoolOperator;
+        using variable_type = Variable;
+        using tree_type = flat_binary_tree<operator_type, variable_type>;
+        using size_type = typename tree_type::size_type;
+
+        flat_bool_expr_tree() = default;
```
Klaim, line 210: What about copy and move? I feel like they should be set explicitly as default, as this seems to be a normal value-type to me.

libmamba/include/mamba/util/flat_bool_expr_tree.hpp
```diff
@@ -0,0 +1,738 @@
...
+        auto evaluate_impl(UnaryFunc& var_evaluator, idx_type idx) const -> bool;
+
+        tree_type m_tree = {};
+    };
+}
+
+#include <cassert>
+#include <iterator>
+#include <stdexcept>
+#include <type_traits>
```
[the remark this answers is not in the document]
Klaim: Also, the meaning of the files might have changed because of the code preceding these includes. They really need to be at the top. 

Note for the future: the only way to prevent code preceding an include (inline or from an other include) is to use modules, they completely remove the issue. They just are not ready for production for cross-platform projects.

libmamba/include/mamba/util/flat_bool_expr_tree.hpp
```diff
@@ -0,0 +1,738 @@
...
+    }
+
+    template <typename V, typename O, typename C>
+    template <typename Op>
+    void InfixParser<V, O, C>::push_operator_impl(Op&& op)
+    {
+        // Input check
+        if (!m_expects_op)
+        {
+            throw std::invalid_argument("Unexpected operator");
```
Klaim, line 570: Consider adding the operator in the error report.

libmamba/include/mamba/util/flat_bool_expr_tree.hpp
```diff
@@ -0,0 +1,738 @@
...
+    }
+
+    template <typename V, typename O, typename C>
+    template <typename Var>
+    void InfixParser<V, O, C>::push_variable_impl(Var&& var)
+    {
+        // Input check
+        if (m_expects_op)
+        {
+            throw std::invalid_argument("Unexpected variable");
```
Klaim, line 544: Consider adding the variable in the error report.

libmamba/include/mamba/util/flat_bool_expr_tree.hpp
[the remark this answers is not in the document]
AntoinePrv: Ok, there are some other files to fix in libmamba then :sweat_smile:

libmamba/include/mamba/util/flat_bool_expr_tree.hpp
AntoinePrv: We already have a flat_set.
This is a binary tree meant to be used as a binary tree, *i.e.* not a tree in the sense *the implementation of a set-like interface based on a tree*. The goal is to query who is a child of whom, not make element unique (they don't have to).

It is flat in the sense that the nodes are all allocated in a vector instead of individually.

libmamba/include/mamba/util/flat_bool_expr_tree.hpp
AntoinePrv: In a sense, the goal of a Parser is a fat/stateful constructor for the tree.
The parser only goal is to create a tree, and tree are difficult to create without a parser.

Not sure what name would be best.

libmamba/include/mamba/util/flat_bool_expr_tree.hpp
AntoinePrv: The type are based on `std::vector`, which is not `constexpr` constructible until C++20.

Is there a point to make `constexpr` methods of a not `constexpr` constructible object?

libmamba/include/mamba/util/flat_bool_expr_tree.hpp
AntoinePrv: I agree, but yes clang-format does it that way. Cannot find an option to configure it...

libmamba/include/mamba/util/flat_bool_expr_tree.hpp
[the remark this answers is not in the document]
Klaim: I didnt spot them but there are then I agree 😬

libmamba/include/mamba/util/flat_bool_expr_tree.hpp
Klaim: Ok that looks useful indeed!

libmamba/include/mamba/util/flat_bool_expr_tree.hpp
AntoinePrv: Copy and move are defaulted. All these classes have normal value semantic, would you explicitly mark them `= default` in all these classes.

libmamba/include/mamba/util/flat_bool_expr_tree.hpp
AntoinePrv: I would like to as well, but `Operator` is an unconstrained template type parameter. Should I expect a `operator<<` or `fmt::format` specialization?

libmamba/include/mamba/util/flat_bool_expr_tree.hpp
AntoinePrv: Same

JohanMabille (approved):

libmamba/include/mamba/util/flat_bool_expr_tree.hpp
AntoinePrv: IMHO this is the same style as rvalue overloads as `std::optional<T>::value` `std::get`...

Extracting the tree is really the point of the parser anyways.

libmamba/include/mamba/util/flat_bool_expr_tree.hpp
Klaim: >  would you explicitly mark them = default in all these classes.

Yes so that the intention is clear for future people who will modify that code: once you start adding one special function you better be explicit on the rest, otherwise just adding members might change the behavior/categories.

PR #2594, opened 2023-06-14
Added filtering iterators

JohanMabille:
This PR adds filtering iterators (and bidirectional  + random_access iterator base classes), a mechanical tool convenient for the refactoring of the downloader.

### DIFF
--- a/libmamba/include/mamba/util/iterator.hpp
+++ b/libmamba/include/mamba/util/iterator.hpp
@@ -1,0 +1,427 @@
+// Copyright (c) 2019, QuantStack and Mamba Contributors
+//
+// Distributed under the terms of the BSD 3-Clause License.
+//
+// The full license is in the file LICENSE, distributed with this software.
+
+#ifndef MAMBA_UTIL_ITERATOR_HPP
+#define MAMBA_UTIL_ITERATOR_HPP
+
+#include <cstddef>
+#include <iterator>
+#include <optional>
+#include <type_traits>
+
+namespace mamba::util
+{
+    /********************************
+     * class xforward_iterator_base *
+     ********************************/
+
+    template <class I, class T, class D = std::ptrdiff_t, class P = T*, class R = T&>
+    class xforward_iterator_base
+    {
+    public:
+
+        using derived_type = I;
+        using value_type = T;
+        using reference = R;
+        using pointer = P;
+        using difference_type = D;
+        using iterator_category = std::forward_iterator_tag;
+
+        friend derived_type operator++(derived_type& d, int)
+        {
+            derived_type tmp(d);
+            ++d;
+            return tmp;
+        }
+
+        friend bool operator!=(const derived_type& lhs, const derived_type& rhs)
+        {
+            return !(lhs == rhs);
+        }
+    };
+
+    template <class Iterator, class Traits>
+    using xforward_iterator_base_from_traits = xforward_iterator_base<
+        Iterator,
+        typename Traits::value_type,
+        typename Traits::difference_type,
+        typename Traits::pointer,
+        typename Traits::reference>;
+
+    /**************************************
+     * class xbidirectional_iterator_base *
+     **************************************/
+
+    template <class I, class T, class D = std::ptrdiff_t, class P = T*, class R = T&>
+    class xbidirectional_iterator_base : public xforward_iterator_base<I, T, D, P, R>
+    {
+    public:
+
+        using derived_type = I;
+        using value_type = T;
+        using reference = R;
+        using pointer = P;
+        using difference_type = D;
+        using iterator_category = std::bidirectional_iterator_tag;
+
+
+        friend derived_type operator--(derived_type& d, int)
+        {
+            derived_type tmp(d);
+            --d;
+            return tmp;
+        }
+    };
+
+    template <class Iterator, class Traits>
+    using xbidirectional_iterator_base_from_traits = xbidirectional_iterator_base<
+        Iterator,
+        typename Traits::value_type,
+        typename Traits::difference_type,
+        typename Traits::pointer,
+        typename Traits::reference>;
+
+    /********************************
+     * xrandom_access_iterator_base *
+     ********************************/
+
+    template <class I, class T, class D = std::ptrdiff_t, class P = T*, class R = T&>
+    class xrandom_access_iterator_base : public xbidirectional_iterator_base<I, T, D, P, R>
+    {
+    public:
+
+        using derived_type = I;
+        using value_type = T;
+        using reference = R;
+        using pointer = P;
+        using difference_type = D;
+        using iterator_category = std::random_access_iterator_tag;
+
+        reference operator[](difference_type n) const
+        {
+            return *(*static_cast<const derived_type*>(this) + n);
+        }
+
+        friend derived_type operator+(const derived_type& it, difference_type n)
+        {
+            derived_type tmp(it);
+            return tmp += n;
+        }
+
+        friend derived_type operator+(difference_type n, const derived_type& it)
+        {
+            derived_type tmp(it);
+            return tmp += n;
+        }
+
+        friend derived_type operator-(const derived_type& it, difference_type n)
+        {
+            derived_type tmp(it);
+            return tmp -= n;
+        }
+
+        friend bool operator<=(const derived_type& lhs, const derived_type& rhs)
+        {
+            return !(rhs < lhs);
+        }
+
+        friend bool operator>=(const derived_type& lhs, const derived_type& rhs)
+        {
+            return !(lhs < rhs);
+        }
+
+        friend bool operator>(const derived_type& lhs, const derived_type& rhs)
+        {
+            return rhs < lhs;
+        }
+    };
+
+    template <class Iterator, class Traits>
+    using xrandom_access_iterator_base_from_traits = xrandom_access_iterator_base<
+        Iterator,
+        typename Traits::value_type,
+        typename Traits::difference_type,
+        typename Traits::pointer,
+        typename Traits::reference>;
+
+    /************************
+     * select_iterator_base *
+     ************************/
+
+    namespace detail
+    {
+        template <class Iterator>
+        using iterator_category_t = typename std::iterator_traits<Iterator>::iterator_category;
+
+        template <class Iterator>
+        using is_random_access_iterator = std::
+            is_same<iterator_category_t<Iterator>, std::random_access_iterator_tag>;
+
+        template <class Iterator>
+        inline constexpr bool is_random_access_iterator_v = is_random_access_iterator<Iterator>::value;
+
+        template <class Iterator>
+        using is_bidirectional_iterator = std::disjunction<
+            std::is_same<iterator_category_t<Iterator>, std::bidirectional_iterator_tag>,
+            is_random_access_iterator<Iterator>>;
+
+        template <class Iterator>
+        inline constexpr bool is_bidirectional_iterator_v = is_bidirectional_iterator<Iterator>::value;
+    }
+
+    template <class Iterator>
+    struct select_iterator_base
+    {
+        template <class I, class T, class D = std::ptrdiff_t, class P = T*, class R = T&>
+        using type = std::conditional_t<
+            detail::is_random_access_iterator_v<Iterator>,
+            xrandom_access_iterator_base<I, T, D, P, R>,
+            std::conditional_t<
+                detail::is_bidirectional_iterator_v<Iterator>,
+                xbidirectional_iterator_base<I, T, D, P, R>,
+                xforward_iterator_base<I, T, D, P, R>>>;
+
+        template <class I, class T>
+        using from_traits = std::conditional_t<
+            detail::is_random_access_iterator_v<Iterator>,
+            xrandom_access_iterator_base_from_traits<I, T>,
+            std::conditional_t<
+                detail::is_bidirectional_iterator_v<Iterator>,
+                xbidirectional_iterator_base_from_traits<I, T>,
+                xforward_iterator_base_from_traits<I, T>>>;
+    };
+
+    template <class Iterator, class R = void>
+    using enable_bidirectional_iterator = std::enable_if_t<detail::is_bidirectional_iterator_v<Iterator>, R>;
+
+    template <class Iterator, class R = void>
+    using enable_random_access_iterator = std::enable_if_t<detail::is_random_access_iterator_v<Iterator>, R>;
+
+    /*******************
+     * filter_iterator *
+     *******************/
+
+    template <class Predicate, class Iterator>
+    class filter_iterator;
+
+    template <class Predicate, class Iterator>
+    struct select_filter_iterator_base
+    {
+        using type = typename select_iterator_base<Iterator>::
+            template from_traits<filter_iterator<Predicate, Iterator>, std::iterator_traits<Iterator>>;
+    };
+
+    template <class Predicate, class Iterator>
+    using select_filter_iterator_base_t = typename select_filter_iterator_base<Predicate, Iterator>::type;
+
+
+    template <class Predicate, class Iterator>
+    class filter_iterator : public select_filter_iterator_base_t<Predicate, Iterator>
+    {
+    public:
+
+        using self_type = filter_iterator<Predicate, Iterator>;
+        using base_type = select_filter_iterator_base_t<Predicate, Iterator>;
+
+        using reference = typename base_type::reference;
+        using pointer = typename base_type::pointer;
+        using difference_type = typename base_type::difference_type;
+        using iterator_category = typename base_type::iterator_category;
+
+        filter_iterator() = default;
+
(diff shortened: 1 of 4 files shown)


libmamba/include/mamba/util/iterator.hpp
```diff
@@ -0,0 +1,400 @@
...
+        template <class It = Iterator>
+        enable_random_access_iterator<It, bool>
+        less_than(const self_type& rhs) const
+        {
+            return m_pred == rhs.m_pred && m_iter < rhs.m_iter;
+        }
+
+    private:
+
+        void advance(difference_type n)
```
Klaim, line 289: Why not use `std::advance` instead?
also, it's the same code with both directions so I would gather step_back and advance together, the only difference is the direction, which is already given by `n`

libmamba/include/mamba/util/iterator.hpp
```diff
@@ -0,0 +1,400 @@
...
+        }
+
+        pointer operator->() const
+        {
+            return m_iter.operator->();
+        }
+
+        bool equal(const self_type& rhs) const
+        {
+            return m_pred == rhs.m_pred && m_iter == rhs.m_iter;
```
Klaim, line 277: This means that the predicate must be comparable, I'm not sure that's a good idea, it contstrains a lot what you can use as a predicate?

libmamba/include/mamba/util/iterator.hpp
```diff
@@ -0,0 +1,400 @@
...
+            It tmp = rhs;
+            difference_type res = 0;
+            while (tmp++ != *this)
+            {
+                ++res;
+            }
+            return res;
+        }
+
+        reference operator*() const
```
Klaim, line 265: I'm not sure (not writing iterators often) but shouldnt this be provided as both const and non-const? With the const version providing a const reference?

libmamba/include/mamba/util/iterator.hpp
```diff
@@ -0,0 +1,400 @@
...
+                step_back(n);
+            }
+            return *this;
+        }
+
+        template <class It = Iterator>
+        enable_random_access_iterator<It, self_type>
+        operator-=(difference_type n)
+        {
+            if (n >= 0)
```
Klaim, line 241: As mentionned below, this check of direction should be done in `advance` which should handle both directions, also matchnig how `std::advance` does.

libmamba/include/mamba/util/iterator.hpp
```diff
@@ -0,0 +1,400 @@
...
+
+        using self_type = filter_iterator<Predicate, Iterator>;
+        using base_type = select_filter_iterator_base_t<Predicate, Iterator>;
+
+        using reference = typename base_type::reference;
+        using pointer = typename base_type::pointer;
+        using difference_type = typename base_type::difference_type;
+        using iterator_category = typename base_type::iterator_category;
+
+        filter_iterator() = default;
```
Klaim, line 184: Is this iterator usable in any way when initialized by default? Other than through re-assigning it?

libmamba/include/mamba/util/iterator.hpp
```diff
@@ -0,0 +1,400 @@
...
+            }
+            return res;
+        }
+
+        reference operator*() const
+        {
+            return *m_iter;
+        }
+
+        pointer operator->() const
```
Klaim, line 270: Same quesiton as operator*

libmamba/include/mamba/util/iterator.hpp
```diff
@@ -0,0 +1,400 @@
...
+        }
+
+        bool equal(const self_type& rhs) const
+        {
+            return m_pred == rhs.m_pred && m_iter == rhs.m_iter;
+        }
+
+        template <class It = Iterator>
+        enable_random_access_iterator<It, bool>
+        less_than(const self_type& rhs) const
```
Klaim, line 282: `less_than` and `equal` should be private, they are implementtion details for the operators (also why even have them named? shouldnt they be called operator< and operator==?)

libmamba/include/mamba/util/iterator.hpp
```diff
@@ -0,0 +1,400 @@
...
+    enable_random_access_iterator<Iterator, bool>
+    operator<(const filter_iterator<Predicate, Iterator>& lhs,
+              const filter_iterator<Predicate, Iterator>& rhs)
+    {
+        return lhs.less_than(rhs);
+    }
+
+    template <class Predicate, class Iterator>
+    filter_iterator<std::decay_t<Predicate>, Iterator>
+    make_filter_iterator(Predicate&& pred, Iterator iter, Iterator begin, Iterator end)
```
Klaim, line 336: In both the constructor and here, I dont understand what the first iterator is? The current position?

Klaim: Also note that in C++17 there is [CTAD](https://en.cppreference.com/w/cpp/language/class_template_argument_deduction) making this kind of `make_*` functions obsolete. You might want to delete this function.

libmamba/include/mamba/util/iterator.hpp
```diff
@@ -0,0 +1,400 @@
...
+        return return_type(
+            std::forward<Predicate>(pred),
+            iter,
+            begin,
+            end);
+    }
+
+    template <class Predicate, class C>
+    filter_iterator<std::decay_t<Predicate>, typename C::iterator>
+    filter_begin(Predicate&& pred, C& container)
```
Klaim, line 348: This header needs more comments :) mainly descriptions of functions which are not mandated by the standard ,like this wone.
Here I would also like you to add some clarification/constraints/pre-conditions like C must be a Range (which kind?), and Predicate must be a callable such as <insert call expression here with return type> is valid. It helps and also will help with expressing it with concepts.

libmamba/include/mamba/util/iterator.hpp
```diff
@@ -0,0 +1,400 @@
...
+    {
+        using return_type = filter_iterator<std::decay_t<Predicate>, typename C::iterator>;
+        return return_type(
```
Klaim, line 351: You can actually remove the `return return_type(...);` expression and replace them by just
`return {...};`
This will also make sure that conversions are explicit, not implicit.

libmamba/include/mamba/util/iterator.hpp
```diff
@@ -0,0 +1,400 @@
...
+            std::forward<Predicate>(pred),
+            container.begin(),
+            container.begin(),
+            container.end()
+        );
+    }
+
+    template <class Predicate, class C>
+    filter_iterator<std::decay_t<Predicate>, typename C::iterator>
+    filter_end(Predicate&& pred, C& container)
```
Klaim, line 361: I'm not totally sure what's the need for `filter_begin` and `filter_end`. 
I would just write a class for example `filtered<C>` where C is a range. Then the constructor would take a range and a predicate and expose a range interface. That way users can easilly obtain the begin and end filtered, but lazilly (dont go through the range if you dont need the end immediately).
That way (and benefiting from CTAD) users can just use `filtered{ mycontainer, some_predicate}` and use that as a range too.

libmamba/tests/src/util/test_iterator.cpp
```diff
@@ -0,0 +1,198 @@
...
+    {
+        return i > 10;
+    }
+
+    bool second_greater_than_10(const std::pair<std::string, int>& p)
+    {
+        return p.second > 10;
+    }
+
+    auto get_test_data()
```
Klaim, line 31: Why make a tuple of these? Why not just set them as constant data? They can be easilly copied in the tests suites. Or wrapped in a srtuct, to not have to re-declare names in tests, just declar an object of that struct type and you're good to go.

libmamba/tests/src/util/test_iterator.cpp
```diff
@@ -0,0 +1,198 @@
...
+        return std::make_tuple(input_sequence, result_sequence, input_map, result_map);
+
+    }
+}
+
+TEST_SUITE("filter_iterator")
+{
+    TEST_CASE("forward_iterator_api")
+    {
+        auto [input_seq, res_seq, input_map, res_map] = get_test_data();
```
Klaim, line 64: `const` ?

libmamba/tests/src/util/test_iterator.cpp
```diff
@@ -0,0 +1,198 @@
...
+            auto citer = cfilter_begin(greater_than_10, input_seq);
+            auto citer_end = cfilter_end(greater_than_10, input_seq);
+
+            auto res_iter = res_seq.end();
+            
+            auto iter2 = iter_end, iter3 = iter_end;
+
+            for (size_t i = 0; i < res_seq.size(); ++i)
+            {
+                --iter_end, --citer_end, --res_iter;
```
Klaim, line 131: 😬 
please prefer 1 line per operation, it's not a problem if it takes vertical space, it's just easier to read/reason/not-misread

libmamba/tests/src/util/test_iterator.cpp
```diff
@@ -0,0 +1,198 @@
...
+                CHECK_EQ(*iter_end, *res_iter);
+                CHECK_EQ(*iter_end, *citer_end);
+            }
+            CHECK_EQ(iter, iter_end);
+            CHECK_EQ(citer, citer_end);
+            --iter_end, --citer_end;
+            CHECK_EQ(iter_end.operator->(), (--input_seq.begin()).operator->());
+
+            CHECK(iter2-- != --iter3);
```
Klaim, line 140: Is order of invocation important here? The order of invocation have changed in C++17 if my memory is correct, it is now guaranteed that the left side is executed before the right side. To clarify: this was not guaranteed in C++14 and before. 
It's also cryptic to read, so I would prefer with the 2 operations separated, as we're in tests it's ok if we name temporaries for clarity.

libmamba/tests/src/util/test_iterator.cpp
```diff
@@ -0,0 +1,198 @@
...
+            CHECK_EQ(iter2->second, res_map.begin()->second);
+            CHECK(iter2++ != ++iter3);
+        }
+    }
+
+    TEST_CASE("bidirectional_iterator_api")
+    {
+        auto [input_seq, res_seq, input_map, res_map] = get_test_data();
+        {
+            auto iter = filter_begin(greater_than_10, input_seq);
```
Klaim, line 119: `const`. 


`const` everywhere! 🤡

libmamba/tests/src/util/test_iterator.cpp
```diff
@@ -0,0 +1,198 @@
...
+    bool second_greater_than_10(const std::pair<std::string, int>& p)
+    {
+        return p.second > 10;
+    }
+
+    auto get_test_data()
+    {
+        std::vector<int> input_sequence = { 1, 12, 2, 3, 14, 4, 18, 20, 4 };
+        std::vector<int> result_sequence = { 12, 14, 18, 20 };
+        std::map<std::string, int> input_map =
```
Klaim, line 35: Except if you need the values to be stable in memory, prefer unordered_map by default, or flat_map when there is such low number of small (in memory) values. Here it will help with test performance.

libmamba/include/mamba/util/iterator.hpp
```diff
@@ -0,0 +1,400 @@
...
+
+        using derived_type = I;
+        using value_type = T;
+        using reference = R;
+        using pointer = P;
+        using difference_type = D;
+        using iterator_category = std::bidirectional_iterator_tag;
+
+        inline friend derived_type operator++(derived_type& d, int)
+        {
```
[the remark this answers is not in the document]
JohanMabille: `operator++()` has to be defined in the inheriting class, it is used for implementing other methods in this CRTP base class to avoid the same boilerplate code in every "final" iterator.

libmamba/include/mamba/util/iterator.hpp
```diff
@@ -0,0 +1,400 @@
...
+
+        void step_back(difference_type n)
+        {
+            while (m_iter != m_begin_limit && n != 0)
+            {
+                --(*this), --n;
+            }
+        }
+
+        void satisfy_predicate()
```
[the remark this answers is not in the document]
JohanMabille: Agreed, I will change it.

libmamba/include/mamba/util/iterator.hpp
JohanMabille: Because `std::advance` implementation calls `operator+=` for random access iterators, leading tooooo:
Stack bomb, stack bomb, yoooooou're my stack bomb:D!

libmamba/include/mamba/util/iterator.hpp
JohanMabille: `operator*` nevers changes the state of the iterator itself, so it should always be const. The `reference` type will actually be a `const reference` (in the sense const value_type&) for const_iterator only.

libmamba/include/mamba/util/iterator.hpp
JohanMabille: It is not, but iterators must be default constructible according to the standard.

libmamba/include/mamba/util/iterator.hpp
JohanMabille: I usually prefer to define operators outside of the class when possible (especially binary operators), these functions just avoid having to declare these operators as friends.

libmamba/include/mamba/util/iterator.hpp
JohanMabille: Indeed, the first iterator is the current position, I can give it a more meaningful name. And indeed for the CTAD, I just don't have the habit to use it yet.

libmamba/include/mamba/util/iterator.hpp
JohanMabille: Will do!

libmamba/tests/src/util/test_iterator.cpp
JohanMabille: sorry :sweat_smile:

libmamba/tests/src/util/test_iterator.cpp
JohanMabille: I want to test non const iterators, for that I need non const sequences.

libmamba/include/mamba/util/iterator.hpp
JohanMabille: The idea behind `filter_begin` and `filter_end` is to be able to write one liner like `std::copy(filter_begin(pred, container), filter_end(pred, conainer), dst.begin())`.

I agree the repetition is not ideal and can be avoided with the solution you propose, but this requires to instantiate the filter before using it.

libmamba/tests/src/util/test_iterator.cpp
JohanMabille: Indeed, I started with a struct and then changed for this and I cannot remember why :sweat_smile: Let's go back to the struct.

libmamba/tests/src/util/test_iterator.cpp
JohanMabille: The order order of invocation is not important, the idea is to test that `operator--` and `operator--(int)` give different results.

libmamba/tests/src/util/test_iterator.cpp
JohanMabille: Not sure to get it, `iter` is changed after, and I want to test non const iterators so the sequances / maps cannot be const.

libmamba/tests/src/util/test_iterator.cpp
JohanMabille: `flat_map` is C++23 only, and `unordered_map` provide forward iterators, not bidirectional ones. That said, I should probably improve the `filter_iterator` so that it can handle forward iterators too.

libmamba/include/mamba/util/iterator.hpp
JohanMabille: Indeed, I'll remove this comparison.

libmamba/include/mamba/util/iterator.hpp
```diff
@@ -0,0 +1,400 @@
...
+        {
+            ++m_iter;
+            satisfy_predicate();
+            return *this;
+        }
+
+        self_type& operator--()
+        {
+            --m_iter;
+            while (m_iter != m_begin_limit && !m_pred(*m_iter))
```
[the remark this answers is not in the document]
JohanMabille: The `satisfy_predicate` is already also used in the constructors. Making it a function that accepts a lambda will leads to duplicate the lambda implementation many times.

libmamba/tests/src/util/test_iterator.cpp
```diff
@@ -0,0 +1,162 @@
...
+
+#include <doctest/doctest.h>
+
+#include "mamba/util/iterator.hpp"
+
+using namespace mamba::util;
+
+namespace
+{
+    bool greater_than_10(int i)
```
Klaim, line 22: You might also want to test with ca callable object, both cases must work.

Klaim (approved):

PR #2603, opened 2023-06-19
Replace MTransaction::m_remove with Solution

AntoinePrv:
- Based on #2584
- Refactoring to use `Solution` in `MTransaction` instead of libsolv `Tansaction*`.

### DIFF
--- a/libmamba/include/mamba/core/transaction.hpp
+++ b/libmamba/include/mamba/core/transaction.hpp
@@ -58,7 +58,7 @@ namespace mamba
         void print();
         bool execute(PrefixData& prefix);
 
-        std::pair<std::string, std::string> find_python_version();
+        [[deprecated]] std::pair<std::string, std::string> py_find_python_version() const;
 
     private:
 
--- a/libmamba/include/mamba/core/transaction.hpp
+++ b/libmamba/include/mamba/core/transaction.hpp
@@ -26,7 +26,6 @@
 
 namespace mamba::solv
 {
-    class ObjTransaction;
     class ObjSolvableViewConst;
 }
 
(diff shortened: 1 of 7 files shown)


libmamba/include/mamba/core/transaction.hpp
```diff
@@ -55,12 +42,8 @@ namespace mamba
         // Only use if the packages have been solved previously already.
         MTransaction(MPool& pool, const std::vector<PackageInfo>& packages, MultiPackageCache& caches);
 
-        ~MTransaction();
-
         MTransaction(const MTransaction&) = delete;
         MTransaction& operator=(const MTransaction&) = delete;
-        MTransaction(MTransaction&&) = delete;
-        MTransaction& operator=(MTransaction&&) = delete;
```
JohanMabille, line 63: This class should have a clear semantics:
- either everything can de defaulted, and we can remove all these methods.
- or we keep the declaration of the destructor and move semantics methods and we default them

libmamba/src/core/transaction.cpp
```diff
@@ -82,6 +84,185 @@ namespace mamba
             }
             return out;
         }
+
+        auto specs_names(const MSolver& solver)
```
JohanMabille, line 88: Would be nice to specify the return type

libmamba/src/core/transaction.cpp
```diff
@@ -141,19 +322,21 @@ namespace mamba
...
-        init();
 
+        m_solution = transaction_to_solution(m_pool, trans);
+
+        m_history_entry.remove.reserve(specs_to_remove.size());
         for (auto& s : specs_to_remove)
         {
             m_history_entry.remove.push_back(s.str());
         }
+        m_history_entry.update.reserve(specs_to_remove.size());
```
JohanMabille, line 339: I guess it should reserve `specs_to_install.size()` instead?

libmamba/src/core/transaction.cpp
```diff
@@ -1060,11 +1109,9 @@ namespace mamba
             remove
         };
         auto format_row =
-            [this,
-             &ctx,
-             &total_size](rows& r, solv::ObjSolvableViewConst s, Status status, std::string diff)
+            [this, &ctx, &total_size](rows& r, const PackageInfo& s, Status status, std::string diff)
         {
-            std::size_t const dlsize = s.size();
+            std::size_t const dlsize = s.size;
```
JohanMabille, line 1114: `const size_t` ;)

JohanMabille (approved):

PR #2611, opened 2023-06-21
Wrap curl functions in `url` files

Hind-M:


libmamba/src/core/curl.cpp
```diff
@@ -628,4 +669,88 @@ namespace mamba
...
+        {
+            throw std::runtime_error(
+                "Could not set URL (code: " + std::to_string(uc) + " - url = " + url + ")"
+            );
+        }
+    }
+
+    namespace
+    {
+        std::map<std::string, CURLUPart> to_curlu_part{ { "CURLUPART_URL", CURLUPART_URL },
```
wolfv, line 713: I am not sure I get the motivation for this change. Why do we need this mapping from string to integers? Didn't we have a decent URL API already?

AntoinePrv: Since this is not meant to change (apart through development), this could also be a `constexpr` function like the following. Slightly more annoying to write, but it has no global state and can be inlined / optimized away. 
```cpp
constexpr auto to_curlu_part(std::string_view name) {
    if(name == "CURLUPART_URL"){
        return CURLUPART_URL;
    } else if (...) {
       ...
   }
}
```

wolfv: But we already have encapsulation with the `get_scheme` `get_path`, ... functions?
And if we go for strings, why keep the names?

But in general, using the functions should be much preferred IMO, and then the get_part function doesn't need to be exposed at all.

Hind-M: > And if we go for strings, why keep the names?

We keep them to be explicit.

> But in general, using the functions should be much preferred IMO, and then the get_part function doesn't need to be exposed at all.

Yes indeed, so the exposure here is not the problem, this mapping is more intended to avoid including specific curl header  (`curl/urlapi.h`) in url files.

Hind-M: Any thoughts on this mapping? @JohanMabille @Klaim

wolfv: if it's really necessary, why not make an enum?

PR #2615, opened 2023-06-22
Context: not a singleton

Klaim:
Notes:
- I'll try to keep the PR buildable, but we need to remove the commit marked to be removed before merging.
- for now this is not finished, it's in an half-state where it builds but Context can be biult several time and one instance is still available as a singleton - therefore we cannot merge that as-is, yet.
- I'll try to keep 1 commit per system de-singleton-usagified to help with reviewing.


Klaim:
Also note that this is not enough to make easier the use of libmamba as a library, that is, you can't have multiple context working correctly right now because of the code that makes Context the handler of the logging system. But this can be fixed in another PR, once we review the output system changes.

libmamba/src/api/c_api.cpp
```diff
@@ -16,12 +16,26 @@
...
+{
+    return new mamba::Context;
+}
+
+void
+mamba_delete_context(mamba::Context* context)
+{
+    delete context;
+}
+
```
Hind-M, line 33: Why are we allocating pointers here? Maybe I missed something but I didn't find any call of these allocaters/deleters. Are we giving the responsibility of deleting to the user or is it controlled somehow? Why aren't we using smart pointers instead?

libmamba/include/mamba/api/configuration.hpp
```diff
@@ -644,6 +656,8 @@ namespace mamba
         template <class T>
         void ConfigurableImpl<T>::compute(int options, const ConfigurationLevel& level)
         {
+            assert(m_config);  // REVIEW: should this be a if & throw?
```
Hind-M, line 659: I guess it's safer yes (at least until all this refactoring is stabilized?).

libmamba/src/api/info.cpp
```diff
@@ -86,7 +89,8 @@ namespace mamba
 
         void print_info(ChannelContext& channel_context, const Configuration& config)
         {
-            auto& ctx = Context::instance();
+            assert(&channel_context.context() == &config.context());
```
Hind-M, line 92: I think it's safer to just replace all the asserts checking all former singletons with if & throw...

AntoinePrv (commented):
If I understand correctly, we have `TransationContext` and `ChannelContext` capturing a `Context*`.
Shouldn't the `ChannelContext` live withing the `Context`? Or both independent?

libmamba/src/api/c_api.cpp
Klaim: This is the C API so there is no smart pointers, only pointers 🥲 
Note that I dont know if the C api is used, but I needed to reflect  the change to make Context usable there.
If we find that we dont need to provide this API, I vote for deleting `c_api.hpp` and `c_api.cpp`

libmamba/src/api/info.cpp
Klaim: This assertion should only fail if we made a mistake in the implementation of the library, user code cannot make this fail (from what I searched) so I prefer an assertion in these kind of situations (project's devs error, not user error). The issue is that debug builds are not tested if my memory is correct.

libmamba/src/api/c_api.cpp
Hind-M: Ohh I didn't realize there is one 😮
Yeah I also vote for removing these... ^^

libmamba/src/api/info.cpp
Hind-M: Yeah I totally agree (I suggested replacing the asserts because we never test debug mode... but yeah maybe we just should).

Klaim:
> If I understand correctly, we have `TransationContext` and `ChannelContext` capturing a `Context*`. Shouldn't the `ChannelContext` live withing the `Context`? Or both independent?

`ChannelContex` uses `Context` so it cannot be in the `Context` unless we manage to remove that dependency.
Note that many classes indeed could be stored in the `Context`, including `Configuration`, assuming we reverse the dependnecy.

Now, either we do that, OR we provide a convenience-type which gather all the things together in a "usual" way, similar to how the `mambapy::Singletons`  struct is setup.

libmamba/include/mamba/api/c_api.h
```diff
@@ -17,7 +18,10 @@ extern "C"
 {
 #endif
 
-    mamba::Configuration* mamba_new_configuration();
+    mamba::Context* mamba_new_context();
```
JohanMabille, line 21: Should expose the `ContextOptions` structure and have this function take it as an argument?

libmamba/include/mamba/core/link.hpp
```diff
@@ -22,10 +23,12 @@
 
 namespace mamba
 {
+    class Context;
```
JohanMabille, line 26: I don't think we need this one here (I don't see any usage of `Context` in this header)

libmamba/include/mamba/core/url.hpp
```diff
@@ -41,6 +42,16 @@ extern "C"
...
 namespace mamba
 {
+    namespace conda_urls
+    {
+        // usernames on anaconda.org can have a underscore, which influences the
+        // first two characters
+        inline const std::regex token_regex{ "/t/([a-zA-Z0-9-_]{0,2}[a-zA-Z0-9-]*)" };
+        inline const std::regex http_basicauth_regex{ "(://|^)([^\\s]+):([^\\s]+)@" };
+        inline const std::regex scheme_regex{ "[a-z][a-z0-9]{0,11}://" };
+    }
```
JohanMabille, line 52: Should be move these definitions to the cpp file?

libmamba/include/mamba/core/util.hpp
```diff
@@ -323,22 +397,35 @@ namespace mamba
...
-        bool dev_mode,
-        bool debug_wrapper_scripts,
-        const std::vector<std::string>& arguments
+        const std::vector<std::string>& arguments,  // TODO: c++20 replace by std::span
+        WrappedCallOptions options = {}
     );
 
-    std::tuple<std::vector<std::string>, std::unique_ptr<TemporaryFile>>
-    prepare_wrapped_call(const fs::u8path& prefix, const std::vector<std::string>& cmd);
+    std::tuple<std::vector<std::string>, std::unique_ptr<TemporaryFile>> prepare_wrapped_call(
```
JohanMabille, line 418: Maybe an alias for this return type could be helpful?

libmamba/src/api/c_api.cpp
JohanMabille: The C API is used by Rhumba, so it cannof be deleted yet. If we want to get rid of it, we must provide R bindings through RCPP, but that's mean an additional library.

libmamba/src/api/info.cpp
JohanMabille: Agreed on the asserts here, let's add a debug build in a dedicated PR once this one is merged.

JohanMabille (commented):
First pass done, I need to make a second one (probably later today).

libmamba/include/mamba/api/c_api.h
Klaim: Yes, I forgot to do that.

BTW did you check the other discussion with Hind where we consider the option to just remove the C API? Or do you know if it is actually used by anyone? I think it would be best provided as a separate library (at least a separate target)

libmamba/include/mamba/core/link.hpp
Klaim: There is probably a lot of these leftovers from when I changed apis one way and then another way. I'll try to do a cleanup. 👍🏽

libmamba/include/mamba/core/url.hpp
Klaim: Unfortunately that would mean potential static initialization fiasco, that's why I opted with `inline const` as it guarantees uniqueness in the binary (although it's possible 2 binaires have their own copy, it should always be the same value).

IFF we had `static constexpr` working for that type then it would be ideal (`constexpr` implies `inline` but it would also guarantee that these objects are built before runtime). Or `constinit` but that's C+20.
Another way to simplify all this would be to use [CTRE](https://compile-time-regular-expressions.readthedocs.io/en/latest/#) instead where we know the regex expression at compile-time, which would be ideal (in all aspects, including speed) but would mean an additional dependency.

libmamba/src/api/c_api.cpp
Klaim: Welp... ok I'll make the necessary fixes.

libmamba/src/api/info.cpp
Klaim: I think I started the work in one of my previous branches, but I hit an issue... anyway ok let's focus on that next.

libmamba/include/mamba/core/url.hpp
AntoinePrv: I'm all for CTRE! `std::regex` has bad rep

JohanMabille: @Klaim not sure I get it, these variables seem to be used in `url.cpp` functions only (although I might have missed some uses), and don't depend on each other, so I don't see how we can run in a static initialization fiasco.

Klaim: They were originally in `context.hpp` so I assumed that users want to have access to these?
If not then ok I'll move them in the cpp.

Klaim: See for details: https://github.com/mamba-org/mamba/pull/2615/files/536d56f2a48960678e5a7e21b392ed4bd29a072c#diff-b13b39da2b425585073993cc5e342e9b973c3863efbbc0fd760790d74cd89831L283-L290

libmamba/src/core/output.cpp
```diff
@@ -267,6 +267,13 @@ namespace mamba
     {
     public:
 
+        ConsoleData(const Context& ctx)
+            : context(ctx)
+        {
+        }
+
+        const Context& context;
```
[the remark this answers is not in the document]
AntoinePrv: There is a Clang-tidy check for that, will look into setting that up!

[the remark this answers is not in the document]
Klaim: @AntoinePrv I think the clang-tidy rule forces the same pattern for all members, be they private or public, whatever the type - if that's the case then we cant use that as it's not the norm everywhere in this code and would break public members in many classes.
If it does just follow the pattern existing in a class, then that would work.

libmamba/include/mamba/core/url.hpp
Klaim: These regex are also used in `util.cpp` so I'll have to keep them that way.

libmambapy/src/main.cpp
```diff
@@ -114,20 +114,48 @@ bind_NamedList(PyClass pyclass)
...
+        }
+        mamba::Context& context()
+        {
+            return m_context;
+        }
+        mamba::Console& console()
+        {
+            return m_console;
+        }
+        mamba::ChannelContext& channel_context()
```
Klaim, line 133: Unfortunately this function needs to be thread-safe to work properly, it is not right now.
Here is one way to do it: https://github.com/mamba-org/mamba/blob/main/libmamba/src/core/singletons.cpp#L160-L167

another way is to use atomic pointers or shared atomics lilke in C++20 (https://en.cppreference.com/w/cpp/memory/shared_ptr/atomic2)

JohanMabille (approved):

PR #2618, opened 2023-06-23
Add topological sort explict export tests

AntoinePrv:


jonashaag:
Thanks a lot, this will ensure we won't reintroduce the bug :)

jonashaag (approved):

Issue #2547, opened 2023-05-28
Incompatibility with conda 23.5.0 ?


noamgat:
### Troubleshooting docs

- [X] My problem is not solved in the Troubleshooting docs

### How did you install Mamba?

conda install mamba

### Search tried in issue tracker

conda 23.5 + python 3.11 incompatibility?

### Latest version of Mamba

- [X] My problem is not solved with the latest version

### Tried in Conda?

I do not have this problem with Conda, just with Mamba

### Describe your issue

We use mamba in our CI/CD pipelines. 
We start from the continuum anaconda docker image (for other reasons). The installation happens using the line

```conda install mamba ```

Afterwards we run

```mamba env create -f environment.yml```

And have started getting the following error in runs in the last 2-3 days despite not changing anything (see the logs).
Forcing the conda version to 23.3.1 (previous version released) fixes the issue.
Conda 23.5.0 was released a few days ago as well, raising my suspicion.

### Logs

```shell
Traceback (most recent call last):
  File "/opt/conda/bin/mamba", line 7, in <module>
    from mamba.mamba import main
  File "/opt/conda/lib/python3.10/site-packages/mamba/mamba.py", line 49, in <module>
    import libmambapy as api
  File "/opt/conda/lib/python3.10/site-packages/libmambapy/__init__.py", line 7, in <module>
    raise e
  File "/opt/conda/lib/python3.10/site-packages/libmambapy/__init__.py", line 4, in <module>
    from libmambapy.bindings import *  # noqa: F401,F403
ImportError: libarchive.so.13: cannot open shared object file: No such file or directory
Cleaning up project directory and file based variables
```

jonashaag:
Starting from Anaconda is not recommended, it sometimes causes problems, see

https://github.com/mamba-org/mamba/issues/1826
https://github.com/mamba-org/mamba/issues/1728
https://github.com/mamba-org/mamba/issues/1724

itamarst:
I was able to work around this by installing mamba like so:

`conda install --override-channels -c conda-forge mamba`

This ensures the libarchive comes from conda-forge.

itamarst:
I also had issues with PyPy being installed, so ended up with this:

`conda install --override-channels -c conda-forge mamba 'python_abi=*=*cp*'`

Perhaps this should be the recommended installation command in the docs.

jonashaag:
[...]

PR #2619, opened 2023-06-25
Force conda-forge in Anaconda install

jonashaag:
Taken from here https://github.com/mamba-org/mamba/issues/2547#issuecomment-1602613370

fyi @itamarst

Are we sure we always have `python_abi` on all platforms?


JohanMabille:
> Are we sure we always have python_abi on all platforms?

I think so. Otherwise, we will fix the doc if someone reports an issue ;)

JohanMabille (approved):

PR #2624, opened 2023-06-26
Direcly call uname for linux detection

AntoinePrv:
Call `uname` directly instead of through a process.
Ideally the whole OS specific should be better wrapped to avoid `#ifdef` and be pure (Context, `env::get`...).


JohanMabille (approved):

Klaim (approved):

wolfv:
This looks ready to go?

JohanMabille:
Yup.

PR #2631, opened 2023-06-28
Not reading hook file

### LINKED ISSUE #2628: micromamba reads `~/micromamba/etc/profile.d/micromamba.sh` from older versions
When using the latest macOS homebrew build of micromamba (https://github.com/Homebrew/homebrew-core/pull/135110), some micromamba commands don't work anymore

```
$ eval "$(micromamba shell hook)" # works
$ micromamba create -n testenv python # works
[...]
$ micromamba run -n testenv python
The following argument was not expected: --shell
/var/folders/rz/q1r4tv0n6ll38q18fbk419mc0000gn/T/mambafCZWLOVkCne: line 5: exec: python: not found
$ micromamba activate testenv
The following argument was not expected: --shell
```

AntoinePrv:
Fix #2628. The file is still written (being conservative) but not read (edit: done ~~expect perhaps by `csh` which as a different initalization script~~).

### DIFF
--- a/libmamba/src/core/activation.cpp
+++ b/libmamba/src/core/activation.cpp
@@ -717,27 +717,15 @@ namespace mamba
 
     std::string Activator::hook(const std::string& shell_type)
     {
-        std::stringstream builder;
-        builder << hook_preamble() << "\n";
-        if (!hook_source_path().empty())
-        {
-            if (fs::exists(hook_source_path()))
-            {
-                builder << read_contents(hook_source_path()) << "\n";
-            }
-            else
-            {
-                std::string contents = get_hook_contents(shell());
-                builder << contents << "\n";
-            }
-        }
-
         // special handling for cmd.exe
-        if (!fs::exists(hook_source_path()) && shell() == "cmd.exe")
+        if (shell() == "cmd.exe")
         {
             get_hook_contents(shell());
             return "";
         }
+
+        std::stringstream builder;
+        builder << hook_preamble() << "\n" << get_hook_contents(shell()) << "\n";
 
         if (Context::instance().shell_completion)
         {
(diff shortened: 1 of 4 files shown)


libmamba/src/core/activation.cpp
```diff
@@ -717,28 +717,28 @@ namespace mamba
...
+        if (is_cmd(this))
         {
             get_hook_contents(shell());
             return "";
         }
 
+        std::stringstream builder;
+        if (is_powershell(this) && fs::exists(hook_source_path()))
+        {
+            builder << read_contents(hook_source_path()) << "\n";
```
JohanMabille, line 735: So the file is still read with powershell and we still have the upgrade issue?

micromamba/tests/test_shell.py
```diff
@@ -86,7 +86,7 @@ def test_hook(self, shell_type):
         elif shell_type == "cmd.exe":
             assert res == ""
         elif shell_type == "tcsh":
-            assert res.count(mamba_exe) == 2
+            assert res.count(mamba_exe) == 3
```
JohanMabille, line 89: Laybe this case could be put wiht "zsh", "bash" and "posix" above? Also, what does this return code mean?

JohanMabille (approved):

libmamba/src/core/activation.cpp
AntoinePrv: No it's more complicated there. I haven't quite understand what is going on.

jonashaag:
Hooray! I'm happy to see this go.

We are still creating those files though, right? We probably can get rid of that code as well.

AntoinePrv:
> Hooray! I'm happy to see this go.

Still missing Windows.

> We are still creating those files though, right?

Yes.

>We probably can get rid of that code as well.

I am unsure if anyone relies on them.

PR #2632, opened 2023-06-28
Try fixing missing subdirs on mamba.py

AntoinePrv:
- Use install repo from libmamba in mamba.py
- Regenerate stubgens


libmambapy/src/main.cpp
```diff
@@ -807,6 +807,7 @@ PYBIND11_MODULE(bindings, m)
         .def_readwrite("version", &PackageInfo::version)
         .def_readwrite("build_string", &PackageInfo::build_string)
         .def_readwrite("build_number", &PackageInfo::build_number)
+        .def_readwrite("noarch", &PackageInfo::noarch)
```
Hind-M, line 810: I don't know how important the `noarch` could be, but it seems to be missing in some of the `PackageInfo` methods compared to the other members. So maybe we should add it there as well? (for consistency and comprehensiveness).

Hind-M (approved):

libmambapy/src/main.cpp
AntoinePrv: Noarch is definitely important. Where do you see it missing?

PR #2627, opened 2023-06-27
Change brace space in clang-format

AntoinePrv:
Removing the space in [this option](https://clang.llvm.org/docs/ClangFormatStyleOptions.html#cpp11bracedliststyle).

PR #2637, opened 2023-06-28
Attempt to fix CI

AntoinePrv:
- Revert "Change brace space in clang-format (#2627)"
- Revert "Improvement: Support ZST in mamba and enable ZST by default (#2404)"


JohanMabille:
SO this fixes the issues we see on the main branch, but triggers a new failure on linux (and probably OSX).

JohanMabille (approved):

PR #2645, opened 2023-06-30
Add Version::starts_with and Version::compatible_with

AntoinePrv:
More ``spec::Version`` comparison functions needed to implement ``VersionSpec``.
It also removes one level of indentation in `test_version.cpp`, sorry for the big diff.

### DIFF
--- a/libmamba/src/specs/version.cpp
+++ b/libmamba/src/specs/version.cpp
@@ -222,17 +222,17 @@ namespace mamba::specs
          */
         template <typename Iter1, typename Iter2, typename Empty1, typename Empty2, typename Cmp>
         constexpr auto lexicographical_compare_three_way_trailing(
-            const Iter1 first1,
-            const Iter1 last1,
-            const Iter2 first2,
-            const Iter2 last2,
-            const Empty1 empty1,
-            const Empty2 empty2,
+            Iter1 first1,
+            Iter1 last1,
+            Iter2 first2,
+            Iter2 last2,
+            const Empty1& empty1,
+            const Empty2& empty2,
             Cmp comp
         ) -> std::pair<strong_ordering, std::size_t>
         {
-            assert(first1 <= last1);
-            assert(first2 <= last2);
+            assert(std::distance(first1, last1) >= 0);
+            assert(std::distance(first2, last2) >= 0);
 
             auto iter1 = first1;
             auto iter2 = first2;
@@ -279,7 +279,7 @@ namespace mamba::specs
             Iter1 last1,
             Iter2 first2,
             Iter2 last2,
-            Empty empty,
+            const Empty& empty,
             Cmp comp
         ) -> std::pair<strong_ordering, std::size_t>
         {
@@ -371,6 +371,14 @@ namespace mamba::specs
         struct AlwaysEqual
         {
         };
+
+        [[maybe_unused]] auto starts_with_three_way(const AlwaysEqual&, const AlwaysEqual&)
+            -> strong_ordering
+        {
+            // This comparison should not happen with the current usage.
+            assert(false);
+            return strong_ordering::equal;
+        }
 
         template <typename T>
         auto starts_with_three_way(const AlwaysEqual&, const T&) -> strong_ordering
--- a/libmamba/src/specs/version.cpp
+++ b/libmamba/src/specs/version.cpp
@@ -47,7 +47,6 @@ namespace mamba::specs
         {
             return compare_three_way(std::strcmp(a.c_str(), b.c_str()), 0);
         }
-
     }
 
     /***************************************
@@ -218,13 +217,14 @@ namespace mamba::specs
          * Similarily ``[1, 1] is less than ``[1, 2, 0]`` but more than ``[1, 1, -1]``
          * because ``-1 < 0``.
          */
-        template <typename Iter1, typename Iter2, typename T, typename Cmp>
+        template <typename Iter1, typename Iter2, typename Empty1, typename Empty2, typename Cmp>
         constexpr auto lexicographical_compare_three_way_trailing(
             Iter1 first1,
             Iter1 last1,
             Iter2 first2,
             Iter2 last2,
-            T empty,
+            Empty1 empty1,
+            Empty2 empty2,
             Cmp comp
         ) -> strong_ordering
         {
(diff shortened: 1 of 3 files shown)


libmamba/src/specs/version.cpp
```diff
@@ -217,78 +216,108 @@ namespace mamba::specs
...
+            const Iter2 last2,
+            const Empty1 empty1,
+            const Empty2 empty2,
             Cmp comp
-        ) -> strong_ordering
+        ) -> std::pair<strong_ordering, std::size_t>
         {
-            for (; (first1 != last1) && (first2 != last2); ++first1, ++first2)
+            assert(first1 <= last1);
+            assert(first2 <= last2);
```
JohanMabille, line 235: This restricts the implementation to random access iterators only

libmamba/src/specs/version.cpp
```diff
@@ -217,78 +216,108 @@ namespace mamba::specs
...
-            Iter1 last1,
-            Iter2 first2,
-            Iter2 last2,
-            T empty,
+            const Iter1 first1,
+            const Iter1 last1,
+            const Iter2 first2,
+            const Iter2 last2,
+            const Empty1 empty1,
+            const Empty2 empty2,
```
JohanMabille, line 230: `const` are unnecessary for variables passed by value.

libmamba/src/specs/version.cpp
```diff
@@ -217,78 +216,108 @@ namespace mamba::specs
...
-                b.cend(),
-                VersionPartAtom{},
-                [](const auto& x, const auto& y) { return compare_three_way(x, y); }
-            );
+                       a.cbegin(),
+                       a.cend(),
+                       b.cbegin(),
+                       b.cend(),
+                       VersionPartAtom{},
+                       [](const auto& x, const auto& y) { return compare_three_way(x, y); }
```
JohanMabille, line 306: why this extra indentation?

libmamba/src/specs/version.cpp
```diff
@@ -217,78 +216,108 @@ namespace mamba::specs
...
-                b.cend(),
-                VersionPart{},
-                [](const auto& x, const auto& y) { return compare_three_way(x, y); }
-            );
+                       a.cbegin(),
+                       a.cend(),
+                       b.cbegin(),
+                       b.cend(),
+                       VersionPart{},
+                       [](const auto& x, const auto& y) { return compare_three_way(x, y); }
```
JohanMabille, line 319: Same question here

libmamba/src/specs/version.cpp
```diff
@@ -337,6 +366,105 @@ namespace mamba::specs
...
+        auto starts_with_three_way(const AlwaysEqual&, const T&) -> strong_ordering
+        {
+            return strong_ordering::equal;
+        }
+
+        template <typename T>
+        auto starts_with_three_way(const T&, const AlwaysEqual&) -> strong_ordering
+        {
+            return strong_ordering::equal;
+        }
```
JohanMabille, line 385: There is an ambiguity if someone calls `starts_with_three_way(AlwaysEqual(), AlwaysEqual());` I guess this combination does not make sense, but it might be worth precising it in a comment to avoid the future definition of a useless overload.

libmamba/src/specs/version.cpp
AntoinePrv: How does the std lib algo deal with this? Should we simply not check anything? I know this could be part of the contract, but I think it can be nice to have extra checks (seeing how tricky it is right now to debug in libmamba).

How about `assert(std::distance(first1, last1) >=0 );` ?

libmamba/src/specs/version.cpp
AntoinePrv: They are not wrong either :) The previous version would `++first1` `++first2`, now with `iter1` and `iter2` there are six iterator with potentially the same type. I though I would be a good idea to make sure we only change the one we mean to.

Alternatively, the arguments were taken by value to imitate the stdlib but I wonder if `const&` would be better suited.

libmamba/src/specs/version.cpp
AntoinePrv: Not sure why clang-format decided to change that. The function is now returning a pair and only the first member is used (note the `.first` on the next line).
We could `auto [cmp, idx] = lexicographical_compare_three_way_trailing(...); return cmp;` if you think it is better.

libmamba/src/specs/version.cpp
AntoinePrv: How about adding
```cpp
[[maybe_unused]] auto starts_with_three_way(const AlwaysEqual&, const AlwaysEqual&) -> strong_ordering {
    static_assert(false, "This comparison should not happen with the current implementation.");
}
```

JohanMabille: Even better

libmamba/src/specs/version.cpp
JohanMabille: I don't think the STL implementation checks anything; I agree that extra checks are a nice-to-have, the `assert(std::distance(first1, last1) >=0 );` akes it work for any iterator so let's go for it.

libmamba/src/specs/version.cpp
JohanMabille: I think the current implementation is fine, I was just surprised by this double indentation of arguments in a function call. We can try to fix the clang-format configuration later, no need to change the code for that imo.

libmamba/src/specs/version.cpp
JohanMabille: Iterator arguments are usually passed by copy because they are usually modified inside the function, so this avoid to code an extra copy in it. Also, they are supposed to be lightweight object, so that the copy does not hurt performance.

> I though I would be a good idea to make sure we only change the one we mean to.

This is from the implementation point-of-view, from the API one, since they are passed by copy, the caller does not really care wht is changed and what is not.

libmamba/src/specs/version.cpp
AntoinePrv: Yes, clang-format does very obscure things sometimes. I could not find any configuration for this.

libmamba/src/specs/version.cpp
AntoinePrv: It doesn't work since this is not a template, I made a regular `assert`.

Issue #2577, opened 2023-06-09
Discussion: `(micro)mamba` should always be in `PATH`


jonashaag:
(This is another example of why I suggest to get rid of shell activation/integration entirely)

If you install Micromamba through http://micro.mamba.pm/install.sh it will be installed to `~/.local/bin/micromamba`. Not everyone has `~/.local/bin` on their `PATH` and Micromamba is fine with this since `micromamba foo` in the shell will not call the `micromamba` binary but the `micromamba` shell function instead.

However if you use any other tool that wants to call Micromamba as a subprocess, eg. pre-commit, this will fail since Micromamba is not on the `PATH`. To users this is really confusing because if you don't know about the shell function you'd think that the `micromamba` binary works fine, while it doesnt'! (Only the shell function works, which has its own way to find the binary.)

This also applies to Mamba.

I think we should make sure that `(micro)mamba` is always on the `PATH`. Maybe we can emit a warning in the shell function if it isn't.

jonashaag:
fyi @pavelzw

pavelzw:
Does the shell function or the binary on `PATH` take priority if the shell function exists and the binary is on `PATH`?

xhochy:
The shell function takes priority.

pavelzw:
Also, I think the current shell initialization could be simplified further. We could just put `eval "(/path/to/micromamba shell hook)"` in the bashrc instead of the whole shell initialization block thing. Then we could also skip the creation of the `$MAMBA_ROOT_PREFIX/etc/profile.d` files since they are not needed.

AntoinePrv:
In fact I don't think there is a user directory which is always gonna be on the Path on Unix, so we should add it to the Path in the initialization script.

Not sure how I feel about automatically adding `~/.local/bin`. That would be my setup, but it seem to be too general of a location for Mamba to add it to the Path itself. What do you think?

allefeld:
Is there any way to get rid of this intricate shell magic entirely? I find it highly opaque and thereby irritating and I am not aware of any other software which does such things. Adding `~/.local/bin` to `PATH` should be all the configuration necessary to use mamba, or at least the minimalist micromamba.

If I understand correctly, the reason for this shell magic is that subprocesses cannot modify the parent shell's environment. Couldn't this be solved by making the primary executable a shell script, which takes care of activation / deactivation itself, but delegates everything else to the binary? Sorry if I'm complete clueless...

jonashaag:
[...]

PR #2649, opened 2023-06-30
No profile.d fallback in rc files

AntoinePrv:
Continuation to #2631. 

@jonashaag @pavelzw does the error handling seem good to you, or will it be incompatible with #2577?

For readability, current output on my system is:
```bash
# >>> mamba initialize >>>
export MAMBA_EXE='/home/antoine/workspace/github.com/mamba-org/mamba/build/cmake/micromamba/micromamba';
export MAMBA_ROOT_PREFIX='/home/antoine/.local/share/mamba';
__mamba_setup="$("$MAMBA_EXE" shell hook --shell bash --root-prefix "$MAMBA_ROOT_PREFIX" 2> /dev/null)"
if [ $? -eq 0 ]; then
    eval "$__mamba_setup"
else
    alias micromamba='echo "Failed to initialize shell for micromamba.\nTry running micromamba shell reinit." 1>&2; "$MAMBA_EXE" "$@"'
fi
unset __mamba_setup
# <<< mamba initialize <<<
```


jonashaag:
Does this even work? If you set `micromamba` to this dummy function, how are you supposed to `shell reinit`?

AntoinePrv:
It would because the message is added at the top, then the alias calls into `$MAMBA_EXE`.
But you make me realize that perhaps we could simply fallback to `alias micromamba="$MAMBA_EXE`, since activating with the binary will give the message:
```txt
'micromamba' is running as a subprocess and can't modify the parent shell.
Thus you must initialize your shell before using activate and deactivate.

To initialize the current  shell, run:
    $ eval "$(micromamba shell hook --shell )"
and then activate or deactivate with:
    $ micromamba activate
To automatically initialize all future () shells, run:
    $ micromamba shell init --shell  --root-prefix=~/micromamba

Supported shells are {bash, zsh, csh, xonsh, cmd.exe, powershell, fish}.
```

jonashaag:
Oh sorry I didn't see the call into `$MAMBA_EXE` 🙈 I would actually prefer to see error messages. Maybe we can remove the `2>/dev/null` above?

jonashaag (approved):

AntoinePrv:
> Somewhat relatedly, I would actually prefer to see error messages. Maybe we can remove the 2>/dev/null above?

That's a good idea! I think we should delay it unit the user calls `micromamba` and not when the rc file is sourced.

AntoinePrv:
I propose:
```bash
# >>> mamba initialize >>>
export MAMBA_EXE='/home/antoine/workspace/github.com/mamba-org/mamba/build/cmake/micromamba/micromamba';
export MAMBA_ROOT_PREFIX='/home/antoine/workspace/github.com/mamba-org/mamba/foo';
__mamba_setup="$("$MAMBA_EXE" shell hook --shell bash --root-prefix "$MAMBA_ROOT_PREFIX" 2> /dev/null)"
if [ $? -eq 0 ]; then
    eval "$__mamba_setup"
else
    alias micromamba="$MAMBA_EXE"  # Fallback on help from mamba activate
fi
unset __mamba_setup
# <<< mamba initialize <<<
```
And the message
```
'micromamba' is running as a subprocess and can't modify the parent shell.
Thus you must initialize your shell before using activate and deactivate.

To initialize the current  shell, run:
    $ eval "$(micromamba shell hook --shell )"
and then activate or deactivate with:
    $ micromamba activate
To automatically initialize all future (zsh) shells, run:
    $ micromamba shell init --shell zsh --root-prefix=~/micromamba
If your shell was already initialized, reinitialize your shell with:
    $ micromamba shell reinit --shell
Otherwise, this may be an issue. In the meantime you can run commands. See:
    $ micromamba run --help

Supported shells are {bash, zsh, csh, xonsh, cmd.exe, powershell, fish}.
```

JohanMabille (approved):

PR #2653, opened 2023-07-05
Use boltons directly instead of using deprecated vendored version from conda

duncanmmacleod:
This PR resolves the deprecation warning currently emitted by mamba:

```console
$ mamba --version
/home/duncan/opt/mambaforge/lib/python3.10/site-packages/mamba/utils.py:12: PendingDeprecationWarning: conda._vendor.boltons is pending deprecation and will be removed in 24.3. Use `boltons` instead.
  from conda._vendor.boltons.setutils import IndexedSet
mamba 1.4.6
conda 23.5.0
```

by patching to use `boltons` directly.

As far as I can tell (which isn't bulletproof by any means) any version of `boltons` is fine, but recent versions of `conda` itself have a fairly aggressive version requirement on `boltons` which `mamba` will pick up anyway.

### DIFF
--- a/mamba/mamba/utils.py
+++ b/mamba/mamba/utils.py
@@ -9,7 +9,7 @@ import tempfile
 import urllib.parse
 from collections import OrderedDict
 
-from conda._vendor.boltons.setutils import IndexedSet
+from boltons.setutils import IndexedSet
 from conda.base.constants import ChannelPriority
 from conda.base.context import context
 from conda.common.serialize import json_dump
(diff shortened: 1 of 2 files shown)


jonashaag:
Thanks for the PR!

IMO we should change this as follows:

If we have boltons we should use that, otherwise fall back to importing from Conda.

We’ll have to add the dependency to some other places as well (all environment.yml files, Conda recipe, maybe other places)

jjerphan:
Mamba has changed a lot and does not rely on conda and python anymore. Thank you for the intend!

PR #2655, opened 2023-07-05
call `init_console` to prevent UTF8 errors when extracting packages

wolfv:


JohanMabille (approved):

JohanMabille:
We probably need to do the same in mamba for the fallback case.

PR #2658, opened 2023-07-06
All headers at the top

AntoinePrv:


JohanMabille (approved):

AntoinePrv:
Sorting still happens among headers that are next to each others. In fact you can use comments to make separation, which is convenient when some incomplete headers need to be included last.

AntoinePrv:
Merging since the failure is unrelated.

PR #2662, opened 2023-07-08
Improve warning when package record not found

maresb:
We just triggered this warning in https://github.com/conda/conda-lock/pull/448. We don't know what it means, but it seems it's some sort of package to be unlinked, but there's no mention of the package name, and the output is to `stdout` which is problematic when capturing JSON.

### DIFF
--- a/mamba/mamba/utils.py
+++ b/mamba/mamba/utils.py
@@ -5,6 +5,7 @@ from __future__ import absolute_import, division, print_function, unicode_litera
 
 import json
 import os
+import sys
 import tempfile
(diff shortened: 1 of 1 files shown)


riccardoporreca:
See https://github.com/conda/conda-lock/pull/448#issuecomment-1627546017 showing what lead to `No package record found!`

jonashaag (approved):

jonashaag:
Thank you!

jonashaag:
There are some other prints that should go to stderr, maybe worth changing those as well

PR #2671, opened 2023-07-13
Added upper bound to fmt to avoid weird failure on ci (windows only)

JohanMabille:
Merge this one instead of #2669 if that works


jonashaag:
I had this/a similar problem on macOS as well

JohanMabille:
Oh, good to know. To fmt is the culprit?

jonashaag:
idk

PR #2685, opened 2023-07-18
Use generic_string for path on Windows unix shells

### LINKED ISSUE #2683: Micromamba >v1.4.5 broken with cygwin
### Troubleshooting docs

Micromamba

yes

### Latest version of Mamba

### Tried in Conda?

Not applicable

### Describe your issue

I'm using micromamba with cygwin. My .bash_profile looks like

```
# >>> mamba initialize >>>
export MAMBA_EXE='/cygdrive/c/Users/USER/.bin/micromamba.exe';
export MAMBA_ROOT_PREFIX='/cygdrive/c/Users/USER/micromamba_prefix';
eval "$("$MAMBA_EXE" shell hook --shell bash --root-prefix "$MAMBA_ROOT_PREFIX")"
# <<< mamba initialize <<<
````

With micromamba <=1.4.5 `shell hook --shell bash` checked if `!hook_source_path().empty()` and if true just emitted the file shell hook without substituting `$MAMBA_EXE`. Only if `hook_source_path()` was empty would  `get_hook_contents` be used to emit the shell hook. Crucially `get_hook_contents` replaced `$MAMBA_EXE` with the path to the binary on its own. However, this produces a broken config, as the replaced path is in windows format which cant be opened by cygwins bash. PR #2631 changed the behavior of `shell hook` so that  `get_hook_contents` is always used which breaks my cygwin config.

I'm not really sure how to tackle this problem, but i think a `--no-replace` option for shell hook would solve my issue. 

_No response_

### Logs

_No response_

### environment.yml

_No response_

### ~/.condarc

_No response_

### PR DESCRIPTION
**Pull request by AntoinePrv**
Fix #2683

### DIFF
--- a/libmamba/include/mamba/core/mamba_fs.hpp
+++ b/libmamba/include/mamba/core/mamba_fs.hpp
@@ -335,6 +335,12 @@ namespace fs
             return this->wstring();
         }
 
+        // Returns an utf-8 string using the ``/`` on all systems.
+        std::string generic_string() const
+        {
+            return to_utf8(m_path.generic_string());
+        }
+
         // Implicit conversion to standard path.
         operator std::filesystem::path() const
         {
--- a/libmamba/src/core/shell_init.cpp
+++ b/libmamba/src/core/shell_init.cpp
@@ -531,19 +531,22 @@ namespace mamba
         if (shell == "zsh" || shell == "bash" || shell == "posix")
         {
             std::string contents = data_micromamba_sh;
-            util::replace_all(contents, "$MAMBA_EXE", exe.string());
+            // Using /unix/like/paths on Unix shell (even on Windows)
+            util::replace_all(contents, "$MAMBA_EXE", exe.generic_string());
             return contents;
         }
         else if (shell == "csh")
         {
             std::string contents = data_micromamba_csh;
-            util::replace_all(contents, "$MAMBA_EXE", exe.string());
+            // Using /unix/like/paths on Unix shell (even on Windows)
+            util::replace_all(contents, "$MAMBA_EXE", exe.generic_string());
             return contents;
         }
         else if (shell == "xonsh")
         {
             std::string contents = data_mamba_xsh;
-            util::replace_all(contents, "$MAMBA_EXE", exe.string());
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by jonashaag** [comment]
Can we add a test for this?

**Comment by AntoinePrv** [comment]
Hmm, bash should already be tested on Windows so it is weird that we had a bug to start with...
Unless there are options to allow Windows paths on Windows bash?

**Comment by pavelzw** [comment]
I tested it on cygwin:

```
~/Documents $ ./micromamba.exe shell hook --shell=bash | grep "__mamba_exe()" -A 5
__mamba_exe() (
    "C:\Users\pavel\Documents\micromamba.exe" "${@}"
)

__mamba_hashr() {
    if [ -n "${ZSH_VERSION:+x}" ]; then
```

Unfortunately, it still uses the windows paths. (got the executable from this commit https://github.com/0xbe7a/mamba/commit/32846b1c663079122356fdfb45a209649533f8da)

**Comment by AntoinePrv** [comment]
I'm confused. @pavelzw can you try the last commit? The [standard](https://en.cppreference.com/w/cpp/filesystem/path/generic_string) clearly states that the separator should be a `/`. Or am I not modifying the right function?

**Comment by pavelzw** [comment]
Just tried it with a brand-new windows server instance with cygwin installed. There, it still uses the windows paths.

```
admin@windows-server /cygdrive/c/Users/admin/Downloads
$ ./micromamba shell hook -s bash

# Copyright (C) 2012 Anaconda, Inc
# SPDX-License-Identifier: BSD-3-Clause

__mamba_exe() (
    "C:\Users\admin\Downloads\micromamba.exe" "${@}"
)
...
```

I used the executable from [`73b0f54` (#2685)](https://github.com/mamba-org/mamba/pull/2685/commits/73b0f54e78fbd9fd858574d18a526b2f60ef55b3).

Klaim (approved):
I'm approving this but worry that the paths displayed might be misleading to users. Maybe displaying both on windows would be best?

AntoinePrv:
@Klaim there is no displaying in this PR. This are activation scripts being run when using `micromamba activate` (the ones that set env variables).

AntoinePrv:
Let's give it a try

PR #2690, opened 2023-07-20
Ignore format changes in git blame

jonashaag:


AntoinePrv (approved):
Can you add a comment in the file indicating if this is read automatically (by who) or how ot use it? Thanks!

PR #2695, opened 2023-07-21
New apis for downloading

JohanMabille:


libmamba/include/mamba/util/iterator.hpp
```diff
@@ -267,7 +297,7 @@ namespace mamba::util
         enable_bidirectional_iterator<It, self_type&> operator--()
         {
             --m_iter;
-            while (m_iter != m_begin_limit && !m_pred(*m_iter))
+            while (m_iter != m_begin_limit && !(m_pred.value()(*m_iter)))
```
Klaim, line 300: Note that using  `*m_pred` works too, both are fine to me.

libmamba/include/mamba/util/iterator.hpp
```diff
@@ -340,13 +370,17 @@ namespace mamba::util
...
-            while (m_iter != m_end && !m_pred(*m_iter))
+            while (m_iter != m_end && !(m_pred.value()(*m_iter)))
             {
                 ++m_iter;
             }
         }
 
-        Predicate m_pred;
+        // Trick to enable move and copy assignment: since lambdas are
+        // not assignable, we encpsulate them in an std::optional and
```
Klaim, line 380: encapsulate

libmamba/include/mamba/util/iterator.hpp
```diff
@@ -340,13 +370,17 @@ namespace mamba::util
...
+        // rely on it to implement assignment operators. The optional
+        // should be replaced with a dedicated wrapper.
```
Klaim, line 382: Also note that most std::fucntion implementations actually use small-buffer optimization so while std::function is heavyweight and expensive for big callable objects, it should be fine for lambdas which dont capture big objects by value, so most lambdas - as  in dynamic allocations will be rare. But optional is indeed not allocating at all as it's based on the real type of the predicate.

libmamba/src/core/compression.hpp
```diff
@@ -14,6 +14,42 @@
...
+
+        writer m_writer;
+    };
+
+    std::unique_ptr<CompressionStream> make_compression_stream(
+        const std::string& url,
+        CompressionStream::writer&& func
+    );
+
+    // TODO: remove the following when switching to new CompressionStream
```
Klaim, line 51: remove ZstdStream?

libmamba/src/core/context.cpp
```diff
@@ -179,6 +179,11 @@ namespace mamba
         return m_authentication_info;
     }
 
+    const std::map<std::string, AuthenticationInfo>& Context::authentication_info() const
```
Klaim, line 182: I think it would help future maintenance (and switching to unordered_map) if this map was a type alias.

libmamba/src/core/curl.hpp
```diff
@@ -64,6 +64,38 @@ namespace mamba
         bool m_serious;
     };
 
+    class CURLId
```
Klaim, line 67: Did you consider making this type definition header-only/inline? It seems to be a good candidate.

libmamba/src/core/curl.hpp
```diff
@@ -64,6 +64,38 @@ namespace mamba
...
+        std::size_t hash() const noexcept;
+
+    private:
+
+        CURL* p_handle;
+    };
+}
+
+template <>
+struct std::hash<mamba::CURLId>
```
Klaim, line 89: Not totally sure but I htink `::std::hash` have to be used to make sure it's the standard and not a sub-namesapce.

libmamba/src/core/download.cpp
```diff
@@ -0,0 +1,609 @@
...
+
+
+namespace mamba
+{
+
+    /**********************************
+     * DownloadAttempt implementation *
+     **********************************/
+
+    DownloadAttempt::DownloadAttempt(CURLHandle& handle, const DownloadRequest* request)
```
Klaim, line 18: Capture `request` as a const reference, store it as a pointer.

libmamba/src/core/download.cpp
```diff
@@ -0,0 +1,609 @@
...
+        m_success_callback = std::move(success);
+        m_failure_callback = std::move(error);
+        downloader.add_handle(*p_handle);
+    }
+
+    namespace
+    {
+        bool can_retry_download(int http_status, const std::string& url)
+        {
+            return (http_status == 413 || http_status == 429 || http_status >= 500)
```
Klaim, line 44: It would be more readable with named constants (so that we get the meaning of these numbers without having to google)

libmamba/src/core/download.cpp
```diff
@@ -0,0 +1,609 @@
...
+    }
+
+    void DownloadAttempt::configure_handle(const Context& context)
+    {
+        // TODO: we should probably store set_low_speed_limit and set_ssl_no_revoke in RemoteFetchParams
+        // if the request is slower than 30b/s for 60 seconds, cancel.
+        std::string no_low_speed_limit = std::getenv("MAMBA_NO_LOW_SPEED_LIMIT")
+                                             ? std::getenv("MAMBA_NO_LOW_SPEED_LIMIT")
+                                             : "0";
+        bool set_low_speed_opt = (no_low_speed_limit == "0");
```
Klaim, line 154: const

libmamba/src/core/download.cpp
```diff
@@ -0,0 +1,609 @@
...
+                << size << " bytes";
+            return ss.str();
+        }
+    }
+
+    void DownloadAttempt::configure_handle(const Context& context)
+    {
+        // TODO: we should probably store set_low_speed_limit and set_ssl_no_revoke in RemoteFetchParams
+        // if the request is slower than 30b/s for 60 seconds, cancel.
+        std::string no_low_speed_limit = std::getenv("MAMBA_NO_LOW_SPEED_LIMIT")
```
Klaim, line 151: const

libmamba/src/core/download.cpp
```diff
@@ -0,0 +1,609 @@
...
+                                             ? std::getenv("MAMBA_NO_LOW_SPEED_LIMIT")
+                                             : "0";
+        bool set_low_speed_opt = (no_low_speed_limit == "0");
+
+        std::string ssl_no_revoke_env = std::getenv("MAMBA_SSL_NO_REVOKE")
```
Klaim, line 156: const

libmamba/src/core/download.cpp
```diff
@@ -0,0 +1,609 @@
...
+                                            ? std::getenv("MAMBA_SSL_NO_REVOKE")
+                                            : "0";
+        bool set_ssl_no_revoke = context.remote_fetch_params.ssl_no_revoke || (ssl_no_revoke_env != "0");
```
Klaim, line 159: const

libmamba/src/core/download.cpp
```diff
@@ -0,0 +1,609 @@
...
+        {
+            p_handle->add_header("If-Modified-Since:" + p_request->if_modified_since.value());
+        }
+       
+        p_handle->set_opt_header();
+    }
+
+    size_t DownloadAttempt::write_data(char* buffer, size_t size)
+    {
+        if (!m_file.is_open())
```
Klaim, line 243: At this point `m_file` needs to be set right? Maybe add an assertion?

libmamba/src/core/download.cpp
```diff
@@ -0,0 +1,609 @@
...
+            return size + 1;
+        }
+        return size;
+    }
+
+    size_t DownloadAttempt::curl_header_callback(char* buffer, size_t size, size_t nbitems, void* self)
+    {
+        auto* s = reinterpret_cast<DownloadAttempt*>(self);
+
+        size_t buffer_size = size * nbitems;
```
Klaim, line 270: const

libmamba/src/core/download.cpp
```diff
@@ -0,0 +1,609 @@
...
+        std::string_view header(buffer, buffer_size);
```
Klaim, line 271: const

libmamba/src/core/download.cpp
```diff
@@ -0,0 +1,609 @@
...
+        auto colon_idx = header.find(':');
+        if (colon_idx != std::string_view::npos)
+        {
+            std::string_view key, value;
```
Klaim, line 275: Please initialize key directly and value where it's initialized.

libmamba/src/core/download.cpp
```diff
@@ -0,0 +1,609 @@
...
+            key = header.substr(0, colon_idx);
+            colon_idx++;
+            // remove spaces
+            while (std::isspace(header[colon_idx]))
+            {
+                ++colon_idx;
+            }
+
+            // remove \r\n header ending
+            auto header_end = header.find_first_of("\r\n");
```
Klaim, line 285: const

libmamba/src/core/download.cpp
```diff
@@ -0,0 +1,609 @@
...
+            value = header.substr(colon_idx, (header_end > colon_idx) ? header_end - colon_idx : 0);
+
+            // http headers are case insensitive!
+            std::string lkey = to_lower(key);
```
Klaim, line 289: const

libmamba/src/core/download.cpp
```diff
@@ -0,0 +1,609 @@
...
+            static_cast<std::size_t>(total_to_download),
+            static_cast<std::size_t>(now_downloaded)
+        });
+        return 0;
+    }
+
+    TransferData DownloadAttempt::get_transfer_data() const
+    {
+        return {
+            /* .http_status = */ p_handle->get_info<int>(CURLINFO_RESPONSE_CODE).value_or(10000),
```
Klaim, line 324: Consider naming this constant, name it ARBITRARY if it is arbitrary.

libmamba/src/core/download.cpp
```diff
@@ -0,0 +1,609 @@
...
+        , p_context(&context)
+        , m_curl_handle(context.threads_params.download_threads)
+        , m_trackers()
+    {
+        if (m_options.sort)
+        {
+            std::sort(
+                m_requests.requests.begin(),
+                m_requests.requests.end(),
+                [](DownloadRequest& a, DownloadRequest& b) -> bool
```
Klaim, line 514: these arguments should be const references, otherwise it's UB

libmamba/src/core/download.cpp
```diff
@@ -0,0 +1,609 @@
...
+            update_downloads();
+        }
+
+        return build_result();
+    }
+
+    void Downloader::prepare_next_downloads()
+    {
+        size_t running_attempts = m_completion_map.size();
+        size_t max_parallel_downloads = p_context->threads_params.download_threads;
```
Klaim, line 541: const

libmamba/src/core/download.cpp
```diff
@@ -0,0 +1,609 @@
...
+
+    void Downloader::update_downloads()
+    {
+        while(auto resp = m_curl_handle.pop_message())
+        {
+            const auto& msg = resp.value();
+            if (!msg.m_transfer_done)
+            {
+                // We are only interested in messages about finished transfers
+                continue;
```
Klaim, line 565: does `pop_message()` handles the waiting of the current thread? I would use `std::yield()` here otherwise, to avoid saturating the cpu.

libmamba/src/core/download.cpp
```diff
@@ -0,0 +1,609 @@
...
+    }
+
+    bool Downloader::download_done() const
+    {
+        return m_waiting_count == 0;
+    }
+
+    MultiDownloadResult Downloader::build_result() const
+    {
+        DownloadResultList l;
```
Klaim, line 594: please avoid 1 letters or abreviations names in non-maths code🥲

libmamba/src/core/download_impl.hpp
```diff
@@ -0,0 +1,149 @@
...
+    class DownloadAttempt
+    {
+
+    public:
+
+        using completion_function = std::function<bool(CURLMultiHandle&,CURLcode)>;
+        using on_success_callback = std::function<bool(DownloadSuccess)>;
+        using on_failure_callback = std::function<bool(DownloadError)>;
+
+        DownloadAttempt(CURLHandle& handle, const DownloadRequest* request);
```
Klaim, line 38: As pointed before, request should be taken by const reference, passing nullptr should not be authorized.

libmamba/src/core/download_impl.hpp
```diff
@@ -0,0 +1,149 @@
...
+    };
+
+    class DownloadTracker
+    {
+    public:
+
+        using completion_function = DownloadAttempt::completion_function;
+        using completion_map_entry = std::pair<CURLId, completion_function>;
+        
+        DownloadTracker(const DownloadRequest* request, std::size_t max_retries, bool fail_fast);
```
Klaim, line 86: same remark about request should be a const ref here

Klaim: Also to avoid the bool in the interface of DownloadTRacker, consider grouping both value arguments in a DownloadTrackerOptions struct with them and default values.

Klaim (commented):
Good job!

Some missing things

1. this really needs to be tested in the c++ tests as once plugged it will be a fundamental piece of the whole library. I understand it is not easy to test because it's using curl internally and this cannot be replaced by some networking fake api, but maybe we can at least test with files in local directory?

2. documentation of the types and functions apis in core/download.hpp

libmamba/include/mamba/util/iterator.hpp
```diff
@@ -256,6 +256,36 @@ namespace mamba::util
...
+            }
+            m_iter = rhs.m_iter;
+            m_begin_limit = rhs.m_begin_limit;
+            m_end = rhs.m_end;
+            return *this;
+        }
+
+        self_type& operator=(self_type&& rhs)
+        {
+            m_pred.reset();
```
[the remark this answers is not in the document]
JohanMabille: `std::optional<T>::operator=` internally makes use of `T::operator=`, which is not available for lambdas, thus the usage of `reset` and `emplace_back`, to call the move constructor of T instead.

libmamba/include/mamba/util/iterator.hpp
JohanMabille: Also it avoids pulling the <functional> header. And I really plan to replace `std::optional` with a dedicated wrapper.

libmamba/src/core/compression.hpp
JohanMabille: Yes it will be replaced with the hidden ZstdCompressionStream defined in the cpp.

libmamba/src/core/download_impl.hpp
JohanMabille: Actually I did it in the first version but I found it a bit overkilled for only two members. I will reintroduce it, I agree the boolean here is kind of ugly.

libmamba/src/core/download.cpp
JohanMabille: Not necessarily, here we rely on lazy opening

libmamba/src/core/download.cpp
JohanMabille: nope this should be done in `m_curl_handle->perform()` actually, which is missing from this function.

libmamba/src/core/curl.hpp
JohanMabille: Actually this is not possible, the compiler complains with: `error: global qualification of class name is invalid before '{' token`

libmambapy/src/main.cpp
```diff
@@ -578,6 +577,7 @@ PYBIND11_MODULE(bindings, m)
         .def_readwrite("retry_backoff", &Context::RemoteFetchParams::retry_backoff)
         .def_readwrite("user_agent", &Context::RemoteFetchParams::user_agent)
         // .def_readwrite("read_timeout_secs", &Context::RemoteFetchParams::read_timeout_secs)
+        .def_readwrite("proxy_servers", &Context::RemoteFetchParams::proxy_servers)
```
Hind-M, line 580: Btw, if we still want backward compatibility with python API, we need to add something like this:
https://github.com/mamba-org/mamba/blob/main/libmambapy/src/main.cpp#L605-L621

.github/workflows/main.yml
```diff
@@ -188,6 +188,7 @@ jobs:
         shell: bash -l {0}
         run: |
           pybind11-stubgen libmambapy.bindings
+          pre-commit run --files stubs/libmambapy/binginds-stubds/__init__.pyi
```
wolfv, line 191: Is this a typo? binginds-stubds -> bindings-stubs?

JohanMabille: Yes definitely! Tahnks for catching it

Issue #2651, opened 2023-07-04
Powershell "MambaModuleArgs" Variable not found.


Mukul1127:
### Troubleshooting docs

- [X] My problem is not solved in the Troubleshooting docs

### How did you install Mamba?

Micromamba

### Search tried in issue tracker

MambaModuleArgs

### Latest version of Mamba

- [X] My problem is not solved with the latest version

### Tried in Conda?

Not applicable

### Describe your issue

When Powershell Core starts, it gives this error:
```pwsh
Remove-Variable:
Line |
   3 |  Remove-Variable MambaModuleArgs
     |  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     | Cannot find a variable with the name 'MambaModuleArgs'.
```
When Powershell 5.1 starts, it still gives this error:
```pwsh
Remove-Variable : Cannot find a variable with the name 'MambaModuleArgs'.
At line:3 char:1
+ Remove-Variable MambaModuleArgs
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : ObjectNotFound: (MambaModuleArgs:String) [Remove-Variable], ItemNotFoundException
    + FullyQualifiedErrorId : VariableNotFound,Microsoft.PowerShell.Commands.RemoveVariableCommand
```

### mamba info / micromamba info

```shell
libmamba version : 1.4.6
     micromamba version : 1.4.6
           curl version : libcurl/8.1.2-DEV Schannel zlib/1.2.13
     libarchive version : libarchive 3.6.2 zlib/1.2.13 liblzma/5.4.3 bz2lib/1.0.8 liblz4/1.9.4 libzstd/1.5.5
       envs directories : C:\Users\mukul\micromamba\envs
          package cache : C:\Users\mukul\micromamba\pkgs
                          C:\Users\mukul\.mamba\pkgs
                          C:\Users\mukul\AppData\Roaming\.mamba\pkgs
            environment : None (not found)
           env location : -
      user config files : C:\Users\mukul\.mambarc
 populated config files :
       virtual packages : __win=0=0
                          __archspec=1=x86_64
                          __cuda=12.2=0
               channels :
       base environment : C:\Users\mukul\micromamba
               platform : win-64
```

### Logs

```shell
Not Applicable
```

### environment.yml

```yaml
Not Applicable
```

### ~/.condarc

```yaml
Not Applicable
```

AntoinePrv:
[...]

PR #2696, opened 2023-07-21
#2651 fail quietly when removing a non-existing variable

phreed:
This was discussed in issue #2651 

When Powershell starts it runs the $PROFILE script.
`micromamba init` updates these profiles (e.g. $profile.CurrentUserAllHosts) with something like the following:
```powershell
#region mamba initialize
# !! Contents within this block are managed by 'mamba shell init' !!
$Env:MAMBA_ROOT_PREFIX = "C:\Users\me\micromamba"
$Env:MAMBA_EXE = "C:\Users\me\.bin\micromamba.exe"
(& $Env:MAMBA_EXE 'shell' 'hook' -s 'powershell' -p $Env:MAMBA_ROOT_PREFIX) | Out-String | Invoke-Expression
#endregion
```
When run the following error is produced.

```
Remove-Variable:
Line |
   3 |  Remove-Variable MambaModuleArgs
     |  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     | Cannot find a variable with the name 'MambaModuleArgs'.
```

The cause of the error is revealed by running the following command.

```powershell
(& $Env:MAMBA_EXE 'shell' 'hook' -s 'powershell' -p $Env:MAMBA_ROOT_PREFIX) | Out-String
```
 
Which produces:
```powershell
Import-Module "$Env:MAMBA_ROOT_PREFIX\condabin\Mamba.psm1" -ArgumentList $MambaModuleArgs
Remove-Variable MambaModuleArgs
```

While `Mamba.psm1`  accepts a missing/empty `ArgumentList`, the subsequent `Remove-Variable` on the undefined variable `MambaModuleArgs` produces the error.

I suspect the idea is that `MambaModuleArgs` is presumed to be set outside of the `region mamba initialize`.  Therefore, I think the fix should accommodate that inten by changing the powershell shell hook...

```powershell
Remove-Variable -Name MambaModuleArgs -ErrorAction SilentlyContinue
```

A related issue is that the documentation is weak on the purpose for `MambaModuleArgs`.

It is used in `Mamba.psm1` and where seems to only have one defined entry.
```powershell
$MambaModuleArgs = @{"ChangePs1"=$True};
```

### DIFF
--- a/libmamba/src/core/activation.cpp
+++ b/libmamba/src/core/activation.cpp
@@ -1058,7 +1058,7 @@ namespace mamba
 
     std::string PowerShellActivator::hook_postamble()
     {
-        return "Remove-Variable MambaModuleArgs";
+        return "Remove-Variable -Name MambaModuleArgs -ErrorAction SilentlyContinue";
     }
 
     fs::u8path PowerShellActivator::hook_source_path()
--- a/micromamba/tests/test_activation.py
+++ b/micromamba/tests/test_activation.py
@@ -951,7 +951,7 @@ class TestActivation:
                 f'$Env:MAMBA_EXE="{mamba_exe}"',
                 "$MambaModuleArgs = @{ChangePs1 = $True}",
                 f'Import-Module "{tmp_root_prefix}\\condabin\\Mamba.psm1" -ArgumentList $MambaModuleArgs',
-                "Remove-Variable MambaModuleArgs",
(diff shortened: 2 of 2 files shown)


AntoinePrv (approved):
Thanks

AntoinePrv:
@phreed I am failing to push to this PR, cold you rebase?

Issue #1787, opened 2022-07-11
Support XDG directories


AntoinePrv:
The idea would be to extend and allow customization of directories used by `micromamba` (to some extend for `mamba`). This would be done according to the [`XDG`](https://wiki.archlinux.org/title/XDG_Base_Directory) specification.

Supported options are listed from highest to lowest priority
### A - Configs
- `$XDG_CONFIG_HOME/mamba/mambarc` if set
- `$HOME/.config/mamba/mambarc` new
- `$HOME/.mambarc` legacy

### B - Environments
- `$CONDA_ENVS_PATH` if set
- `$MAMBA_ROOT_PREFIX/envs/` if set
- `$XDG_DATA_HOME/mamba/envs` if set
- `$HOME/.local/share/mamba/envs`

### C - Packages
- `$CONDA_PKGS_DIRS` if set
- `$MAMBA_ROOT_PREFIX/pkgs/` if set
- `$XDG_CACHE_HOME/mamba/pkgs` if set
- `$HOME/.cache/mamba/pkgs`

### D - Authentication
Not sure if `authentication.json` should be with configs or data.

-----

At least *A* and *D* would be useful for further isolating tests (which I believe is one of the reason I'm still having trouble running tests on my local machine).

bepoli:
I think it's worth noting that `conda` already implements this (https://github.com/conda/conda/pull/10982), which means that people storing condarc in `XDG_CONFIG_HOME/conda` will be required to move or link it if they switch to mamba. It would be nice to have the same behaviour in `mamba` and `micromamba`.

anuramat:
what about root_prefix? it's still not xdg compliant, is it?
sorry for necroing

PR #2714, opened 2023-07-26
 Use xdg schemas for config saving/reading (minified)

danpf:
This is a 'mini' version of https://github.com/mamba-org/mamba/pull/1836

I tried not to rock the boat too much with this one.

This PR added some helper functions in order to support the xdg directory schema, as well as pollute the users home directory less.

Fixes: https://github.com/mamba-org/mamba/issues/1787

Maybe I shouldn't have, but it felt weird intercalating the mamba and conda configs, so i separated them and just made all mamba configs superseed the conda configs.

as an aside, (sorry for the un-requested opinion), I totally understand what this org is doing, but I think trying to act as a drop in replacement for conda while at the same time building out its own configs (mambarc etc) is very confusing.    I think it should just be one or the other.  Or at least only condarc etc until the project is mature enough, and then switch to mambarc after a major version bump when it's ready.

### DIFF
--- a/libmamba/src/api/configuration.cpp
+++ b/libmamba/src/api/configuration.cpp
@@ -1776,11 +1776,22 @@ namespace mamba
                                          ctx.prefix_params.root_prefix / "condarc.d",
                                          ctx.prefix_params.root_prefix / ".mambarc" };
 
-        std::vector<fs::u8path> home = { env::home_directory() / ".conda/.condarc",
-                                         env::home_directory() / ".conda/condarc",
-                                         env::home_directory() / ".conda/condarc.d",
-                                         env::home_directory() / ".condarc",
-                                         env::home_directory() / ".mambarc" };
+        // Precedence is least to most. Configuration::set_rc_values iterates over this vector
+        // setting config values as they're found, overwriting preceeding ones.
+        // Because user_config_dir default=$XDG_CONFIG_HOME/mamba but we wanted to allow
+        // $XDG_CONFIG_HOME/conda and '..' seems like the best way to make it conda/mamba
+        // compatible. Otherwise I would have to set user_config_dir to either be just
+        // $XDG_CONFIG_HOME and always supply mamba after calling it, or I would have to give
+        // it a mamba argument, all so I can supply conda in a few default cases. it seems
+        // like ../conda is an easier solution
+        std::vector<fs::u8path> home = {
+            env::user_config_dir() / "../conda/.condarc", env::user_config_dir() / "../conda/condarc", env::user_config_dir() / "../conda/condarc.d",
+            env::home_directory() / ".conda/.condarc",   env::home_directory() / ".conda/condarc", env::home_directory() / ".conda/condarc.d",
+            env::home_directory() / ".condarc",
+            env::user_config_dir() / ".mambarc", env::user_config_dir() / "mambarc", env::user_config_dir() / "mambarc.d",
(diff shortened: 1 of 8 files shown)


libmamba/src/api/configuration.cpp
```diff
@@ -1776,11 +1786,30 @@ namespace mamba
...
+            env::home_directory() / ".conda/condarc.d",
+            env::home_directory() / ".condarc",
+        };
+        if (env::get("CONDARC"))
+        {
+            conda_user.push_back(fs::u8path(env::get("CONDARC").value()));
+        }
+
+        std::vector<fs::u8path> mamba_user = {
+            env::user_config_dir() / ".mambarc",      env::user_config_dir() / "mambarc",
```
jonashaag, line 1804: @pavelzw do you think we should be adding these? I don't have a strong opinion. Nobody will use them anyways.

libmamba/src/core/environment.cpp
```diff
@@ -267,7 +269,49 @@ namespace mamba
...
+            return fs::u8path(maybe_home);
+        }
+
+        fs::u8path user_config_dir()
+        {
+            std::string maybe_user_config_dir = env::get("XDG_CONFIG_HOME").value_or("");
+            if (maybe_user_config_dir.empty())
+            {
+#ifdef _WIN32
+                maybe_user_config_dir = ::mamba::win::get_folder("roamingappdata");
```
jonashaag, line 281: @pavelzw see discussion in the other PR. I'm fine with the decision to use this as the Windows default folder; people can always set the XDG vars.

jonashaag: On Cygwin I'd be surprised if `~/.config` wasn't read by default though...

jonashaag (approved):
Thanks a bunch, perfect work :)

jonashaag:
@pavelzw

pavelzw:
Sorry, I forgot to look at this. I'll do it today 😅

libmamba/src/api/configuration.cpp
pavelzw: I think actually @0xbe7a uses them 😄 so yes

libmamba/src/core/environment.cpp
pavelzw: I'm also fine with this. If people want to have XDG like stuff on cygwin, they should just set it in their bashrc/bash_profile.

pavelzw (approved):
Thanks! 💜 LGTM

pavelzw:
> trying to act as a drop in replacement for conda while at the same time building out its own configs (mambarc etc) is very confusing

I agree!
I personally also think that the main reason for existence for mamba is now gone since conda also implemented the libmamba parser.

jonashaag:
🎉

jonashaag:
@danpf Interestingly for me this doesn't work if I don't create an empty `~/.config/mamba` folder:

```
$ pwd
/Users/j/.config

$ micromamba info | grep populated
 populated config files : /Users/j/.config/mamba/../conda/.condarc

$ rmdir mamba/
$ micromamba info | grep populated
 populated config files :
```

danpf:
@jonashaag It looks like the code has been significantly refactored since this was merged. Although my long comment is still there, (seems like it should be deleted now) it looks like someone expanded the config/environment.  So I don't think you should still see the `../conda` filepaths at all.

jonashaag:
Not sure what you are referring to. Your tests still seem to be there. So if the tests used to worked we should have realized if something broke. I'll try to investigate.

jonashaag:
Sorry, now I understand where the changes are coming from: https://github.com/mamba-org/mamba/pull/2983

Issue #2697, opened 2023-07-21
How to see effective yaml configuration?


chebee7i:
The documentation at https://mamba.readthedocs.io/en/latest/user_guide/configuration.html is very useful.  As part of it, there is a segment which says:

-----

The resulting configuration written using YAML syntax is:
```console
channels:
  - my-channel  # 'CLI'
  - conda-forge  # '~/.mambarc'
always_yes: true  # 'MAMBA_ALWAYS_YES' > '~/.mambarc'
```

----

Is there any way that we can generate such a yaml file (bonus: with attribution included)?

danpf:
https://github.com/mamba-org/mamba/issues/2697
`micromamba config list --sources`

run `micromamba config list --help` for other arguments to alter the output with more or less descriptors.

PR #2717, opened 2023-07-28
Add command to docs for completeness

danpf:
Very simple doc change, re: https://github.com/mamba-org/mamba/issues/2697
makes sense, given previous commands in this doc show their commandline to produce the shown stdout


jonashaag (approved):
Thanks!

PR #2725, opened 2023-08-02
Update installation.rst

JacekHoleczek:
Fix a "micromamba config" command.


JacekHoleczek:
I can now see that the same fix has already been proposed last week: https://github.com/mamba-org/mamba/pull/2703

PR #2727, opened 2023-08-02
Remove warnings from test_activation

AntoinePrv:


micromamba/tests/test_activation.py
```diff
@@ -279,710 +268,697 @@ def shvar(v, interpreter):
...
-
-    @pytest.mark.parametrize("interpreter", get_interpreters())
-    def test_shell_init(
-        self,
-        tmp_home,
-        winreg_value,
-        tmp_root_prefix,
-        tmp_path,
-        interpreter,
+def env_to_dict(out, interpreter="bash"):
```
AntoinePrv, line 271: All this diff is from making the tests *not* class method.
I think this is better so there is not temptation to use class attributes.

Hind-M (approved):
Okay! And what were the warnings about?

AntoinePrv:
Linter warnings such as "_reference to <...> may be undefined_" because of the `from helpers import *`.

PR #2734, opened 2023-08-04
Safe Curl opt in url.cpp

AntoinePrv:
Using RAII to prevent leaking memory

### DIFF
--- a/libmamba/src/core/url.cpp
+++ b/libmamba/src/core/url.cpp
@@ -21,11 +21,15 @@ namespace mamba
     namespace
     {
         /**
-         * A RAII `CURL*` createed from `curl_easy_handle`.
+         * A RAII ``CURL*`` created from ``curl_easy_handle``.
          *
          * Never null, throw exception at construction if creating the handle fails.
+         * This wrapper is not meant to handle more cases, it remains an implementation detail
+         * for url parsing, which we should ideally do without Curl.
+         * This differs from ``CURLHanlde`` in "curl.hpp" which is a fully-featured CURL wrapper
+         * for accessing the internet.
          */
-        class CurlEasyHandle
+        class CURLEasyHandle
         {
         public:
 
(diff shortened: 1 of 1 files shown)


libmamba/src/core/url.cpp
```diff
@@ -15,6 +17,152 @@
...
+
+    namespace
+    {
+        /**
+         * A RAII `CURL*` createed from `curl_easy_handle`.
+         *
+         * Never null, throw exception at construction if creating the handle fails.
+         */
+        class CurlEasyHandle
+        {
```
JohanMabille, line 29: There is already such a handle in curl.hpp that could be used instead.

libmamba/src/core/url.cpp
```diff
@@ -15,6 +17,152 @@
...
+
+        CurlStr::CurlStr(pointer data, size_type len)
+            : m_data(data)
+            , m_len(len)
+        {
+        }
+
+        CurlStr::~CurlStr()
+        {
+            free_data_if_present();
```
JohanMabille, line 119: There is an issue here: if a CurlStr is copied or moved, you have a double delete.

libmamba/src/core/url.cpp
```diff
@@ -15,6 +17,152 @@
...
+        }
+
+        auto CurlStr::raw_input() -> input_pointer
+        {
+            free_data_if_present();
```
JohanMabille, line 124: I think this should not be called here

libmamba/src/core/url.cpp
```diff
@@ -15,6 +17,152 @@
...
+                {
+                    return { { m_data } };
+                }
+            }
+            return std::nullopt;
+        }
+
+        void CurlStr::free_data_if_present()
+        {
+            if (m_data)
```
JohanMabille, line 156: This is not required, calling free (or ant curl-equivalent) on a null pointer is ok (it's a noop)

JohanMabille (changes requested):

libmamba/src/core/url.cpp
AntoinePrv: Should we throw instead if there is data present?

libmamba/src/core/url.cpp
AntoinePrv: Will delete those.

libmamba/src/core/url.cpp
AntoinePrv: There are some differences:
 - `~CurlHandle` calls ` curl_slist_free_all(p_headers)` which was not the case here.
 - There is no `CURL* raw();` which was the point here.
 
 Overall `curl.hpp` seems to be intended for actually using CURL to connect to the Internet, whereas here it is only a implementation detail for parsing urls (or at least what I'm trying to move it forward)

Klaim: If we keep that new RAII type, please add a comment over the class to explain why this exist and how it is different from the CurlHandle, like you did here.

libmamba/src/core/url.cpp
Klaim: If it's a development mistake when the data is not present, put an assert.

libmamba/src/core/url.cpp
JohanMabille: - freeing the p_headers is not  an issue, the pointer is null constructed so it is a noop if you don't explicitely add any header.
- curl should wrapped instead of beeing exposed through a raw accessor, thus the usage of the wrapper in curl.


Anyway, if you prefer to keep.this tony wrapper (which is not a problem asnlong as it is not exposed in any header) you need to implement proper valu semantics; the current implementation has the same issue as CurlStr regarding the destructor.

PR #2738, opened 2023-08-07
Remove created prefix if aborted with `--platform`

### LINKED ISSUE #2736: "Non-conda folder exists at prefix" after aborted env creation with `--platform`
With Micromamba 1.4.9, on osx-arm64:

```
$ micromamba create -p /tmp/env --platform osx-64 zlib
Confirm changes: [Y/n] n  <=== abort this
$ micromamba create -p /tmp/env --platform osx-64 zlib
error    libmamba Non-conda folder exists at prefix
```

Hind-M:
Fix #2736

### DIFF
--- a/libmamba/include/mamba/api/create.hpp
+++ b/libmamba/include/mamba/api/create.hpp
@@ -19,8 +19,11 @@ namespace mamba
 
     namespace detail
     {
-        void
-        store_platform_config(const fs::u8path& prefix, const std::string& platform, bool& remove_prefix);
+        void store_platform_config(
+            const fs::u8path& prefix,
+            const std::string& platform,
+            bool& remove_prefix_on_failure
+        );
     }
 }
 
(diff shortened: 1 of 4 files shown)


jonashaag:
Thanks, LGTM, although I wonder if we can't create the file only after successful creation?

Otherwise instead of prompting the user which might be confusing, can we detect if the env folder was created by Mamba and if yes delete it?

Hind-M:
> Thanks, LGTM, although I wonder if we can't create the file only after successful creation?

I suppose we theoretically can, but it would probably be more prone to future bugs? The prefix directory is created to update the given platform info in the `.mambarc` file which could be required somehow, somewhere...

> Otherwise instead of prompting the user which might be confusing, can we detect if the env folder was created by Mamba and if yes delete it?

Yeah I changed it now to detect creation and avoid passing `config` too...

jonashaag (approved):
Thanks! I suggest renaming to `remove_prefix_on_failure` or something like that because we don't remove it in all cases

PR #2740, opened 2023-08-08
Revert "Remove created prefix if aborted with `--platform` (#2738)"

Hind-M:
This reverts commit beebd7ee4720e6d972c54c4fd95a7ade821b1af6.

### DIFF
--- a/libmamba/include/mamba/api/create.hpp
+++ b/libmamba/include/mamba/api/create.hpp
@@ -19,11 +19,7 @@ namespace mamba
 
     namespace detail
     {
-        void store_platform_config(
-            const fs::u8path& prefix,
-            const std::string& platform,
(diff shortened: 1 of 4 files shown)


jonashaag:
I don't think the CI failure is related

Hind-M:
Reverting here is intended to see if the failure persists... We'll see (maybe a bad merge with the previous [commit](https://github.com/mamba-org/mamba/commit/3c3fa084e23270bb3e93ad1616169331d46996f3)?)

Hind-M:
@AntoinePrv maybe we should wait until we see how it goes with #2741?

JohanMabille (changes requested):
We should close this, see my comment in #2741

PR #2741, opened 2023-08-08
Try fixing #2736 again

Hind-M:


Hind-M:
The failing test seems to be `test_update.py` with an error in the solver `nothing provides requested "pkg"`. It also shows `ERROR at setup` on all the failing tests and I suspect that this would be related to the changes on the setup of `micromamba` tests using fixtures and not classes anymore? `test_update` still uses classes...

Hind-M:
Closing this as it showed its purpose.

PR #2744, opened 2023-08-10
Isolate URL object

AntoinePrv:
- Hide CURL
- Refactor `URLHandler` as a `URL` class with strong invariant.
- Manage '%' encoding.
- Rewrite a number of function to make them independent of CURL / <regex>
- Add many tests
- Close #2611

### DIFF
--- a/libmamba/include/mamba/util/url.hpp
+++ b/libmamba/include/mamba/util/url.hpp
@@ -41,7 +41,9 @@ namespace mamba::util
          * The fields of the URL must be percent encoded, otherwise use the individual
          * field setters to encode.
          * For instance, "https://user@email.com@mamba.org/" must be passed as
-         *"https://user%40email.com@mamba.org/".
+         *"https://user%40email.com@mamba.org/".The first '@' charater is part of the username
+         * "user@email.com" whereas the second is the URL specification for separating username
+         * and hostname.
          *
          * @see Encode
          * @see mamba::util::url_encode
--- a/libmamba/include/mamba/util/url.hpp
+++ b/libmamba/include/mamba/util/url.hpp
@@ -44,19 +44,19 @@ namespace mamba::util
         [[nodiscard]] auto scheme() const -> const std::string&;
 
         /** Set a non-empty scheme. */
-        URL& set_scheme(std::string_view scheme);
+        auto set_scheme(std::string_view scheme) -> URL&;
 
         /** Return the user, or empty if none. */
         [[nodiscard]] auto user() const -> const std::string&;
 
         /** Set or clear the user. */
-        URL& set_user(std::string_view user);
+        auto set_user(std::string_view user) -> URL&;
 
         /** Return the password, or empty if none. */
         [[nodiscard]] auto password() const -> const std::string&;
 
         /** Set or clear the password. */
-        URL& set_password(std::string_view password);
+        auto set_password(std::string_view password) -> URL&;
 
         /** Return the basic authetification string. */
         [[nodiscard]] auto authentication() const -> std::string;
@@ -65,13 +65,13 @@ namespace mamba::util
         [[nodiscard]] auto host() const -> const std::string&;
 
         /** Set a non-empty host. */
-        URL& set_host(std::string_view host);
+        auto set_host(std::string_view host) -> URL&;
 
         /** Return the port, or empty if none. */
         [[nodiscard]] auto port() const -> const std::string&;
 
         /** Set or clear the port. */
-        URL& set_port(std::string_view port);
+        auto set_port(std::string_view port) -> URL&;
 
         /** Return the autority part of the URL. */
         [[nodiscard]] auto authority() const -> std::string;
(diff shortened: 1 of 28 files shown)


libmamba/tests/src/util/test_string.cpp
```diff
@@ -72,6 +72,34 @@ namespace mamba::util
...
+            CHECK_EQ(remove_prefix("", ""), "");
+            CHECK_EQ(remove_prefix("hello", ""), "hello");
+            CHECK_EQ(remove_prefix("hello", "hello"), "");
+            CHECK_EQ(remove_prefix("", "hello"), "");
+            CHECK_EQ(remove_prefix("https://localhost", "https://"), "localhost");
+            CHECK_EQ(remove_prefix("https://localhost", "http://"), "https://localhost");
+            CHECK_EQ(remove_prefix("aabb", "a"), "abb");
+            CHECK_EQ(remove_prefix("", 'a'), "");
+            CHECK_EQ(remove_prefix("a", 'a'), "");
+            CHECK_EQ(remove_prefix("aaa", 'a'), "aa");
```
Hind-M, line 86: Does trying to remove a suffix with `remove_prefix` and vice versa work?
i.e: `CHECK_EQ(remove_prefix("aabb", "b"), "aabb");`

libmamba/tests/src/util/test_url.cpp
```diff
@@ -0,0 +1,432 @@
...
+            CHECK_EQ(url.path(), "/");
+            CHECK_EQ(url.pretty_path(), "/");
+            CHECK_EQ(url.user(), "userx123");
+            CHECK_EQ(url.password(), "üúßsajd");
+            CHECK_EQ(url.port(), "");
+            CHECK_EQ(url.query(), "");
+            CHECK_EQ(url.fragment(), "");
+        }
+
+        SUBCASE("http://user%40email.com:test@localhost:8000")
```
Hind-M, line 161: Maybe add a test for `http://user@email.com:test@localhost:8000`?

libmamba/tests/src/util/test_url.cpp
```diff
@@ -0,0 +1,432 @@
...
+            CHECK_EQ(url.scheme(), "ftp");
+            CHECK_EQ(url.host(), "[2001:db8:85a3:8d3:1319:0:370:7348]");
+            CHECK_EQ(url.path(), "/page");
+            CHECK_EQ(url.pretty_path(), "/page");
+            CHECK_EQ(url.user(), "user");
+            CHECK_EQ(url.password(), "pass");
+            CHECK_EQ(url.port(), "9999");
+            CHECK_EQ(url.query(), "");
+            CHECK_EQ(url.fragment(), "");
+        }
```
Hind-M, line 262: Should we add a subcase test with `fragment` not empty?

libmamba/include/mamba/util/url.hpp
```diff
@@ -0,0 +1,164 @@
...
+        inline static constexpr std::string_view https = "https";
+        inline static constexpr std::string_view localhost = "localhost";
+
+        /**
+         * Create a URL from a string.
+         *
+         * The fields of the URL must be percent encoded, otherwise use the individual
+         * field setters to encode.
+         * For instance, "https://user@email.com@mamba.org/" must be passed as
+         *"https://user%40email.com@mamba.org/".
```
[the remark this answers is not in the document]
AntoinePrv: The second `@` must not be encoded in this example, it's the URL separators between user/password and hostname, whereas the first one is a character of the username.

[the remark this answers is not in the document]
AntoinePrv: Adding more doc.

libmamba/tests/src/util/test_url.cpp
AntoinePrv: It will be a parsing error. Adding a negative test.

Hind-M (approved):

AntoinePrv:
I'm merging for moving forward on other PR and fixing the CI, but feel free to add reviews, I will implement them in another PR.

AntoinePrv:
@Klaim @JohanMabille Do you think it is a good idea to use `fs::path` as the path component of this `URL`?
I like the idea of being able to manipulate a path as a path and not a string, but this come with a number of issues on Windows...
If only we had a split `Unixpath`/`WindowsPath`/`Path` like in Python...

Hind-M:
> @Hind-M @Klaim @JohanMabille Do you think it is a good idea to use `fs::path` as the path component of this `URL`? I like the idea of being able to manipulate a path as a path and not a string, but this come with a number of issues on Windows... If only we had a split `Unixpath`/`WindowsPath`/`Path` like in Python...

Hmm I guess you won't know if you don't give it a try ^^

JohanMabille:
> @Hind-M @Klaim @JohanMabille Do you think it is a good idea to use `fs::path` as the path component of this `URL`? I like the idea of being able to manipulate a path as a path and not a string ...

I think this would be a wrong good idea ;) Especially when you have API like this one:

```cpp
        /**
         * Append a sub path to the current path.
         *
         * Contrary to `std::filesystem::path::append`, this always append and never replace
         * the current path, even if @p subpath starts with a '/'.
         */
        auto append_path(std::string_view subpath) -> URL&;
```

If the path is stored (and returned) as a `fs::path` object, this would definitely create confusion. Also it is quite easy to build a `fs::path` from the path component of the URL when it is required / makes sense.

JohanMabille (commented):
Post merge review: LGTM , nothing to change so far.

Issue #2692, opened 2023-07-20
Make virtual package handling consistent with Conda


jonashaag:
- `__archspec`:
  - Missing `CONDA_OVERRIDE_ARCHSPEC` support
  - Missing support for arm64 and others
  - See [get_archspec_name](https://github.com/search?q=repo%3Aconda%2Fconda%20get_archspec_name&type=code)
- `__cuda`:
  - Not sure, see https://github.com/conda/conda/blob/main/conda/plugins/virtual_packages/cuda.py
- `__linux`:
  - Setting `CONDA_SUBDIR=linux-64` should make `__linux` available (if no kernel info can be parsed, should be 0). See https://github.com/conda/conda/blob/main/conda/plugins/virtual_packages/linux.py
  - Same for `__win` and `__osx`.

PR #2749, opened 2023-08-14
Fix __linux virtual package default version

jonashaag:
This fixes the last bullet point in https://github.com/mamba-org/mamba/issues/2692.


jonashaag:
I'd like to get this into today's next release

AntoinePrv (approved):

jonashaag:
Thank you!❤️

JohanMabille:
Actually I think the test is not correct, it does not set `Context::platform` (via `CONDA_SUBDIR` for instance) to `linux-64`, thus the failure on Windows.

PR #2753, opened 2023-08-16
Read repodata.json using nl::json (rerun)

AntoinePrv:
Rerun #2688


AntoinePrv:
So this is still timing out... nlohmann json is too slow for our case.

AntoinePrv:
In the end
- Parsing directly from simdjson into libsolv (experimental)
- Remove `.solv` file on windows since it is slower that parsing the json.

JohanMabille (approved):

PR #2755, opened 2023-08-16
Fix and improve static builds

AntoinePrv:


jonashaag:
@pavelzw

AntoinePrv:
This is back to the error we get on `main` (well other PRs since `main` is older), which I believe means we need to update the patch.

As an alternative, in https://github.com/conda-forge/micromamba-feedstock/pull/136 I try to allow a way to use a local source. It works locally.

AntoinePrv:
@pavelzw any idea why we are seeing https://github.com/mamba-org/setup-micromamba/issues/68 on OSX builds?

AntoinePrv:
Great, this works!

AntoinePrv:
The remaining failures are unrelated

Hind-M (approved):
Great thanks! Is the `umamba_tests` failure on macos unrelated too?

jonashaag:
> @pavelzw any idea why we are seeing [mamba-org/setup-micromamba#68](https://github.com/mamba-org/setup-micromamba/issues/68) on OSX builds?

I created https://github.com/mamba-org/setup-micromamba/issues/130

PR #2762, opened 2023-08-18
[CI] Add `Debug` build type

Hind-M:


AntoinePrv (commented):
I agree we should find a solution for this, but will this not make the CI very slow?

IMHO what we need is to run `assert`, what if we replace them with a custom `MAMBA_ASSERT` for which we can control when it is ON or OFF?

Hind-M:
> I agree we should find a solution for this, but will this not make the CI very slow?

Yeah most likely, or we could only run with Debug instead of Release (at least for `unix`).

> IMHO what we need is to run `assert`, what if we replace them with a custom `MAMBA_ASSERT` for which we can control when it is ON or OFF?

Could be a solution too.

JohanMabille (approved):

Issue #2724, opened 2023-08-01
[`win/powershell`] Installing into a different environment throws an error


dhirschfeld:
### Troubleshooting docs

- [X] My problem is not solved in the Troubleshooting docs

### How did you install Mamba?

Micromamba

### Search tried in issue tracker

-

### Latest version of Mamba

- [X] My problem is not solved with the latest version

### Tried in Conda?

I didn't try

### Describe your issue

If you try to install into a a named environment when no current environment is active the powershell script throws an error trying to activate an empty string (after the install completes successfully)
```powershell
micromamba create -n bug-2724
# ensure no environment is active
micromamba deactivate
micromamba deactivate
micromamba install -n bug-2724 --yes bat
```
```powershell
<snip>
Transaction finished

To activate this environment, use:

    micromamba activate bug-2724

Or to execute a single command in this environment, use:

    micromamba run -n bug-2724 mycommand

Invoke-Expression : Cannot bind argument to parameter 'Command' because it is an empty string.
At line:120 char:48
+                     Invoke-Expression -Command $activateCommand;
+                                                ~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidData: (:) [Invoke-Expression], ParameterBindingValidationException
    + FullyQualifiedErrorId : ParameterArgumentValidationErrorEmptyStringNotAllowed,Microsoft.PowerShell.Commands.InvokeExpressionCommand
```

### mamba info / micromamba info

```shell
libmamba version : 1.4.9
     micromamba version : 1.4.9
           curl version : libcurl/8.1.2-DEV Schannel zlib/1.2.13
     libarchive version : libarchive 3.6.2 zlib/1.2.13 liblzma/5.4.3 bz2lib/1.0.8 liblz4/1.9.4 libzstd/1.5.5
       envs directories : C:\conda\envs
          package cache : C:\conda\pkgs
                          C:\conda\pkgs
            environment : None (not found)
           env location : -
      user config files : C:\Users\hirdav.FFMA\.mambarc
 populated config files : C:\conda\.mambarc
       virtual packages : __win=0=0
                          __archspec=1=x86_64
               channels : https://conda.anaconda.org/conda-forge/win-64
                          https://conda.anaconda.org/conda-forge/noarch
       base environment : C:\conda
               platform : win-64
```

dhirschfeld:
The offending code:
[...]

PR #2763, opened 2023-08-19
fix: only reactivate current environment

chawyehsu:
Fixes: #2724

### DIFF
--- a/libmamba/data/Mamba.psm1
+++ b/libmamba/data/Mamba.psm1
@@ -115,11 +115,11 @@ function Invoke-Mamba() {
                 # reactivate environment
                 if (@("install", "update", "remove").contains($Command))
                 {
-                   $activateCommand = (& $Env:MAMBA_EXE shell reactivate -s powershell | Out-String);
-                   if ($activateCommand) {
-                       Write-Verbose "[micromamba shell reactivate --shell powershell]`n$activateCommand";
-                       Invoke-Expression -Command $activateCommand;
-                   }
+                    $activateCommand = (& $Env:MAMBA_EXE shell reactivate -s powershell | Out-String);
+                    if ($activateCommand) {
+                        Write-Verbose "[micromamba shell reactivate --shell powershell]`n$activateCommand";
+                        Invoke-Expression -Command $activateCommand;
+                    }
                 }
             }
         }
--- a/libmamba/data/Mamba.psm1
+++ b/libmamba/data/Mamba.psm1
@@ -113,26 +113,13 @@ function Invoke-Mamba() {
                 & $Env:MAMBA_EXE $Command @OtherArgs;
 
                 # reactivate environment
-                if (@("install", "update", "remove").contains($Command)) {
-                    if ($Env:CONDA_DEFAULT_ENV) {
-                        # A dead simple way to get the possible environment name,
-                        # check for '-n' or '--name' and grab the next argument.
-                        # It is then used to compare against the current environment
-                        # and the environment reactivates only if the name matches.
-                        $argEnv = $null;
-                        for ($i = 0; $i -lt $OtherArgs.Count; $i++) {
-                            if ($OtherArgs[$i] -eq "-n" -or $OtherArgs[$i] -eq "--name") {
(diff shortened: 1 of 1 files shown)


jonashaag:
Thank you for working on this. Why is this necessary for Powershell but not the other shells?

chawyehsu:
Theoretically, if it's not the current environment, then don't reactivate is the proper behavior for all shells. But it seems all other shells except for powershell can trigger reactivation and ignore the error of no-op silently. The check of matching current environment can be added to all other shells but I doubt the necessity.

jonashaag (changes requested):
We should instead change the C++ code that handles `shell reactivate -s powershell` to return a noop that doesn't cause an error in Powershell

jonashaag:
> The check of matching current environment can be added to all other shells

Alternatively, we can do this for all shells. But it's confusing to have this implemented for a single shell only

chawyehsu:
> We should instead change the C++ code that handles shell reactivate -s powershell to return a noop that doesn't cause an error in Powershell.

In fact the c++ code already did the right part, returning empty for shells. The problem is on the powershell side, powershell throws error trying to execute the noop empty script block, and that's reason it needs the patch.

Another way to fix this it what [dhirschfeld](https://github.com/dhirschfeld) commented in https://github.com/mamba-org/mamba/issues/2724#issuecomment-1663081812, which is simpler (just ignore the return) but less semantic context.

jonashaag:
Can't we make the C++ code output something that's a noop but syntactically correct?

chawyehsu:
Maybe. But what's the consideration of specializing the output from C++ for one shell, which from another point of view can also be a confusion? Fixing the issue on the shell side for the shell is the simplest and safe way I could propose.

jonashaag:
I consider it a bug in the Powershell C++ implementation. The interface is something along the lines of: "Given an environment, `reactivate` outputs a script that can be run in the shell to refresh the shell." Right now the Powershell code doesn't implement this interface correctly, while other shells do.

chawyehsu:
> "Given an environment, reactivate outputs a script that can be run in the shell to refresh the shell."

I see you point, but let me illustrate the current behavoir. The abstract `reactivate` C++ interface returns empty when there is no activating environment, this is correct since you won't need to reactivate when you don't have one env activated.

https://github.com/mamba-org/mamba/blob/6e4d91e2eca3fd913ebf5e91f86c58749f0be179/libmamba/src/core/activation.cpp#L369-L373

The returned `EnvironmentTransform` then is wrapped by [`script()`](https://github.com/mamba-org/mamba/blob/6e4d91e2eca3fd913ebf5e91f86c58749f0be179/libmamba/src/core/activation.cpp#L668), which is a shell-specific function to turn `envt` into shell script. The one of PowerShell:

https://github.com/mamba-org/mamba/blob/6e4d91e2eca3fd913ebf5e91f86c58749f0be179/libmamba/src/core/activation.cpp#L1045-L1080

Nothing special for the PowerShell C++ implementaion, it will return nothing when the `EnvironmentTransform` argument is empty, like `script()` of other shells.

Shells like bash [`eval`](https://github.com/mamba-org/mamba/blob/main/libmamba/data/micromamba.sh#L21) the empty script without issue. However, mamba's PowerShell hook is using the [`Invoke-Expression -Command`](https://github.com/mamba-org/mamba/blob/main/libmamba/data/Mamba.psm1#L120) which can not process empty script silently, `Invoke-Expression` requries a non-empty script block, and this is #2724.

Hence, either you fix it by ignoring the empty return (the shell isde), or returning a non-empty script even if it's a no-op (the C++ side). But what non-empty script should be returned when you want to fix it on the C++ side...

jonashaag:
Thanks for the detailed analysis. I'm fine with keeping the C++ as is. Would you mind reducing your PR to only the non empty check? We can do the other check in another PR that updates all shells.

jonashaag (approved):

jonashaag:
Thank you!

AntoinePrv:
@jonashaag Can you please stop merging PR when the CI is not green.

PR #2764, opened 2023-08-20
fix: Shell completion section title missing

wy-luke:
The `Shell completion` section title is missing in the latest docs.  I'm not sure whether my edits is right or not. I just follow the style above.


jonashaag (approved):
Thank you!

wy-luke:
> Thank you!

You're welcom. It's my pleasure to assist!

PR #2769, opened 2023-08-22
Fix fish completion

soraxas:
The fish completion for environments completion is currently broken.

That's due to some changes to `micromamba`'s output.

I've tested that in previous `micromamba` version:
```sh
$ ./bin/micromamba --version
1.2.0

$ ./bin/micromamba env list

                                           __
      /_/

  Name     Active  Path                             
──────────────────────────────────────────────────────
  base             /home/soraxas/micromamba             
  my_other_env     /home/soraxas/micromamba/envs/octomap
...
```

Current latest version:
```sh
$ micromamba --version
1.4.6

$ micromamba env list
  Name     Active  Path                             
──────────────────────────────────────────────────────
  base             /home/soraxas/micromamba             
  my_other_env     /home/soraxas/micromamba/envs/octomap
...
```

Notice the missing $\mu$-mamba logo

----------------------------

We can fix the completion by changing the `tail` argument from `11` to `2`, but I opt to a more "robust" approach using `sed` that currently works in both case.

But this is based on the assumption that
1. `base` environment always exists, and
2. `base` environment is always the first environment.

Are these assumption always correct?

### DIFF
--- a/libmamba/data/mamba.fish
+++ b/libmamba/data/mamba.fish
@@ -77,7 +77,7 @@ end
 # Autocompletions below
 
 function __fish_mamba_envs
-  micromamba env list | sed -n '/^[[:space:]]\+base[[:space:]]\+/,$p' | cut -d' ' -f3
+  micromamba env list | tail -n +3 | cut -d' ' -f3
 end
 
 function __fish_mamba_packages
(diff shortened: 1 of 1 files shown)


jonashaag (approved):

jonashaag:
Thanks for the fix! Unfortunately both assumptions are incorrect I think

PR #2777, opened 2023-08-23
Revert failing test

jonashaag:
Revert a wrong test I introduced in https://github.com/mamba-org/mamba/pull/2749.

I'm very sorry about this. I should have waited for CI to be green. We had a red CI for the previous commits and I blindly assumed that my changes weren't responsible for the red CI.


JohanMabille:
No worries ;) I think the test can be fixed by setting the `Context::platform` (via the `CONDA_SUBDIR` environment variable for instance) to linux-64. This can be done after the release though.

JohanMabille (approved):

PR #2779, opened 2023-08-24
Splitted GHA workflow

JohanMabille:
This is to avoid Unix jobs interruptions due to failure on Windows and vice versa (Tranxen200 TM).


.github/workflows/unix.yml
```diff
@@ -0,0 +1,254 @@
+name: Unix build
```
AntoinePrv, line 1: Nit: should we change the name to something other than "build"? Otherwise it is confusing what is the difference between "Unix build" and Unix build in "static builds". Maybe "Unix tests"?
"CI" was a terrible name, thanks for changing it.

JohanMabille: I agree, XXX tests sounds better

AntoinePrv (approved):

PR #2786, opened 2023-08-25
Fix pins

### LINKED ISSUE #2737: Python pin is ignored
```
micromamba create -p $PWD/env 'python<3' --platform osx-64
```

```
micromamba install -p $PWD/env black
```

There is no version of black that supports Python 2 so this will suggest to upgrade to Python 3.

I think we should fail hard instead.

AntoinePrv:
- Add libsolv test scenarios
- Fix pins
- Close #2737

### DIFF
--- a/libmamba/src/core/solver.cpp
+++ b/libmamba/src/core/solver.cpp
@@ -234,10 +234,12 @@ namespace mamba
         // Necessary for attributes to be properly stored
         installed->internalize();
 
-        // Lock the dummy solvable so that it stays install.
+        // WARNING keep separate or libsolv does not understand
         // Force verify the dummy solvable dependencies, as this is not the default for
         // installed packages.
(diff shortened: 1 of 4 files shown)


AntoinePrv:
@jonashaag would you mind giving this one a try? I'll add a umamba_test.

jonashaag:
Thanks, this works!

```
error    libmamba Could not solve for environment specs
    The following packages are incompatible
    ├─ numpy * py311* is installable and it requires
    │  └─ python >=3.11,<3.12.0a0 , which can be installed;
    └─ pin-1 is not installable because it requires
       └─ python 3.10** , which conflicts with any installable versions previously reported.
```

The `pin-1 is not installable` line is a bit confusing though

AntoinePrv:
I know but this will be easier to do with the new solver API.

Issue #2705, opened 2023-07-25
Access denied when using `self-update`


jonashaag:
```
micromamba --version # 1.4.9
# Downgrade to 1.4.5
micromamba self-update --version '1.4.5'
```

```
critical libmamba rename: Access is denied.: "C:\Users\USER\.bin\m.exe", "C:\Users\USER\.bin\m.exe.bkup"
```

jonashaag:
I removed `m.exe.bkup`, then it worked but I got

```
The following arguments were not expected: C:\Users\USER\micromamba -p
Run with --help for more information.
```

jonashaag:
Similar issue on Linux

```
error    libmamba Could not lock non-existing path '/home/USER/micromamba/conda-meta'

Transaction starting
Download only - packages are downloaded and extracted. Skipping the linking phase.
The following arguments were not expected: /home/USER/micromamba -p
```

wolfv:
Hmm – the first issue looks like a Windows issue about not making it easy to replace a file that is in use. I need to check, but I think we wrote some additional wrapper stuff in the cmd.exe hooks to take care of that, and I didn't think of the "bash / cygwin" use case ... 

Regarding the issue on Linux I am wondering if it has to do with some of the `shell` command modifications that were recently done (cc @AntoinePrv)

AntoinePrv:
#2787 should do for the extra argument. I think this currently working as intended, `reinit` seems to be a guard againt possible change in the initialization script.

It the Windows still an issue?

PR #2787, opened 2023-08-25
Fix extra agrument in self-update reinit

AntoinePrv:
Fix comment mentioned in #2705

This should do it, I am not aware we need to take any more arguments in `reinit`.


jonashaag:
Shouldn't it be `-r <root_prefix>`?

AntoinePrv:
We could, but that I think this is othogonal, here we would still call it without any option.
Also `reinit` is basically looping over `init` commands, so we are not missing much.

JohanMabille (approved):

PR #2790, opened 2023-08-28
Bring back repodata_use_zst

jonashaag:
Some enterprise Conda channel sources like Artifactory don't properly invalidate the `repodata.json.zst`. This adds back the flag to disable reading the `zst` file

### DIFF
--- a/libmamba/src/core/subdirdata.cpp
+++ b/libmamba/src/core/subdirdata.cpp
@@ -600,30 +600,33 @@ namespace mamba
             auto& ctx = Context::instance();
             if (!ctx.offline || forbid_cache())
             {
-                bool has_value = m_metadata.has_zst.has_value();
-                bool is_expired = m_metadata.has_zst.has_value()
-                                  && m_metadata.has_zst.value().has_expired();
-                bool has_zst = m_metadata.check_zst(channel_context, p_channel);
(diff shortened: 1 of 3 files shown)


libmamba/src/core/subdirdata.cpp
```diff
@@ -600,30 +600,33 @@ namespace mamba
             auto& ctx = Context::instance();
             if (!ctx.offline || forbid_cache())
             {
-                bool has_value = m_metadata.has_zst.has_value();
-                bool is_expired = m_metadata.has_zst.has_value()
-                                  && m_metadata.has_zst.value().has_expired();
-                bool has_zst = m_metadata.check_zst(channel_context, p_channel);
-                if (!has_zst && (is_expired || !has_value))
+                if (ctx.repodata_use_zst)
```
AntoinePrv, line 603: No else clause? What happens when we deactivate it?

jonashaag: Consider this the same case as `!has_zst && !is_expired && has_value)` in the old code. No zst checking is done. Not sure exactly how the code that runs after this works, but if I'm not mistaken it should not change behaviour.

AntoinePrv (approved):

JohanMabille (approved):

PR #2791, opened 2023-08-28
Drop unused code

jonashaag:


JohanMabille:
Not sure why you are removing these functions. They are part of the public C++ API, so they might be used by external projects based on libmamba. Besides, they come with other functions related to rc confugration that you keep, so I don't understand.

jonashaag:
It doesn't have a single test though 🤨

JohanMabille:
Indeed, that something we should add.

PR #2798, opened 2023-08-29
Parse subdirs in CLI match specs

### LINKED ISSUE #2792: Subdir in match spec does not take multichannel into account
### Troubleshooting docs

Micromamba

subdir multichannel

### Latest version of Mamba

### Tried in Conda?

I didn't try

### Describe your issue

If you have this config

```
custom_multichannels:
  my-conda-forge:
    - conda-forge
```

This works:

```
micromamba create -p /tmp/env 'my-conda-forge::xz' --dry-run
```

But this doesn't; it tries to download the `my-conda-forge` repodata:

```
micromamba create -p /tmp/env 'my-conda-forge/osx-arm64::xz' --dry-run
```

I wonder if this is related to #2788

_No response_

### Logs

_No response_

### environment.yml

_No response_

### ~/.condarc

_No response_

jonashaag:
Fixes #2792

For example, the channel in `install conda-forge/noarch::tzdata` would previously be parsed as `conda-forge/noarch` while it should be parsed as `conda-forge` with `noarch` subdir.

### DIFF
--- a/libmamba/include/mamba/core/channel.hpp
+++ b/libmamba/include/mamba/core/channel.hpp
@@ -18,6 +18,14 @@
 
 namespace mamba
 {
+    // ATTENTION names with substrings need to go longer -> smalle
+    // otherwise linux-ppc64 matches for linux-ppc64le etc!
+    const std::vector<std::string> KNOWN_PLATFORMS = {
+        "noarch",       "linux-32",      "linux-64",    "linux-aarch64", "linux-armv6l",
+        "linux-armv7l", "linux-ppc64le", "linux-ppc64", "osx-64",        "osx-arm64",
+        "win-32",       "win-64",        "win-arm64",   "zos-z"
+    };
+
     // Note: Channels can only be created using ChannelContext.
     class Channel
     {
--- a/libmamba/include/mamba/util/url_manip.hpp
+++ b/libmamba/include/mamba/util/url_manip.hpp
@@ -39,13 +39,8 @@ namespace mamba::util
         bool with_credential
     );
 
-    void split_platform(
-        const std::vector<std::string>& known_platforms,
-        const std::string& url,
-        const std::string& context_platform,
-        std::string& cleaned_url,
-        std::string& platform
-    );
+    std::pair<std::string, std::optional<std::string>>
+    split_platform(const std::string& url, const std::vector<std::string>& known_platforms);
 
     /**
      * If @p url starts with a scheme, return it, otherwise return empty string.
--- a/libmamba/src/core/channel.cpp
+++ b/libmamba/src/core/channel.cpp
@@ -42,14 +42,6 @@ namespace mamba
 
         const char LOCAL_CHANNELS_NAME[] = "local";
         const char DEFAULT_CHANNELS_NAME[] = "defaults";
-
-        // ATTENTION names with substrings need to go longer -> smalle
-        // otherwise linux-ppc64 matches for linux-ppc64le etc!
-        const std::vector<std::string> KNOWN_PLATFORMS = {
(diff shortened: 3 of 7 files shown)


libmamba/src/core/channel.cpp
```diff
@@ -231,19 +223,14 @@ namespace mamba
...
-                    std::string cleaned_url = "", platform = "";
-                    util::split_platform(
-                        KNOWN_PLATFORMS,
-                        value,
-                        Context::instance().platform,
-                        cleaned_url,
-                        platform
-                    );
-                    if (!platform.empty())
+                    auto all_platforms = KNOWN_PLATFORMS;
```
jonashaag, line 226: IIUC this creates a copy of `KNOWN_PLATFORMS` with the context channel added. Is that correct? I don't want to modify `KNOWN_PLATFORMS`.

JohanMabille: Indeed, this creates a copy. If you keep `all_platforms` and `context_platform` argument in the `split_platform` API, you don't need to do this.

libmamba/include/mamba/core/channel.hpp
```diff
@@ -18,6 +18,14 @@
 
 namespace mamba
 {
+    // ATTENTION names with substrings need to go longer -> smalle
+    // otherwise linux-ppc64 matches for linux-ppc64le etc!
+    const std::vector<std::string> KNOWN_PLATFORMS = {
```
JohanMabille, line 23: I think it would be better to keep the definition of this variable in the cpp file, and provide a function to get it in the header.

JohanMabille (changes requested):
I think we should keep "Fixing issues" PRs as minimal as possible, and avoid API change as much as we can.

libmamba/include/mamba/util/url_manip.hpp
```diff
@@ -39,13 +39,8 @@ namespace mamba::util
...
     );
 
-    void split_platform(
-        const std::vector<std::string>& known_platforms,
-        const std::string& url,
-        const std::string& context_platform,
-        std::string& cleaned_url,
-        std::string& platform
-    );
+    std::pair<std::string, std::optional<std::string>>
```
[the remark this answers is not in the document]
jonashaag: Break what API? I changed all code that uses this. Is this an interface used by external users?

The previous interface had an implicit flag to signal if it found a platform: The output `platform` variable is set to `""` or not. I find that much less obvious than using an optional.

I agree with restoring `context_platform`.

libmamba/include/mamba/core/channel.hpp
jonashaag: OK! I don't understand what difference that makes but not a C++ expert

libmamba/include/mamba/util/url_manip.hpp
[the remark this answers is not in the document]
JohanMabille: Indeed that's an internal API, my bad.
I agree that it's better to use an `optional` than an implicit flag for the `platform` output variable, my point was that keeping the output names was more expressive than returning a complex composed type. But as I said, a typedef on the return type can help reading it.

libmamba/include/mamba/core/channel.hpp
JohanMabille: If in the future we want to update the `KNOWN_PLATFORMS` variable:
- if `KNOWN_PLATFORMS` is defined in the header, any binary using it will have to recompile when updating, meaning we break the ABI compatibility.
- if it is defined in a source file, the client can simply upgrade the dependency with no further action required on its side.

I know this header is not really a "public" API yet (in that it's not in `include/api`), but it will likely become one after the ongoing refactoring. We want libmamba to be usable as a framework and not just be a geared towards the CLI.

jonashaag:
Superseded by #2799

PR #2799, opened 2023-08-29
Parse subdirs in CLI match specs

### LINKED ISSUE #2792: Subdir in match spec does not take multichannel into account

jonashaag:
https://github.com/mamba-org/mamba/pull/2798 without the interface changes.

Fixes #2792

For example, the channel in `install conda-forge/noarch::tzdata` would previously be parsed as `conda-forge/noarch` while it should be parsed as `conda-forge` with `noarch` subdir.

### DIFF
--- a/libmamba/src/core/channel.cpp
+++ b/libmamba/src/core/channel.cpp
@@ -231,18 +231,17 @@ namespace mamba
                 // that already contains the platform
                 else
                 {
-                    std::string cleaned_url = "", platform = "";
+                    std::string platform = "";
                     util::split_platform(
-                        KNOWN_PLATFORMS,
+                        get_known_platforms(),
                         value,
                         Context::instance().platform,
-                        cleaned_url,
+                        value,
                         platform
                     );
                     if (!platform.empty())
                     {
                         platforms.push_back(std::move(platform));
-                        value = cleaned_url;
                     }
                 }
             }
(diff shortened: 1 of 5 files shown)


libmamba/src/core/channel.cpp
```diff
@@ -231,18 +231,17 @@ namespace mamba
...
                         value,
                         Context::instance().platform,
-                        cleaned_url,
+                        value,
                         platform
                     );
                     if (!platform.empty())
                     {
                         platforms.push_back(std::move(platform));
-                        value = cleaned_url;
```
jonashaag, line 245: I changed this because I think we always want the `rstrip()` from `split_platform`.

JohanMabille: I think all usecases should be consistent, so let's rstrip all the time.

JohanMabille (approved):

PR #2805, opened 2023-09-01
Add CondaURL

AntoinePrv:


libmamba/include/mamba/specs/platform.hpp
```diff
@@ -33,12 +40,24 @@ namespace mamba::specs
...
+        // For reflexion purposes only
+        count_,
     };
 
+    constexpr auto known_platforms_count() -> std::size_t
+    {
+        return static_cast<std::size_t>(Platform::count_);
+    }
+
+    constexpr auto known_platforms() -> std::array<Platform, known_platforms_count()>;
```
JohanMabille, line 53: There is already the notion of `known_platforms` in [channel.hpp](https://github.com/mamba-org/mamba/blob/main/libmamba/include/mamba/core/channel.hpp#L23), it would make sense to rebind it to your changes (or replace its usage with your), to avoid duplication.

libmamba/include/mamba/specs/url.hpp
```diff
@@ -0,0 +1,141 @@
+// Copyright (c) 2023, QuantStack and Mamba Contributors
```
JohanMabille, line 1: I think it would be more intuitive if this file was named `conda_url.hpp`

libmamba/include/mamba/util/url.hpp
```diff
@@ -57,99 +61,156 @@ namespace mamba::util
         [[nodiscard]] auto scheme() const -> const std::string&;
 
         /** Set a non-empty scheme. */
-        auto set_scheme(std::string_view scheme) -> URL&;
+        void set_scheme(std::string_view scheme);
```
JohanMabille, line 64: It's annoying to lose the chained construction of the URL. I wonder if it could make sense to use a CRTP here, or even to merge the CondaURL and the URL classes? (do we have usages of URL where we don't want to have access to the additiona methods form CondaURL?

libmamba/include/mamba/specs/platform.hpp
JohanMabille: Yes, I meant to use the code from this PR in the channel file; but this can be done in a dedicated PR.

libmamba/include/mamba/util/url.hpp
JohanMabille: After discussion in person:
- having two distinct abstractions for URL and CondaURL is better thatn a single class; it helps to keep isolation of the dependency on CURL
- CRTP is a bit of overkill just for keeping the chained construction syntax.

JohanMabille (approved):

PR #2802, opened 2023-08-31
add `context.register_envs` to control whether environments are registered to `environments.txt` or not

jaimergp:
Closes #2783 

xref https://github.com/conda/conda/pull/12924
xref https://github.com/conda/constructor/pull/716

PR #2806, opened 2023-09-01
Wired up register_envs flag

JohanMabille:
This completes #2802
cc @wolfv


jonashaag:
Do we want to add a test for this?

wolfv:
From the python side we could also set the variable but would have to check if the car is available on conda with hasattr or so

AntoinePrv (approved):

JohanMabille:
> Do we want to add a test for this?

Yes I need to add them

Issue #2758, opened 2023-08-17
Micromamba does not read `default_channels` unless `defaults` is in the channels


jonashaag:
If you have no channels configured, Mamba defaults to whatever is in `default_channels`, and Micromamba defaults to an empty list of channels.

Is that a bug or a feature?

JohanMabille:
Sounds like a bug to me, `default_channels` should be used by default.

AntoinePrv:
Now defaults to conda-forge.

Issue #2140, opened 2022-11-29
`micromamba config set` doesn't work with empty list


jonashaag:
Micromamba 0.25:

```
micromamba config set default_channels []
vector::_M_range_check: __n (which is 1) >= this->size() (which is 1)
```

Micromamba 1:

```
$ micromamba config set default_channels []
$ grep default_channels ~/.condarc
default_channels: "\x83\x83"
```

PR #2809, opened 2023-09-05
Fix nodefaults in documentation

jonashaag:
`nodefaults` is supported only in environment.yml files

### DIFF
--- a/docs/source/micromamba-installation.rst
+++ b/docs/source/micromamba-installation.rst
@@ -126,8 +126,7 @@ An exclusive `conda-forge <https://conda-forge.org/>`_ setup can be configured w
 
 .. code-block:: bash
 
-   micromamba config append channels conda-forge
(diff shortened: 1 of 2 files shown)


docs/source/micromamba-installation.rst
```diff
@@ -126,8 +126,7 @@ An exclusive `conda-forge <https://conda-forge.org/>`_ setup can be configured w
 
 .. code-block:: bash
 
-   micromamba config append channels conda-forge
-   micromamba config append channels nodefaults
+   micromamba config append channels conda-forge`
```
jonashaag: This works only until https://github.com/mamba-org/mamba/issues/2758 is fixed. After that we need to set `default_channels: []` but that doesn't work because of https://github.com/mamba-org/mamba/issues/2140

jonashaag: It's a no-op

Issue #2666, opened 2023-07-11
Mamba 1.4.8: not a SOLV file


jonashaag:
I'm getting this with Micromamba 1.4.8, I'll need some time to come up with a reproducer

```
critical libmamba Unable to read repo solv file 'conda-forge/osx-64', error was: not a SOLV file
```

jonashaag:
I'm getting a bunch of other errors, maybe this is due to broken locking (running a test suite with concurrency)

```
mapping pread: No such file or directory
mapping pread: No such file or directory
mapping pread: No such file or directory
mapping pread: No such file or directory
mapping pread: No such file or directory
mapping pread: No such file or directory
warning  libmamba Could not validate package '/Users/runner/micromamba/pkgs/libcxx-15.0.7-h71dddab_0/info/repodata_record.json': md5 and sha256 sum unknown.
    Set safety_checks to disabled to override this warning.
warning  libmamba Extracted package cache '/Users/runner/micromamba/pkgs/libcxx-15.0.7-h71dddab_0' has no checksum
warning  libmamba Could not validate package '/Users/runner/micromamba/pkgs/libcxx-15.0.7-h71dddab_0.conda': md5 and sha256 sum unknown.
    Set safety_checks to disabled to override this warning.
warning  libmamba Could not validate package '/Users/runner/micromamba/pkgs/xtl-0.7.5-hb8565cd_0/info/repodata_record.json': md5 and sha256 sum unknown.
    Set safety_checks to disabled to override this warning.
warning  libmamba Extracted package cache '/Users/runner/micromamba/pkgs/xtl-0.7.5-hb8565cd_0' has no checksum
warning  libmamba Could not validate package '/Users/runner/micromamba/pkgs/xtl-0.7.5-hb8565cd_0.conda': md5 and sha256 sum unknown.
    Set safety_checks to disabled to override this warning.
warning  libmamba Could not validate package '/Users/runner/micromamba/pkgs/xtensor-0.23.10-h940c156_0/info/repodata_record.json': md5 and sha256 sum unknown.
    Set safety_checks to disabled to override this warning.
warning  libmamba Extracted package cache '/Users/runner/micromamba/pkgs/xtensor-0.23.10-h940c156_0' has no checksum
warning  libmamba Could not validate package '/Users/runner/micromamba/pkgs/xtensor-0.23.10-h940c156_0.tar.bz2': md5 and sha256 sum unknown.
    Set safety_checks to disabled to override this warning.
[...]

PR #2811, opened 2023-09-06
Adding locks for cache directories in case multiple processes are try…. 

rmittal87:
fixes 2666

Change-Id: Ib71f328f345569868602de32945f56d6f65096fd

### DIFF
--- a/libmamba/src/api/install.cpp
+++ b/libmamba/src/api/install.cpp
@@ -589,9 +589,16 @@ namespace mamba
             );
         }
 
-        MTransaction trans(pool, solver, package_caches);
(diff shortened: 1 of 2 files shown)


rmittal87:
fixes #2666

jonashaag:
Cool! Again, is there a way to test this? I know race conditions are hard to test, but maybe we can build something that spawns a few concurrent workers for a couple of times and if no crash is observed, it is accepted.

JohanMabille (approved):

JohanMabille:
Tahnks!

Can you run the linter to fix the CI? `pre-commit run --all-files` should do it.

Issue #1429, opened 2022-02-01
Package dir lock doesn't work as expected


grst:
From  https://github.com/mamba-org/mamba/issues/739 I understood that in https://github.com/mamba-org/mamba/pull/894 a lockfile mechanism was introduced that should enable multiple `mamba create` commands to run in parallel. 

However, this doesn't seem to work for me. When I create two environments in parallel, both processes start downloading packages concurrently instead of waiting for the lock to be released which leads to inconsistencies and finally, a failure:

```
error    File not valid: SHA256 sum doesn't match expectation "/home/sturm/anaconda3/pkgs/r-testthat-3.1.2-r41h03ef668_0.tar.bz2"
    Expected: f472c93f5fcc8ede1c4789369ca5c357d5fde5a98de963bb6be6f14651142b23
    Actual: 8f62d4455a15508487548baff427409acf1fa67b0dbcb012c4a700f052bfbd0e
[...]

>>>>>>>>>>>>>>>>>>>>>> ERROR REPORT <<<<<<<<<<<<<<<<<<<<<<

    Traceback (most recent call last):
      File "/home/sturm/anaconda3/lib/python3.7/site-packages/conda/exceptions.py", line 1080, in __call__
        return func(*args, **kwargs)
      File "/home/sturm/anaconda3/lib/python3.7/site-packages/mamba/mamba.py", line 921, in exception_converter
        raise e
      File "/home/sturm/anaconda3/lib/python3.7/site-packages/mamba/mamba.py", line 915, in exception_converter
        exit_code = _wrapped_main(*args, **kwargs)
      File "/home/sturm/anaconda3/lib/python3.7/site-packages/mamba/mamba.py", line 869, in _wrapped_main
        result = do_call(args, p)
      File "/home/sturm/anaconda3/lib/python3.7/site-packages/mamba/mamba.py", line 738, in do_call
        exit_code = create(args, parser)
      File "/home/sturm/anaconda3/lib/python3.7/site-packages/mamba/mamba.py", line 614, in create
        return install(args, parser, "create")
      File "/home/sturm/anaconda3/lib/python3.7/site-packages/mamba/mamba.py", line 562, in install
        downloaded = transaction.prompt(repos)
    RuntimeError: Found incorrect download: proj. Aborting
```

I am running mamba 0.20.0, see also the full error messages below. 

### To reproduce
```bash
# clear package cache to avoid all packages are downloaded already
conda clean -af

# Run in terminal 1
[...]

PR #2813, opened 2023-09-06
Softlinks were not getting created properly earlier. --always-softlin…

rmittal87:
…k option is now fully functional

Change-Id: I21e1c9ac4214065b551272bef788611c91917796


jonashaag:
Is there a way to test this using a unit test?

rmittal87:
fixes  #1429

rmittal87:
@jonashaag .. need to add elfutils in the reipe of micromamba-feedstock recipe for unix.  Without that build is failing.

jonashaag:
Why do we need elfutils to check if a file is executable?

rmittal87:
We are using elf library to check if a file is execulatble or not
        Elf* elf = elf_begin(fd, ELF_C_READ, NULL);
        if (!elf)
        {
            LOG_DEBUG << path.string() << " error calling elf_begin";
            close(fd);
            return false;
        }

        Elf_Kind ek = elf_kind(elf);
        if (ek != ELF_K_ELF)
        {
            LOG_DEBUG << path.string() << " not an ELF file";
            close(fd);
            return false;
        }

        Elf64_Ehdr* elfhdr = elf64_getehdr(elf);
        if (!elfhdr)
        {
            LOG_DEBUG << path.string() << " error reading ELF header";
            close(fd);
            return false;
        }
        bool rc = elfhdr->e_type == ET_EXEC;

jonashaag:
Why not just look at the executable bit or the elf header bytes?

rmittal87:
fixed the crash, added the testcase/ package for which it was crashing. @jonashaag kidnly review

PR #2815, opened 2023-09-06
Drop CONDA_OVERRIDE_WIN environment variable

jonashaag:
Unsupported by Conda and not useful in Mamba.


libmamba/include/mamba/core/util_os.hpp
```diff
@@ -38,7 +38,6 @@ namespace mamba
 
     void run_as_admin(const std::string& args);
     bool enable_long_paths_support(bool force, Palette palette = Palette::no_color());
-    std::string windows_version();
```
JohanMabille, line 41: I think this should not be removed.

jaimergp:
Hm, apparently [`conda` does not query](https://github.com/conda/conda/blob/e9870ccaaa1ee1c57ff38b075f308e6c051c72a7/conda/plugins/virtual_packages/windows.py#L13C18-L13C18) for any version number for Windows. It just reports it as `win=0=0`. It makes sense to not allow overrides for something that is always the same but the libmamba implementation does query for `ver`. Is that necessary or, conversely, should `conda` do the same?

jonashaag:
@jaimergp I'm also fine with changing the behavior in Conda. I just want both implementations to be the same.

jaimergp:
It looks like `__win` (or `__unix`) were [never](https://github.com/conda/conda/blob/20ea9d5a7a658ab95acd2abcd5b7484f11686110/conda/core/index.py#L204-L209) versioned in `conda`, so the easiest way to make both behave the same is to disable versioning in libmamba. It would also save the subprocess call on every process.

My main question is whether there are packaging strategies that require `__win` versions in the same way we depend on `__glibc` or `__cuda`. I am not aware of any.

jonashaag:
Not that I am aware of.

AntoinePrv:
> My main question is whether there are packaging strategies that require __win versions in the same way we depend on __glibc or __cuda. I am not aware of any.

Well if it was never versioned, how could they?

Maybe we should do the right thing and make it versioned in both?

jaimergp:
FYI with the potential approval of https://github.com/conda/ceps/pull/103, this might not be needed after all.

PR #2816, opened 2023-09-06
Fix and improve Python stubs

AntoinePrv:
Stubgen tests are failing on multiple PRs.
It came to my attention that stubgens would depend on the Python version (e.g. `set[T]` came with Python 3.9) with no easy way to configure it.
Developpers also often to generate them.

The result is to generate them dynamically, simplifying the tests.
To generate in the binary folder:
```bash
cmake --build build/ --target stubgen
```
To "install" them in the source tree
```bash
cmake --install build/ --component PythonBindingsStubgen
```
Them proceed as usual with
```bash
python -m pip install libmambapy
```


JohanMabille:
The idea of having the stubgens in the repo was to be able to detect Python API changes by comparing the repo stubgens with those generated from the new libmambapy. If we remove both the comparison and the stored subgens, I don't see the point of still generating them in the CI.

I find the stubgens tests more stable now that we lint the generated stubgens in the CI.

AntoinePrv:
> The idea of having the stubgens in the repo was to be able to detect Python API changes by comparing the repo stubgens with those generated from the new libmambapy.

I don't really like committing files that are generated, I think it is flaky. Besides the files should depend on the Python version. But I see you point.

> If we remove both the comparison and the stored subgens, I don't see the point of still generating them in the CI.

The point is to see if their generation works :)

JohanMabille:
> The point is to see if their generation works :)

That's a point ;)

Issue #2819, opened 2023-09-07
[Micromamba] Missing features / mamba - to be implemented


Hind-M:
Based on the `mamba` tests, here are some features missing in `micromamba` compared to `mamba`:

- [ ] Missing `env update` command
- [ ] Missing `env config` command
- [ ] Missing local channels: `-c local` or `--use-local` (exists with `repoquery`)
- [ ] Check `repoquery` behavior differences (related open [issue](https://github.com/mamba-org/mamba/issues/2326))
- [ ] Returned `version` is not tested with `micromamba info`
- [ ] Missing `--mamba-experimental` (not sure if this is really useful/used...)
If set, supposed to disable `--json` if it's there, and to still execute a transaction even with `--dry-run`

jjerphan:
@Hind-M: do you think this can be closed? :slightly_smiling_face:

jjerphan:
I do not really see the benefits of `--mamba-experimental` given its description.

Do you have more information about it?

Hind-M:
> Do you have more information about it?

No, therefore keeping track of it here.

PR #2827, opened 2023-09-12
[Micromamba] Add `env update`

Hind-M:
Fix part of #2819

### DIFF
--- a/libmamba/src/api/update.cpp
+++ b/libmamba/src/api/update.cpp
@@ -139,11 +139,7 @@ namespace mamba
                         remove_specs.push_back(it.second.name);
                     }
                 }
-                solver.add_jobs(remove_specs, SOLVER_ERASE | SOLVER_CLEANDEPS);  // TODO it doesn't
-                                                                                 // seem that the
-                                                                                 // SOLVER_CLEANDEPS
-                                                                                 // is working
-                                                                                 // here...
+                solver.add_jobs(remove_specs, SOLVER_ERASE | SOLVER_CLEANDEPS);
             }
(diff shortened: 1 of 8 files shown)


libmamba/src/api/update.cpp
```diff
@@ -127,6 +127,24 @@ namespace mamba
...
+                std::vector<std::string> remove_specs;
+                for (auto& it : hist_map)
+                {
+                    if (std::find(update_specs.begin(), update_specs.end(), it.second.name)
+                        == update_specs.end())
+                    {
+                        remove_specs.push_back(it.second.name);
+                    }
+                }
+                solver.add_jobs(remove_specs, SOLVER_ERASE | SOLVER_CLEANDEPS);  // TODO it doesn't
```
Hind-M, line 142: The wanted behavior here is to remove `remove_specs` **and** their dependencies, but only the specs are removed here.
Adding `SOLVER_CLEANDEPS` doesn't seem to be working... 
Any ideas @AntoinePrv ?

AntoinePrv: Does it work with only `SOLVER_ERASE`?

Hind-M: Yes, adding `SOLVER_CLEANDEPS` doesn't change anything on that matter (I just left it for now to illustrate).

Hind-M: I think that it was just some mixing with the installation steps locally when I first tested it but is working fine now with the `micromamba` tests 👍

JohanMabille (approved):

PR #2848, opened 2023-09-18
Print error code if `run` fails

jonashaag:


AntoinePrv (commented):
Actually I am wondering if it's a good idea to modify the output of the subprocess.
Then it becomes difficult to parse the output for user, _i.e._ `micromamba run cmd` is not longer a substitute for simply running `cmd`.

I think we should do whatever `bash -c <cmd>` is doing, which I believe is `stdout > stdout`, `stderr > stderr` and same return code.

jonashaag:
But this about errors in Mamba, not in the subprocess, no? If bash fails to start a process, or a process segfaults or whatever, it prints stuff.

AntoinePrv (approved):
> But this about errors in Mamba, not in the subprocess, no?

I think you're right

PR #2853, opened 2023-09-19
Add local channels test

Hind-M:


libmamba/tests/src/core/test_channel.cpp
```diff
@@ -311,6 +312,38 @@ namespace mamba
             ctx.custom_channels.clear();
         }
 
+        TEST_CASE("custom_local_channels")
+        {
+            auto& ctx = mambatests::context();
+
+            // Create conda-bld directory to enable testing
+            auto conda_bld_dir = env::home_directory() / "conda-bld";
+            bool to_be_removed = fs::create_directories(conda_bld_dir);
```
AntoinePrv, line 321: Do we have a way to trigger the test without side effect?

Hind-M: Well to trigger the test, we need this folder (either in the home directory, the target prefix or the root prefix). And since this is supposed to be testing `libmamba`, the home directory is the only option left.
So I don't see how we can trigger this without creating this folder, otherwise we would have a `micromamba` integration test instead...

AntoinePrv: That's sad :( I hope we can work around it with the next design

AntoinePrv (approved):

Issue #2595, opened 2023-06-15
Discussion: micromamba releases should be built and released in this repository


pavelzw:
Xref 
https://github.com/mamba-org/micromamba-releases/issues/5
#2536 

Right now, the micromamba release workflow is the following:

- somebody releases a new version
- the Conda Forge Auto tick bot doesn’t realize that there is a new version here because of the tag structure of this repository (see https://conda-forge.org/status/ -> version updates -> errored)
- Somebody needs to manually create a PR incrementing the version of the recipe 
- the conda-forge build fails (https://github.com/conda-forge/micromamba-feedstock/pull/120)
- Somebody needs to fix the build 
- The build gets merged
- The build is available on anaconda.org
- micro.mamba.pm points to the latest anaconda.org build
- mamba/micromamba-releases also gets their releases from there

I think this release workflow is highly inefficient and would propose this alternative:
- on push of a tag, the CI builds its own static build versions (the windows build got way less flakey with #2520 
- using softprops/action-gh-release, a release with the binaries attached is automatically drafted
- Somebody needs to press publish (this also could be skipped if we say `draft: false`)
- …
- Profit

Like this, we could guarantee that there is always the most recent version available.

pavelzw:
@AntoinePrv @JohanMabille what do you think of this proposal? I could create a PR implementing this in CI if you are on board with it.

jonashaag:
I strongly support this proposal. The current release process needs too much human interaction and is scattered over multiple repositories. I'm also in favour of hosting the binaries here rather than in a separate repository.

pavelzw:
> Would that be using the recipe from the feedstock?

I would suggest keeping the recipe for the static build in this repository.

JohanMabille:
It would be weird to not have micromamba available on conda-forge. I agree the releasing process is error-prone and implies too many human interactions, but deprecating the micromamba feedstock looks a bit extreme to me.I think we can have both the automatic release on this repo with the process you described, and continue to update the feedstock repo. Ideally we should be able to use a bot / GHA to automagically update the feedstock upon release tag until we get back to a more traditional workflow when we split this repo and the conda-forge bot can detect releases.

pavelzw:
> deprecating the micromamba feedstock looks a bit extreme to me

I agree, maybe it came out wrong. I didn't mean to deprecate the `micromamba-feedstock` but the static build on conda-forge.

[...]

PR #2859, opened 2023-09-21
Build micromamba win with feedstock

AntoinePrv:


AntoinePrv:
@pavelzw any idea how to get the "Micromamba static builds" started?
\

pavelzw:
My suggestion would be to use _only_ vcpkg for static builds and avoid going through the pain of static builds with conda-build for windows 😅

The static builds with only vcpkg have cause way fewer issues than trying with conda-build. I don't see the value of trying to port this to conda-build. 

Personally, I would just leave the static CI with windows be and try to migrate the release workflow to actually use this build instead of the conda-forge static binaries as discussed in #2595.

AntoinePrv:
This is not DRY, we need to sync the feedstock and this CI constantly. VCPKG has caused a lot of outages too. Conda-Forge takes more pain to setup I agree  but once it is setup it's stable and simple, no need for extra cache etc. We also need to work with two systems, which is not convenient, because were are not gonna stop shipping `libmambapy`.

pavelzw:
But how about trying to get rid of the vcpkg dependency in the recipe first? You already did the (hopefully 😅) most painful part by moving the libsolv dependency to conda-forge.

AntoinePrv:
Fair enough. How about trying adding a light environment and starting to also used the libs that are properly packaged statically?

AntoinePrv:
Great this seems to be working. The point is I am adding some more dependencies and don't want to do the work twice.

@pavelzw
- Do you know if these are deployed anywhere? If so I might be more reluctant to do mixed builds here (but I would also encourage to take the builds from conda-forge).
- Can we remove entries from `vcpkg.json`, or should we keep that file complete because it is used elsewhere?

JohanMabille (approved):

PR #2862, opened 2023-09-22
Build micromamba with /MD

AntoinePrv:


JohanMabille (approved):

AntoinePrv:
@pavelzw we've switched to using `/MD` (DLL C runtime on Windows) in `micromamba-feedstock` on @Klaim's advice and doing the same here. Let us know if you see an issue with it.

pavelzw:
image

I think we need to put a timestamp on this cache or else it will always have the old cache files and will never update the cache.

PR #2873, opened 2023-09-27
Pin yaml-cpp <0.8

JohanMabille:
yaml-cpp 0.8 is broken on Windows: `find_package` set the library to `yaml-cpp.lib` instead of the full path.


JohanMabille:
Let's wait for #2874 that would be a better fix.

PR #2879, opened 2023-09-28
Filesystem library

AntoinePrv:


libmamba/include/mamba/api/configuration_impl.hpp
```diff
@@ -304,21 +304,21 @@ namespace YAML
     };
 
     template <>
-    struct convert<fs::u8path>
+    struct convert<mamba::fs::u8path>
```
Klaim, line 307: Why was using the canonical name necessary? We are already in the `mamba` namespace.

libmamba/include/mamba/filesystem/u8path.hpp
```diff
@@ -1151,18 +1138,7 @@ namespace fs
...
-        if (utimensat(AT_FDCWD, path.string().c_str(), NULL, 0) == -1)
-        {
-            ec = std::error_code(errno, std::generic_category());
-        }
-#else
-        auto new_time = fs::file_time_type::clock::now();
-        std::filesystem::last_write_time(path, new_time, ec);
-#endif
-    }
+    void last_write_time(const u8path& path, now, std::error_code& ec) noexcept;
```
Klaim, line 1141: Note that we used inlined functions specifically so that `<filesystem>` was used directly seemlessly at compilation for the user. While this specific function is not matching that intent (it was added afterwards), keep in mind that moving the code of functions in general from this file into cpp will have some (probably negative) effects on performance. As you didnt move other functions I suppose you realize all that already, just stating this "for the record".

libmamba/src/filesystem/u8path.cpp
```diff
@@ -6,21 +6,37 @@
...
-#include "mamba/core/mamba_fs.hpp"
-#include "mamba/util/string.hpp"
+// We can use the presence of UTIME_OMIT to detect platforms that provide
+// utimensat.
+#if defined(UTIME_OMIT)
+#define USE_UTIMENSAT
+#endif
+#endif
+
+#include "mamba/filesystem/u8path.hpp"
```
Klaim, line 20: BTW the usual way to organize header is that in the cpp matching a header we put the related header as first. Currently our clang-format will reorganize them in alphabetical order. I was wondering: is there a way to have both logics?

libmamba/src/filesystem/u8path.cpp
```diff
@@ -6,21 +6,37 @@
...
     std::filesystem::path normalized_separators(std::filesystem::path path)
     {
         auto native_string = path.native();
-        static constexpr auto platform_separator = L"\\";
-        static constexpr auto other_separator = L"/";
-        mamba::util::replace_all(native_string, other_separator, platform_separator);
+        static constexpr wchar_t platform_separator = L'\\';
+        static constexpr wchar_t other_separator = L'/';
+
+        auto pos = native_string.find(other_separator);
```
Klaim, line 33: Why was `util::replace_all` not good enough here? Also I would prefer this loop to be a function.

micromamba/src/login.cpp
```diff
@@ -64,13 +64,13 @@ set_logout_command(CLI::App* subcom)
         []()
         {
             static auto path = mamba::env::home_directory() / ".mamba" / "auth";
-            fs::u8path auth_file = path / "authentication.json";
+            mamba::fs::u8path auth_file = path / "authentication.json";
```
Klaim, line 67: `const`

micromamba/src/login.cpp
```diff
@@ -158,15 +158,15 @@ set_login_command(CLI::App* subcom)
             }
 
             static auto path = mamba::env::home_directory() / ".mamba" / "auth";
```
Klaim, line 160: const

I suspect the `static` doesnt help performance here but oh well.

micromamba/src/login.cpp
```diff
@@ -158,15 +158,15 @@ set_login_command(CLI::App* subcom)
...
-            fs::create_directories(path);
+            mamba::fs::create_directories(path);
 
 
             nlohmann::json auth_info;
-            fs::u8path auth_file = path / "authentication.json";
+            mamba::fs::u8path auth_file = path / "authentication.json";
```
Klaim, line 165: const

libmamba/include/mamba/filesystem/u8path.hpp
AntoinePrv: Yes, I move this one in particular because it had OS specific dependencies that I though were best hidden.

libmamba/src/filesystem/u8path.cpp
```diff
@@ -6,21 +6,37 @@
 
 #include <filesystem>
 #include <string>
+#if !defined(_WIN32)
```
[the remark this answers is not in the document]
AntoinePrv: Agreed, but this we need to adjust in clang-format

libmamba/src/filesystem/u8path.cpp
AntoinePrv: Yes, I think clang-format does that if the header is written as `"header.hpp"` (no directory)

AntoinePrv: But then it seems to not always compile...

libmamba/include/mamba/api/configuration_impl.hpp
AntoinePrv: No, here here are specializing in the `YAML::` namespace.

libmamba/src/filesystem/u8path.cpp
AntoinePrv: So I am trying to get to a state of multiple internal "libraries" where there is no cyclic dependencies between  them.
The goal is to restrict what code can be used where, so that for instance "filesystem" or "util" does not start calling code form "core". These restrictions would then restrict the opportunities for spaghetti code.

Or maybe I am being too dogmatic?

libmamba/src/filesystem/u8path.cpp
Klaim: 😬 not a good sign then... will be fixed by modules one day 🥲

libmamba/src/filesystem/u8path.cpp
Klaim: It's fine but it's not clear to me how this is related to this code, is it because otherwise there is a cyclic dependency between the filesystem and utils? If so, maybe isolating that specific replace function in it's header would solve the issue?

AntoinePrv: @Klaim I wanted to use CMake `add_library(mamba-fs OBJECT filesystem.cpp)` to literally **forbid** including code from the rest of mamba in this library (and then do the same with `mamba-util`, `mamba-solv-cpp`...).
With more investigation, it turns out it's not so easy to do: since the headers are public, they still need to be added to libmamba to be installed, so I'm not sure it's worth the trouble.

For this particular issue, it turns out that since `L"\\"` and `L"/"` are in fact single characters, a simple call to `std::replace` will do (and is probably more performant). What do you think?

Klaim: Anyway if it's tested and `std::replace` works in that specific function, I'm fine by it.

Klaim: OK then I guess we can use replace

libmamba/include/mamba/core/subdirdata.hpp
```diff
@@ -17,11 +16,12 @@
 #include "mamba/core/context.hpp"
 #include "mamba/core/download.hpp"
 #include "mamba/core/error_handling.hpp"
-#include "mamba/core/mamba_fs.hpp"
+#include "mamba/core/fetch.hpp"
```
JohanMabille, line 19: Note: subdirdata has been refactored and does not depend anymore on this file, so we should not include it. This can be removed later, when `fetch.hpp` is removed from the code base.

libmamba/include/mamba/fs/filesystem.hpp
```diff
@@ -4,27 +4,14 @@
 //
 // The full license is in the file LICENSE, distributed with this software.
 
-#ifndef MAMBA_CORE_FS_HPP
-#define MAMBA_CORE_FS_HPP
+#ifndef MAMBA_CORE_FS_FILESYSTEM_HPP
+#define MAMBA_CORE_FS_FILESYSTEM_HPP
```
JohanMabille, line 8: Nitpicking: should the inclusion guard be `MAMBA_FS_FILESYSTEM`?

Issue #2882, opened 2023-10-02
conda 23.9.0 breaks mamba 1.5.1


aqc-carlodri:
### Troubleshooting docs

- [X] My problem is not solved in the Troubleshooting docs

### How did you install Mamba?

Mambaforge

### Search tried in issue tracker

ensure

### Latest version of Mamba

- [X] My problem is not solved with the latest version

### Tried in Conda?

I do not have this problem with Conda, just with Mamba

### Describe your issue

using mamba 1.5.1 and upgrading to conda [23.9.0](https://github.com/conda/conda/blob/main/CHANGELOG.md#2390-2023-09-27) breaks mamba:
```
$ mamba info
Traceback (most recent call last):
  File "C:\Users\cpi-eng\miniconda3\Scripts\mamba-script.py", line 6, in <module>
    from mamba.mamba import main
  File "C:\Users\cpi-eng\miniconda3\Lib\site-packages\mamba\mamba.py", line 16, in <module>
    from conda.cli.common import (
ImportError: cannot import name 'ensure_name_or_prefix' from 'conda.cli.common' (C:\Users\cpi-eng\miniconda3\Lib\site-packages\conda\cli\common.py)
```

### mamba info / micromamba info

```shell
n/a
```

KrOstir:
Having the same issue. After updating mamba with `mamba update --all` all mamba commands give the `ImportError`. conda commands work as expected.

aqc-carlodri:
Downgrading `conda` to 23.7.4 fixes the issue.

jonashaag:
See https://github.com/mamba-org/mamba/pull/2878

sweettyler:
same happened to me.

kaczmarj:
>Soo, what is the solution? Downgrade to previous version or?


the solution at this point is to downgrade conda ([mentioned above](https://github.com/mamba-org/mamba/issues/2882#issuecomment-1742524151))

```bash
conda install -n base conda==23.7.4
mamba --help  # works now!
```

JohanMabille:
The repodata has been [patched](https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/pull/563), mamba 1.5.1 now pins conda < 23.9.0, so a fresh install should work out of the box.

In the meantime mamba  has been [patched to support latest conda](https://github.com/mamba-org/mamba/pull/2878), we will release 1.5.2 with this fix in a few days.

JohanMabille:
[...]

PR #2878, opened 2023-09-28
compatibility fixes for conda 23.9 and conda-libmamba-solver 23.9

jaimergp:
There have been some changes that break `mamba`'s CLI:

- `conda` has deprecated `ensure_name_or_prefix`. We do that manually now.
- `conda` does not take `context` as an argument in `init_loggers` anymore.
- `conda-libmamba-solver` exposes the same `repoquery` subcommand through the subcommand plugin system, which causes a [conflict in Python 3.11](https://github.com/conda/conda-libmamba-solver/issues/303). Interestingly only in that version.

PR #2884, opened 2023-10-02
Update GitHub Actions steps to open Issues for failed scheduled jobs

jdblischak:
The recent conda release 23.9.0 broke the interface with mamba (#2882, #2878). This issue was first detected by the [nightly job against conda-canary](https://github.com/mamba-org/mamba/blob/0fbb0e4fab2d65a68c1bcd388b377feaa20f5370/.github/workflows/conda_canary.yml) back on [August 27th](https://github.com/mamba-org/mamba/actions/runs/5990241039/job/16247352587#step:9:43). However, an Issue was never opened because the path to the issue template file was incorrect. Similarly, the [scheduled jobs to test micro.mamba.pm](https://github.com/mamba-org/mamba/actions/workflows/test_micro.mamba.pm.yml?query=event%3Aschedule) have been failing for weeks, but never open an Issue because the repository is not cloned in that job.

This PR does the following:

* `conda_canary.yml`
  * Fix path to issue template file
  * Only open an Issue for a scheduled job on the mamba-org owned repo (this avoids opening Issues on forks, and also opening unnecessary issues from manually triggered workflow runs)
* `test_micro.mamba.pm.yml`
  * Separates the issue opening step into its own job that will be executed if any of the upstream matrix of jobs fails
  * Checkout the repo to obtain the issue template file
  * Only open an Issue for a scheduled job on the mamba-org owned repo

### DIFF
--- a/.github/workflows/conda_canary.yml
+++ b/.github/workflows/conda_canary.yml
@@ -73,8 +73,8 @@ jobs:
           mamba env update -f mamba/tests/update_env.yml
 
       - uses: JasonEtco/create-an-issue@1a16035489d05041b9af40b970f02e301c52ffba
-        if: failure()
+        if: failure() && github.repository_owner == 'mamba-org' && github.event_name == 'schedule'
         with:
-          filename: mamba/.github/workflows/bot_issue_template.md
+          filename: .github/workflows/bot_issue_template.md
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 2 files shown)


jonashaag:
> However, an Issue was never opened because the path to the issue template file was incorrect

Thanks for debugging this; I was wondering why it wasn't working...

jonashaag (approved):
Thank you!

jdblischak:
Reading over my changes again, I realized I forgot to update the path to the issue template in the micromamba workflow. To be safe, I also added conditional triggers so that the workflows are always run if they are edited (but they won't open an Issue since they weren't "schedule" events)

jdblischak:
Following up: the issue opening has been fixed (at least for the micromamba test). Last night's scheduled job [failed](https://github.com/mamba-org/mamba/actions/runs/6391511064) and opened Issue #2887

PR #2886, opened 2023-10-03
Add FindLibsolv.cmake

AntoinePrv:


JohanMabille:
do you think ti could be worth fixing the cmake packaging of libsolv upstream?

AntoinePrv:
@JohanMabille there is no `libsolv-config.cmake`. I would have added this `FindLibsolv` in the package itself, but from what I understood, `FindModule.cmake` are not looked in `lib/cmake/...`

JohanMabille:
What I meant by fixing it upstream was to add the generation on the libsolv-config cmake (through a .cmake.in cofigurable file probably) andnhave it installed aside the other components when installing libsolv

PR #2917, opened 2023-10-17
Added PackageFetcher

JohanMabille:


libmamba/src/core/transaction.cpp
```diff
@@ -881,152 +876,86 @@ namespace mamba
...
+            /*std::packaged_task<void(std::size_t)> task{[eit](std::size_t downloaded_size) {
+                eit->run(downloaded_size);
+            }};*/
+            auto task = std::make_shared<std::packaged_task<void(std::size_t)>>([eit](std::size_t downloaded_size) {
+                eit->run(downloaded_size);
+            });
+            extract_tasks.push_back(task->get_future());
+            download_requests.push_back(fit->build_download_request([extract_task=std::move(task)](std::size_t downloaded_size) {
+                MainExecutor::instance().schedule([t = std::move(extract_task)](std::size_t ds) {
+                    (*t)(ds); }, downloaded_size);
```
JohanMabille, line 909: @Klaim I find this part really cumbersome, `std::packaged_task` and `std::function` do not play well together. Having the `future` and the `promise` in the PackageExtractRequest would simplify this code a lot (although I admit it would expose synchronization primitives in the header :s).

Klaim: Yeah that's verbose because it's callback-based. Chaining continuations would be a bit less verbose but still noisy. The ultimate syntax to express the same thing would be through coroutines or task-graphs.

>  `std::packaged_task` and `std::function` do not play well together

I dont get what you mean here? `packaged_task` is just a callable object, it works well with `std::function`. Or do you mean the callback-based continuations is not ideal?

JohanMabille: `std::function` can store copiable objects only, while `std::package_task` (and therefore the lambda capturing it) is movable only. So we have to store the task in a `smart_pointer` and I really don't like it.

Klaim: Ah yes I'm used to have `std::move_only_function` around. In this case, dont use a smaprt pointer, just replaced `std::packaged_task` by a lambda wrapping a `std::promise`. Then if you need to, put the promise in the smart pointer. That avoids complicating the callback handling system.

```
auto p = std::make_shared<std::promise<void>>();
std::future<void> f = p->get_future();
auto continuation = [p = std::move(p), callback]{
   auto maybe_error = safe_invoke(callback);
   ... // error handling
   p->set_value();
 };
 ```
 
 Either that or we need an equivalent to `std::move_only_function` , there are many relyable implementations on github.

Comment (JohanMabille) [reply libmamba/src/core/transaction.cpp:909]: let's go with the promise/future for now

Comment (Klaim) [reply libmamba/src/core/transaction.cpp:909]: BTW if it's too much work, storing the `packaged_task` in a `shared_ptr` to pass it around should work too.

Comment (JohanMabille) [reply libmamba/src/core/transaction.cpp]: Actually the version with promise instead of future does not simplify that much; we have 2 lambda instead of 3? but need to capture two elements and set manually the promise. Besides, we will get back to packaged_task when we introduce move_only_functions, so let's keep the packaged_task.

Comment (Klaim) [reply libmamba/src/core/transaction.cpp]: You mean shared packated task? Works for me 👍🏽

File: libmamba/src/core/package_fetcher.cpp [hunk 2]
```diff
@@ -0,0 +1,428 @@
...
+    }
+
+    namespace
+    {
+        std::mutex urls_txt_mutex;
+    }
+
+    void PackageFetcher::update_urls_txt() const
+    {
+        std::lock_guard<std::mutex> lock(urls_txt_mutex);
```
Klaim, line 400: I'm not sure that's necessary? Isnt `ofstream` already writing in 1 locked operation if there is only one value to write?

libmamba/src/core/transaction.cpp
```diff
@@ -825,208 +828,231 @@ namespace mamba
...
-        );
-        auto& aggregated_pbar_manager = dynamic_cast<AggregatedBarManager&>(pbar_manager);
-
-        auto& channel_context = m_pool.channel_context();
-        auto& ctx = channel_context.context();
-        DownloadExtractSemaphore::set_max(ctx.threads_params.extract_threads);
-
-        if (ctx.experimental && ctx.validation_params.verify_artifacts)
+        using FetcherList = std::vector<PackageFetcher>;
+        // Free functions instead of private method to avoid exposing downloaders
```
Klaim, line 834: Nah, free functions are very good for this, no need for pimpl. 👍🏽

libmamba/src/core/transaction.cpp
```diff
@@ -825,208 +828,231 @@ namespace mamba
...
+        MultiDownloadRequest build_download_requests(
+            FetcherList& fetchers,
+            ExtractRequestList& extract_requests,
+            ExtractTaskList& extract_tasks,
+            std::size_t download_size
+        )
+        {
+            MultiDownloadRequest download_requests;
+            download_requests.reserve(download_size);
+            for (auto [fit, eit] = std::tuple{ fetchers.begin(), extract_requests.begin() };
```
Klaim, line 908: 😬

libmamba/src/core/transaction.cpp
```diff
@@ -825,208 +828,231 @@ namespace mamba
...
-                if (!t->finished())
+                PackageExtractRequest::Result res = eit->get_result();
+                if (!res.valid || !res.extracted)
                 {
-                    all_finished = false;
-                    break;
+                    fit->clear_cache();
+                    all_valid = false;
+                    // TODO: check if we can remove this
+                    throw std::runtime_error(
```
Klaim, line 976: Throwing here means we never return `false` so the return type doesnt help at all the caller.
I agree with the comment that this throw should be removed and let the caller decide what to do once they have checked.

Basically this function does not just check vaility, it also updates a cache, so maybe there is a need for 2 separate fnuctions.
The validity check could be done with `std::all_of` but then you would need extract requests to know their fetchers.
Not saying to improve now, but maybe add a comment to improve this function later?

libmamba/src/core/transaction.cpp
```diff
@@ -825,208 +828,231 @@ namespace mamba
...
+        auto extract_end = std::partition(
+            download_end,
+            fetchers.end(),
+            [](const auto& f) { return f.needs_extract(); }
+        );
+
+        auto download_size = static_cast<std::size_t>(std::distance(fetchers.begin(), download_end));
+        auto extract_size = static_cast<std::size_t>(std::distance(fetchers.begin(), extract_end));
+
+        // At this point:
```
Klaim, line 1011: Very nice 👍🏽

libmamba/src/core/transaction.cpp
```diff
@@ -825,208 +828,231 @@ namespace mamba
...
+        // - [fetchers.begin(), download_end) contains packages that need to be downloaded,
+        // validated and extracted
+        // - [download_end, extract_end) contains packages that need to be extracted only
+        // - [extract_end, fecthers.end()) contains packages already installed and extracted
+
+        ExtractRequestList extract_requests = build_extract_requests(ctx, fetchers, extract_size);
```
Klaim, line 1017: I think this can be `const` now?

libmamba/src/core/transaction.cpp
```diff
@@ -825,208 +828,231 @@ namespace mamba
...
-                all_valid = false;
-                throw std::runtime_error(
-                    std::string("Found incorrect download: ") + t->name() + ". Aborting"
-                );
-            }
+            task.wait();
         }
 
-        return !is_sig_interrupted() && downloaded && all_valid;
+        bool all_valid = check_all_valid(fetchers, extract_requests);
```
Klaim, line 1054: Welp, `all_valid` can only be `true`. 😂 
Also `const` 🤓

Klaim (commented):
First pass of review, mostly nitpicking.

libmamba/include/mamba/core/package_fetcher.hpp
```diff
@@ -0,0 +1,158 @@
...
+
+        const std::string& name() const;
+        bool needs_download() const;
+
+        void set_progress_callback(progress_callback_t cb);
+
+        void run();
+        void run(std::size_t downloaded_size);
+
+        Result get_result() const;
```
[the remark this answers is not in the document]
JohanMabille: No, since all the synchronization is done out of this class, we have no mean to have a blocking call here.

libmamba/include/mamba/core/download_progress_bar.hpp
```diff
@@ -52,14 +53,61 @@ namespace mamba
...
+
+        PackageDownloadMonitor(const PackageDownloadMonitor&) = delete;
+        PackageDownloadMonitor& operator=(const PackageDownloadMonitor&) = delete;
+
+        PackageDownloadMonitor(PackageDownloadMonitor&&) = delete;
+        PackageDownloadMonitor& operator=(PackageDownloadMonitor&&) = delete;
+
+        // Requires extract_tasks.size() >= requests.size()
+        // Requires for i in [0, dl_requests.size()), extract_tasks[i].needs_download()
+        void observe(
```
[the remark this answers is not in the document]
JohanMabille: Theoretically it could, also there would be no gain compared to instantiate a new one.

libmamba/include/mamba/core/package_fetcher.hpp
```diff
@@ -0,0 +1,158 @@
...
+    {
+        validate_update,
+        validate_success,
+        validate_failure,
+        extract_update,
+        extract_success,
+        extract_failure
+    };
+
+    class PackageExtractRequest
```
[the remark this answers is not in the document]
JohanMabille: Yeah I agree the name is confusing, I will change that.

libmamba/src/core/download.cpp
```diff
@@ -847,4 +854,35 @@ namespace mamba
...
+    }
+
+    bool check_resource_exists(const std::string& url, const Context& context)
+    {
+        if (!context.remote_fetch_params.curl_initialized)
+        {
+            // TODO: Move this into an object that would be automatically initialized
+            // upon construction, and passed by const reference to this function instead
+            // of context.
+            Context& ctx = const_cast<Context&>(context);
```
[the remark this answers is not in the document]
JohanMabille: Yeah I know. Thus the comment ;)

libmamba/src/core/download_progress_bar.cpp
```diff
@@ -4,32 +4,158 @@
...
+                r.separator.deactivate();
+
+                auto console_stream = Console::stream();
+                r.print(console_stream, 0, false);
+            }
+        }
+
+        std::function<void(ProgressBarRepr&)> download_repr(ProgressProxy& progress_bar)
+        {
+            // TODO: remove lambda capture and retrieve the progress bar via r.progress_bar()
```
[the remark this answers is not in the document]
JohanMabille: Not sure, I'm gonna check.

libmamba/src/core/download_progress_bar.cpp
```diff
@@ -63,148 +189,311 @@ namespace mamba
...
+                dl_requests[i].progress = [this, i](const DownloadEvent& e)
+                { update_progress_bar(i, e); };
+            }
         }
-        else
+
+        init_aggregated_download();
+        init_aggregated_extract();
+
+        auto& pbar_manager = dynamic_cast<AggregatedBarManager&>(
```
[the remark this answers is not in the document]
JohanMabille: Indeed, good catch!

libmamba/src/core/download_progress_bar.cpp
```diff
@@ -63,148 +189,311 @@ namespace mamba
...
+    void PackageDownloadMonitor::init_aggregated_extract()
+    {
+        auto& pbar_manager = dynamic_cast<AggregatedBarManager&>(
+            Console::instance().progress_bar_manager()
+        );
+        auto* extract_bar = pbar_manager.aggregated_bar("Extract");
+        if (extract_bar)
         {
-            complete_checking_progress_bar(index);
+            // TODO: remove lambda capture and retrieve the progress bar via repr.progress_bar()
```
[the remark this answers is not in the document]
JohanMabille: Same answer as before, I need to check.

libmamba/src/core/package_fetcher.cpp
```diff
@@ -0,0 +1,428 @@
...
+    {
+        auto res = ValidationResult::VALID;
+        if (expected_size() && expected_size() != downloaded_size)
+        {
+            res = ValidationResult::SIZE_ERROR;
+            LOG_ERROR << "File not valid: file size doesn't match expectation " << m_tarball_path
+                      << "\nExpected: " << expected_size() << "\nActual: " << downloaded_size
+                      << "\n";
+            Console::instance().print(filename() + " tarball has incorrect size");
+            // TODO: terminate monitor
```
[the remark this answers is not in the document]
JohanMabille: Nope, this has been implemented in the transaction.

libmamba/src/core/transaction.cpp
JohanMabille: > Basically this function does not just check vaility, it also updates a cache, so maybe there is a need for 2 separate fnuctions.

That would mean iterating twice, but I agree, that would be cleaner. Or we could rename the method `clean_invalid_cache` ?

> The validity check could be done with std::all_of but then you would need extract requests to know their fetchers.

Actually the requests already know their fetchers, so we could do this now. This would allow to iterate on the extract requests only. My only concerns, if we don't split the method, or if we split it but still want to iterate on the extract requests only, is that PackageExtractRequest should expose the clean_cache metho to forward the call to the fetcher. A bit weird, but I can live with it.

libmamba/src/core/transaction.cpp
JohanMabille: Not really, we call non const method of the extract request later.

libmamba/include/mamba/core/package_fetcher.hpp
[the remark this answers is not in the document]
Klaim: To me "task" only means something that runs like a fonction, might be embarking it's data to work on. It's an element processed by an scheduling and/or asynchronous or synchronous system, like a value is to a container. So it would work well in that context.
In general I dont like "runner" for tasks, it's ambiguous with the role of a scheduler and/or executor and/or execution context.

Other alternatives: operation, "PackageExtracter" (as in it's role is to do the extract)... 
You decide 👍🏽

libmamba/src/core/package_fetcher.cpp
JohanMabille: I have no idea. I'll add a comment so we can check it later, I just took the previous code as is.

libmamba/include/mamba/core/package_fetcher.hpp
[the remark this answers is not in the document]
JohanMabille: OK, then I'm probably biased on this ;) Let's use PackageExtractTask then.

libmamba/include/mamba/core/package_fetcher.hpp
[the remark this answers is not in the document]
JohanMabille: We're doomed ;) We could also return the result from the `run` methods, and change the `future` type used in the transaction class.

libmamba/src/core/package_fetcher.cpp
```diff
@@ -0,0 +1,428 @@
...
+        else
+        {
+            m_url = pkg_info.url;
+        }
+
+        fs::u8path extracted_cache = caches.get_extracted_dir_path(m_package_info);
+        if (extracted_cache.empty())
+        {
+            fs::u8path tarball_cache = caches.get_tarball_path(m_package_info);
+            auto& cache = caches.first_writable_cache(true);
```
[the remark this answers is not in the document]
JohanMabille: Not possible here, we call a non const method on cache later.

libmamba/src/core/transaction.cpp
JohanMabille: thanks ;)

SylvainCorlay (approved):

libmamba/src/core/package_fetcher.cpp
```diff
@@ -0,0 +1,419 @@
...
+        bool is_valid = true;
+        bool is_extracted = p_fetcher->extract(m_options);
+        return { is_valid, is_extracted };
+    }
+
+    auto PackageExtractTask::run(std::size_t downloaded_size) -> Result
+    {
+        using ValidationResult = PackageFetcher::ValidationResult;
+        ValidationResult validation_res = p_fetcher->validate(downloaded_size, get_progress_callback());
+        bool is_valid = validation_res == ValidationResult::VALID;
```
Klaim, line 46: const

libmamba/src/core/package_fetcher.cpp
```diff
@@ -0,0 +1,419 @@
...
+        LOG_DEBUG << "Waiting for decompression " << m_tarball_path;
+        update_monitor(cb, PackageExtractEvent::extract_update);
+
+        {
+            std::lock_guard<counting_semaphore> lock(PackageFetcherSemaphore::semaphore);
+            interruption_point();
+            LOG_DEBUG << "Decompressing '" << m_tarball_path.string() << "'";
+            try
+            {
+                fs::u8path extract_path = get_extract_path(filename(), m_cache_path);
```
Klaim, line 269: const

libmamba/src/core/package_fetcher.cpp
```diff
@@ -0,0 +1,419 @@
...
+
+    PackageExtractTask PackageFetcher::build_extract_task(ExtractOptions options)
+    {
+        return { this, std::move(options) };
+    }
+
+    void PackageFetcher::clear_cache() const
+    {
+        fs::remove_all(m_tarball_path);
+        fs::u8path dest_dir = strip_package_extension(m_tarball_path.string());
```
Klaim, line 300: const

libmamba/src/core/transaction.cpp
```diff
@@ -825,208 +828,227 @@ namespace mamba
...
 
-            auto* extract_bar = aggregated_pbar_manager.aggregated_bar("Extract");
-            if (extract_bar)
+        void schedule_extractions(
+            ExtractTaskList& extract_tasks,
+            ExtractTrackerList& extract_trackers,
+            std::size_t download_size
+        )
+        {
+            for (auto it = extract_tasks.begin() + static_cast<std::ptrdiff_t>(download_size);
```
Klaim, line 936: Prefer using `std::next`? Also, you're not guaranteed that `.begin()` gives a pointer (see debug build in msvc as an example).

Actually it's not clear to me why `download_size` is used here? Is it actually an index?

libmamba/src/core/transaction.cpp
```diff
@@ -825,208 +828,227 @@ namespace mamba
...
+            MultiDownloadRequest requests,
+            const Context& context,
+            DownloadOptions options,
+            PackageDownloadMonitor* monitor
+        )
         {
-            LOG_ERROR << "Download didn't finish!";
-            return false;
+            auto result = download(std::move(requests), context, options, monitor);
+            bool all_downloaded = std::accumulate(
```
Klaim, line 954: I think you actually wanted `std::all_of`?

Klaim (approved):
Minor improvs (and a question) listed otherwise LGTM

libmamba/src/core/transaction.cpp
JohanMabille: > Also, you're not guaranteed that .begin() gives a pointer

For random access iterators, we are guaranteed the `it + offset` is valid.

> Actually it's not clear to me why download_size is used here? Is it actually an index?

Wd want to scheldule extractions for packages that don't need to be downloaded, because downloading a package already triggers its extraction.

Klaim: > For random access iterators, we are guaranteed the it + offset is valid.

Ok, then my question seems to be more like why the cast to `ptrdiff_t`? `size_t` doesnt work here?

> We want to scheldule extractions for packages that don't need to be downloaded, because downloading a package already triggers its extraction.

I think it's just the name `download_size` with a `size_t ` as type that's misleading to me?

JohanMabille: `operator+` for random access iterators accept a `ptrdiff_t` only, using a `size_t` results in a conversion warning.

Klaim: Ok got it. 👍🏽

Issue #2129, opened 2022-11-21
Error with package `__osx` when removing package `ipykernel` on macOS


errubec:
I was trying to remove an unneeded installation of `ipykernel` from my environment with `mamba remove ipykernel`. However, the process reports the following error (reformatted):

```
File "/Users/<myname>/mambaforge/lib/python3.10/site-packages/conda/_vendor/boltons/setutils.py", line 249, in remove
        raise KeyError(item)
    KeyError: PackageRecord(_hash=-7575337989302587721,
        name='__osx', version='13.0.1', build='0', build_number=0,
        channel=Channel("@"), subdir='osx-arm64', fn='__osx',
        md5='12345678901234567890123456789012',
        package_type='virtual_system'
    )
```

An inspection of `ipykernel` dependencies list the following package named `__osx`:

| Package | Version | Build | Channel |
| - | - | - | - |
| `__osx` | 13.0.1 | 0 | installed |

Apparently this package is some representation of the underlying operating system and cannot be 'removed' like an ordinary package, but the `remove` script doesn't appear to be handling it differently, causing the error and failing to uninstall `ipykernel` at all.

Is there a way to fix this?

jonashaag:
What’s your mamba version?

errubec:
Thank you @jonashaag for coming back to me so quickly.

> What’s your mamba version?

`mamba --version` prints

```
mamba 1.0.0
conda 22.9.0
```

Would this be helpful?

bstadlbauer:
Also just ran into this. Not sure how tricky this would be to implement, but in case it's just a small change happy to help out

wolfv:
@bstadlbauer yes, just filtering out packages with `__` in front would do it :)

beenje:
I got the same issue on Linux with `__unix`:

```
$ mamba --version
mamba 1.1.0
conda 22.11.1

$ mamba create -y -n foo flask python=3.10
$ mamba remove -n foo flask

  Removing specs:

   - flask


  Package               Version  Build              Channel         Size
──────────────────────────────────────────────────────────────────────────
  Remove:
──────────────────────────────────────────────────────────────────────────

  - __unix                    0  0                  installed
  - click                 8.1.3  unix_pyhd8ed1ab_2  conda-forge
  - flask                 2.2.2  pyhd8ed1ab_0       conda-forge
  - importlib-metadata    6.0.0  pyha770c72_0       conda-forge
[...]

Issue #1794, opened 2022-07-14
Bug in mamba remove [package] when [package] depends on cuda 


jamartinh:
Remove:
───────────────────────────────────────────────────────────────────────

  - __cuda       11.4  0                         installed

KeyError: PackageRecord(_hash=5352623804000103275, name='__cuda', version='11.4', build='0', build_number=0, channel=Channel("@"), subdir='linux-64', fn='__cuda', md5='12345678901234567890123456789012', package_type='virtual_system')

jonashaag:
Same as https://github.com/mamba-org/mamba/issues/1770?

Kaixhin:
I have also come across this issue when trying to `mamba uninstall pytorch`, with this cryptic message about the above `__cuda` package appearing when it fails.

Fissium:
Is there still no solution?

jjerphan:
I cannot reproduce the problem with mamba 1.5.10 and mamba 2.0.2.

Feel free to comment with a reproducer if you observe this issue again.

Issue #1770, opened 2022-06-29
`mamba uninstall cudatoolkit` fails (conda equivalent works)


pierreglaser:
Steps to reproduce: in a shell, run:

```sh
mamba create -n condaforgecuda cudatoolkit
mamba uninstall -n condaforgecuda cudatoolkit
```

First line works fine, second fails with:
```sh
Removing specs: ['cudatoolkit']
Transaction

  Prefix: /nfs/nhome/live/pierreg/.local/miniforge/envs/condaforgecuda

  Removing specs:

   - cudatoolkit


  Package          Version  Build        Channel         Size
───────────────────────────────────────────────────────────────
  Remove:
───────────────────────────────────────────────────────────────

  - __glibc           2.31  0            installed
  - _libgcc_mutex      0.1  conda_forge  conda-forge
  - _openmp_mutex      4.5  2_gnu        conda-forge
  - cudatoolkit     11.7.0  hd8887f6_10  conda-forge
  - libgcc-ng       12.1.0  h8d9b700_16  conda-forge
  - libgomp         12.1.0  h8d9b700_16  conda-forge
  - libstdcxx-ng    12.1.0  ha89aaad_16  conda-forge

  Summary:

  Remove: 7 packages

  Total download: 0 B

───────────────────────────────────────────────────────────────

Confirm changes: [Y/n] Y

# >>>>>>>>>>>>>>>>>>>>>> ERROR REPORT <<<<<<<<<<<<<<<<<<<<<<

    Traceback (most recent call last):
      File "/nfs/nhome/live/pierreg/.local/miniforge/lib/python3.9/site-packages/conda/_vendor/boltons/setutils.py", line 247, in remove
        didx = self.item_index_map.pop(item)
    KeyError: PackageRecord(_hash=-1097009890971377200, name='__glibc', version='2.31', build='0', build_number=0, channel=Channel("@"), subdir='linux-64', fn='__glibc', md5='12345678901234567890123456789012', package_type='virtual_system')

    During handling of the above exception, another exception occurred:

    Traceback (most recent call last):
      File "/nfs/nhome/live/pierreg/.local/miniforge/lib/python3.9/site-packages/conda/exceptions.py", line 1114, in __call__
        return func(*args, **kwargs)
      File "/nfs/nhome/live/pierreg/.local/miniforge/lib/python3.9/site-packages/mamba/mamba.py", line 935, in exception_converter
        raise e
[...]

PR #2923, opened 2023-10-19
Fix support for the PackageType.VIRTUAL_SYSTEM enum

ogrisel:
While trying to use conda-lock with the latest conda and mamba on scikit-learn, I got errors due to the use of recently introduced `conda.models.enums.PackageType`.

Here is a PR to attempt to fix it while preserving backward compat.

Probably fixes: #2129, #1794, #1770.


ogrisel:
Note: since some of the linked issues are really old. The commit in conda that introduced the `PackageType.VIRTUAL_SYSTEM` enum is 4 years old. So maybe it has always been an enum rather than a string and the backward compat protection in my PR is useless:

- https://github.com/conda/conda/commit/f32904bef0606fab12ee9d87380ca4201c5c4996

mamba/mamba/utils.py
```diff
@@ -26,6 +26,19 @@
...
+try:
+    from conda.models.enums import PackageType
+except ImportError:
+    PackageType = None
+
+
+def _is_virtual_system(rec):
+    if PackageType is None:
+        # Backward compat for older conda versions
+        return rec.package_type == "virtual_system"
```
ogrisel, line 38: Based on my latest comment above, I suppose that this is not needed and we could always import the `PackageType` enum instead. Let me know what you think.

ogrisel:
I am not sure how to craft a minimal reproducer from the linked issues that wouldn't require access to the network and I am familiar with how you write tests for this project.

Would you prefer a minimal unit test? Or an integration test, for instance for the case reported in #1770 using `check_output` in `test_all.py`?

jonashaag:
Any test is fine really. We have lots of tests that need the network.

ogrisel:
I am having trouble setting up a dev environment for mamba following the instructions found here: https://mamba.readthedocs.io/en/latest/developer_zone/build_locally.html:

```
$ mamba env create --name mambadev --file ./mamba/environment-dev.yml
...
Installing collected packages: securesystemslib
Successfully installed securesystemslib-0.30.0

done
Unable to install package for sel(win).

Please double check and ensure your dependencies file has
the correct spelling.  You might also try installing the
conda-env-sel(win) package to see if provides the required
installer.
```

ogrisel:
@jonashaag I have pushed a test. I am could not get my test setup to actually use my dev `mamba/mamba/utils.py` (even when following the instructions of the contributor guide), it would still use the `mamba/mamba/utils.py` from my `miniforge3/lib/python3.10/site-packages` for some reason.

Still I think the PR contents is working as intended and hopefully your CI can confirm this.

jonashaag:
Perfect thank you! Would you mind also adding a Linux test? Could be a copy paste or a parameteized test

jonashaag (approved):

ogrisel:
> Could be a copy paste or a parameterized test

I updated the test to parameterize it for Linux and Windows (in addition to macOS).

AntoinePrv:
Thanks for this ! It should also be backported to the `1.x` branch if you want to see it a `1.5.3` release (not planned for now).

JohanMabille:
We should fix the build first.

ogrisel:
Indeed, I hardcoded the use of bash because the tests was originally written only for macOS. Now that it was parameterized for windows, we need to unhardcode the shell. I will do a follow-up PR.

ogrisel:
@jonashaag @AntoinePrv @JohanMabille hopefully #2924 should fix the Windows CI.

PR #2924, opened 2023-10-20
Fix test_remove_virtual_package on Windows

ogrisel:
Follow-up fix for the recently merged #2923.


AntoinePrv (approved):

JohanMabille:
This fix is for mamba on Windows, which passed the CI, so no need to wait for other jobs.

PR #2929, opened 2023-10-22
Fix broken tcsh aliases

bruchim-cisco:
Fix broken csh and tcsh multiline aliases, in order for tcsh hook to be run as separate file. The hook still needs to be written to be written to an xternal file then sourced because the parentheses in the first row of the comments will cause an error, but this is a temporary fix until the tcsh script is ported from conda.

### DIFF
--- a/libmamba/data/micromamba.csh
+++ b/libmamba/data/micromamba.csh
@@ -5,40 +5,40 @@ alias __mamba_exe '"$MAMBA_EXE" "\!*"'
 
 alias __mamba_hashr 'rehash'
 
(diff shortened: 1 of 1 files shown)


SylvainCorlay:
cc @wolfv

SylvainCorlay:
Should this be backported to 1.x?

JohanMabille (approved):

JohanMabille:
Thanks!

JohanMabille:
> Should this be backported to 1.x?

It think so.

PR #2946, opened 2023-10-28
Fix ctrl-c handling issues in libmambapy

### LINKED ISSUE #1594: Instantiating `libmambapy` objects blocks `KeyboardInterrupt` exceptions
Originally raised by @jakirkham at https://github.com/conda/conda/issues/11361

---

Stdin prompts cannot be Ctrl+C'd out if a `libmambapy` object was instantiated. I can reproduce this with `Pool()`.

This affects `mamba` and, subsequently, `conda_libmamba_solver`. I guess any Python process is affected by this, with or without conda.

Reproducer:

```python
from libmambapy import Pool

if __name__ == "__main__":
    Pool()
    while True:
        answer = input("Type X to exit: ")
        if answer.lower() == "x":
            break
```

If you try to Ctrl+C out of this, you can't because `Pool()` was instantiated. You can only press `x` and Enter. However, if you comment out the `Pool()` line , it works as expected: you can Ctrl+C because an exception is raised. Note that `Pool()` was not stored in a Python reference at all; it should have been garbage collected, but somehow it still lingers around.

jaimergp:
Closes #1594


jonashaag (commented):
I don't like that this special cases Python. Could we add a flag to enable/disable the default signal handler instead?

AntoinePrv (approved):

dholth:
Do we also need to add set_interrupted etc. to allow Python to drive mamba's signal handling?

Will mamba get confused when it tries to replace the signal handler for certain i/o?

AntoinePrv:
Our attempts with @jaimergp to drive signal handling from Python were not successful this is an improvement for now.

PR #2952, opened 2023-10-30
Refactor CI and libamambapy tests

AntoinePrv:
- Add simple libmambapy test
- Use CMakePresets in CI


pavelzw:
I've never done this sort of stuff in CI, only with my local PAT 👀

jonashaag:
Can you give more context please? What exactly are you trying to do that doesn't work? Is it a solution for you to manually delete the cache through GitHub's UI?

AntoinePrv:
I just want to divide the workflow into `build` > `test-libmamba` | `test-libmambapy` | `test-micromamba`, but apparently there is no easy way to share data between jobs.

I don't understand how using artifacts is Github's recomended solution: it's very annoying to get a proper key, and it will definitly blow up the storage data, when all you want is to share data for the short lived duration of the workflow :shrug:

jonashaag:
Hmm yeah probably there is not good way to do this in GHA.

AntoinePrv:
Permission denied, now. How user unfriendly does this have to be :roll_eyes:

JohanMabille (approved):

PR #2953, opened 2023-10-31
Channel initialization

AntoinePrv:
- Set default_channels at config level
- Resolve default multi channel at the config level
- Fix Yaml source printers
- Resolve local multi channel at the config level
- Remove ChannelContext::make_simple_channel


libmamba/include/mamba/core/channel.hpp
```diff
@@ -26,24 +26,36 @@ namespace mamba
...
 
-        ~Channel();
 
-        const std::string& location() const;
-        const std::string& name() const;
+        Channel(
+            specs::CondaURL url,
+            std::string canonical_name,
+            util::flat_set<std::string> platforms = {}
+        );
```
JohanMabille, line 57: Now that this constructor is public, do we still need the `make_channel` method in the ChannelContext?

JohanMabille (approved):

PR #2960, opened 2023-11-06
Modularize libmambapy

AntoinePrv:
- Modularize libmambapy
- Add test-libmambapy to Taskfile


JohanMabille (changes requested):
I find it confusing to have the file `main.cpp` in the core subfolder (implying it's not the main for the whole library), defining the bindings, and having the main macro `PYDBIND11_MODULE` called in the file `core.cpp` file. It looks to me like the logic is inverted.

Even if we split the bindings in many files (which I agree is far better than a huge monolithic file), we will keep a single global module (libmambapy), and the name passed to the `PYBIND11_MODULE` macro should be the same as that of the extension library (dll/dylib/so) to avoid [this issue](https://pybind11.readthedocs.io/en/stable/faq.html?highlight=PYBIND11_MODULE#importerror-dynamic-module-does-not-define-init-function) (although the tests seem to pass, so I might have missed something).

A more natural code organization would be one pair of hpp/cpp per libmamba subfolder that we want to bind, and the `main.cpp` defining the module (i.e. calling PYBIND11_MODULE) and calling the `bind_submodule` functions defined in the other files.

JohanMabille (approved):

PR #2963, opened 2023-11-07
Fixed move semantics of DownloadAttempt

JohanMabille:


libmamba/src/core/download_impl.hpp
```diff
@@ -34,9 +34,11 @@ namespace mamba
         using on_success_callback = std::function<bool(DownloadSuccess)>;
         using on_failure_callback = std::function<bool(DownloadError)>;
 
-        explicit DownloadAttempt(const DownloadRequest& request);
+        DownloadAttempt() = default;
```
Klaim, line 37: I'm not totally sure but I think to link properly you need `DownloadAttempt::DownloadAttempt()=default;`  in the cpp (moving the definition in the cpp).
If you get no link issues in msvc, ignore this comment.

libmamba/src/core/download_impl.hpp
```diff
@@ -47,43 +49,60 @@ namespace mamba
...
+            DownloadError build_download_error(CURLcode code) const;
+            DownloadError build_download_error(TransferData data) const;
+            DownloadSuccess build_download_success(TransferData data) const;
+
+            CURLHandle* p_handle = nullptr;
+            const DownloadRequest* p_request = nullptr;
+            on_success_callback m_success_callback;
+            on_failure_callback m_failure_callback;
+            std::size_t m_retry_wait_seconds = std::size_t(0);
+            std::unique_ptr<CompressionStream> p_stream = nullptr;
```
Klaim, line 98: BTW can this be a normal member now? or an optional?

libmamba/src/core/download_impl.hpp
```diff
@@ -47,43 +49,60 @@ namespace mamba
...
+            std::ofstream m_file;
+            std::string m_cache_control;
+            std::string m_etag;
+            std::string m_last_modified;
+        };
+
+        std::unique_ptr<Impl> p_impl = nullptr;
```
Klaim, line 105: Note that it's already `nullptr` by default.

libmamba/src/core/download_impl.hpp
```diff
@@ -47,43 +49,60 @@ namespace mamba
...
+
+            bool can_retry(CURLcode code) const;
+            bool can_retry(const TransferData& data) const;
+
+            TransferData get_transfer_data() const;
+            DownloadError build_download_error(CURLcode code) const;
+            DownloadError build_download_error(TransferData data) const;
+            DownloadSuccess build_download_success(TransferData data) const;
+
+            CURLHandle* p_handle = nullptr;
```
Klaim, line 93: Isnt handle already a value/pointer? Do you need to point to it?

libmamba/src/core/download.cpp
```diff
@@ -126,33 +126,47 @@ namespace mamba
...
+    )
+        : p_impl(std::make_unique<
+                 Impl>(handle, request, downloader, context, std::move(success), std::move(error)))
+    {
+    }
+
+    auto DownloadAttempt::create_completion_function() -> completion_function
     {
-        m_retry_wait_seconds = std::size_t(0);
+        return [impl = p_impl.get()](CURLMultiHandle& handle, CURLcode code)
```
Klaim, line 144: If you find issues later with this lambda being called after `impl` have been destroyed, consider using task_synchronizer to guarantee that this does not ever happen.

Klaim (approved):
LGTM, mostly have questions for details.

libmamba/src/core/download_impl.hpp
JohanMabille: With this change, the handle is now a member (value) of DownloadTracker, to avoid recreating it each time we build an attempt. BUt the attempt still needs it, thus this pointer.

libmamba/src/core/download_impl.hpp
JohanMabille: It cannot because CompressionStream is a polymorphic type.

PR #2969, opened 2023-11-12
Prototype two-file "index plus updates" repodata format

dholth:
Attempt to add libmamba support for https://github.com/conda/conda/blob/b8ef64e4ad27b36d39e519188203d0e2ff43925b/conda/gateways/repodata/jlap/README.md

Which collects patches into a second, smaller repodata.json overlay to avoid the problem of having to re-serialize the entire file all the time. If a package exists in the overlay, we skip loading it from the upstream repodata.

My C++ is rusty. I haven't yet had luck importing the mamba refactor's Python bindings. I appreciate the C++ parser compared to what libsolv is doing.


jjerphan:
Closing this PR as sharded repodata is now the preferred option over JLAP; please let us know if JLAP is actually used.

PR #2981, opened 2023-11-15
Add weakening_map

AntoinePrv:
A type of map that weakens key to find a match.
The generic implementation of `AuthenticationDataBase`, also needed for `custom_channels`.

### DIFF
--- a/libmamba/include/mamba/util/weakening_map.hpp
+++ b/libmamba/include/mamba/util/weakening_map.hpp
@@ -1,0 +1,160 @@
+// Copyright (c) 20123 QuantStack and Mamba Contributors
+//
+// Distributed under the terms of the BSD 3-Clause License.
+//
+// The full license is in the file LICENSE, distributed with this software.
+
+#ifndef MAMBA_UTIL_WEAKENING_MAP_HPP
+#define MAMBA_UTIL_WEAKENING_MAP_HPP
+
+#include <stdexcept>
+
+#include <fmt/format.h>
+
+namespace mamba::util
+{
+    /**
+     * A Map wrapper that can weaken a key to find more matches.
+     *
+     * The API of a standard map is unmodified, only methods ending with "weaken" look for
+     * multiple keys.
+     * This can be understood as an extreme generalization of defaults: when a key is not found,
+     * the behaviour is to look for another key.
+     * The behaviour for generating the sequence of weaken keys is controled by the Weakener.
+     */
+    template <typename Map, typename Weakener>
+    class weakening_map : private Map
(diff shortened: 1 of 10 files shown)


libmamba/include/mamba/util/weakening_map.hpp
```diff
@@ -0,0 +1,149 @@
...
+        using typename Base::mapped_type;
+        using typename Base::value_type;
+        using typename Base::size_type;
+        using typename Base::iterator;
+        using typename Base::const_iterator;
+        using weakener_type = Weakener;
+
+        using Base::Base;
+        template <typename... Args>
+        weakening_map(weakener_type weakener, Args&&... args);
```
JohanMabille, line 38: Should be declared as explicit

libmamba/include/mamba/util/weakening_map.hpp
```diff
@@ -0,0 +1,149 @@
...
+        using Base::swap;
+        using Base::extract;
+        using Base::merge;
+
+        using Base::reserve;
+
+        using Base::at;
+
+        [[nodiscard]] auto at_weaken(const key_type& key) -> mapped_type&;
+        [[nodiscard]] auto at_weaken(const key_type& key) const -> const mapped_type&;
```
JohanMabille, line 65: Should this replace the `at` API? Asked differently, does the user needs both methods? Is there a need to distinguish between strict finding and weaken finding?

libmamba/include/mamba/util/weakening_map.hpp
```diff
@@ -0,0 +1,149 @@
...
+
+        using Base::find;
+
+        auto find_weaken(const key_type& key) -> iterator;
+        auto find_weaken(const key_type& key) const -> const_iterator;
```
JohanMabille, line 70: Same question here.

libmamba/include/mamba/util/weakening_map.hpp
```diff
@@ -0,0 +1,149 @@
...
+
+        [[nodiscard]] auto contains(const key_type& key) const -> bool;
+
+        [[nodiscard]] auto contains_weaken(const key_type& key) const -> bool;
```
JohanMabille, line 74: And here.

libmamba/include/mamba/util/weakening_map.hpp
AntoinePrv: I'd say it's still relevant, for instance if users want to remove an entry, they would more likely want an exact match.

JohanMabille (approved):

PR #2985, opened 2023-11-16
Fail fast except on debug runs

AntoinePrv:


AntoinePrv:
No need to wait for tests to complete.

PR #3000, opened 2023-11-21
Move core/channel > specs/channel

AntoinePrv:


libmamba/include/mamba/core/subdirdata.hpp
```diff
@@ -12,18 +12,22 @@
...
-#include "mamba/core/pool.hpp"
 #include "mamba/core/repo.hpp"
 #include "mamba/core/util.hpp"
 #include "mamba/fs/filesystem.hpp"
 
 namespace mamba
 {
+    namespace specs
+    {
+        class Channel;
```
JohanMabille, line 26: Why moving this class to the `specs` folder (and inside the specs namespace)? I agree with the overall architecture, but for me Channel is the result of resolving the spec, and therefore is a core component. O do you plan to have core depending on `specs` and you want to avoid circular dependencies between layers?

libmamba/src/core/channel_context.cpp
```diff
@@ -0,0 +1,96 @@
...
+    ChannelContext::ChannelContext(Context& context)
+        : m_context(context)
+    {
+        m_channel_params.channel_alias = specs::CondaURL::parse(
+            util::path_or_url_to_url(m_context.channel_alias)
+        );
+        m_channel_params.home_dir = util::user_home_dir();
+        m_channel_params.current_working_dir = fs::current_path();
+        m_channel_params.authentication_db = m_context.authentication_info();
+        m_channel_params.platforms = create_paltforms(m_context.platforms());
```
JohanMabille, line 43: - These members should been initialized with a member initializer list
- typo: create_paltforms => create_platforms (I guess the declaration and the definition have the typo too, since it compiles)

libmamba/src/core/channel_context.cpp
```diff
@@ -0,0 +1,96 @@
...
+                for (auto& chan : make_channel(location))
+                {
+                    channels.push_back(std::move(chan));
+                }
+            }
+            m_channel_params.custom_multichannels.emplace(multi_name, std::move(channels));
+        }
+    }
+
+    auto ChannelContext::make_channel(std::string_view name) -> channel_list
```
JohanMabille, line 72: Why returning by value instead of ref?

libmamba/src/core/subdirdata.cpp
```diff
@@ -693,8 +699,8 @@ namespace mamba
                 m_name + " (check zst)",
                 m_repodata_url + ".zst",
                 "",
-                /*head_only = */ true,
-                /*ignore_failure = */ true
+                /* lhead_only = */ true,
+                /* lignore_failure = */ true
```
JohanMabille, line 703: Why these changes? The members of DownloadRequests are named `head_only` and `ignore_failure`, this is for preparing C++20 aggregate initialization (actually, we are missing the . in front of each member)

AntoinePrv: C++20 aggregate initialization works for aggregate initialization, here we have a regular constructor call so nothing will change in C++20. Adding the `.` won't change anything.

The names of the parameters used in the constructor are prefix with `l`, I made it match because clang-tidy can check the comment and emit a warning when function parameters do not match the one in the comment.

libmamba/include/mamba/core/subdirdata.hpp
AntoinePrv: Maybe the `specs` namespace is not adequately named. It is intended to hold all abstractions that could in theory be exchanged with Conda. All the abstraction are *descriptive*: they are functional (no side effects), and hold only data. A protocol in a sense. It has `VersionSpec` and `Version`, `ChannelSpec`, `Platform`...

`ChannelContext` is a glue/machinery component and does not belong there.

libmamba/src/core/subdirdata.cpp
JohanMabille: > here we have a regular constructor call so nothing will change in C++20. Adding the . won't change anything

Yeah, the idea was to remove this call in C++20, but I agree having the clang-tidy check is better.

JohanMabille (approved):

PR #3001, opened 2023-11-22
Bind Channel

AntoinePrv:


libmamba/include/mamba/specs/channel_spec.hpp
```diff
@@ -89,13 +89,16 @@ namespace mamba::specs
         ChannelSpec(std::string location, dynamic_platform_set filters, Type type);
 
         [[nodiscard]] auto type() const -> Type;
+        void set_type(Type t);
```
JohanMabille, line 92: I really don't like having setters for all members. This breaks the encapsulation and make it equivalent to have everything as public members.

libmamba/include/mamba/specs/channel_spec.hpp
```diff
@@ -89,13 +89,16 @@ namespace mamba::specs
...
 
         [[nodiscard]] auto location() const& -> const std::string&;
         [[nodiscard]] auto location() && -> std::string;
+        void set_location(std::string location);
```
JohanMabille, line 96: Same here

libmamba/include/mamba/specs/channel_spec.hpp
```diff
@@ -89,13 +89,16 @@ namespace mamba::specs
...
         auto clear_location() -> std::string;
 
         [[nodiscard]] auto platform_filters() const& -> const dynamic_platform_set&;
         [[nodiscard]] auto platform_filters() && -> dynamic_platform_set;
+        void set_platform_filters(dynamic_platform_set filters);
```
JohanMabille, line 101: Same here

libmamba/include/mamba/specs/conda_url.hpp
```diff
@@ -35,7 +35,7 @@ namespace mamba::specs
         explicit CondaURL(util::URL&& url);
         explicit CondaURL(const util::URL& url);
 
-        auto base() const -> const util::URL&;
+        [[nodiscard]] auto generic() const -> const util::URL&;
```
JohanMabille, line 38: I find `base` more meaningful

libmambapy/src/libmambapy/bindings/specs.cpp
```diff
@@ -0,0 +1,466 @@
...
+            .def(
+                py::init<std::string, ChannelSpec::dynamic_platform_set, ChannelSpec::Type>(),
+                py::arg("location"),
+                py::arg("platform_filters"),
+                py::arg("type") = ChannelSpec::Type::Unknown
+            )
+            .def("__copy__", &copy<ChannelSpec>)
+            .def("__deepcopy__", &deepcopy<ChannelSpec>, pybind11::arg("memo"))
+            .def_property("type", &ChannelSpec::type, &ChannelSpec::set_type)
+            .def_property("location", &ChannelSpec::location, &ChannelSpec::set_location)
```
JohanMabille, line 291: I guess these are the culprits for the setters in the C++ code. Do we really need that on the Python side? Can't we require the objects to be constructed once and for good?

libmamba/include/mamba/specs/conda_url.hpp
AntoinePrv: The fact the it the base is not relevant here. What matters is that there is a generic alternative to this class to which we can convert.

JohanMabille (approved):

Issue #3005, opened 2023-11-23
Improve new `search` output for PyTorch queries


jonashaag:
> The new output style is awful for AI developers. The info of `build` column is essential to install `pytorch` or `tensorflow`, because we need to choose the correct python version and cuda version. 

The old style:

```sh
$ ./micromamba-1.5.0-1 search pytorch=1.9.1
Getting repodata from channels...

conda-forge/linux-64                                        Using cache
conda-forge/noarch                                          Using cache


 Name    Version Build                 Channel
───────────────────────────────────────────────────
 pytorch 1.9.1   cuda110py38hf84197b_3 conda-forge
 pytorch 1.9.1   cuda112py39h4e14dd4_3 conda-forge
 pytorch 1.9.1   cuda112py38h4f2a933_3 conda-forge
 pytorch 1.9.1   cuda112py37h3bec1eb_3 conda-forge
 pytorch 1.9.1   cuda112py36h36e649e_3 conda-forge
 pytorch 1.9.1   cuda111py39hb4a4491_3 conda-forge
 pytorch 1.9.1   cuda111py38h2f85826_3 conda-forge
 pytorch 1.9.1   cuda111py37he371307_3 conda-forge
 pytorch 1.9.1   cuda111py36h8a2106e_3 conda-forge
 pytorch 1.9.1   cuda110py39h5cf7045_3 conda-forge
 pytorch 1.9.1   cpu_py36ha8b20dc_3    conda-forge
 pytorch 1.9.1   cuda110py37h4a33b93_3 conda-forge
 pytorch 1.9.1   cuda110py36he570edd_3 conda-forge
 pytorch 1.9.1   cuda102py39h2fcd037_3 conda-forge
 pytorch 1.9.1   cuda102py38ha031fbe_3 conda-forge
 pytorch 1.9.1   cuda102py37h98b7ee3_3 conda-forge
 pytorch 1.9.1   cuda102py36h3d4679f_3 conda-forge
 pytorch 1.9.1   cpu_py39hc5866cc_3    conda-forge
 pytorch 1.9.1   cpu_py38h1ee18c8_3    conda-forge
 pytorch 1.9.1   cpu_py37hf3cc979_3    conda-forge
``` 

The new style:
```sh
$ ./micromamba-1.5.3-0 search pytorch=1.9.1
Getting repodata from channels...

conda-forge/linux-64                                        Using cache
conda-forge/noarch                                          Using cache


[...]

PR #2823, opened 2023-09-08
Further improve `micromamba search` output

delsner:
# Motivation

Minor fixes and changes to aesthetics in the current `micromamba search` output.

# Changes

- Fix duplicate build SHAs for packages from multiple channels 
- Fix whitespace padding in single package view
- Change underscore to unicode dash `─`
- Add version collapsing in single package view for >5 versions
- Use `-v` to show all builds as requested in #2847

# Examples

<details>
  <summary>Single package view</summary>

## Single package view

```
❯ ./build/mamba-shared-debug/micromamba/micromamba search "pandas"

pandas 2.1.0 py39h425d09f_0 (+ 2 builds)
────────────────────────────────────────

 Name            pandas
 Version         2.1.0
 Build           py39h425d09f_0
 Size            11984 kB
 License         BSD-3-Clause
 Subdir          osx-arm64
 File Name       pandas-2.1.0-py39h425d09f_0.conda
 URL             https://*****conda.anaconda.org/conda-forge/osx-arm64/pandas-2.1.0-py39h425d09f_0.conda
 MD5             3f819b036f93bdf1bc7c84f5e6ce3d06
 SHA256          22ca3c15e5212e4c00ce190ef8c935406d4150e32277ee784fb7e0362ef2a30e

 Dependencies:
  - python_abi 3.9.* *_cp39
  - python >=3.9,<3.10.0a0
  - python >=3.9,<3.10.0a0 *_cpython
  - libcxx >=15.0.7
  - numpy >=1.22.4,<2.0a0
  - python-dateutil >=2.8.1
  - pytz >=2020.1
  - python-tzdata >=2022a

 Other Versions (28):

  Version Build
 ───────────────────────────────────────────
  2.0.3   py39h6b13a34_1       (+ 7 builds)
  2.0.2   py39h6b13a34_0       (+ 3 builds)
  ...     (24 hidden versions)          ...
  1.1.4   py39ha4bedbf_0       (+ 1 builds)
  1.1.3   py39h88a2970_2       (+ 1 builds)
```

</details>

<details>
  <summary>Single package verbose view with all builds</summary>

## Single package verbose view with all builds (`-v`)

```
❯ ./build/mamba-shared-debug/micromamba/micromamba search "pandas" -v
Getting repodata from channels...
...
info     libmamba Metadata from solv are valid, loading successful


      pandas 2.1.0 py39h425d09f_0
────────────────────────────────────────

 Name            pandas
 Version         2.1.0
 Build           py39h425d09f_0
[...]

PR #3006, opened 2023-11-23
Backport: Further improve search output (#2823)

delsner:
This backports patch #2823 to 1.x in order to address https://github.com/mamba-org/mamba/issues/3005.

### DIFF
--- a/libmamba/src/core/query.cpp
+++ b/libmamba/src/core/query.cpp
@@ -173,15 +173,10 @@ namespace mamba
             fmt::print(out, fmtstring, "Subdir", pkg.subdir);
             fmt::print(out, fmtstring, "File Name", pkg.fn);
 
-            using CondaURL = typename specs::CondaURL;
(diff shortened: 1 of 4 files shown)


libmamba/src/core/query.cpp
```diff
@@ -800,12 +881,13 @@ namespace mamba
...
             }
-
             for (const auto& entry : packages)
             {
                 print_solvable(
+                    out,
                     entry.second[0],
-                    std::vector(entry.second.begin() + 1, entry.second.end())
+                    std::vector(entry.second.begin() + 1, entry.second.end()),
+                    Context::instance().output_params.verbosity > 0
```
delsner, line 890: This is a slight difference to the original patch #2823; since the `context` singleton isn't available in the entrypoint in libmamapy yet, we're retrieving the context singleton here directly.

delsner:
CI seems to fail due to `pytest.skip` issue for which I backported another patch (#2952) in #3007

PR #3034, opened 2023-12-05
Bind ChannelContext

AntoinePrv:
- Bind ChannelContext
- Deprecate Context() in Python

### DIFF
--- a/libmambapy/src/libmambapy/bindings/legacy.cpp
+++ b/libmambapy/src/libmambapy/bindings/legacy.cpp
@@ -715,9 +715,23 @@ bind_submodule_impl(pybind11::module_ m)
         .def("has_zst", &ChannelContext::has_zst);
 
     py::class_<Context, std::unique_ptr<Context, py::nodelete>> ctx(m, "Context");
-    ctx.def(py::init(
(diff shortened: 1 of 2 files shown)


libmambapy/src/libmambapy/bindings/legacy.cpp
```diff
@@ -696,10 +696,42 @@ bind_submodule_impl(pybind11::module_ m)
         .value("CRITICAL", mamba::log_level::critical)
         .value("OFF", mamba::log_level::off);
 
+    py::class_<ChannelContext>(m, "ChannelContext")
+        .def_static(
+            // Still need a singleton as long as mambatest::singleton::channel_context is used
```
AntoinePrv, line 701: Wondering whether to leave this, or to take the opportunity to get rid of the ChannelContext singleton altogether.

There aren't that many breakage. The most annoying one is `MatchSpec` which now requires an explicit `ChannelContext` while this is something I wanted to get rid of (but we have so many things to do).

JohanMabille: I think we can keep it for now, but it would be nice to remove it before releasing 2.0

JohanMabille (approved):

PR #3050, opened 2023-12-08
tuf

AntoinePrv:
- RepoCheckerStore skeleton
- Simplify Context sub params
- Const context in validation
- RepoChecker reference_wrapper
- noexcept u8path move


SylvainCorlay:
Do you think this should get in @AntoinePrv @Hind-M ?

AntoinePrv:
IMHO no, this was only a draft.

Hind-M:
> Do you think this should get in @AntoinePrv @Hind-M ?

No because it needs tweaking. But I'm working on something based on it anyway...

PR #3054, opened 2023-12-10
Shouldn't we rather deprecate mamba?

jonashaag:
Maybe controversial change? I'd like to propose that we're not recommending new Mamba installations anymore. Will help keep the support volume down.

How do we trigger a RTD build?

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -1,24 +1,18 @@
 Welcome to Mamba's documentation!
 =================================
 
-Mamba is a fast, robust, and cross-platform package manager.
-
-It runs on Windows, OS X and Linux (ARM64 and PPC64LE included) and is fully compatible with ``conda`` packages and supports most of conda's commands.
+Mamba is a fast, robust, and cross-platform package manager for Conda packages.
 
 The ``mamba-org`` organization hosts multiple Mamba flavors:
 
-- ``mamba``: a Python-based CLI conceived as a *drop-in* replacement for ``conda``, offering higher speed and more reliable environment solutions
-- ``micromamba``: a pure C++-based CLI, self-contained in a single-file executable
-- ``libmamba``: a C++ library exposing low-level and high-level APIs on top of which both ``mamba`` and ``micromamba`` are built
+- Micromamba: a pure C++-based CLI, self-contained in a single-file executable.
+- libmamba: a C++ library exposing low-level and high-level APIs on top of which both Mamba and Micromamba are built.
+- (Mamba: Legacy software; a Python-based CLI conceived as a *drop-in* replacement for Conda, offering higher speed.)
 
-.. note::
-   In this documentation, ``Mamba`` will refer to all flavors while flavor-specific details will mention ``mamba``, ``micromamba`` or ``libmamba``.
-
-.. note::
-    :ref:`micromamba<micromamba>` is especially well fitted for the CI use-case but not limited to that!
+In this documentation, "Mamba" will refer to all flavors while flavor-specific details will mention Mamba, Micromamba, or libmamba.
 
 You can try Mamba now by visiting the installation for
-:ref:`mamba<mamba-install>` or :ref:`micromamba<umamba-install>`
+:ref:`mamba<mamba-install>` or :ref:`micromamba<umamba-install>`.
 
 
 .. toctree::
--- a/docs/source/installation/mamba-installation.rst
+++ b/docs/source/installation/mamba-installation.rst
@@ -4,11 +4,23 @@
 Mamba Installation
 ==================
 
-Fresh install (recommended)
-***************************
+.. warning::
+   With the release of Conda 23.10, **Mamba is deprecated**.
 
-We recommend that you start with the `Miniforge distribution <https://github.com/conda-forge/miniforge>`_ >= `Miniforge3-22.3.1-0`.
-If you need an older version of Mamba, please use the Mambaforge distribution.
+   Please do not create any new installations of Mamba.
+
+   Please use :ref:`Micromamba <umamba-install>` or `Miniforge <https://github.com/conda-forge/miniforge>`_ >= `23.3.1-0` instead.
+
+
+Legacy installation methods (not recommended)
+---------------------------------------------
+
+Mambaforge/Miniforge
+*********************
+
+You can use the `Miniforge distribution <https://github.com/conda-forge/miniforge>`_ >= `Miniforge3-23.3.1-0` to install Mamba.
+If you need an older version of Mamba, please use the Mambaforge distribution instead.
+
 Miniforge comes with the popular ``conda-forge`` channel preconfigured, but you can modify the configuration to use any channel you like.
 
 After successful installation, you can use the mamba commands as described in :ref:`mamba user guide<mamba>`.
--- a/docs/source/installation/micromamba-installation.rst
+++ b/docs/source/installation/micromamba-installation.rst
@@ -5,19 +5,21 @@
 Micromamba Installation
 =======================
 
-``micromamba`` is a fully statically-linked, self-contained, executable.
-This means that the ``base`` environment is completely empty.
-The configuration for ``micromamba`` is slighly different, namely all environments and cache will be
-created by default under the ``MAMBA_ROOT_PREFIX`` environment variable.
-There is also no pre-configured ``.condarc``/``.mambarc`` shipped with micromamba
-(they are however still read if present).
+Micromamba is a fully statically-linked, self-contained, executable written in C++.
+
+Micromamba is not require Conda to be installed on your system; it is entirely self-contained.
+
+The ``base`` environment is completely empty.
+
+The configuration for Micromamba is slighly different from Conda:
+
+- All environments and cache will be created by default under the ``MAMBA_ROOT_PREFIX`` environment variable.
+- There is no pre-configured ``.condarc``/``.mambarc`` shipped with micromamba (they are however still read if present).
 
 .. _umamba-install-automatic-installation:
 
-Operating System package managers
-*********************************
 Homebrew
-^^^^^^^^
+********
 
 On macOS, you can install ``micromamba`` from `Homebrew <https://brew.sh/>`_:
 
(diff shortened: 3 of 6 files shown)


JohanMabille:
Actually we still want to provide mamba, but as a dynamic build of micromamba. So maybe advertising that it is not Python anymore would be better?

Regarding the RTD build, it is triggered upon merge only, not PR.

jonashaag:
> but as a dynamic build of micromamba

What does that mean?

> Regarding the RTD build, it is triggered upon merge only, not PR.

So we don't have any docs CI?

AntoinePrv:
Personally I also don't see the point for a `mamba` dynamic build in the conda ecosystem.
This is more confusing than ever now since `micromamba` and would be `mamba` are in fact the same executable.
We can provide a `mamba-forge` with `micromamba` (that we could rename mamba) in it to preserve the distribution feeling, without risking user breaking their base environment, and reducing our workload.

The only point of building it dynamically is for packagers such as `brew`, `apt`, etc. and local development.

Having two names will just maintain the confusion in the ecosystem, all the while making the code base more buggy.

jonashaag:
Unless we quickly make a decision on this and add documentation on the current situation, I predict that there will be a lot of support tickets created.

AntoinePrv:
What's wrong with the current doc? 1.x is still the stable release of Mamba.
After which, the `MiniForge` will still be the default way of installing a dynamically linked executable called `mamba`.
`micromamba` will change name but this is not done for now.

Am I missing something? Sorry if we did not make the best job at communicating our intents (they still moving targets). Happy to hop on a call with you to discuss the next steps.

jonashaag:
Nothing problematic per se. It would be nice to reduce the amount of confusion in the community by deprecating Mamba. But if you are planning to do something other than deprecating Mamba then my recommendation probably doesn't make any sense. 

Let me say again though that I think it is a terrible idea to create something out of Micromamba that is supposed to be like Conda and is called Mamba. Because it will never be exactly like Conda or the old Mamba.

It is already really confusing to a lot of users, even very technically savvy ones. Any complexity we add to the selection of "Conda-like" tools will drive people away from Mamba/Micromamba, and probably also Conda.

AntoinePrv:
>  It would be nice to reduce the amount of confusion in the community

On this we completely agree.

I think our view is that it will not be completely the same, hence the major version bump. To us it seems making a single thing out of `mamba` and `micromamba` reduces the complexity in the ecosystem (mamba just becomes a single client). In my experience (in which I don't make the most advance use of mamba TBH), there is little to no discrepancy.

If I understand correctly, you're saying that the breaking changes between mamba 1.0 (`mamba.py`) and 2.0 (`micromamba` dynamically linked) is not worth it? But then what do we offer/recommend those user? Isn't deprecating and suggesting they move to another client a bigger change for them?

I think `mamba.py` never matched all `conda` features and its users were aware of it. While breaking changes are annoying, they are necessary to move forward.
We're committed to doing this move with the community. Are there some unaddressed incompatibilities you are worried of? Do you have other migration ideas you'd like to discuss (upper pin in `mambaforge`? Separate package names like `mamba2`? ...).

cc: @isuruf @chrisburr @wolfv @jezdez @jaimergp @JohanMabille @Hind-M @Klaim

jonashaag:
> I think mamba.py never matched all conda features and its users were aware of it

That's not my experience. Most people don't care about package managers and just want to get packages installed. I think a large part of Mamba's audience does not know that Mamba can have incompatibilities with Conda (until something doesn't work). This used to be worse with the different solvers, but there are still some annoying incompatibilities eg when it comes to proxies and authentication. 

Frankly I suggest that some of the decision makers spend a couple of weeks doing customer support here on GitHub to learn about the level of understanding that is in the community. Be aware that only the most capable/motivated people even open tickets typically, and the rest just gives up or finds a workaround somewhere else. 

My proposal:

- Mamba.py users should switch to Conda. The most pressing issue, the slow solver, has been fixed and with that the pain of using Conda has been reduced greatly. I think this will automatically happen anyways because there's less of an incentive to use Mamba now.
- Deprecate Mamba.py
- Keep Micromamba mostly as-is and focus fixing its bugs and rough edges. Improve integration with the ecosystem (eg PyCharm, VS Code, Jupyter, ...). Improve documentation.

I'm very skeptical that you/we can pull off a Conda-compatible Micromamba. Already now most Mamba bugs never get fixed. How are we going to cope with an even more difficult challenge?

AntoinePrv:
We can indeed recommend that users wanting an exact `conda` experience switch to `conda`, mentioning that the fast solver is now the default there.

JohanMabille:
I thought that statically linked executables were highly discouraged on conda-forge (or even against the conda-forge policy), for security reasons (a security fix in a dependency would required rebuilding the executable while it would benefit from the fix automagically if it was dynamically linked). But I might be wrong here (I mean, micromamba is already hosted on conda-forge), and I appreciate the fact that a pure statically linked executable would prevent an update of libmamba / libmambapy from breaking the package manager itself.

So if that's not against conda-forge policy, I tend to agree with @jaimergp here:
- rename micromamba into mamba
- build it as a pure static binary and distribute it through conda-forge by default 
- provide a new micromamba package so that `micromamba` is an alias for mamba

@AntoinePrv @Klaim @Hind-M @jonashaag @wolfv any thought on this before I break it all :smiling_imp: ?

Klaim:
> So if that's not against conda-forge policy, I tend to agree with @jaimergp here:
> 
> * rename micromamba into mamba
> * build it as a pure static binary and distribute it through conda-forge by default
> * provide a new micromamba package so that `micromamba` is an alias for mamba

What about `libmamba`, the library?

If the intention here is to publish a package with a `mamba` executable standalone with all it's dependencies static (except the runtime, which is the dependency that should never be statically embedded, except for special cases) - and also provide the shared library of `libmamba` in the same package, then I'm against it.

AntoinePrv:
> and also provide the shared (or static, or both) library of libmamba in the same package, then I'm against it.

That would be dynamic and in a separate `libamamba` (`libmambapy`) package.

Klaim:
So packages would be

- `mamba` -> only the exe, static dependencies (how are they provided?)
- `micromamba` -> alias to `mamba` with alials to the exe , for compatibility
- `libmamba(py)` -> shared library (and therefore we dont provide the static versions for `mamba` through this?)

am I correct?

JohanMabille:
@Klaim almost:

- libmamba (dynamic library)
- libmamba-static (static library)
- libmambapy (dynamic library)
- mamba (executable statically linked, static dependencies comes form conda-forge (and vcpkg for missing dependencies on Windows, but we plan to fix that)
- micromamba, an "alias" package.

libmamba, limamba-static and libmambapy are built via the same recipe, but this produces 3 distinct packages.

Klaim:
Ok that seems fine, just to clarify:

- package mamba would depends on libmamba-static
- micromamba as alias would still mean the executable is named... mamba?

Hind-M:
I don't really see the point of having `micromamba` as an alias of `mamba` (if the latter is just a renaming of the old static `micromamba`). I don't think compatibility is a good enough reason to bother with all this. IMHO, we should only keep one (`mamba`?) and advertise that the other (`micromamba`) doesn't exist anymore. Otherwise, as a user, I would find it very confusing to have two different packages supposed to be the same but with different names just for historical reasons...

Klaim:
I agree with @Hind-M

jaimergp:
> I thought that statically linked executables were highly discouraged on conda-forge

The spirit is there, specially for libraries, but I think we do exceptions, as you mention:

> micromamba is already hosted on conda-forge

I think it would be good to discuss this in a community meeting. For example, we have one in 30 mins. If you have the time, feel free to check [the calendar](https://conda.org/community/calendar) and add the item to the [agenda](https://hackmd.io/621Ewr3dS4KCC_m89TexOQ?both). I can also bring it up on your behalf to the best of my ability.

Klaim:
I cannot join but I'm fine with the plan, assuming the 2 points I listed would be answered as "yes" and "let's avoid the alias package if possible".

jaimergp:
> Otherwise, as a user, I would find it very confusing to have two different packages supposed to be the same but with different names just for historical reasons...

The alias could inject a little deprecation warning to notify users of the change. This could also be done in a v1.x release so avoid the maintenance cost in the v2 branch, though. I am thinking of mostly CI usage where they might be blindly running micromamba.

But if micromamba is _never_ published for v2, then the problem is smaller. You might leave a few users behind because they never realized the "new micromamba" is actually "mamba" now 🤷 

But that's just my suggestion not an order, obviously :)

Klaim:
Maybe this could be resolved in reactive steps? step one, not publish micromamba v2 package, step 2, gather feedback as to if that is needed. If so, step 3 publish the alias package. That way also spread the effort where it's the most important through time.

JohanMabille:
> package mamba would depends on libmamba-static

Yes

> micromamba as alias would still mean the executable is named... mamba

The package would just install aliases so that typing or invoking micromamba forwards to mamba.

> I don't think compatibility is a good enough reason to bother with all this

There are GHA and scripts that rely on micromamba, other on mamba. Not providing an alias package means some of these GHA / scripts won't upgrade to mamba 2.0

Klaim:
Welp, that clarifies it, it's not a realistic option to avoid the alias package then haha

jonashaag:
I think if we make a release to the `mamba` package that's actually Micromamba under the hood, all hell will break loose. People need to be prepared for this at least a couple of months long with warnings to upper-pin their Mamba versions so as to not get an automatic "upgrade" to Micromamba

AntoinePrv:
One option could be to keep the name of the **package** as `micromamba` so that `mamba` never upgrades to micromamba.

JohanMabille:
That would advocate for keeping the micromamba command (having the mamba command in the micromamba package sounds a bit weird to me).

SylvainCorlay:
The initial idea for Version 2.0:

 - we continue producing the micromamba static binary and retain the micromamba name for the executable in this case,
 - we continue producing the libmamba shared object as it is a core dependency of conda,
 - we produce a mamba executable as part of the mamba package linking dynamically with libmamba.

This way:

 - all CI workflows and other set usage in which people invoke mamba or micromamba continue working, and so do the download scripts for micromamba installs.
 - mamba follows the policy of conda-forge of dynamicaly linking with its dependencies, and therefore it would use the same libmamba as conda if both are installed in the same environment.

SylvainCorlay:
Closing since this is not the way we are moving forward.

pavelzw:
what is the way now?

AntoinePrv:
Mamba 2.0 is a dynamic link of micromamba released in mini-forge. 
Micromamba is unchanged.

PR #3066, opened 2023-12-16
lock solv file before writing

jchorl:
I started running a bunch of micromamba in parallel and got weird errors. I added PIDs to the log messages and was able to correlate a few logs:

```
780 [16:21:49] info     libmamba Writing libsolv solv file "/home/ubuntu/micromamba/pkgs/cache/497deca9.solv" for repo conda-forge/linux-64
```

```
2491 [16:21:49] debug    libmamba Unlocking '/home/ubuntu/micromamba/pkgs/cache/497deca9.solv'
2491 [16:21:49] trace    libmamba Removing file '/home/ubuntu/micromamba/pkgs/cache/497deca9.solv.lock'
2491 [16:21:49] critical libmamba Unable to read repo solv file 'conda-forge/linux-64', error was: unexpected EOF, depth = 3 
```

I was able to reliably reproduce this issue, but only in a complex, parallelized build setup.

I poked through the libsolv write code to try to figure out why writes are not atomic, but it's too complex for me: https://github.com/openSUSE/libsolv/blob/94536ca4e74fa8d624998beb8a57a08559ea8eac/src/repo_write.c#L1223

Anyway, locking on libsolv write reliably fixed the issue for me.


jonashaag (approved):
Cool!

jchorl:
Can we merge this?

AntoinePrv (approved):
Thanks!

AntoinePrv:
Arf, still some handling to do when the solv file does not exist

jchorl:
> Arf, still some handling to do when the solv file does not exist

I see the test that failed. I tried to understand why it was locking a non-existent file - presumably because the solv failed. Using a binary from this branch, I tried to reproduce with no luck:
```
(mamba) root@48d0101d08de:/work# ./build/cmake/micromamba/mamba create -n env1 pkgs/main/noarch::python --dry-run -y -c conda-forge --no-rc       
conda-forge/linux-64                                        Using cache                                                                           
conda-forge/noarch                                          Using cache                                                                           
critical libmamba The package "pkgs/main[noarch]::python" is not found in any loaded channels. Try adding more channels or subdirs.
```

I suspect the error message about locking a non-existent file might be due to the env vars/home that the test sets. It's possible if the solv fails, the directory structure never gets created. I'm not sure.

Anyway, it doesn't seem to change behaviour. At worst, it could cause a spurious error message.

Do you have any pointers on how to chase this down?

I have updated the test such that it passes by changing the check from strict equals to contains.

jchorl:
Actually, while I don't think that log causes any harm, it seems to be more prevalent and spurious than I anticipated. I'll need to dive into what's causing it.

AntoinePrv:
@jchorl I'm undertaking a major refactoring of `Mrepo`  in #3118 so future change may not be able to be merged.
I'm curious t know if you find something.
I did notice there is no lock on writing the solv file 
https://github.com/mamba-org/mamba/blob/13d034147f44dad041b4c85e38c6848738c04651/libmamba/src/core/repo.cpp#L601

PR #3098, opened 2024-01-03
Add BuildNumberSpec

AntoinePrv:


libmamba/src/specs/build_number_spec.cpp
```diff
@@ -0,0 +1,331 @@
...
+            return std::holds_alternative<U>(lhs) && std::holds_alternative<U>(rhs)
+                   && equal(std::get<U>(lhs), std::get<U>(rhs));
+        }
+
+        template <typename... T>
+        auto equal(const std::variant<T...>& lhs, const std::variant<T...>& rhs) -> bool
+        {
+            return (equal_with<T>(lhs, rhs) || ...);
+        }
+    }
```
JohanMabille, line 91: I don't get the need for all this complexity. From what I understand (but I might be wrong), you want to ensure that the variant hold objects of the same type, which you can check with `lhs.index() == rhs.index()`

libmamba/src/specs/build_number_spec.cpp
```diff
@@ -0,0 +1,331 @@
...
+
+    BuildNumberPredicate::BuildNumberPredicate(BuildNumber ver, BinaryOperator op)
+        : m_build_number(std::move(ver))
+        , m_operator(std::move(op))
+    {
+    }
+
+    auto operator==(const BuildNumberPredicate& lhs, const BuildNumberPredicate& rhs) -> bool
+    {
+        return (lhs.m_build_number == rhs.m_build_number) && equal(lhs.m_operator, rhs.m_operator);
```
JohanMabille, line 150: Actually you could even perform the variant comparison here and get rid of the `  auto equal(const std::variant<T...>& lhs, const std::variant<T...>& rhs)` function.

AntoinePrv: Again, I cannot specialized `operator==(std::equal_to<int>, std::equal_to<int>)` without UB.

libmamba/src/specs/build_number_spec.cpp
JohanMabille: From the definitions I read here, it looks like objects are equal as soon as they have the same type. Did I miss something?

libmamba/src/specs/build_number_spec.cpp
JohanMabille: Is there a case where `operator==(std::equal_to<int>, std::equal_to<int>)` should return `false`? Otherwise, comparing the type is enough.

libmamba/src/specs/build_number_spec.cpp
AntoinePrv: Ha I see what you mean. The code written does a proper variant comparison, but from the types included here (none have members), it is equivalent to only checking for type equality.
This is not what I wrote because this does not describe the proper intent. It may lead to subtle bugs should one modify/add a predicate.

JohanMabille (approved):

PR #3102, opened 2024-01-04
windows installation scripts

yuxi-liu-wired:
EDIT: The `micromamba_2.ps1` script should have this added at the top!

```ps1
# Setting up the MAMBA_ROOT_PREFIX environment variable
$Env:MAMBA_ROOT_PREFIX="$HOME\micromambaenv"
```


AntoinePrv (changes requested):
I think we'd rather want to get rid of the whole "Manual installation section" and add the [Windows downloader script](https://github.com/mamba-org/micromamba-releases#windows) in the "Install script" section. What do you think?

AntoinePrv:
I'll close since there has been changes to the doc since. Feel free to reopen if needed.

PR #3105, opened 2024-01-05
Add getter and setters in PackageInfo

AntoinePrv:
Turns out we need some getter/setter to keep unlikely data.


JohanMabille (commented):
Why do we need getters and setters here?

PR #3117, opened 2024-01-09
Rename ChannelSpec > UndefinedChannel

AntoinePrv:
Feels a bit like running in circle...

I realised ``ChannelSpec`` anme could be confusing wrt ``MatchSpec``, when in fact it is intended
to describe "a channel string whose location has not yet be fully determined".

I prefer changing the name before v2 than sticking with a bad name.


JohanMabille (commented):
Maybe "UnresolvedChannel" would be better? Undefined sounds like "something went wrong"

JohanMabille (approved):

PR #3119, opened 2024-01-10
Warning around manual install and add ref to conda-libmamba

AntoinePrv:
The manual installation take a lot of visual space (and distract inexperienced users) while they provide little value.
I cannot imagine someone wanting a manual install that cannot simply go lookup the steps in `install.sh` and adapt as needed, as suggested in the new `:: note:`.

### DIFF
--- a/docs/source/developer_zone/changes-2.0.rst
+++ b/docs/source/developer_zone/changes-2.0.rst
@@ -33,6 +33,7 @@ Micromamba
 Micromamba recieves all new features and its CLI remains mostly unchanged.
 
 Breaking changes include:
+
 - ``micromamba shell init`` root prefix parameter ``--prefix`` (``-p``) was renamed
   ``--root-prefix`` (``-r``).
   Both options were supported in version ``1.5``.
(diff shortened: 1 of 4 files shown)


Hind-M:
> The manual installation take a lot of visual space (and distract inexperienced users) while they provide little value. I cannot imagine someone wanting a manual install that cannot simply go lookup the steps in `install.sh` and adapt as needed, as suggested in the new `:: note:`.

Hum, I kind of find it useful since the steps are clearer and more human readable (the user won't need to go analyze the install script to know what he needs to do), but I agree that I don't really know which kind of users would go for the manual installation.

JohanMabille:
I agree with @Hind-M , instructions in the doc have a narrative you don't find in the script and could be used by "advanced" users who don't want to read the script or would srtruggle to adapt it. However, we could move this section to the bottom of the documentation (in a "advanced user" section for instance), and just provide a link (with a warning) in the install section.

AntoinePrv:
In this case I'd say less is more, but it seems I'm the only with this point of view.

JohanMabille (approved):

PR #3135, opened 2024-01-17
Add expected caster to Union

AntoinePrv:
Cast `tl::expected<T, E>` to a union of either `T` or `E`.
This is lesser effort that could later be enriched with method of `and_then`, `value`...

Even if the type may be unconventional in Python, casting it has several advatanges over throwing
exceptions.
- Exception remains for expceptional cases and are not used for general control flow,
- The API is closer to `libmamba`'s,
- It dimishes the glue code required to bind functions (as well as possible erorrs).


JohanMabille (approved):

AntoinePrv:
Does not interfere with `micromamba` tests.

PR #3141, opened 2024-01-19
Solver Request

AntoinePrv:
A new libsolv independent interface for passing jobs to the Solver.

### DIFF
--- a/libmamba/include/mamba/api/install.hpp
+++ b/libmamba/include/mamba/api/install.hpp
@@ -41,10 +41,10 @@ namespace mamba
         PrefixData& prefix_data,
         std::vector<std::string> specs,
         bool freeze_installed
-    ) -> Request;
+    ) -> solver::Request;
 
     void add_pins_to_request(
-        Request& request,
+        solver::Request& request,
         const Context& ctx,
         PrefixData& prefix_data,
         std::vector<std::string> specs,
@@ -52,7 +52,7 @@ namespace mamba
         bool no_py_pin
     );
 
-    void print_request_pins_to(const Request& request, std::ostream& out);
+    void print_request_pins_to(const solver::Request& request, std::ostream& out);
 
     void install_explicit_specs(
         Context& ctx,
--- a/libmamba/include/mamba/api/install.hpp
+++ b/libmamba/include/mamba/api/install.hpp
@@ -35,8 +35,7 @@ namespace mamba
         const Configuration& config,
         const std::vector<std::string>& specs,
         bool create_env = false,
-        bool remove_prefix_on_failure = false,
-        int solver_flag = SOLVER_INSTALL
+        bool remove_prefix_on_failure = false
     );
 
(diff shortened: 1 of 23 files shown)


libmamba/include/mamba/api/install.hpp
```diff
@@ -35,11 +34,26 @@ namespace mamba
...
+        bool freeze_installed
+    ) -> solver::Request;
+
+    void add_pins_to_request(
+        solver::Request& request,
+        const Context& ctx,
+        PrefixData& prefix_data,
+        std::vector<std::string> specs,
+        bool no_pin,
+        bool no_py_pin
```
JohanMabille, line 52: - What does `no_py_pin` means, and why do we have a specific handling for it?
- Can we imagine that in the future, more specific pinning handling would be required? In that case, maybe passing a "pin" structure instead of a bunch of boolean values coulb be safer.
- no_pin => no_gin =====>[] :smile:

libmamba/src/api/install.cpp
```diff
@@ -396,18 +396,101 @@ namespace mamba
...
+                        }
+                        out << "  - " << item.spec.str() << '\n';
+                    }
+                },
+                req
+            );
+        }
+    }
+
+    void install_specs_impl(
```
JohanMabille, line 486: This function should be moved into an anonymous namespace

libmamba/include/mamba/core/solver.hpp
```diff
@@ -94,8 +92,6 @@ namespace mamba
 
         [[nodiscard]] const std::vector<specs::MatchSpec>& install_specs() const;
         [[nodiscard]] const std::vector<specs::MatchSpec>& remove_specs() const;
```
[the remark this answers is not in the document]
AntoinePrv: This was already here, planning to remove it in another PR.

libmamba/include/mamba/api/install.hpp
AntoinePrv: It's one of the CLI option ` --no-py-pin,--py-pin{false} Do not automatically pin Python`. Why? Conda.
Yes, we should definitly handle all that Python stuff in a more generic way, also true for `pip_as_python_dependency` things which should really be a `recommend` or something.

For now I'm just trying to change the solver.

libmamba/src/api/install.cpp
```diff
@@ -396,18 +396,101 @@ namespace mamba
...
+        {
+            bool first = true;
+            std::visit(
+                [&](const auto& item)
+                {
+                    if constexpr (std::is_same_v<std::decay_t<decltype(item)>, solver::Request::Pin>)
+                    {
+                        if (first)
+                        {
+                            out << "\nPinned packages:\n\n";
```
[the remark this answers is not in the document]
AntoinePrv: Going for the _status quo_ here since it was not the point of the PR.

JohanMabille (approved):

Issue #1820, opened 2022-07-25
`micromamba run` should pass signals to child


jonashaag:
I need to double check this but it seems like signals are not (always?) passed to the child. To reproduce, run `micromamba run bash` and press Ctrl+C, it will stop the child (but only on Windows?).

PR #3152, opened 2024-01-24
Forward signals to child process with micromamba run

MihaiBSony:
Hi! This MR tries to fix issue #1820, in which signals are not forwarded to the child process. I think this should count as a first draft - in the final version it's probably a good idea to save the state of the signal handlers and restore it once the call is over, instead of just overwriting them.  Looking forward to your feedback!

P.S.: I don't know much about Windows, so I didn't even try to handle that case.

### DIFF
--- a/libmamba/src/core/run.cpp
+++ b/libmamba/src/core/run.cpp
@@ -427,7 +427,7 @@ namespace mamba
                 );
             }
 #endif
-            PID pid;
+            static PID pid;
             std::error_code lec;
             static reproc::process proc;
 
@@ -445,19 +445,21 @@ namespace mamba
             MainExecutor::instance().schedule(
                 []()
                 {
-                    signal(
-                        SIGTERM,
-                        [](int /*signum*/)
+                    auto forward_signal = [](int signum)
+                    {
+                        kill(pid, signum);
+                    };
+
+                    for (int signum = 1; signum < NSIG; signum++)
+                    {
+                        if (signum != SIGCHLD)
                         {
-                            LOG_INFO << "Received SIGTERM on micromamba run - terminating process";
-                            reproc::stop_actions sa;
-                            sa.first = reproc::stop_action{ reproc::stop::terminate,
-                                                            std::chrono::milliseconds(3000) };
-                            sa.second = reproc::stop_action{ reproc::stop::kill,
-                                                             std::chrono::milliseconds(3000) };
-                            proc.stop(sa);
+                            struct sigaction sa;
+                            sa.sa_flags = 0;
+                            sa.sa_handler = forward_signal;
+                            sigaction(signum, &sa, nullptr);
                         }
-                    );
+                    }
                 }
             );
 #endif
@@ -468,9 +470,12 @@ namespace mamba
                 opt.redirect.parent = true;
             }
 
+
             ec = reproc::drain(proc, reproc::sink::null, reproc::sink::null);
 
-            std::tie(status, ec) = proc.stop(opt.stop);
+            do {
+                std::tie(status, ec) = proc.stop(opt.stop);
+            } while (ec.value() == 4); // we get error code 4 in case of a signal; Just call stop again
 
             if (ec)
             {


jonashaag:
Do have more understanding of why signal forwarding works on Linux and macOS but not Windows? Should this patch be Windows-only instead?

MihaiBSony:
My changes don't touch windows at all (almost all changes are inside a `#ifndef _WIN32` guard). Previously, the only signal that was somewhat working was SIGTERM, because micromamba was adding a signal handler for it, which in turn was stopping the child process with a SIGTERM. The PR attempts to forward all signals to the child, except for SIGCHLD

MihaiBSony:
The context where I encountered this as an issue is a setup involving SLURM and pytorch-lightning. There I needed the python interpreter to receive the SIGUSR1 signal, but, whenever I ran it as `micromamba run -n env python script.py`, sending SIGUSR1 to micromamba would just terminate the micromamba process and the child, since that is the default action for SIGUSR1 according to [the signal(7) manpage](https://man7.org/linux/man-pages/man7/signal.7.html).

While working on the issue I got some inspiration from the source code of the Singularity/Apptainer containerization tool, since `micromamba run` seems to roughly have a similar purpose as `singularity exec`. There they seem to handle signals in a similar way as this PR, forwarding them to the child process (if I understood the code correctly). See [here](https://github.com/apptainer/apptainer/blob/bbfc4487c003936c1575b7bed7f123b9717eeba0/internal/pkg/runtime/engine/oci/monitor_linux.go).

jonashaag:
Thanks a lot for the context!

We definitely need to add tests for this feature. I guess it could be as simple as using `micromamba run` to start a Python script that prints whatever signals it receives, and using that output in the test.

That could also help understanding the Windows situation :)

MihaiBSony:
Hi @jonashaag!

I made the code more robust and added unit tests for signal forwarding and process suspension (with CTRL-Z). Unfortunately the code is using a lot of Linux syscalls, so not even the unit tests are usable for windows. Could you take a look at it and maybe tag other people who have more knowledge of signal handling or linux processes? In particular, I'm not sure if the way I'm dealing with the previous signal handler and the signal mask is correct, and if I'm waiting correctly for the child process. Also, it might make sense to move some of the new code to separate functions to keep things tidy.

On the plus side, I already tested the binary in my setup and it works for my use case! 😀

libmamba/src/core/run.cpp
```diff
@@ -283,6 +285,39 @@ namespace mamba
...
+                // suspend entire process group as well
+                kill(0, SIGSTOP);
+            }
+            else
+            {
+                break;
+            }
+        }
+
+        // Compute the return status mimicking the private logic in reproc++
```
jonashaag, line 309: Can you add a link to that logic as a comment please

libmamba/src/core/run.cpp
```diff
@@ -442,24 +477,38 @@ namespace mamba
             }
 
 #ifndef _WIN32
-            MainExecutor::instance().schedule(
```
jonashaag, line 445: Why this change?

libmamba/src/core/run.cpp
```diff
@@ -442,24 +477,38 @@ namespace mamba
...
+
+            // Stop the normal signal handler so that SIGINT is forwarded as well,
+            // and unblock all signals, while saving the previous state
+            // TODO: Is this the right place to change the mask and is this the right order?
+            stop_receiver_thread();
+            pthread_sigmask(SIG_SETMASK, &empty_sigmask, &prev_sigmask);
+
+            std::map<int, struct sigaction> prev_actions;
+
+            // Signal handler lambda forwarding all signals except CHLD to the child
```
jonashaag, line 491:
```suggestion
            // Signal handler that forwards all signals except CHLD to the child
```

libmamba/src/core/run.cpp
```diff
@@ -470,12 +519,23 @@ namespace mamba
 
             ec = reproc::drain(proc, reproc::sink::null, reproc::sink::null);
 
-            std::tie(status, ec) = proc.stop(opt.stop);
+#ifndef _WIN32
+            status = wait_for_process(pid);
```
jonashaag, line 523: Between the place where we installed the signal handlers up until after this line, is it possible to get any exceptions? If so, should we put the cleanup/restoring code into a `finally`?

jonashaag (commented):
Great work <3

libmamba/src/core/run.cpp
MihaiBSony: I added a link to the main part of the logic, which is [here](https://github.com/DaanDeMeyer/reproc/blob/1c07bdbec3f2e/reproc/src/process.posix.c#L455). Do you want a second link for the `wpid < 1` part? That would be [here](https://github.com/DaanDeMeyer/reproc/blob/1c07bdbec3f2ecba7125b9499b9a8a77bf9aa8c7/reproc/src/process.posix.c#L471)

libmamba/src/core/run.cpp
MihaiBSony: I think I forgot this change from one of the experiments I did. I'll revert it.

Do you know why the signal handler was being set with the `MainExecutor` instead of directly?

libmamba/src/core/run.cpp
MihaiBSony: I did not check, but that sounds like a good idea. Unfortunately I'm not sure how to do that in C++

libmamba/src/core/run.cpp
```diff
@@ -281,6 +283,40 @@ namespace mamba
...
+                kill(0, SIGSTOP);
+            }
+            else
+            {
+                break;
+            }
+        }
+
+        // Compute the return status mimicing the private logic in reproc++
+        // https://github.com/DaanDeMeyer/reproc/blob/1c07bdbec3f2ecba7125b9499b9a8a77bf9aa8c7/reproc/src/process.posix.c#L455
```
jonashaag, line 309:
```suggestion
        // Compute the return status mimicing the private logic in reproc++:
        // https://github.com/DaanDeMeyer/reproc/blob/1c07bdbec3f2ecba7125b9499b9a8a77bf9aa8c7/reproc/src/process.posix.c#L471
        // https://github.com/DaanDeMeyer/reproc/blob/1c07bdbec3f2ecba7125b9499b9a8a77bf9aa8c7/reproc/src/process.posix.c#L455
```

jonashaag: Or

```suggestion
        // Compute the return status mimicing the private logic in reproc++
        // https://github.com/DaanDeMeyer/reproc/blob/1c07bdbec3f2ecba7125b9499b9a8a77bf9aa8c7/reproc/src/process.posix.c#L465-L478
```

libmamba/src/core/run.cpp
jonashaag: No 🤷

jonashaag:
What signal handler do we have installed in this loop?

```
// Set handlers for all signals and save the previous state.
```

Ie. what are the previous handlers? If this always the fixed list from `set_signal_handler`? If so, I guess we could simplify the code by making it less generic?

libmamba/src/core/run.cpp
MihaiBSony: I traced this back to 60d4b3a, but still don't know why the signal was being set in a separate thread.

With the new changes, it seems more difficult to do the modifications correctly in a separate thread. The MainExecutor doesn't seem to have functionality to check if a scheduled task has finished, meaning that it wouldn't be possible to check that modifying and restoring the signal masks happen before the end of the `run_in_environment` call without using some sort of synchronization primitives. At the same time, the signal mask is set on a per-thread basis, so the calls to `pthread_sigmask` would have to anyhow be done on the main thread. What do you think about leaving this as is?

MihaiBSony:
> What signal handler do we have installed in this loop?
> 
> ```
> // Set handlers for all signals and save the previous state.
> ```
> 
> Ie. what are the previous handlers? If this always the fixed list from `set_signal_handler`? If so, I guess we could simplify the code by making it less generic?

When used in our code, I think only the so-called default handler is set. If, on the other hand, `run_in_environment` is used from third-party code (which could be the case, since it's in libmamba?), then we can't make any assumptions about the signals.

Also, there's actually one corner case which I don't handle and which seems impossible to handle without reworking other parts of the code. If someone uses `set_signal_handler` to set a non-default handler, I can't restore it at the end of the `run_in_environment` call, and my code just sets the default handler instead.

The so-called signal handlers from `set_signal_handler` are actually just threads blocking on a `sigwait` call. We could save the function that is passed as an argument, but if it's third-party code and has any side-effects before getting to `sigwait`, the behaviour would be pretty hard to predict. This seems to rather be a design flaw in the `set_signal_handler` functionality, but maybe there's a reason why we don't just use the normal linux signal handlers instead?

jonashaag:
It seems that `set_signal_handler` was introduced in https://github.com/mamba-org/mamba/pull/1193. Can you please check @MihaiBSony's last comment and let us know if there is a way to simplify the signal handling here? Especially for the `run_in_environment` API. @adriendelsalle  @AntoinePrv

libmamba/src/core/run.cpp
jonashaag: @wolfv do you recall why:

AntoinePrv:
I don't know much about the signal handlers but I know they must be forwarded to python when used in `libmambapy`.

PR #3153, opened 2024-01-25
Refactor solver flags

AntoinePrv:


JohanMabille (approved):

AntoinePrv:
MacOS runners are under heavy load... merging without the last job since we already have MacOS coverage.

PR #3157, opened 2024-01-25
Implemented support for mirrors

JohanMabille:


libmamba/src/download/curl.cpp
```diff
@@ -28,6 +29,9 @@ namespace mamba
             const std::string& ssl_verify
         )
         {
+            std::ofstream out("trace.txt", std::ios_base::app);
+            out << "configuring CURL handle:"
+                << "\n\turl = " << url << std::endl;
```
AntoinePrv, line 34: What is this trace.txt?

libmamba/include/mamba/specs/channel.hpp
```diff
@@ -109,6 +109,7 @@ namespace mamba::specs
         auto clear_platforms() -> platform_list;
         void set_platforms(platform_list platforms);
 
+        [[nodiscard]] auto id() const -> const std::string&;
```
AntoinePrv, line 112: Can we a add a doc saying what this is (a cross URL id) and its limitations (dependent on `channel_alias`).

libmamba/src/specs/channel.cpp
```diff
@@ -45,6 +45,7 @@ namespace mamba::specs
     Channel::Channel(std::vector<CondaURL> mirror_urls, std::string display_name, platform_list platforms)
         : m_mirror_urls(std::move(mirror_urls))
         , m_display_name(std::move(display_name))
+        , m_id(m_display_name)
```
AntoinePrv, line 48: That's still not making it a proper ID, but its something.
```suggestion
        , m_id(util::rstrip(m_display_name, '/'))
```

libmamba/src/specs/channel.cpp
```diff
@@ -155,6 +156,11 @@ namespace mamba::specs
         m_platforms = std::move(platforms);
     }
 
+    auto Channel::id() const -> const std::string&
+    {
+        return m_id;
+    }
+
```
AntoinePrv, line 163: Should we consider
```cpp
auto operator==(Channel const& lhs, Channel const& rhs) -> bool {
return lhs.id() == rhs.id();
}
```

AntoinePrv (commented):
My first few comments to get started since this seems to still be a moving target.

libmamba/src/download/curl.cpp
JohanMabille: A leftover, I need to remove them, I just wanted to get all the tests green before removing it.

libmamba/src/specs/channel.cpp
JohanMabille: Ok let's revert this change for now, the current Channel's API is not really consistent with that comparison operator (especially because we can change attributes of an already created channel, without having its ID changed.).

JohanMabille:
@AntoinePrv I've addressed the comment but the last one, I think this one is ready to go.

SylvainCorlay:
Bravo ! This is getting close.

libmamba/tests/src/core/test_mirror.cpp
```diff
@@ -10,7 +10,7 @@
 
 namespace mamba
 {
-    TEST_SUITE("mirror")
+    /*TEST_SUITE("mirror")
```
Hind-M, line 13: Is there a reason the tests are commented here?

JohanMabille: Yes, I have started to partially move the mirror implementation to the download folder / namespace. Since I want to make incremental steps, I have not port everything yet, but I had to remove `core/mirror.cpp` from the CMakeLists.txt to avoid symbol duplication, and I had to comment this test until the missing functions are ported to `download`

Hind-M (approved):

libmamba/include/mamba/download/downloader.hpp
```diff
@@ -98,48 +30,49 @@ namespace mamba
...
-    DownloadResult download(
-        DownloadRequest request,
+    Result download(
+        Request request,
+        const mirror_map& mirrors,
         const Context& context,
-        DownloadOptions options = {},
-        DownloadMonitor* monitor = nullptr
+        Options options = {},
+        Monitor* monitor = nullptr
```
AntoinePrv, line 72: Is optional reference meant here?

libmamba/include/mamba/download/mirror.hpp
```diff
@@ -0,0 +1,127 @@
...
+
+#include "mamba/download/request.hpp"
+
+namespace mamba::download
+{
+    class MirrorID
+    {
+    public:
+
+        explicit MirrorID(std::string_view v);
```
AntoinePrv, line 26: Allow to move a string into the class.
```suggestion
        explicit MirrorID(std::string v);
```

libmamba/include/mamba/download/mirror.hpp
```diff
@@ -0,0 +1,127 @@
...
+#include <string>
+#include <vector>
+
+#include <fmt/core.h>
+
+#include "mamba/download/request.hpp"
+
+namespace mamba::download
+{
+    class MirrorID
```
AntoinePrv, line 22: Should this be a plain struct?

libmamba/include/mamba/download/mirror.hpp
```diff
@@ -0,0 +1,127 @@
...
+
+        std::string m_value;
+    };
+
+    struct MirrorRequest : RequestBase
+    {
+        using header_list = std::vector<std::string>;
+
+        std::string url;
+        std::optional<header_list> headers;
```
AntoinePrv, line 43: Could this be only `header_list` with empty means no headers?

libmamba/include/mamba/download/request.hpp
```diff
@@ -0,0 +1,179 @@
...
+
+        explicit string_view_alias(std::string_view s);
+        operator std::string_view() const;
+
+    private:
+
+        std::string_view m_wrapped;
+    };
+
+    using MirrorName = string_view_alias<0>;
```
AntoinePrv, line 137: Why not?
```suggestion
struct MirrorName
{
    std::string_view name;
}
```

libmamba/src/download/downloader_impl.hpp
```diff
@@ -105,9 +103,77 @@ namespace mamba
...
+
+        enum class State
+        {
+            WAITING_SEQUENCE_START,
+            PREPARING_DOWNLOAD,
+            RUNNING_DOWNLOAD,
+            LAST_REQUEST_FINISHED,
+            LAST_REQUEST_FAILED,
+            SEQUENCE_FINISHED,
+            SEQUENCE_FAILED,
```
AntoinePrv, line 152: Nit: are we doing full caps for enum members?

libmamba/src/download/mirror.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+    MirrorID::MirrorID(std::string_view v)
+        : m_value(v)
+    {
+    }
+
+    std::string MirrorID::to_string() const
+    {
+        return fmt::format("MirrorID <{}>", m_value);
+    }
```
AntoinePrv, line 19: Should we write a custom `fmt::formatter` instead?

libmamba/src/download/mirror.cpp
```diff
@@ -0,0 +1,169 @@
+#include <fmt/format.h>
```
AntoinePrv, line 1:
```suggestion
// Copyright (c) 2023, QuantStack and Mamba Contributors
//
// Distributed under the terms of the BSD 3-Clause License.
//
// The full license is in the file LICENSE, distributed with this software.

#include <fmt/format.h>
```

libmamba/src/download/mirror_map.cpp
```diff
@@ -0,0 +1,51 @@
+#include "mamba/download/mirror_map.hpp"
```
AntoinePrv, line 1:
```suggestion
// Copyright (c) 2023, QuantStack and Mamba Contributors
//
// Distributed under the terms of the BSD 3-Clause License.
//
// The full license is in the file LICENSE, distributed with this software.

#include "mamba/download/mirror_map.hpp"
```

libmamba/tests/src/download/test_downloader.cpp
```diff
@@ -17,8 +17,9 @@ namespace mamba
         TEST_CASE("file_does_not_exist")
         {
 #ifdef __linux__
```
AntoinePrv, line 19: Does not work elsewhere?

AntoinePrv (approved):

libmamba/include/mamba/download/downloader.hpp
JohanMabille: Yes, but Monitor is a polymorphic type, and we cannot use `std::optional` with `references`.

libmamba/src/download/mirror.cpp
```diff
@@ -0,0 +1,169 @@
...
+     ************************************/
+
+    PassThroughMirror::PassThroughMirror()
+        : Mirror(PassThroughMirror::id())
+    {
+    }
+
+    MirrorID PassThroughMirror::id()
+    {
+        return MirrorID("");
```
Klaim, line 148: Consider making a named constant for `MirrorID("")` and using that instead, to clarify what this means.

Klaim (approved):
This needs more documentation and maybe some tweaks, but LGTM in general.

libmamba/include/mamba/download/mirror.hpp
JohanMabille: No, because the internal representation should never been accessed by client code.

libmamba/include/mamba/download/downloader.hpp
Klaim: Ah I just posted a review with a similar remark.
`Monitor` being polymorphic is not a problem as long as a reference or pointer is passed. However indeed `std::optional` does not support references (it's being discussed to enable it).
Question: arent we already using `tl::optional` which does support references?

libmamba/include/mamba/download/mirror.hpp
Klaim: Yes, we need this to not change interface in the future but stilll be optimizable.

libmamba/src/download/downloader_impl.hpp
Klaim: My understanding is that it's just for names of states, not specific to enums.

libmamba/include/mamba/download/request.hpp
JohanMabille: Because initially I also wanted to have aliases for the `url_path` and why not the`name` fields of the request, and I found this solution more elegant than repeating the same code for wrapping the same type. I had some problem with wrapping the url that I postponed to a future PR. Ideally, this structure should be moved to some util files, and accept an additional template argument (for the type it should alias). But I agree it may seem overkill actually.

A minor difference is that the structure forces a {} initialization while the alias lets the choice.

libmamba/src/download/downloader_impl.hpp
JohanMabille: It makes it consistent with the full caps used for constant (either in mamba or in its dependencies), and that's a convention I'm used too. Bu honestly I don't know if we have a convention for that in mamba, I just sticked with what was here before the PR.

libmamba/include/mamba/download/downloader.hpp
JohanMabille: Nope, we only use `tl::expected`, and I'm a bit reluctant to add a new dependency just for handling optional references in a few cases :s

Klaim: Ah yes I mixed these up. Ok then.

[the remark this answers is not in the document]
JohanMabille: Yes, but I am a bit reluctant to add another dependency just for this here.

libmamba/include/mamba/download/mirror.hpp
```diff
@@ -0,0 +1,127 @@
...
+    // for everything to ensure a smooth transition
+    class PassThroughMirror : public Mirror
+    {
+    public:
+
+        using request_gerenrator_list = Mirror::request_generator_list;
+
+        PassThroughMirror();
+
+        static MirrorID id();
```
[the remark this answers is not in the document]
JohanMabille: Each mirror is responsible for creating a `MirrorID` based on the arguments it is given (most of the time, it will be passed an URL only). In the case of this Mirror, passing an argument does not really make sense, since there should be a single PassThroughMirror only. I will rename it into `make_id`.

libmamba/src/download/downloader.cpp
```diff
@@ -515,50 +524,212 @@ namespace mamba
...
+     *     - set_state(true) => LAST_REQUEST_FINISHED
+     *     - set_state(false) => LAST_REQUEST_FAILED
+     *     - set_state(Error with wait_next_retry) => LAST_REQUEST_FAILED
+     *     - set_state(Error no wait_next_retry  ) => SEQUENCE_FAILED
+     * LAST_REQUEST_FINISHED:
+     *     - m_step == m_request_generators.size() ? => SEQUENCE_FINISHED
+     * LAST_REQUEST_FAILED:
+     *     - m_retries == p_mirror->max_retries ? => SEQUENCE_FAILED
+     */
+    MirrorAttempt::MirrorAttempt(Mirror* mirror)
```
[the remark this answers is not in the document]
JohanMabille: Mmh not sure about the assertion here, if the constructor takes it by reference, then it has a valid address and it is clear the pointer cannot be null.

libmamba/tests/src/download/test_downloader.cpp
JohanMabille: It looks like a leftover from a previous PR, let's test on other platforms.

PR #3165, opened 2024-01-31
Proper sorting of display actions

AntoinePrv:
Previous sort was sorting strings, and so was sorting according to the first symbol `+`, `-`...


JohanMabille:
Do we really need to expose `variant_cmp.hpp` as a public header?

AntoinePrv:
Not really, no. I'd say this is true for >50% of `util` headers.

But currently, I don't think we don't have a good way of both namespacing AND making headers private.
This header gets reused in a follow up PR so it cannot be hidden down a compilation unit.

JohanMabille:
I think namespacing and making heades private are orthogonal. If headers are not inclduded in header under the `include` folder (and if we don't want to provide them to uses), they can be moved to the `src` folder, whatever the namespace they use, and that would be enough to make them private headers (actually almost, we also need to hide symbols).

PR #3167, opened 2024-01-31
Improve Query API

AntoinePrv:
- Clean Query
- Bind Query alike to C++


AntoinePrv:
Not much to see here, just cleaning some things.

Issue #2994, opened 2023-11-20
repoquery not working anymore


h-vetinari:
### Troubleshooting docs

- [X] My problem is not solved in the Troubleshooting docs

### How did you install Mamba?

Mambaforge

### Search tried in issue tracker

repoquery

### Latest version of Mamba

- [X] My problem is not solved with the latest version

### Tried in Conda?

Not applicable

### Describe your issue

`mamba repoquery search` is somehow not functional anymore. It prints the help string, but any non-trivial use reports that the command isn't found:
```
>mamba repoquery search -h
usage: mamba repoquery search [-p PLATFORM] [--no-installed] [--pretty] [-a] [-h] [-c CHANNEL] [--use-local]
                              [--override-channels] [--repodata-fn REPODATA_FNS] [--experimental {jlap,lock}]
                              [--no-lock] [-C] [-k] [--offline] [--json] [-v] [-q]
                              package_query
[...]
>mamba repoquery search boa=0.16
Currently, only install, create, list, search, run, info, clean, remove, update, repoquery, activate and deactivate are supported through mamba.
```

Regular mamba operations (update / install / create) work fine.

### mamba info / micromamba info

```shell
>mamba info

          mamba version : 1.5.3
     active environment : None
            shell level : 0
       user config file : C:\Users\[...]\.condarc
 populated config files : C:\ProgramData\miniforge\.condarc
                          C:\Users\[...]\.condarc
          conda version : 23.10.0
    conda-build version : not installed
         python version : 3.11.6.final.0
       virtual packages : __archspec=1=x86_64
                          __win=0=0
       base environment : C:\ProgramData\miniforge  (read only)
      conda av data dir : C:\ProgramData\miniforge\etc\conda
  conda av metadata url : None
           channel URLs : https://conda.anaconda.org/conda-forge/win-64
                          https://conda.anaconda.org/conda-forge/noarch
          package cache : E:\miniforge\pkgs
                          C:\ProgramData\miniforge\pkgs
       envs directories : E:\miniforge\envs
                          C:\ProgramData\miniforge\envs
                          C:\Users\[...]\.conda\envs
                          C:\Users\[...]\AppData\Local\conda\conda\envs
               platform : win-64
             user-agent : conda/23.10.0 requests/2.31.0 CPython/3.11.6 Windows/10 Windows/10.0.22621 solver/libmamba conda-libmamba-solver/23.11.1 libmambapy/1.5.3
          administrator : False
             netrc file : None
[...]

PR #3170, opened 2024-02-02
1.x: Fix repoquery subcommand for Python 3.11/3.12 

mbargull:
We had a small typo in gh-3036 which caused the `args.func == ".repoquery"` case to not be caught.
For Python versions prior to 3.11 the `args._plugin_subcommand == ".main_repoquery"` case has been handled correctly which is why we didn't see problems with `mamba repoquery` for those versions.

Our tests did not catch this since `mamba` returned a zero exit code in case of not recognized commands.
This PR changes it to return exit code `1` in these cases now.
(N.B.: Cases like `mamba completely-unknown-command` are handled with non-zero exit code by `argparse`.)

Fixes gh-2994 .
Fixes gh-3033 .

### DIFF
--- a/mamba/mamba/mamba.py
+++ b/mamba/mamba/mamba.py
@@ -769,7 +769,7 @@ def do_call(args, parser):
         exit_code = update(args, parser)
     elif relative_mod == ".main_init":
         exit_code = shell_init(args)
-    elif relative_mod in (".main_repoquery", "repoquery"):
+    elif relative_mod in (".main_repoquery", ".repoquery"):
         exit_code = repoquery(args, parser)
     else:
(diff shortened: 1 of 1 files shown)


JohanMabille:
Thanks!

jezdez:
Good catch @mbargull!

mamba/mamba/mamba.py
```diff
@@ -778,7 +778,7 @@ def format_param(nm, val):
             " deactivate are supported through mamba."
         )
 
-        return 0
+        return 1
```
jezdez, line 781: Wow, yeah, that's better.

PR #3171, opened 2024-02-02
Clean libsolv use in Transaction

AntoinePrv:
Based on https://github.com/mamba-org/mamba/pull/3168

- Remove direct usage of libsolv in transaction;
- Remove varrious captures.


AntoinePrv:
Not much to see here, just moving things and removing captures.

PR #3178, opened 2024-02-05
Added HTTP Mirrors

JohanMabille:


libmamba/include/mamba/specs/package_info.hpp
```diff
@@ -45,6 +52,7 @@ namespace mamba::specs
         NoArchType noarch = NoArchType::No;
         std::size_t size = 0;
         std::size_t timestamp = 0;
+        PackageType package_type = PackageType::Unknown;
```
AntoinePrv, line 55:
```suggestion
        // FIXME this is a temporary hack to accommodate Python wheels but wheels and conda PackageInfo,
        // should really be split in different types. 
        PackageType package_type = PackageType::Unknown;
```

libmamba/include/mamba/api/channel_loader.hpp
```diff
@@ -19,11 +19,24 @@ namespace mamba
...
+    // repodata.json files if they are not
+    // cached locally.
     auto load_channels(  //
         Context& ctx,
         ChannelContext& channel_context,
         solver::libsolv::Database& pool,
         MultiPackageCache& package_caches
     ) -> expected_t<void, mamba_aggregated_error>;
+
+    // Creates channels and mirrors objects,
```
AntoinePrv, line 33: Creates them where?

libmamba/include/mamba/api/channel_loader.hpp
```diff
@@ -19,11 +19,24 @@ namespace mamba
     class ChannelContext;
     class MultiPackageCache;
 
+    // Creates channels and mirrors objects
+    // and loads channels, i.e. download
+    // repodata.json files if they are not
+    // cached locally.
```
AntoinePrv, line 25: use Doxygen
`
```cpp
/**
 * Brief.
 *
 * Details.
 */
  ```
  For documentation.

File: libmamba/include/mamba/solver/libsolv/database.hpp [hunk 4]
```diff
@@ -71,6 +71,7 @@ namespace mamba::solver::libsolv
         auto add_repo_from_repodata_json(
             const fs::u8path& path,
             std::string_view url,
+            const std::string& channel_id,
```
AntoinePrv, line 74: Shouldn't we rather take a `specs::Channel`?

libmamba/include/mamba/specs/archive.hpp
```diff
@@ -15,10 +15,9 @@
 
 namespace mamba::specs
 {
-    inline static constexpr auto ARCHIVE_EXTENSIONS = std::array{
-        std::string_view(".tar.bz2"),
-        std::string_view(".conda"),
-    };
+    inline static constexpr auto ARCHIVE_EXTENSIONS = std::array{ std::string_view(".tar.bz2"),
```
AntoinePrv, line 18: Should we split that in two? Or even just hard code it where we need it?
I fear adding pypi code in Conda abstactions will just make it harder on the long run.

libmamba/src/api/channel_loader.cpp
```diff
@@ -286,4 +285,44 @@ namespace mamba
...
+            {
+                for (auto channel : channel_context.make_channel(location))
+                {
+                    create_mirrors(channel, context.mirrors);
+                }
+            }
+        }
+    }
+
+    void init_channels_from_specs(
```
AntoinePrv, line 312:
```suggestion
    void init_channels_from_package_urls(
```

libmamba/src/core/package_fetcher.cpp
```diff
@@ -317,9 +322,33 @@ namespace mamba
...
+        {
+            return "";
+        }
+    }
+
+    std::string PackageFetcher::url_path() const
+    {
+        if (m_package_info.package_type == specs::PackageType::Conda)
+        {
+            return m_package_info.subdir + "/" + m_package_info.filename;
```
AntoinePrv, line 341:
```suggestion
            return util::concat(m_package_info.subdir, '/', m_package_info.filename);
```

libmamba/tests/src/core/test_env_lockfile.cpp
```diff
@@ -4,6 +4,8 @@
 //
 // The full license is in the file LICENSE, distributed with this software.
 
+#include <iostream>
+
```
AntoinePrv, line 8:
```suggestion
```

micromamba/tests/channel_a/osx-64/repodata.json
AntoinePrv, line 1: I this added on purpose?

JohanMabille: Not at all, and this might explain the failure on OSX actually

AntoinePrv: There are many other data files commited.

libmamba/include/mamba/solver/libsolv/database.hpp
JohanMabille: We could but that would require to store the channel in the SubdirData (which is where we get the `channel_id`), which means either duplicate it from the `ChannelContext`, or ensure the `ChannelContext` will live longer than the subdirs.
Besides, I'm not sure we need the entire channel information here, the `channel_id` is used to retrieve the mirrors stored in the Context; or we could refactor that to have the mirrors stored in the subdir directly (but again, this refactoring is beyond the scope of this PR).

AntoinePrv: The channel have value semantics so it does not need to be stored anywhere.
We may not need all the channel information but that make the API more clear as to what to expected and where to get it.

JohanMabille: > The channel have value semantics so it does not need to be stored anywhere.

I think these are unrelated. Unless you want to resolve channels every time you need them, you have to store the channels somewhere once you have created them. And if f we want to avoid passing the `Context` and the `ChannelContext` everywhere, we should definitely create the channels once and for good at the "beginning".

AntoinePrv: Aren't channels already stored in the channel context for caching purposes? Plus they are not that expensive to resolve.
One of the point of the Channel refactor was to make them first class citizen, I feel this is going backwards, putting strings everywhere.

One difficulty though is that this is not even the URL to a channel but a subdir, for which we do not have information...

JohanMabille: > Aren't channels already stored in the channel context for caching purposes? Plus they are not that expensive to resolve.

The issue is not they are expensive to resolve, but rather that you need the ChannelContext to resolve them (and then you need to pass it everywhere).

> One of the point of the Channel refactor was to make them first class citizen, I feel this is going backwards, putting strings everywhere.

The string is only for the mapping channel <=> mirrors. We could actually change that for a "strong" typedef (ChannelId), but we definitely don't need the entire channel object here; and actually, a subdir is a "sub" channel, for a given platform,. If we store a channel in the subdir, we then may have a consistency issue with those in the ChannelContext; avoiding this consistency means either storing a reference (but then we need to ensure the values live longer than the references), or removing all the setters from the channels (which actually would be my preference). 

So it's not going backward, but rather an intermediate state before we refactor the SubdirData and the cache in the ChannelContext.

SylvainCorlay:
273a7ba520f3cc9239b13bc6cee72841

PR #3179, opened 2024-02-06
Automate releases (`CHANGELOG.md` updating)

Hind-M:
This adds a script to automate updating `CHANGELOG.md` for releases.
This needs all the upcoming merged PRs to have exactly one of the following labels:
`release::enhancements`, `release::bug_fixes`, `release::ci_docs`

@JohanMabille @AntoinePrv @jonashaag @Klaim please make sure it's the case when merging in the future.

PS1: Feel free to tag here anyone who can merge in this repo to be notified. Thanks!
PS2: If anyone is aware of a way to make these labels mandatory in order to prevent from merging if missing, that would be great!

### DIFF
--- a/.github/workflows/label_check.yml
+++ b/.github/workflows/label_check.yml
@@ -1,0 +1,43 @@
+name: Check release label
+
+on:
+  pull_request:
+    types:
+      - synchronize
+      - opened
+      - reopened
+      - labeled
+      - unlabeled
+
+jobs:
+  label_check:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Check labels
+        run: |
+          NUMBER_OF_LABELS=$(jq '.pull_request.labels | length' "$GITHUB_EVENT_PATH")
+          if [ $NUMBER_OF_LABELS -eq 0 ]; then
+            echo "PR has no labels. Please add at least one label of release type."
+            exit 1
+          fi
+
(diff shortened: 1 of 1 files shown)


JohanMabille:
I think we can add a CI job to require such a label (I know JupyterLab has it)

AntoinePrv:
Can we get a `chore` for cleaning that change nothing for the user?

Alternatively (but this may be more work), we could look into [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/). There are tools to generate automated changelogs, as well as pre-commits [hooks](https://github.com/compilerla/conventional-pre-commit) for enforcing we use them.

Hind-M:
> Alternatively (but this may be more work), we could look into [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/). There are tools to generate automated changelogs, as well as pre-commits [hooks](https://github.com/compilerla/conventional-pre-commit) for enforcing we use them.

Yes indeed, it's another way of doing things. I actually first thought of using a tool to generate the changelog but figured that we may want to keep the current format of the changelog file (with contributor name etc), but maybe not? The script here is really some solution to avoid all the manual cumbersome work with the release. This could be short term of course and we may want to think about the whole process and conventional commits more thoroughly considering what can be involved and improved while we're at it?

> Can we get a `chore` for cleaning that change nothing for the user?

I'm not sure I get what you mean. Is this related to the conventional commits?

JohanMabille (approved):

PR #3184, opened 2024-02-08
Simple logging fix

### LINKED ISSUE #3112: Trace and debug log levels don't work
### Troubleshooting docs

### Anaconda default channels

log-level trace debug

### Latest version of Mamba

### Tried in Conda?

### Describe your issue

Providing `--log-level trace` to micromamba doesn't seem to get any trace or debug logs. I tried changing [this line](https://github.com/mamba-org/mamba/blob/main/libmamba/src/core/subdirdata.cpp#L581) to a `LOG_DEBUG` and `LOG_TRACE` just to make sure and the result is that nothing is logged.

_No response_

### Logs

_No response_

### environment.yml

_No response_

### ~/.condarc

_No response_

AntoinePrv:
Fix logging when libmamba is used as a lib without context.
The proper way would be to set logging sinks that get configured in micromamba or any other client code.

Close #2713
Close #3112

### DIFF
--- a/libmamba/include/mamba/core/output.hpp
+++ b/libmamba/include/mamba/core/output.hpp
@@ -124,6 +124,7 @@ namespace mamba
         Console& operator=(Console&&) = delete;
 
         static Console& instance();
+        static bool has_instance();
         static ConsoleStream stream();
         static bool prompt(std::string_view message, char fallback = '_');
         static bool prompt(std::string_view message, char fallback, std::istream& input_stream);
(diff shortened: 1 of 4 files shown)


libmamba/include/mamba/core/output.hpp
```diff
@@ -124,6 +124,7 @@ namespace mamba
         Console& operator=(Console&&) = delete;
 
         static Console& instance();
+        static bool has_instance();
```
Klaim, line 127: I would prefer `is_available` as it would match better the functionality and would be less "I know it's a singleton..."

Klaim (approved):

PR #3186, opened 2024-02-09
Add libsolv.Database Bindings and tests

AntoinePrv:


libmamba/include/mamba/fs/filesystem.hpp
```diff
@@ -303,6 +306,12 @@ namespace mamba::fs
             return to_utf8(m_path);
         }
 
+        // Returns a default encoded string.
+        const std::string& native() const
```
AntoinePrv, line 310: @Klaim I needed to add this function for pybind11, as it calls it and does utf-8 translation itself.
Is that OK?

libmambapy/src/libmambapy/bindings/expected_caster.hpp
```diff
@@ -20,73 +18,40 @@ namespace PYBIND11_NAMESPACE
...
             {
-                return caster_type::cast(expected_to_variant(std::forward<Expected>(src)), policy, parent);
+                if (src)
+                {
+                    return make_caster<T>::cast(std::forward<Expected>(src).value(), policy, parent);
+                }
+                else
+                {
+                    throw std::forward<Expected>(src).error();
+                }
```
AntoinePrv, line 51: tl::expected now throws on unexpected cases.

libmamba/include/mamba/fs/filesystem.hpp
Klaim: As long as it does what the standard does, that is returning the real native encoding of that the path use on that system, yes it's OK

I believe we didnt set this function to discourage trying to use something else than utf-8 with our library, but it's still just reflecting what the sandard does otherwise.

libmamba/include/mamba/solver/problems_graph.hpp
```diff
@@ -62,9 +62,22 @@ namespace mamba::solver
...
     private:
 
+        auto base() const -> const Base&;
+
         auto remove_asym(const key_type& a, const key_type& b) -> bool;
+
+        template <typename TT>
+        friend auto operator==(const conflict_map<TT>& lhs, const conflict_map<TT>& rhs) -> bool;
+        template <typename TT>
+        friend auto operator!=(const conflict_map<TT>& lhs, const conflict_map<TT>& rhs) -> bool;
```
JohanMabille, line 72: Nitpicking: Only `operator==` has to be friend

JohanMabille (approved):

PR #3191, opened 2024-02-16
1.x: Run tests for conda=24.1

mbargull:
Just seeing whether `mamba=1`'s test suite passes with `conda=24.1`.

If it does, could we tag the current `1.x` with `1.5.7` and rebuild on https://github.com/conda-forge/mamba-feedstock with the test and dependency update from https://github.com/conda-forge/mamba-feedstock/pull/208 (excluding its last commit, of course)?

### DIFF
--- a/mamba/environment-dev.yml
+++ b/mamba/environment-dev.yml
@@ -25,7 +25,7 @@ dependencies:
   - pytest >=7.3
   - pyyaml
   - boltons
-  - conda >=23.11.0
+  - conda >=24.1.0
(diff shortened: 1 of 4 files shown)


jaimergp:
Seems to be working on the mamba side. micromamba is not for some reason?

```
AttributeError: 'CallSpec2' object has no attribute 'funcargs'
```

Doesn't seem to be related to mamba itself.

AntoinePrv (approved):

PR #3192, opened 2024-02-19
[mamba content trust] Enable verifying packages signatures

Hind-M:
Based on initial work in #3050 
Making this ready for review and resuming work on another PR (tests and additional tweaks) to avoid frequent rebasing.

### DIFF
--- a/libmamba/include/mamba/solver/libsolv/database.hpp
+++ b/libmamba/include/mamba/solver/libsolv/database.hpp
@@ -74,7 +74,7 @@ namespace mamba::solver::libsolv
             const std::string& channel_id,
             PipAsPythonDependency add = PipAsPythonDependency::No,
             UseOnlyTarBz2 only_tar = UseOnlyTarBz2::No,
-            bool verify_packages = false,
+            VerifyPackages verify_packages = VerifyPackages::No,
             RepodataParser parser = RepodataParser::Mamba
         ) -> expected_t<RepoInfo>;
 
--- a/libmamba/include/mamba/solver/libsolv/database.hpp
+++ b/libmamba/include/mamba/solver/libsolv/database.hpp
@@ -74,6 +74,7 @@ namespace mamba::solver::libsolv
             const std::string& channel_id,
             PipAsPythonDependency add = PipAsPythonDependency::No,
             UseOnlyTarBz2 only_tar = UseOnlyTarBz2::No,
+            bool verify_packages = false,
             RepodataParser parser = RepodataParser::Mamba
         ) -> expected_t<RepoInfo>;
 
--- a/libmamba/src/solver/libsolv/helpers.cpp
+++ b/libmamba/src/solver/libsolv/helpers.cpp
@@ -37,7 +37,7 @@ namespace mamba::solver::libsolv
 {
     // Beyond this value, the timestamp would be in milliseconds and therefore should be converted
     // to seconds.
-    const auto MAX_CONDA_TIMESTAMP = 253402300799ULL;
+    inline constexpr auto MAX_CONDA_TIMESTAMP = 253402300799ULL;
 
     void set_solvable(solv::ObjPool& pool, solv::ObjSolvableView solv, const specs::PackageInfo& pkg)
     {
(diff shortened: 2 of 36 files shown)


libmamba/src/solver/libsolv/helpers.cpp
```diff
@@ -35,6 +35,10 @@
 
 namespace mamba::solver::libsolv
 {
+    // Beyond this value, the timestamp would be in milliseconds and therefore should be converted
+    // to seconds.
+    const auto MAX_CONDA_TIMESTAMP = 253402300799ULL;
```
AntoinePrv, line 40:
```suggestion
    inline constexpr auto MAX_CONDA_TIMESTAMP = 253402300799ULL;
```

libmamba/src/solver/libsolv/helpers.cpp
```diff
@@ -307,6 +324,32 @@ namespace mamba::solver::libsolv
...
+                    LOG_INFO << "Signatures for '" << filename
+                             << "' are set in corresponding solvable.";
+                }
+            }
+            else
+            {
+                LOG_DEBUG << "No signatures available for '" << filename
+                          << "'. Downloading without verifying artifacts.";
+            }
+
```
AntoinePrv, line 352: Should we isolate this in a separate function?

libmamba/src/solver/libsolv/helpers.cpp
```diff
@@ -140,6 +147,15 @@ namespace mamba::solver::libsolv
...
 
+        void
+        get_fake_signatures(simdjson::dom::parser& fake_parser, simdjson::dom::object& fake_signatures)
+        {
+            // Getting a fake simdjson::dom::object to simulate empty signatures
+            // A valid empty simdjson::dom::object is not handled in simdjson
+            const auto fake_signatures_json = R"(  { "fake_signatures": "val" }  )"_padded;
+            fake_signatures = fake_parser.parse(fake_signatures_json).get_object().value();
+        }
+
```
AntoinePrv, line 158: Why fake? Could empty be a better alternative instead instead?

libmamba/src/solver/libsolv/helpers.cpp
```diff
@@ -347,14 +392,70 @@ namespace mamba::solver::libsolv
...
+                get_fake_signatures(fake_parser, fake_signatures);
+                set_repo_solvables(
+                    pool,
+                    repo,
+                    parsed_url,
+                    channel_id,
+                    default_subdir,
+                    packages,
+                    fake_signatures
+                );
```
AntoinePrv, line 436: What if we use a check (`if` or `constexpr if`) around the setting of the signature in `set_solvable`, would that simplify this fake parser?

libmamba/include/mamba/solver/libsolv/database.hpp
```diff
@@ -74,6 +74,7 @@ namespace mamba::solver::libsolv
             const std::string& channel_id,
             PipAsPythonDependency add = PipAsPythonDependency::No,
             UseOnlyTarBz2 only_tar = UseOnlyTarBz2::No,
+            bool verify_packages = false,
```
AntoinePrv, line 77: Let's make this a strongly typed boolean to be consistent with the other two parameters.

libmamba/src/solver/libsolv/helpers.cpp
Hind-M: Hum because it's not really an empty json?

libmamba/src/solver/libsolv/helpers.cpp
Hind-M: Well we are actually checking if the signatures are empty around that indeed by getting the object in [here](https://github.com/mamba-org/mamba/pull/3192/files#diff-e2b0f07cf93b08e4834c9512510e360569da2cddcd70c333eea783269510c173R169). And there is no other way (at least that I could find) to check that with simdjson objects unless we use this fake signatures object workaround.

AntoinePrv: What if passing down a `std::optional<simdjson::dom::object>` as signature to `set_repo_solvables`, would that avoid creating the fake parser?

AntoinePrv:
Could we also add a test related to signature loading in `test_database.cpp`? There is a small test repodata being used under `libmamba/tests/data/repodata/conda-forge-numpy-linux-64.json`, that could modified, or another sample could be created for that purpose.

libmamba/src/solver/libsolv/helpers.cpp
Hind-M: Good point!

AntoinePrv (approved):

SylvainCorlay:
Congrats on landing this @Hind-M

PR #3210, opened 2024-03-04
feat: allow underscores anywhere in token

wolfv:
The token format on anaconda is `wo-1234-asdf-123123` (UUID4 style). The first two letters are derived from the user name (e.g. wolfv in my case).

On prefix.dev, we use something like `pfx_1234asdf123123...`. We are now changing this to use `-` as well (to maximize compatibility). I think `_` is still better because it allows tokens to be easier selected e.g. in a browser / text-editor.

LMK what you think! :)

### DIFF
--- a/libmamba/src/core/util.cpp
+++ b/libmamba/src/core/util.cpp
@@ -1603,7 +1603,7 @@ namespace mamba
     {
         // usernames on anaconda.org can have a underscore, which influences the
         // first two characters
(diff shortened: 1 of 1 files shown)


AntoinePrv:
We can add this but I don't think it's what you want. These regexs only get used in the logging to hide credentials.
What you want is in CondaURL (although I am not sure there are such limitations)

wolfv:
We actually changed it on the prefix.dev side to accomodate conda & mamba. So I think we can also close this one for now.

PR #3219, opened 2024-03-07
Make Taskfile.dist.yml Windows-compatible

carschandler:
PowerShell doesn't set `$env:PWD` automatically, but it can be easily retrieved by making `PWD` a `sh` variable in the Taskfile, which remains POSIX-compliant for other platforms.

If you don't want to modify the Taskfile, then at least a note about this should be made on https://mamba.readthedocs.io/en/latest/developer_zone/dev_environment.html#

### DIFF
--- a/Taskfile.dist.yml
+++ b/Taskfile.dist.yml
@@ -1,6 +1,8 @@
 version: '3'
 
 vars:
+  # Windows doesn't set the PWD environment variable automatically, but we can
+  # assign it using a dynamic variable since pwd does work in PowerShell
   PWD:
     sh: pwd
   BUILD_DIR: 'build'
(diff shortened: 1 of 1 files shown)


Taskfile.dist.yml
```diff
@@ -1,6 +1,8 @@
 version: '3'
 
 vars:
+  PWD:
+    sh: pwd
```
AntoinePrv, line 5: Is `pwd` a command on Windows? Can you add a comment why this change was added.

carschandler: `pwd` is an alias in Windows PowerShell for `Get-Location`. I've verified it works on Windows when using Task. And sure I can add a comment.

AntoinePrv (approved):

PR #3234, opened 2024-03-14
[mamba-content-trust] Add integration test

Hind-M:
- Add integration test for `mamba-content-trust` (mamba client side) using `conda-content-trust` (server side)
- Fix `trusted_channels` in context not working with more than one item
- Remove `repo_signed` (generated when running the script)


libmamba/include/mamba/core/repo_checker_store.hpp
```diff
@@ -32,11 +32,11 @@ namespace mamba
 
         explicit RepoCheckerStore(repo_checker_list checkers);
 
-        [[nodiscard]] auto find_checker(const Channel& chan) const -> const RepoChecker*;
```
JohanMabille, line 35: Why removing the const here? It looks legit. Same question for methods below.

JohanMabille (commented):
sorry for the late review!

libmamba/include/mamba/core/repo_checker_store.hpp
Hind-M: Yes it was legit until I realized that the [trusted_channels](https://github.com/mamba-org/mamba/pull/3234/files#diff-b13b39da2b425585073993cc5e342e9b973c3863efbbc0fd760790d74cd89831R47) in `Context` don't handle more that one item (giving errors if the first one is not the correct/used one). And that's because `generate_index_checker()` is fetching the root and delegation roles files and was called [here](https://github.com/mamba-org/mamba/pull/3234/files#diff-e1c0c3fb3bae3014dd66061fe03db2e7ad0fda578dcafe250786ebf9e44973c2L45) (when populating the `RepoCheckerStore` and before knowing if the corresponding channel will be used) but is now moved to [here](https://github.com/mamba-org/mamba/pull/3234/files#diff-58ac26b46f461c3b18d2f04c7fca096ee04a8e5b00c1f6e00a0c20c023ab4d42R595). And since it's not `const` the `RepoChecker` and these methods can't be anymore.

JohanMabille: :smiling_face_with_tear:

JohanMabille (approved):

Issue #3174, opened 2024-02-03
micromamba env update --prune not working properly


brano543:
### Troubleshooting docs

- [X] My problem is not solved in the Troubleshooting docs

### Anaconda default channels

- [X] I do NOT use the Anaconda default channels (pkgs/* etc.)

### How did you install Mamba?

Micromamba

### Search tried in issue tracker

micromamba env update

### Latest version of Mamba

- [X] My problem is not solved with the latest version

### Tried in Conda?

I didn't try

### Describe your issue

Hello guys,

I have tried the latest release https://github.com/mamba-org/mamba/releases/tag/2023.12.14 to see if command
micromamba env update -f ./conda/linux_env.yaml --prune
works correctly.

Unfortunately after --prune flag is added, command sometimes fails.

**.envrc file**
```
export PATH=$HOME/micromamba:$PATH
export MAMBARC=$(realpath ./.mambarc)
export MAMBA_ROOT_PREFIX="$HOME/micromamba"
eval "$(micromamba shell hook -s posix)"
```

**Steps to reproduce**
1.1) Initial environment looks like
```
[environments.txt](https://github.com/mamba-org/mamba/files/14154424/environments.txt)
```
1.2) Execute **micromamba env create -f ./conda/linux_env.yaml**
Note: Even if I create .mamba/pkgs at this point, at later stage I will still reproduce the issue
```
conda-forge/noarch                                  13.4MB @  47.7MB/s  0.3s
conda-forge/linux-64                                32.2MB @  36.2MB/s  1.0s

error    libmamba Could not lock non-existing path '/home/pandora/.mamba/pkgs'

Transaction

  Prefix: /home/pandora/micromamba/envs/linux

  Updating specs:

   - rattler-build
   - valgrind


  Package             Version  Build        Channel          Size
───────────────────────────────────────────────────────────────────
  Install:
───────────────────────────────────────────────────────────────────

  + _libgcc_mutex         0.1  conda_forge  conda-forge       3kB
  + _openmp_mutex         4.5  2_gnu        conda-forge      24kB
  + ca-certificates  2024.2.2  hbcca054_0   conda-forge     155kB
  + libgcc-ng          13.2.0  h807b86a_5   conda-forge     771kB
  + libgomp            13.2.0  h807b86a_5   conda-forge     420kB
  + libstdcxx-ng       13.2.0  h7e041cc_5   conda-forge       4MB
  + libzlib            1.2.13  hd590300_5   conda-forge      62kB
[...]

PR #3217, opened 2024-03-07
Improvment on how to build from source

ZHG2017:
Having put a bit more details on how to build from source either as a developer or as a user

PR #3243, opened 2024-03-22
A quick fix for the issue of update env with --prune flag

ZHG2017:
This is a quickfix for the issue #3174 

Added install request construction for the update process

Added condition test to avoid conflict for removing and updating the same package at the same time

Added unit test to validate the fix

### DIFF
--- a/libmamba/src/api/update.cpp
+++ b/libmamba/src/api/update.cpp
@@ -55,13 +55,29 @@ namespace mamba
             else
             {
                 request.jobs.reserve(specs.size());
+
+                bool found;
+                auto hist_map = prefix_data.history().get_requested_specs_map();
                 if (remove_not_specified)
                 {
-                    auto hist_map = prefix_data.history().get_requested_specs_map();
                     for (auto& it : hist_map)
                     {
-                        if (std::find(specs.begin(), specs.end(), it.second.name().str())
-                            == specs.end())
+                        // Verify if any package is installed but not present in the target
+                        // environment file
+                        found = false;
+                        for (const auto& sp : specs)
+                        {
+                            if (mamba::specs::MatchSpec::parse(sp)->name().contains(
+                                    it.second.name().str()
+                                ))
+                            {
+                                found = true;
+                                break;
+                            }
+                        }
+                        // Remove any installed package but not present in the target environment
+                        // file
+                        if (!found)
                         {
                             request.jobs.emplace_back(Request::Remove{
                                 specs::MatchSpec::parse(it.second.name().str())
@@ -73,13 +89,41 @@ namespace mamba
                     }
                 }
 
+
                 for (const auto& raw_ms : specs)
                 {
-                    request.jobs.emplace_back(Request::Update{
-                        specs::MatchSpec::parse(raw_ms)
-                            .or_else([](specs::ParseError&& err) { throw std::move(err); })
-                            .value(),
-                    });
+                    found = false;
+                    // if (std::find(specs.begin(), specs.end(),
+                    // mamba::specs::MatchSpec::parse(raw_ms)->name().str())
+                    //     != specs.end()) continue;
+                    for (auto& it : hist_map)
+                    {
+                        // Verify if any package in the target environment file is already installed
+                        if (mamba::specs::MatchSpec::parse(raw_ms)->name().contains(
+                                it.second.name().str()
+                            ))
+                        {
+                            // Update any package in the target environment file that is already
+                            // installed
+                            request.jobs.emplace_back(Request::Update{
+                                specs::MatchSpec::parse(raw_ms)
+                                    .or_else([](specs::ParseError&& err) { throw std::move(err); })
+                                    .value(),
+                            });
+                            found = true;
+                            break;
+                        }
+                    }
+                    // Install any package in the target environment file that is not installed yet
+                    // iff flag --prune is present
+                    if (!found && remove_not_specified)
+                    {
+                        request.jobs.emplace_back(Request::Install{
+                            specs::MatchSpec::parse(raw_ms)
+                                .or_else([](specs::ParseError&& err) { throw std::move(err); })
+                                .value(),
+                        });
+                    }
                 }
             }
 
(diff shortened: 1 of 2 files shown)


Hind-M:
Thanks for the PR.
Can you only include in this one the relevant files for the bug fix? The documentation related files should be in another PR (I think reopening #3217 will do).
For the `Linters` failure in the CI, running `pre-commit run --all` in the mamba source folder locally before committing should fix it.

ZHG2017:
> Thanks for the PR. Can you only include in this one the relevant files for the bug fix? The documentation related files should be in another PR (I think reopening #3217 will do). For the `Linters` failure in the CI, running `pre-commit run --all` in the mamba source folder locally before committing should fix it (it will most likely modify the non-compliant files to follow the linting rules (defined [here](https://github.com/mamba-org/mamba/blob/main/.pre-commit-config.yaml)), so you should add them and commit them).

Hi @Hind-M , Thank you for the suggestion. Let me figure out how to use hte `linter` then. And I think I will create a new branch as well as a new PR for the documentation since I messed up some commits.

Hind-M:
What was the reason the bug is happening, the corresponding [test](https://github.com/mamba-org/mamba/blob/main/micromamba/tests/test_env.py#L159-L175) is supposed to check the `--prune` flag's behavior...
Can you add a test reproducing the issue and showing that this PR patch is fixing it?

ZHG2017:
> Thanks for the PR. Can you only include in this one the relevant files for the bug fix? The documentation related files should be in another PR (I think reopening #3217 will do). For the `Linters` failure in the CI, running `pre-commit run --all` in the mamba source folder locally before committing should fix it (it will most likely modify the non-compliant files to follow the linting rules (defined [here](https://github.com/mamba-org/mamba/blob/main/.pre-commit-config.yaml)), so you should add them and commit them).

Hi, I have tried to use the `pre-commit` but did not find it any where, I checked in the folder `.git/hooks/` but there is nothing related to this CI check.  I wonder if this CI check only happens on the server

ZHG2017:
> What was the reason the bug is happening, the corresponding [test](https://github.com/mamba-org/mamba/blob/main/micromamba/tests/test_env.py#L159-L175) is supposed to check the `--prune` flag's behavior... Can you add a test reproducing the issue and showing that this PR patch is fixing it?

Yes, I have just realized that I did not add any unit test to validate my fix, actually I only manually typed some CLI command to verify if the fix works for the concerned issue.
The issue happens because the update inside the mamba will check if any package is previously installed in the target environment, if so, the update process will mark it as remove but just after that the internal process will mark the same package to be updated which is seemingly a conflict. Meanwhile, the most recent version of mamba will just ignore any installed package for a given environment file and will not install any package if it is missing in a given environment file.

Hind-M:
> > Thanks for the PR. Can you only include in this one the relevant files for the bug fix? The documentation related files should be in another PR (I think reopening #3217 will do). For the `Linters` failure in the CI, running `pre-commit run --all` in the mamba source folder locally before committing should fix it (it will most likely modify the non-compliant files to follow the linting rules (defined [here](https://github.com/mamba-org/mamba/blob/main/.pre-commit-config.yaml)), so you should add them and commit them).
> 
> Hi, I have tried to use the `pre-commit` but did not find it any where, I checked in the folder `.git/hooks/` but there is nothing related to this CI check. I wonder if this CI check only happens on the server

So if you're working within an environment created from the [environment-dev.yml](https://github.com/mamba-org/mamba/blob/main/dev/environment-dev.yml#L48) file, `pre-commit` should be already installed (you can check using `mamba list` if `mamba` is the package manager you're using).
Otherwise, you can install it (`mamba install pre-commit -c conda-forge`).

Then, you only need to run `pre-commit run --all` in your git root folder. If you run `git status` afterwards, you will see the modified files that you'll need to commit.

ZHG2017:
> > > Thanks for the PR. Can you only include in this one the relevant files for the bug fix? The documentation related files should be in another PR (I think reopening #3217 will do). For the `Linters` failure in the CI, running `pre-commit run --all` in the mamba source folder locally before committing should fix it (it will most likely modify the non-compliant files to follow the linting rules (defined [here](https://github.com/mamba-org/mamba/blob/main/.pre-commit-config.yaml)), so you should add them and commit them).
> > 
> > 
> > Hi, I have tried to use the `pre-commit` but did not find it any where, I checked in the folder `.git/hooks/` but there is nothing related to this CI check. I wonder if this CI check only happens on the server
> 
> So if you're working within an environment created from the [environment-dev.yml](https://github.com/mamba-org/mamba/blob/main/dev/environment-dev.yml#L48) file, `pre-commit` should be already installed (you can check using `mamba list` if `mamba` is the package manager you're using). Otherwise, you can install it: `mamba install pre-commit -c conda-forge`
> 
> Then, you only need to run `pre-commit run --all` in your git root folder. If you run `git status` afterwards, you will see the modified files that you'll need to commit.

Hi, I have used the  `pre-commit run --all` to format the code of mamba meanwhile I am still working on how to setup a unit test to prove my modification.

Hind-M:
There is `test_classic_spec`  in `test_update.py` failing [here](https://github.com/mamba-org/mamba/actions/runs/8434394421/job/23098036714?pr=3243). The modifications should have broken it.
This is due to the solution actions in transaction [not empty](https://github.com/mamba-org/mamba/blob/main/libmamba/src/core/transaction.cpp#L828-L831)  and so the ["message" key](https://github.com/mamba-org/mamba/blob/main/libmamba/src/core/transaction.cpp#L374-L378) is missing.

ZHG2017:
> There is `test_classic_spec` in `test_update.py` failing [here](https://github.com/mamba-org/mamba/actions/runs/8434394421/job/23098036714?pr=3243). The modifications should have broken it. This is due to the solution actions in transaction [not empty](https://github.com/mamba-org/mamba/blob/main/libmamba/src/core/transaction.cpp#L828-L831) and so the ["message" key](https://github.com/mamba-org/mamba/blob/main/libmamba/src/core/transaction.cpp#L374-L378) is missing.

Well, I have no idea how to fix this, I tried once using the built executable with my modification and the one built without my modification manually, and I saw no difference for the output. I don't know what can I do to pass this test.

PR #3262, opened 2024-04-04
Use QuantStack vcpkg registry for liblzma/xz

JohanMabille:
This can be reverted when the [xs issue of vcpkg](https://github.com/microsoft/vcpkg/pull/37957) has been fixed.


JohanMabille:
Closing now that the issue has been fixed upstream.

PR #3282, opened 2024-04-25
[CEP-15] Support `base_url` with `repodata_version: 2` using `mamba` parser

### LINKED ISSUE #3276: Implement CEP-15 `base_url` (repodata_version: 2)
[CEP-15](https://github.com/conda-incubator/ceps/blob/main/cep-15.md) landed some time ago. `conda` with `solver=classic` implements it now with https://github.com/conda/conda/pull/13744. It would be nice to have this in libmamba too so we can fully support it with `conda-libmamba-solver`. 🙏

Hind-M:
Fix #3276 
This adds the support only when using `simdjson` as a parser (which is the default).
Using `libsolv` doesn't support the CEP yet (cf. [this issue](https://github.com/openSUSE/libsolv/issues/562)).

### DIFF
--- a/libmamba/src/solver/libsolv/helpers.cpp
+++ b/libmamba/src/solver/libsolv/helpers.cpp
@@ -559,12 +559,6 @@ namespace mamba::solver::libsolv
                 {
                     base_url = std::string(url.value_unsafe());
                 }
-                else
-                {
-                    throw std::invalid_argument(
-                        "'info/base_url' not found in repodata.json file while 'repodata_version': 2"
-                    );
-                }
             }
         }
 
--- a/libmamba/src/solver/libsolv/helpers.cpp
+++ b/libmamba/src/solver/libsolv/helpers.cpp
@@ -540,14 +540,35 @@ namespace mamba::solver::libsolv
         const auto lock = LockFile(filename);
         const auto repodata = parser.load(filename);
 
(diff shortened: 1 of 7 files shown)


libmamba/src/solver/libsolv/helpers.cpp
```diff
@@ -540,14 +540,35 @@ namespace mamba::solver::libsolv
...
+            if (repodata_version.value_unsafe() == 2)
+            {
+                if (auto url = repodata.at_pointer("/info/base_url").get_string(); !url.error())
+                {
+                    base_url = std::string(url.value_unsafe());
+                }
+                else
+                {
+                    throw std::invalid_argument(
+                        "'info/base_url' not found in repodata.json file while 'repodata_version': 2"
```
jaimergp, line 565: I don't think the CEP forces `base_url` to be defined. It can be defined, but it's not required. So maybe this should not be an error.

JohanMabille: Indeed, according to the [CEP](https://github.com/conda/ceps/blob/main/cep-15.md#specification):

> In the absence of base_url, packages are loaded relative to the repodata.json

JohanMabille (changes requested):

JohanMabille (approved):

PR #3296, opened 2024-05-15
Fix OCIMirror use

Hind-M:


libmamba/src/core/package_fetcher.cpp
```diff
@@ -324,26 +325,12 @@ namespace mamba
 
     std::string PackageFetcher::channel() const
     {
-        if (m_package_info.package_type == specs::PackageType::Conda)
```
Hind-M, line 327: Not sure about why this was added but the package type is set to its default value `Unknown` while it should be set to `Conda` when downloading using OCI mirrors.
That said, changing this (in this PR) still makes the tests pass in the CI, so I don't know what's the use case of the `else`...
Same for `url_path()`.

libmamba/src/download/downloader.cpp
```diff
@@ -239,19 +239,24 @@ namespace mamba::download
...
                 case CURLINFO_HEADER_OUT:
+                    log = Console::hide_secrets(std::string_view(data, size));
                     logger->info(fmt::format("> {}", log));
                     break;
                 case CURLINFO_HEADER_IN:
+                    log = Console::hide_secrets(std::string_view(data, size));
                     logger->info(fmt::format("< {}", log));
                     break;
                 default:
+                    // WARNING Using `hide_secrets` here will give a seg fault on linux,
```
Hind-M, line 258: I actually found out that the seg fault was related to the `default` case, so getting `hide_secrets` back in the other cases and adding a warning in the default one.

libmamba/src/download/mirror_impl.cpp
```diff
@@ -250,9 +280,13 @@ namespace mamba::download
...
         std::string url = get_blob_url(split_path, data->sha256sum);
-        return MirrorRequest(initial_request, url);
+        std::vector<std::string> headers = { get_authentication_header(data->token) };
+
+        return MirrorRequest(initial_request, url, std::move(headers));
     }
 
+    // This is not used but could be if we use creds
+    // cf. OCIMirror constructor comment in header
     bool OCIMirror::need_authentication() const
```
Hind-M, line 290: Removing `need_authentication()` from where it was used to avoid confusion as it's not clear how the credentials are supposed to be used. To pull artifacts we request a token, and we don't support uploading like `powerloader`, so AFAIK there is no need for credentials at the moment.
I left the function implementation though even if it's not used (as a hint for maybe later if needed, like `username` and `password`) but we could also decide this is useless and just remove it/them.

SylvainCorlay:
🎉

libmamba/src/core/package_fetcher.cpp
JohanMabille: I don't remember either Oo

libmamba/include/mamba/download/request.hpp
```diff
@@ -140,6 +140,11 @@ namespace mamba::download
     {
         std::string mirror_name;
         std::string url_path;
+        // TODO maybe we would want to use a struct instead
+        // containing the `checksum` and its `type`
+        // (to handle other checksums types like md5...)
+        // cf. `Checksum` struct in powerloader
+        std::string sha256;
```
JohanMabille, line 147: I think we want something even more generic to handle additional parameters specific to a kind of storage. This should be done in a dedicated PR, though.

SylvainCorlay:
Bravo! Congrats on getting this in @Hind-M 🎉 🎉 🎉 🎉 🎉 🎉

Issue #2550, opened 2023-05-29
micromamba: No channels in initial setup/base


kfsone:
### Troubleshooting docs

- [X] My problem is not solved in the Troubleshooting docs

### How did you install Mamba?

forge: Mambaforge
mini: curl micro.mamba.pm/install.sh | bash
windows: powershell instructions: https://mamba.readthedocs.io/en/latest/installation.html#windows

### Search tried in issue tracker

no channels

### Latest version of Mamba

- [X] My problem is not solved with the latest version

### Tried in Conda?

I have this problem with Conda as well, without using Mamba

### Describe your issue

Technically an issue with the documentation, but also a pointless barrier to entry

in docker/sandbox,
- follow installation instructions for mambaforge and/or micromamba: https://mamba.readthedocs.io/en/latest/installation.html and others eg https://prefix.dev/docs/mamba/introduction
- attempt to mamba/micromamba install anything

```
(base) PS C:\Users\WDAGUtilityAccount> mm install wheel

                                           __
          __  ______ ___  ____ _____ ___  / /_  ____ _
         / / / / __ `__ \/ __ `/ __ `__ \/ __ \/ __ `/
        / /_/ / / / / / / /_/ / / / / / / /_/ / /_/ /
       / .___/_/ /_/ /_/\__,_/_/ /_/ /_/_.___/\__,_/
      /_/

warning  libmamba No 'channels' specified
error    libmamba Could not solve for environment specs
    The following package could not be installed
    └─ wheel   does not exist (perhaps a typo or a missing channel).
critical libmamba Could not solve for environment specs
```

in a similar vein, running the Windows mambaforge installer does not ask for or set the MAMBA_FORGE_ROOT, so the user experience is a little odd.

### mamba info / micromamba info

```shell
micro 1.4.3
```

### Logs

```shell
root@625b6930d533:/# micromamba install -v wheel

                                           __
          __  ______ ___  ____ _____ ___  / /_  ____ _
         / / / / __ `__ \/ __ `/ __ `__ \/ __ \/ __ `/
        / /_/ / / / / / / /_/ / / / / / / /_/ / /_/ /
       / .___/_/ /_/ /_/\__,_/_/ /_/ /_/_.___/\__,_/
      /_/

info     libmamba Parsing MatchSpec wheel
warning  libmamba No 'channels' specified
info     libmamba All targets to download are cached
info     libmamba All targets to download are cached
info     libmamba Adding package record to repo __archspec
[...]

PR #3303, opened 2024-05-28
Added instruction on how to add conda-forge to channels for micromamba

sbalk:
Took the bait in https://github.com/mamba-org/mamba/issues/2550#issuecomment-1627248505.
Added instruction to add conda-forge to channels for micromamba

### DIFF
--- a/micromamba/src/update.cpp
+++ b/micromamba/src/update.cpp
@@ -107,7 +107,7 @@ update_self(Configuration& config, const std::optional<std::string>& version)
         else
         {
             throw mamba::mamba_error(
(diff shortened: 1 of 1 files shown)


JohanMabille:
Thanks!

Hind-M (commented):
Hum, I think #2550 is more referring to using `micromamba` without configured channels, while the code here is related to self updating `micromamba`...
Anyway, the docs are giving now the correct command to use [here](https://github.com/mamba-org/mamba/blob/main/docs/source/installation/micromamba-installation.rst?plain=1#L141).

Hind-M:
Thanks for taking the time for this!

PR #3305, opened 2024-05-29
docs: Add note about `ccache`

jjerphan:
The compilation is failing without the presence of `ccache` in this development environment.


docs/source/developer_zone/dev_environment.rst
```diff
@@ -127,7 +127,7 @@ Install development dependencies
 
 .. code:: bash
 
-    micromamba create -n mamba -c conda-forge -f dev/environment.yml
+    micromamba create -n mamba -c conda-forge -f dev/environment-dev-extra.yml
```
Hind-M, line 130: I think `dev/environment-dev.yml` is more appropriate here as it's the basic mandatory environment, and then we can say that when using the provided CMake Preset, we may want to add `dev/environment-dev-extra.yml` as well.

docs/source/developer_zone/dev_environment.rst
```diff
@@ -130,6 +130,17 @@ Install development dependencies
     micromamba create -n mamba -c conda-forge -f dev/environment.yml
```
Hind-M, line 130:
```suggestion
    micromamba create -n mamba -c conda-forge -f dev/environment-dev.yml
```

Hind-M: `dev/environment.yml` doesn't exist anymore so this needs fixing.

docs/source/developer_zone/dev_environment.rst
```diff
@@ -130,6 +130,17 @@ Install development dependencies
...
+
+    You need to have ccache installed on your system.
+    You can either install it via your package manager or via conda-forge
+    in this development environment with:
+
+    .. code:: bash
+
+        micromamba install -n mamba -c conda-forge ccache
+
+
```
Hind-M, line 143: I was more thinking of a `note` just after talking about presets since it's related (just after this [line](https://github.com/mamba-org/mamba/blob/main/docs/source/developer_zone/dev_environment.rst?plain=1#L141)).
Something like:
```
.. note::
    If you do choose to use the provided CMake Preset, you may need to add ``ccache`` as an extra requirement as specified in ``dev/environment-dev-extra.yml``.
```

docs/source/developer_zone/dev_environment.rst
```diff
@@ -127,9 +127,20 @@ Install development dependencies
...
```
Hind-M, line 143: I think we can completely remove this part as `ccache` is related to the provided Preset (so the note below is enough).

Hind-M (approved):
Thanks!
I think this will be merged after `beta2` release (as the corresponding commit is ready), unless we decide to get more PRs in before.

PR #3306, opened 2024-05-30
Fix release scripts

Hind-M:
- Fix requested files limit (100) by using `grapghql` command allowing pagination
- Only write in file when PRs list is not empty in each category
- Fix regex expression to only capture characters between two square brackets without including the brackets themselves

### DIFF
--- a/releaser.py
+++ b/releaser.py
@@ -126,7 +126,7 @@ def main():
             release_start = idx + 1
             break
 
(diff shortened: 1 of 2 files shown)


Hind-M:
The CI failures are not related to this PR.
cf. #3308

JohanMabille (approved):

Hind-M:
Merging this as the CI failure on windows is not related.

PR #3307, opened 2024-05-30
[OCI - Mirrors] Add tests and doc

Hind-M:


libmamba/tests/src/util/test_os_osx.cpp
```diff
@@ -22,7 +22,7 @@ TEST_SUITE("util::os_osx")
         if (util::on_mac)
         {
             REQUIRE(maybe_version.has_value());
-            static const auto version_regex = std::regex(R"r(\d+\.\d+\.\d+)r");
+            static const auto version_regex = std::regex(R"r(\d+\.\d+(\.\d+)?)r");
```
Hind-M, line 25: This should be removed/rebased later after merging #3308 as it's already tackling this.

JohanMabille (approved):

Issue #3271, opened 2024-04-12
v2: Ctrl+C / SIGINT / KeyboardInterrupt handling not working


jaimergp:
### Troubleshooting docs

- [X] My problem is not solved in the Troubleshooting docs

### Anaconda default channels

- [X] I do NOT use the Anaconda default channels (pkgs/* etc.)

### How did you install Mamba?

Other (please describe)

### Search tried in issue tracker

ctrl c handling

### Latest version of Mamba

- [X] My problem is not solved with the latest version

### Tried in Conda?

Not applicable

### Describe your issue

In the latest beta, `Ctrl+C` handling from Python doesn't work correctly. In v1, we added `Context.use_default_signal_handler(...)` as a workaround, but this code doesn't seem to work in v2 anymore.

This simple script reproduces the problem. Run it with `python` and try to `Ctrl+C` out of it:

```python
import time
from libmambapy import Context
# This line below doesn't have any effect;
# it can be commented out and the result is the same
Context.use_default_signal_handler(False)
print("Sleep for 10s... Try to Ctrl+C")
time.sleep(10)
```

It just gets stuck until the end of the 10s sleep:

```
$ python debug_ctrl_c.py 
Sleep for 10s... Try to Ctrl+C
^C^C^C^C^C^C^C^C
```

If you remove the libmambapy imports, the outcome is:

```
$ python debug_ctrl_c.py 
Sleep for 10s... Try to Ctrl+C
^CTraceback (most recent call last):
  File "/workspaces/conda-libmamba-solver/debug_ctrl_c.py", line 6, in <module>
    time.sleep(10)
KeyboardInterrupt
```

### mamba info / micromamba info

```shell
libmamba version : 2.0.0
          mamba version : 2.0.0
           curl version : libcurl/8.7.1 OpenSSL/3.2.1 zlib/1.2.13 zstd/1.5.5 libssh2/1.11.0 nghttp2/1.58.0
     libarchive version : libarchive 3.7.2 zlib/1.2.13 liblzma/5.2.6 bz2lib/1.0.8 liblz4/1.9.3 libzstd/1.5.5
       envs directories : /root/.local/share/mamba/envs
          package cache : /root/.local/share/mamba/pkgs
                          /root/.mamba/pkgs
            environment : /opt/conda (active)
           env location : /opt/conda
      user config files : /root/.mambarc
 populated config files : /opt/conda/.condarc
       virtual packages : __unix=0=0
                          __linux=5.15.49=0
[...]

PR #3309, opened 2024-05-31
libmambapy:  use `Context` explicitly

Klaim:
In `libmambapy`:
- bound singletons lifetime to the first `Context` instance created
- throw an error if more than one `Context` instance are created at any time
(this only applies to python and will be changed at another time)
- therefore the python user must create a `Context` before using any other functions
- when creating the `Context`, users are able to set a `ContextOptions` to change the initial behavior, this allows to set `ContextOptions.enable_logging_and_signal_handling = False` and worksaround https://github.com/mamba-org/mamba/issues/3271

```python
import time
from libmambapy import Context, ContextOptions

# if we want the default behavior (with signal handling from mamba):
mamba_context = Context()

# OR if we do not want signal handling from context:
mamba_context = Context(ContextOptions(enable_logging_and_signal_handling = False))
```


Klaim:
Last issues:
- not sure if I updated all the necessary documentation, feel free to point me if there is more to update
- there seems to be a mismatch with the python api definition, I'm investigating
- there seems to be a kind-of-random crash in the mamba c++ tests, maybe due to the changes related to `mamba::Context` instance destruction (which now removes the loggers)

docs/source/developer_zone/changes-2.0.rst
```diff
@@ -55,11 +55,23 @@ The Python bindings to the C++ ``libmamba`` library remain available through ``i
...
 Changes include:
 
-- The global ``Context``, previously available through ``Context()``, must now be accessed through
-  ``Context.instance()``.
-  What's more, it is required to be passed explicitly in a few more functions.
-  Future version of ``libmambapy`` will continue in this direction until there are no global context.
+- The global ``Context``, previously available through ``Context()``, must now be instantiated at least
+  once before using ``libmambapy`` functions and types. Only one instance can exist at any time,
+  but that instance can be destroyed and another recreated later, for example if you use it in
+  specific functions scopes.
```
Hind-M, line 61: In this case, maybe add `mamba_context = libmambapy.Context()` [here](https://github.com/mamba-org/mamba/blob/main/docs/source/usage/specs.rst?plain=1#L30) and similar lines as well?

docs/source/usage/solver.rst
```diff
@@ -42,6 +42,8 @@ The first way to add a repository is from a list of |PackageInfo| using
 
    import libmambapy
 
+   mamba_context = libmambapy.Context()
```
Hind-M, line 45: I'm actually wondering if this is necessary since the tests in `libmambapy` are working without adding this (if I'm not mistaken)?

AntoinePrv: Indeed, `libmamba.solver`, `libmamba.utils` and `libmamba.specs` work independently from the Context.

docs/source/developer_zone/changes-2.0.rst
Klaim: I'm not completely sure but I guess it's safe to assume as long as we import `libmambapy` that should be a requirement

docs/source/usage/solver.rst
Klaim: Alright ,but then there is no example where I should add the line?

Hind-M: Hmm no I think it's actually independent from the logging output. I didn't find any reference to it in the corresponding folders in `libmamba`.

libmamba/src/api/install.cpp
```diff
@@ -444,7 +444,11 @@ namespace mamba
     {
         using Request = solver::Request;
 
-        request.jobs.reserve(request.jobs.size() + (!no_pin) * ctx.pinned_packages.size() + !no_py_pin);
+        LOG_WARNING << "DEBUG WARNING: no_py_pin = " << (no_py_pin ? "true" : "false");
```
Hind-M, line 447: We can remove this now I guess (or if you think this can be useful, use `LOG_DEBUG` instead?).

Klaim: ah yes I added this to check something and forgot to remove it, will do

JohanMabille:
but_why_jacky_chan

Klaim:
> but_why_jacky_chan

Yeah the ci issue is definitely not related to this PR but not sure if it's a good idea to merge.

Hind-M:
I opened another [PR](https://github.com/mamba-org/mamba/pull/3321) to reproduce the issue again on the CI (after adding some log), but couldn't get it to fail for now. I re-ran the failed job here, let's see if it succeeds this time (if so I suggest that we move forward and merge if this is ready)...

docs/source/developer_zone/changes-2.0.rst
JohanMabille: >  Only one instance can exist at any time

Do we have a mean to guarantee that?

JohanMabille (approved):

Klaim:
@JohanMabille 
> Do we have a mean to guarantee that? If so this can be done in a future PR.

Sorry I didnt see that comment before. Answer is yes it's already enforced but only at runtime: we throw an exception when a `Context` is created from `libmambapy` while there is already one instance alive, see https://github.com/mamba-org/mamba/pull/3309/files#diff-b06477fac9b8a278eaf243edbc7dc3c91c14bfde8db0b974ecd6c5e1edf44f8dR676

PR #3310, opened 2024-06-05
OCI/Conda mapping

Hind-M:
Map packages names starting with `_` to match OCI registries rules (and [CEP](https://github.com/conda/ceps/pull/70))

### DIFF
--- a/libmamba/src/download/mirror_impl.cpp
+++ b/libmamba/src/download/mirror_impl.cpp
@@ -309,13 +309,25 @@ namespace mamba::download
 
     std::string OCIMirror::get_repo(const std::string& repo) const
     {
+        // OCI image names cannot start with `_`
+        // Get the package name and prepend with `zzz` to map to `channel_mirrors` implementation
+        auto parts = util::rsplit(repo, "/", 1);
+        assert(parts.size() == 2);
+        std::string mapped_package_name = parts.back();
+        std::string mapped_repo = repo;
+        if (util::starts_with(mapped_package_name, "_"))
+        {
+            mapped_package_name.insert(0, std::string("zzz"));
+            mapped_repo = fmt::format("{}/{}", parts[0], mapped_package_name);
+        }
+
         if (!m_repo_prefix.empty())
         {
-            return fmt::format("{}/{}", m_repo_prefix, repo);
+            return fmt::format("{}/{}", m_repo_prefix, mapped_repo);
         }
         else
         {
-            return repo;
+            return mapped_repo;
         }
     }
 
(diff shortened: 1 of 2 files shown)


libmamba/src/download/mirror_impl.cpp
```diff
@@ -71,7 +71,7 @@ namespace mamba::download
      * OCIMirror implementation *
      ****************************/
 
-    namespace
+    namespace utils
```
JohanMabille, line 74: Why turning it into a non anonymous namespace? Did you have conflict with functions defined in another namespace?

JohanMabille (commented):
LGTM except for the anonymous namesapce.

wolfv:
Did you do an end-to-end test yet? Because I was pretty confused that this ever worked before (since `_libgcc_mutex` is basically in all Linux environments and starts with an underscore) :)

libmamba/src/download/mirror_impl.cpp
Hind-M: To have access in the [tests](https://github.com/mamba-org/mamba/pull/3310/files#diff-409867cced670ed6abe27e11cb1d7c4a1a4363d90b094eb5f1dd3016bc5ff0f9R15-R54).

Hind-M:
> LGTM except for the anonymous namesapce.

Please review https://github.com/mamba-org/mamba/pull/3307 first, as this PR is based on it.

Hind-M:
> Did you do an end-to-end test yet? Because I was pretty confused that this ever worked before (since `_libgcc_mutex` is basically in all Linux environments and starts with an underscore) :)

I added a check on mamba list output after creating the test env, and it's using `_go_select` instead of `_libgcc_mutex` now.

libmamba/src/download/mirror_impl.cpp
JohanMabille: In that case I think it would be worth having a preprocessor token defining the namespace to util if we build the test, and emtpy if not. Otherwise this symbol (and the other ones included in this namespace) will be exported and visible from other libraries, which is something we wanted to avoid with the anonymous namespace.

@Klaim any thought on this?

Klaim: You are correct that naming the scope makes the symbols visible, anonymous namespaces are essentially similar to marking everything `static` (and more). So this is not a small change.

For testing internal functions there are various usual strategies:
- making the internal functions visible (like here);
- tests link with a "utility library" version of the library, which is essentially like a static library except it's not removing unused functions at linking;
- real proper "unit tests" as in test source file which is specific to that component and runs as a single executable, testing only that sub-component;
- the tests "include" the cpp they test (which can lead to ODR violations, depends on how both are coded);

From the current mamba organisation, looks like the first solution is the easiest to setup at the moment. Alternating the definition of the namespace might not work properly because this is an internal translation unit of libmamba, so you would have to build libmamba twice, once with the visible symbols, once with them hidden.

JohanMabille: Ok then let's keep it as is for now, if it clashes with other symbols when linking, we will simply change its name.

JohanMabille (approved):

Klaim:
@wolfv (When names are "reserved" it implies that the "implementation" is allowed to use them, they are reserved to avoid conflicts mainly, so I suppose `_libgcc_mutex` is considered part of the "implementation" as it's part of the OS API? That would explain it.)

Issue #3312, opened 2024-06-06
Don't try to write SOLV cache in read-only directories


jaimergp:
### Troubleshooting docs

- [X] My problem is not solved in the Troubleshooting docs

### Anaconda default channels

- [X] I do NOT use the Anaconda default channels (pkgs/* etc.)

### How did you install Mamba?

Other (please describe)

### Search tried in issue tracker

writable read-only cache

### Latest version of Mamba

- [X] My problem is not solved with the latest version

### Tried in Conda?

Not applicable

### Describe your issue

Comes from https://github.com/conda/conda-libmamba-solver/issues/522

conda-libmamba-solver + libmamba 1.5.x cannot operate on read-only caches because loading via `api.Repo(...)` always tries to write the SOLV cache file back to disk, and fails to do so in the write permissions are not there:

```
File "/tmp/software/python/3.11.4/lib/python3.11/site-packages/conda_libmamba_solver/index.py", line 293, in _json_path_to_repo_info

    repo = api.Repo(self._pool, noauth_url, str(path_to_use), escape_channel_url(noauth_url))
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

RuntimeError: Permission denied
```

I think this line:

https://github.com/mamba-org/mamba/blob/4f269258b4237a342da3e9891045cdd51debb27c/libmamba/src/core/repo.cpp#L332

should have some error handling because that `write_solv()` call is not critical to the repo loading. Nice to have, sure, but not required.

### mamba info / micromamba info

```shell
libmamba 1.5.8
```

jaimergp:
Just learnt that this is solved in v2, so feel free to close if 1.x is not being maintained anymore.

PR #3313, opened 2024-06-06
[1.x] Ignore SOLV writing exceptions

jaimergp:
Closes https://github.com/mamba-org/mamba/issues/3312

### DIFF
--- a/libmamba/src/core/repo.cpp
+++ b/libmamba/src/core/repo.cpp
@@ -6,6 +6,7 @@
 
 #include <algorithm>
 #include <array>
+#include <exception>
 #include <tuple>
 
 #include <nlohmann/json.hpp>
@@ -329,7 +330,14 @@ namespace mamba
 
         if (name() != "installed")
         {
-            write_solv(solv_file);
+            try
+            {
+                write_solv(solv_file);
+            }
+            catch (const std::exception& e)
+            {
+                LOG_INFO << "Could not write libsolv solv file for " << json_file << ": " << e.what();
+            }
         }
     }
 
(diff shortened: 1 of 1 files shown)


jaimergp:
FWIW, this is implemented in libmamba 2.x as a warning:

https://github.com/mamba-org/mamba/blob/a10a532fd810453d337547f35b43abe78aee5053/libmamba/src/core/package_database_loader.cpp#L117-L127

libmamba/src/core/repo.cpp
```diff
@@ -329,7 +330,14 @@ namespace mamba
...
         if (name() != "installed")
         {
-            write_solv(solv_file);
+            try
+            {
+                write_solv(solv_file);
+            }
+            catch (const std::exception& e)
+            {
+                LOG_INFO << "Could not write libsolv solv file for " << json_file << ": " << e.what();
```
Hind-M, line 339: I think we can emit a warning instead (and to be consistent with mamba 2):
```suggestion
                LOG_WARNING << "Could not write libsolv solv file for " << json_file << ": " << e.what();
```

Hind-M:
I think you can rebase and that will fix the CI :)

jaimergp:
Let's see what happens!

mcg1969:
@jaimergp I'd love to see this fixed, so thank you for working on it. May we consider pushing it forward?

mcg1969:
Use cases: I have a read-only package cache with pre-computed `.solv` files

Hind-M (approved):
Thanks!

PR #3318, opened 2024-06-11
maint: Remove `PrefixData::load`

jjerphan:
Merge its content within `PrefixData::PrefixData`.

### DIFF
--- a/libmamba/src/core/prefix_data.cpp
+++ b/libmamba/src/core/prefix_data.cpp
@@ -45,11 +45,6 @@ namespace mamba
         , m_prefix_path(prefix_path)
         , m_channel_context(channel_context)
     {
-        load();
-    }
-
-    void PrefixData::load()
-    {
         auto conda_meta_dir = m_prefix_path / "conda-meta";
         if (lexists(conda_meta_dir))
         {


JohanMabille (approved):

Hind-M:
We should remove [this](https://github.com/mamba-org/mamba/blob/main/libmamba/include/mamba/core/prefix_data.hpp#L46) as well.

jjerphan:
Yes, I probably should have kept this PR in Draft. 

There's the same pattern with `SubdirData`:

https://github.com/mamba-org/mamba/blob/3e1e97b0e6e49cac042cff7396fbbe5e71a63a9f/libmamba/src/core/subdirdata.cpp#L594

Hind-M:
IMHO, I think we should keep them all, since it's kind of separating the logic and it's easier to read (better that blocks of code at once). And it's not like it's defined somewhere far away in the codebase or in another file... It's just the method right below...

jjerphan:
`PrefixData::load` only was called once in the constructor, in that case I think it would have made sense to merge it to prevent any misuse. No strong opinion. 

Do you propose reverting the merge commit?

Hind-M:
No strong opinion either, I was just expressing mine ^^

> Do you propose reverting the merge commit?

Either that or removing the leftover in the header. But for `SubdirData`, I would keep it the way it is.

jjerphan:
I just have opened https://github.com/mamba-org/mamba/pull/3325.

PR #3330, opened 2024-06-20
build: Use Position-Independent Code

jjerphan:
I needed it to compile libmamba's test suite on my workflow.

Naive: question are there reasons for not using [PIC](https://wiki.gentoo.org/wiki/Hardened/Position_Independent_Code_internals) at the moment?


JohanMabille:
Building shared libraries with CMake automatically turns on the fPIC flag. The only usecase where you wold need this flag is when you mix static and dynamic linked libraries, which is not what we are doing in mamba.

PR #3342, opened 2024-07-04
fix: narrowing conversion in util::compare tests

pbrit:


Hind-M:
Hmm, I would rather do something like:
`CHECK(cmp_equal(signed char{ -1 }, signed char{ -1 }));`
and keep the tests.

Hind-M:
> Hmm, I would rather do something like: `CHECK(cmp_equal(signed char{ -1 }, signed char{ -1 }));` and keep the tests.

@pbrit What do you think?

PR #3352, opened 2024-07-17
Unpin cryptography, python, and add make to environment-dev.yml

jaimergp:
I'm using this in some Docker containers as part of my devcontainer config, and they don't include `make`. Also, cryptography<40 is not needed anymore (see https://github.com/conda/conda-content-trust/pull/57).

I also unpinned `python` so you can use it on any solvable Python environment without overriding to 3.9.

`mitmproxy` and `doctest` are not available on linux-aarch64, so that fails, but the selector syntax only allows OS-level names, not architectures, so I can't omit those.

### DIFF
--- a/dev/environment-dev.yml
+++ b/dev/environment-dev.yml
@@ -6,6 +6,7 @@ dependencies:
   - cmake >=3.16
   - pkg-config  # Used by some CMake dependencies
   - ninja
+  - make  # not always present
   # libmamba dependencies
   - cpp-expected
   - fmt
@@ -34,7 +35,7 @@ dependencies:
   - sel(win): menuinst
   - conda-content-trust
   - conda-package-handling
-  - cryptography<40.0  # Or breaks conda-content-trust
+  - cryptography
   - securesystemslib
   # libmambapy build dependencies
   - scikit-build


dev/environment-dev.yml
```diff
@@ -24,7 +25,7 @@ dependencies:
   # micromamba dependencies
   - cli11 >=2.2
   # micromamba test dependencies
-  - python =3.9  # Some not too recent version
+  - python
```
Hind-M, line 28: I actually don't know the reason of pinning python here... Maybe this is hiding some incompatibilities. Let's see what happens with the CI.

Hind-M:
The failures seem to be related to using `fmt 11`. As https://github.com/mamba-org/mamba/pull/3368 was merged, rebasing on main should fix it.

dev/environment-dev.yml
Hind-M: Ok I think unpinning python makes it get `fmt 11` everywhere while #3368 only handles a part of it in the codebase.
I suggest restoring python pinning for this PR and handle `fmt 11` in another one (I will open an issue for it).
Is it something blocking at the moment or more like an enhancement opening for more possibilities?

That being said, that's only showing one issue of possible other ones when unpinning python, so that would be something to check within the fixing PR.

jjerphan:
Missing elements for the support of fmt 11 are introduced in https://github.com/mamba-org/mamba/pull/3371.

Hind-M:
Now that #3371 is merged, rebasing should hopefully fix the failures this time :)

jjerphan:
#3377 should also have resolved the other problem on the CI, @jaimergp. So this PR might be integrated easily now. :)

Hind-M (approved):
Thanks!

PR #3363, opened 2024-07-26
Make more classes hashable and comparable

jjerphan:
This allows using instances of `MatchSpec` and `PackageInfo` in mappings such as `std::unordered_map`.

### DIFF
--- a/libmamba/include/mamba/specs/build_number_spec.hpp
+++ b/libmamba/include/mamba/specs/build_number_spec.hpp
@@ -125,7 +125,7 @@ namespace mamba::specs
          */
         [[nodiscard]] auto contains(BuildNumber point) const -> bool;
 
-        // TODO: only use the `= default` implementation of `operator==` when we will use C++20.
+        // TODO(C++20): replace by the `= default` implementation of `operator==`
         [[nodiscard]] auto operator==(const BuildNumberSpec& other) const -> bool
         {
             return m_predicate == other.m_predicate;
--- a/libmamba/include/mamba/specs/build_number_spec.hpp
+++ b/libmamba/include/mamba/specs/build_number_spec.hpp
@@ -15,6 +15,7 @@
 #include <fmt/format.h>
 
 #include "mamba/specs/error.hpp"
+#include "mamba/util/tuple_hash.hpp"
 
 namespace mamba::specs
 {
@@ -124,6 +125,16 @@ namespace mamba::specs
          */
         [[nodiscard]] auto contains(BuildNumber point) const -> bool;
 
+        [[nodiscard]] auto operator==(const BuildNumberSpec& other) const -> bool
+        {
+            return m_predicate == other.m_predicate;
+        }
+
+        [[nodiscard]] auto operator!=(const BuildNumberSpec& other) const -> bool
+        {
+            return !(*this == other);
+        }
+
     private:
 
         BuildNumberPredicate m_predicate;
@@ -153,6 +164,18 @@ struct fmt::formatter<mamba::specs::BuildNumberSpec>
 
     auto
     format(const ::mamba::specs::BuildNumberSpec& spec, format_context& ctx) -> decltype(ctx.out());
+};
+
+template <>
+struct std::hash<mamba::specs::BuildNumberSpec>
+{
+    auto operator()(const mamba::specs::BuildNumberSpec& spec) const -> std::size_t
+    {
+        auto seed = std::size_t{ 0 };
+        seed = mamba::util::hash_combine_val(seed, spec.str());
+
+        return seed;
+    }
 };
 
 #endif
--- a/libmamba/include/mamba/specs/build_number_spec.hpp
+++ b/libmamba/include/mamba/specs/build_number_spec.hpp
@@ -125,6 +125,7 @@ namespace mamba::specs
          */
         [[nodiscard]] auto contains(BuildNumber point) const -> bool;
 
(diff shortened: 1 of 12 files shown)


libmamba/include/mamba/specs/build_number_spec.hpp
```diff
@@ -124,6 +125,16 @@ namespace mamba::specs
          */
         [[nodiscard]] auto contains(BuildNumber point) const -> bool;
 
+        [[nodiscard]] auto operator==(const BuildNumberSpec& other) const -> bool
```
Klaim, line 128: Add a `// TODO: ` to replace this by  `= default;` as soon as we use C++20 for mamba, as it will help with maintenance.

Do it for every additional `operator==` in this PR, because this is introducing a potential maintenance headache (I think that's why it was not done before and also because it was not necessary).

libmamba/include/mamba/specs/chimera_string_spec.hpp
```diff
@@ -66,4 +77,15 @@ struct fmt::formatter<mamba::specs::ChimeraStringSpec>
...
 };
 
+template <>
+struct std::hash<mamba::specs::ChimeraStringSpec>
+{
+    auto operator()(const mamba::specs::ChimeraStringSpec& spec) const -> std::size_t
+    {
+        auto seed = std::size_t(0);
+        seed = mamba::util::hash_combine_val(seed, spec.str());
+        return seed;
```
Klaim, line 87: Why not just `return mamba::util::hash_combine_val(0, spec.str());` ?

libmamba/include/mamba/specs/regex_spec.hpp
```diff
@@ -47,6 +47,18 @@ namespace mamba::specs
 
         [[nodiscard]] auto str() const -> const std::string&;
 
+        [[nodiscard]] auto operator==(const RegexSpec& other) const -> bool
+        {
+            return (
```
Klaim, line 52: remove the parenthesis

libmamba/include/mamba/specs/match_spec.hpp
```diff
@@ -150,7 +169,23 @@ namespace mamba::specs
             std::string features = {};
             string_set track_features = {};
             bool optional = false;
+
+            [[nodiscard]] auto operator==(const ExtraMembers& other) const -> bool
+            {
+                return filename == other.filename && subdirs == other.subdirs && md5 == other.md5
```
Klaim, line 175: I feel like one line per comparison would be easier to maintain.

libmamba/include/mamba/specs/package_info.hpp
```diff
@@ -81,4 +82,47 @@ namespace mamba::specs
...
     void from_json(const nlohmann::json& j, PackageInfo& pkg);
 }
+
+template <>
+struct std::hash<mamba::specs::PackageInfo>
+{
+    auto operator()(const mamba::specs::PackageInfo& pkg) const -> std::size_t
+    {
+        auto seed = std::size_t(0);
+        seed = mamba::util::hash_combine_val(seed, pkg.name);
```
Klaim, line 92: All this code would be simplified by using `hash_vals` instead (it's the generalisation of the code pattern here, if my understanding is correct)
Same remark for other hash impls.

libmamba/include/mamba/specs/build_number_spec.hpp
```diff
@@ -125,6 +125,7 @@ namespace mamba::specs
          */
         [[nodiscard]] auto contains(BuildNumber point) const -> bool;
 
+        // TODO: only use the `= default` implementation of `operator==` when we will use C++20.
```
Klaim, line 128:
```suggestion
        // TODO: replace by `= default` implementation of `operator==` when we will use C++20.
```
Same suggestion for all others.

libmamba/include/mamba/specs/match_spec.hpp
```diff
@@ -136,22 +136,26 @@ namespace mamba::specs
...
         [[nodiscard]] auto contains_except_channel(const PackageInfo& pkg) const -> bool;
 
-        auto operator==(const MatchSpec& other) const -> bool
+        // TODO: only use the `= default` implementation of `operator==` when we will use C++20.
+        [[nodiscard]] auto operator==(const MatchSpec& other) const -> bool
         {
-            return m_channel == other.m_channel && m_version == other.m_version
-                   && m_name == other.m_name && m_build_string == other.m_build_string
-                   && m_name_space == other.m_name_space && m_build_number == other.m_build_number
+            return m_channel == other.m_channel               //
```
Klaim, line 142: remove empty comments, same for others

Klaim (approved):
LGTM with some cleanups

libmamba/include/mamba/specs/match_spec.hpp
jjerphan: This allows formatting with one pair of attribute per line. Are there other ways to format such lines?

Klaim: Is this because the automatic formatting forces to do that? Otherwise I dont get why the `//` are necessary at all?

Klaim: If it's because of the automatic formatting, feel free to ignore my remark.

Hind-M:
Can you please add tests in the corresponding files in `libmamba/tests/src/specs` and `libmamba/tests/src/util`? Thanks!

PR #3366, opened 2024-07-29
[CI] Revert "Make more classes hashable and comparable (#3363)"

Hind-M:
This is a test to check CI failure after merging #3363

### DIFF
--- a/libmamba/include/mamba/specs/build_number_spec.hpp
+++ b/libmamba/include/mamba/specs/build_number_spec.hpp
@@ -15,7 +15,6 @@
 #include <fmt/format.h>
 
 #include "mamba/specs/error.hpp"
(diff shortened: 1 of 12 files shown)


jjerphan:
I think this is due to [fmt 11 now being used](https://github.com/mamba-org/mamba/actions/runs/10141377644/job/28038514499#step:7:198). :confused:

Hind-M:
> I have not found other places where it is being used. It is as if [the global pin on `fmt 10`](https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/e0c71c48b56a202c680ffdfa7e83ac1d0a4930e1/recipe/conda_build_config.yaml#L338-L339) was ignored.

Hmm weird...  ☹️

PR #3367, opened 2024-07-29
build: Depend on fmt <11

jjerphan:


Hind-M:
I think [this](https://github.com/mamba-org/mamba/blob/main/.github/workflows/static_build.yml#L145) is where we need to add a constraint. We should keep the other files as they are.

jjerphan:
That's right.

I proposed to instead support fmt 11 directly with https://github.com/mamba-org/mamba/pull/3368.

Issue #3370, opened 2024-07-30
Unpin `Python` in `environment-dev.yml`


Hind-M:
The reasons behind this pinning are not clear, and it's preventing from using other python versions.
One issue, shown in #3352, is getting `fmt 11` which is not handled entirely in the codebase.
When unpinning, the CI will give more info and may generate other issues to handle.

PR #3368, opened 2024-07-29
build: Support fmt 11

jjerphan:
Required changes for `fmt 11`:
 - `fmt::join` is now declared and defined in `fmt/ranges.h` (see https://github.com/fmtlib/fmt/commit/50565f9853926501bd1c9bda07c6a98f4d940691)
 - `fmt::format` has to respect its argument's `const`-qualification
 
Other changes are required as include paths might have change slightly.

PR #3371, opened 2024-07-30
build: Support fmt 11 (follow-up)

jjerphan:
Fixes https://github.com/mamba-org/mamba/issues/3370.

Follow up of #3368.

A few other elements were required, as observed in: https://github.com/mamba-org/mamba/pull/3352.

### DIFF
--- a/.github/workflows/unix_impl.yml
+++ b/.github/workflows/unix_impl.yml
@@ -29,6 +29,8 @@ jobs:
           environment-name: build_env
           cache-environment: true
       - name: Use libcxx <18 (see conda-forge/libcxx-feedstock#162)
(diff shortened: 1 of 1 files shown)


.github/workflows/unix_impl.yml
```diff
@@ -28,6 +28,10 @@ jobs:
           environment-file: ./dev/environment-dev.yml
           environment-name: build_env
           cache-environment: true
+      - name: Use libcxx <18 (see conda-forge/libcxx-feedstock#162)
```
Hind-M, line 31: Can we add a `TODO` comment to remove this part when the issue is fixed?

Hind-M (approved):

PR #3396, opened 2024-08-06
[micromamba] Fix behavior of `env update` (to mimick conda)

### LINKED ISSUE #3388: `micromamba env update` behavior different from `conda`
`micromamba env update` doesn't update the whole environment when given an env file but only specs already installed.
cf. https://github.com/mamba-org/mamba/issues/3354
Specifically https://github.com/mamba-org/mamba/issues/3354#issuecomment-2263391579, https://github.com/mamba-org/mamba/issues/3354#issuecomment-2263606469 and https://github.com/mamba-org/mamba/issues/3354#issuecomment-2264812839

Hind-M:
Fix #3388

### DIFF
--- a/libmamba/include/mamba/api/update.hpp
+++ b/libmamba/include/mamba/api/update.hpp
@@ -11,13 +11,39 @@ namespace mamba
 {
     class Configuration;
 
-    void update(
-        Configuration& config,
-        bool update_all = false,
-        bool prune_deps = false,
-        bool env_update = false,           // Specific to `env update` command
-        bool remove_not_specified = false  // Specific to `env update` command
-    );
+    enum class UpdateAll : bool
+    {
+        No = false,
+        Yes = true,
+    };
+
+    enum class PruneDeps : bool
+    {
+        No = false,
+        Yes = true,
+    };
+
+    enum class EnvUpdate : bool  // Specific to `env update` command
+    {
+        No = false,
+        Yes = true,
+    };
+
+    enum class RemoveNotSpecified : bool  // Specific to `env update` command
+    {
+        No = false,
+        Yes = true,
+    };
+
+    struct UpdateParams
+    {
+        UpdateAll update_all = UpdateAll::No;
+        PruneDeps prune_deps = PruneDeps::No;
+        EnvUpdate env_update = EnvUpdate::No;
+        RemoveNotSpecified remove_not_specified = RemoveNotSpecified::No;
+    };
+
+    void update(Configuration& config, const UpdateParams& update_params = {});
 }
 
 #endif
--- a/libmamba/include/mamba/api/update.hpp
+++ b/libmamba/include/mamba/api/update.hpp
@@ -15,7 +15,8 @@ namespace mamba
         Configuration& config,
         bool update_all = false,
         bool prune_deps = false,
-        bool remove_not_specified = false
+        bool env_update = false,           // Specific to `env update` command
+        bool remove_not_specified = false  // Specific to `env update` command
     );
 }
 
--- a/libmamba/src/api/update.cpp
+++ b/libmamba/src/api/update.cpp
@@ -27,6 +27,7 @@ namespace mamba
             std::vector<std::string> specs,
             bool update_all,
             bool prune_deps,
+            bool env_update,
             bool remove_not_specified
         ) -> solver::Request
         {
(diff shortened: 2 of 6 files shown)


libmamba/include/mamba/api/update.hpp
```diff
@@ -15,7 +15,8 @@ namespace mamba
         Configuration& config,
         bool update_all = false,
         bool prune_deps = false,
-        bool remove_not_specified = false
+        bool env_update = false,           // Specific to `env update` command
```
Klaim, line 18: Potential future improvement: replace these arguments by a struct containing the sequence of booleans, that would help with maintaining the code calling this and make everything less verbose.

libmamba/src/api/update.cpp
```diff
@@ -187,17 +228,17 @@ namespace mamba
         );
 
 
-        auto execute_transaction = [&](MTransaction& transaction)
+        auto execute_transaction = [&](MTransaction& trans)
```
Klaim, line 231: Why shortening such a clear name? 🥲

micromamba/src/env.cpp
```diff
@@ -284,6 +284,14 @@ set_env_command(CLI::App* com, Configuration& config)
 
     update_subcom->callback(
         [&config]
-        { update(config, /*update_all*/ false, /*prune_deps*/ false, remove_not_specified); }
+        {
+            update(
+                config,
+                /*update_all*/ false,
```
Klaim, line 290: It would be worth, in addition to have a struct with these options, to use `enum class` values to set them so that it's clear just by reading, without comments (which could go out of sync). C++20 also allows to `{ .update_all = false, ... }` which would be useful too.

micromamba/tests/test_env.py
```diff
@@ -138,35 +139,39 @@ def test_env_remove(tmp_home, tmp_root_prefix):
 def test_env_update(tmp_home, tmp_root_prefix, tmp_path, prune):
     env_name = "env-create-update"
 
-    # Create env with python=3.11.0 and xtensor=0.25.0
-    helpers.create("python=3.11.0", "xtensor=0.25.0", "-n", env_name, "--json", no_dry_run=True)
+    # Create env with python=3.10.0 and xtensor=0.25.0
+    helpers.create("python=3.10.0", "xtensor=0.25.0", "-n", env_name, "--json", no_dry_run=True)
```
Klaim, line 143: That doesnt seem like a trivial change. Is this to reflect conda's behavior?

libmamba/src/api/update.cpp
Hind-M: Because of the shadowing warnings 🥲
We can also do something like `inner_transaction` but it's longer...

micromamba/tests/test_env.py
Hind-M: So the changes in the test are a kind of an enhancement containing more checks that the basic previous one:

- It now handles a spec containing a name and a version (it was just the spec name before and there was actually a bug in there)
- Add a new package not already in the current env (the main difference with `conda` and the rational behind this PR), so that it will be installed now.
- Ask to update a spec with a different version (wasn't the case before), so it could be `3.10.0` or anything really, to be updated explicitly to something more recent, `3.11.*` for example...

But that doesn't change the previous test checks generally speaking.

Klaim: > Ask to update a spec with a different version (wasn't the case before), so it could be 3.10.0 or anything really, to be updated explicitly to something more recent, 3.11.* for example...

Would it be worth testing separately 2 or 3 different versions?

Hind-M: Hmm yes but it could possibly lead to code duplication, should we rather do that or should I add more comments to explain the test in detail?

libmamba/src/api/update.cpp
Klaim: Oh ok, no it's fine as is 👍🏽 I didnt know about the shadowing warnings.
I personally dont like warnings about shadowing because shadowing is an excellent way to make the local code clear without impacting the global code, but most people disagree hahaha

micromamba/tests/test_env.py
Klaim: Maybe add an issue to do that in at a separate time?

Hind-M: You mean split the tests (even if it generates code duplication) or add the comments?
I think it's fine if we do it here, while it's fresh ^^ (maybe even best unless we want to change something else in addition to that at the same time in the codebase...)

Klaim: I meant adding an additional "python version" parameter to `test_env_update` and that parameter would be set to 2 or 3 python versions, with at least one being recent, one not being recent, at least. (that's assuming the testing context allows that).
Would it make sense to do that?

Hind-M: Ah! No I don't think the versions would make a difference, that's why changing it was not a big deal.

Klaim: I see, ok so nevermind, it's all good 👍🏽

Hind-M: Can I have an approval then? :) Thanks!

Klaim (approved):
LGTM 👍🏽

PR #3398, opened 2024-08-08
[win-64] Remove workaround

Hind-M:
Removing the workaround as the [issue](https://github.com/actions/runner-images/issues/10004) is now fixed.


Hind-M:
The failure is related to `fmt` (issue fixed [here](https://github.com/mamba-org/mamba/pull/3400)).

Klaim (approved):

Issue #3415, opened 2024-08-28
Logging callbacks have too much overhead on libmambapy


jaimergp:
### Troubleshooting docs

- [X] My problem is not solved in the Troubleshooting docs

### Anaconda default channels

- [X] I do NOT use the Anaconda default channels (pkgs/* etc.)

### How did you install Mamba?

Mambaforge or latest Miniforge

### Search tried in issue tracker

logging debug

### Latest version of Mamba

- [X] My problem is not solved with the latest version

### Tried in Conda?

Not applicable

### Describe your issue

2.x API includes logging callbacks so the information reported by libsolv or libmamba can be routed through a Python callable via `libmambapy.solver.libsolv.Database.set_logger()`. However, this becomes very slow when the `libsolv` throughput is too large (e.g. `verbosity=DEBUG`). It causes a slowdown so large that the solver takes minutes to finish.

Are there any ways to have a less resource consuming way? I don't need fancy formatting or full control, just the ability to decide _when_ to send different levels to stdout based on the verbosity chosen by the user.

### mamba info / micromamba info

```shell
This is with libmambapy 2.0.0rc3
```

### Logs

```shell
N/A
```

### environment.yml

```yaml
N/A
```

### ~/.condarc

```yaml
N/A
```

jjerphan:
Thank you for the report.

We have observed a tangential issue for some environments whose resolution via `libsolv` was hanging due to the logging system. Both issue have the same root cause: most of `libsolv`'s logs are debug logs which are emitting by default, even for release builds.

See:
https://github.com/mamba-org/mamba/blob/08d7b4e9cb9b168d05f3bdc57c0708ee80d5a1c9/libmamba/src/solver/libsolv/database.cpp#L90-L108

https://github.com/mamba-org/mamba/blob/08d7b4e9cb9b168d05f3bdc57c0708ee80d5a1c9/libmamba/src/solver/libsolv/database.cpp#L110-L127

An option can be for (release) builds not to emit debug logs unless if explicitly asked by users (with `verbosity=DEBUG` as you propose).

jjerphan:
#3416 addresses the performance regression. We think that the handling of log levels and of verbosity better be implemented in another PR.

jjerphan:
@jaimergp: can you report if you have observed improvements?

jaimergp:
> @jaimergp: can you report if you have observed improvements with 2.0.0rc4?

[...]

PR #3416, opened 2024-08-28
fix: Reduce logging system overhead

jjerphan:
Partially addresses https://github.com/mamba-org/mamba/issues/3415.

This fixes the regression which has been observed in several release candidates of 2.0.0 by restoring [the behavior of `1.x`](https://github.com/mamba-org/mamba/blob/1.x/libmamba/src/core/pool.cpp#L72).

@Hind-M and I believe that we should think of the UX of the verbosity level handling as proposed in https://github.com/mamba-org/mamba/issues/3415 in another PR, especially given this work would necessitate much more work with potentially change to public APIs.


micromamba/tests/test_create.py
```diff
@@ -143,6 +143,20 @@ def test_env_lockfile_different_install_after_create(tmp_home, tmp_root_prefix,
     helpers.install("-p", env_prefix, "-f", install_spec_file, "-y", "--json")
 
 
+@pytest.mark.parametrize("shared_pkgs_dirs", [True], indirect=True)
+def test_env_spdlog_overhead_regression(tmp_home, tmp_root_prefix, tmp_path):
```
JohanMabille, line 147: xeus-cling is not available on Windows and I guess the test on OSX is actually done for ARM64 arch, for which xeus-cling is not available either. We should probably limit this test to Linux only.

JohanMabille (commented):
the failure of `test_remove_orphaned` in this PR is weird, we should confirm that something changed in the dependencies of xtenor-python by running the CI on an empty commit. If  it does not fail, then the test can be considered as flaky since no change in this PR is related to this.

libmamba/src/solver/libsolv/database.cpp
```diff
@@ -109,7 +109,13 @@ namespace mamba::solver::libsolv
...
     void Database::set_logger(logger_type callback)
     {
-        ::pool_setdebuglevel(pool().raw(), std::numeric_limits<int>::max());  // All
+        // We use the penultimate level of verbosity to avoid the most verbose messages
+        // (of type SOLV_DEBUG_RULE_CREATION | SOLV_DEBUG_WATCHES), which might spam the
+        // output and make mamba hang.
+        // See:
+        // https://github.com/openSUSE/libsolv/blob/27aa6a72c7db73d78aa711ae412231768e77c9e0/src/pool.c#L1623-L1637
+        int penultimate_level = 3;
+        ::pool_setdebuglevel(pool().raw(), penultimate_level);
```
Hind-M, line 118: Why don't we do the same as in [1.x](https://github.com/mamba-org/mamba/blob/1.x/libmamba/src/core/pool.cpp#L72)?
Gives more control to the user that way?

micromamba/tests/test_create.py
```diff
@@ -143,6 +143,23 @@ def test_env_lockfile_different_install_after_create(tmp_home, tmp_root_prefix,
...
+    # Non-regression test
+
+    env_prefix = tmp_path / "myenv"
+    create_spec_file = tmp_path / "env-spdlog-callback.yaml"
+
+    shutil.copyfile(__this_dir__ / "env-spdlog-callback.yaml", create_spec_file)
+
+    # Must not hang and complete successfully
+    res = helpers.create("-p", env_prefix, "-f", create_spec_file, "-y", "--json")
+    assert res["success"]
```
Hind-M, line 160: I think that this will succeed at the end, the issue (failure) should be if it takes an abnormal amount of time to finish. So maybe add an assert on the execution time of this test to not exceed a certain limit? (The limit value would be defined based on the local tests that you had for example).

libmamba/src/solver/libsolv/database.cpp
jjerphan: Indeed, it would be much better!

micromamba/tests/test_create.py
jjerphan: That's a good remark. I guess we could use `pytest-timeout` for this so that the test automatically fails after a given period. What do you think?

micromamba/tests/test_create.py
```diff
@@ -143,6 +143,24 @@ def test_env_lockfile_different_install_after_create(tmp_home, tmp_root_prefix,
     helpers.install("-p", env_prefix, "-f", install_spec_file, "-y", "--json")
 
 
+# Only run this test on Linux, as it is the only platform where xeus-cling
+# (which is part of the environment) is available.
+@pytest.timeout(30)
```
Hind-M, line 148: I think it's:
```suggestion
@pytest.mark.timeout(30)
```
But isn't 30 seconds still a lot? (just checking here ^^)

Hind-M: Hum, and is it hanging without the fix even with dry run? (If not the test should be live run).
I would still say that we should have the test actually executed... (my previous question was more asking for confirmation, not really asking for a change).
(And btw, and generally speaking, you can still keep the `assert` even with dry run IINM).

libmamba/src/solver/libsolv/database.cpp
```diff
@@ -15,6 +15,7 @@
 #include <solv/solver.h>
 #include <spdlog/spdlog.h>
 
+#include "mamba/core/context.hpp"
```
Hind-M, line 18: Yeah well that's what we want to avoid, except `core` and `api`, all the other files are context free...

micromamba/tests/test_create.py
jjerphan: > Hum, and is it hanging without the fix even with dry run?

It is not hanging with and without the dry run. It is just that we need nothing else than the dry-run for this test, this would also make it shorter.

I will reintroduce the `assert`.

JohanMabille (approved):

micromamba/tests/test_create.py
```diff
@@ -143,6 +143,24 @@ def test_env_lockfile_different_install_after_create(tmp_home, tmp_root_prefix,
     helpers.install("-p", env_prefix, "-f", install_spec_file, "-y", "--json")
 
 
+# Only run this test on Linux, as it is the only platform where xeus-cling
+# (which is part of the environment) is available.
+# @pytest.mark.timeout(60)
```
jjerphan, line 148: I do not currently understand why this timeouts on the CI, whereas my manual test takes 10 seconds locally.

micromamba/tests/test_create.py
jjerphan: > is if it's hanging with the dry run without the fix

`micromamba` was observed to block during the resolution within `libsolv`. Dry-runs like live-runs (run `libsolv`'s resolution) but just do not perform the transaction then.

I just have rechecked manually and the issue also happens for dry-runs without the fix; why would it not hang in this case?

micromamba/tests/test_remove.py
```diff
@@ -47,10 +47,7 @@ def test_remove_orphaned(tmp_home, tmp_root_prefix, tmp_xtensor_env, tmp_env_nam
     assert keys.issubset(set(res.keys()))
     assert res["success"]
 
-    if sys.platform == "darwin" and platform.machine() == "arm64":
-        assert len(res["actions"]["UNLINK"]) == 12
-    else:
-        assert len(res["actions"]["UNLINK"]) == 11
+    assert len(res["actions"]["UNLINK"]) == 12
```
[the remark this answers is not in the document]
jjerphan: I just have opened https://github.com/mamba-org/mamba/pull/3417 accordingly.

micromamba/tests/test_create.py
Hind-M: Yes it should, thus the "check" in case we are missing something....
Thanks for checking!

libmamba/src/solver/libsolv/database.cpp
```diff
@@ -109,7 +109,14 @@ namespace mamba::solver::libsolv
...
     {
-        ::pool_setdebuglevel(pool().raw(), std::numeric_limits<int>::max());  // All
+        // We guard against the last level of verbosity to avoid the most verbose messages
+        // (of type SOLV_DEBUG_RULE_CREATION | SOLV_DEBUG_WATCHES), which might spam the
+        // output and make mamba hang.
+        // See:
+        // https://github.com/openSUSE/libsolv/blob/27aa6a72c7db73d78aa711ae412231768e77c9e0/src/pool.c#L1623-L1637
+        // TODO: Make `level` configurable once the semantics and UX for verbosity are clarified.
+        int level = 3;  // Context().output_params.verbosity - 1;
+        ::pool_setdebuglevel(pool().raw(), std::min(level, 3));
```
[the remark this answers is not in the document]
jjerphan: In that case, let me make the comment above even more explicit.

micromamba/tests/test_create.py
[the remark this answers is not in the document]
jjerphan: See https://github.com/mamba-org/mamba/pull/3416/files#r1735783066.

Hind-M: And how much time does it take?

libmamba/src/solver/libsolv/database.cpp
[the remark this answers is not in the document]
Hind-M: Well I would say remove it and add it when using verbosity or whatever equivalent, because it doesn't make sense here.

micromamba/tests/test_create.py
jjerphan: Let me check when running the Python test suite.

libmamba/src/solver/libsolv/database.cpp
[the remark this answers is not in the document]
jjerphan: Does 0d8ea53 make it clearer?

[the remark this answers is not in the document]
Hind-M: Yeah definitely better thanks!

libmamba/src/solver/libsolv/database.cpp
```diff
@@ -109,7 +109,13 @@ namespace mamba::solver::libsolv
 
     void Database::set_logger(logger_type callback)
     {
-        ::pool_setdebuglevel(pool().raw(), std::numeric_limits<int>::max());  // All
+        // We use the penultimate level of verbosity of libsolv (which is 3) to avoid the most
+        // verbose messages (of type SOLV_DEBUG_RULE_CREATION | SOLV_DEBUG_WATCHES), which might
+        // spam the output and make mamba hang. See:
+        // https://github.com/openSUSE/libsolv/blob/27aa6a72c7db73d78aa711ae412231768e77c9e0/src/pool.c#L1623-L1637
+        // TODO: Make `level` configurable once the semantics and UX for verbosity are clarified.
+        int level = 3;  // Context().output_params.verbosity - 1;
```
Hind-M, line 117: Ah maybe add a `const` here.

jjerphan:
```suggestion
        const int level = 3;  // Context().output_params.verbosity - 1;
```

micromamba/tests/test_create.py
jjerphan: Honestly, let's just ship?

Hind-M: Why? The test here doesn't do things the right way since we need to check that it doesn't timeout.
If we skip this, I don't see the point of adding it.
And I would rather spend more time doing things the right way than just ship and then still have the same problems again later...

jjerphan: It actually took 93.58s.

Hind-M: Hum, and when it hangs (without the fix) how much time does it take (with dry run)?

jjerphan: It hangs forever (several hours).

If I set the level to 0 (which entirely removes logs), it takes less than 20 seconds.

Hind-M: Ok so let's set 100 for the timeout.

jjerphan: Yes, let me just look at the value of `verbosity` before to know whether the value passed as `level` was 3 or something less as it could still introduce a performance regression.

libmamba/src/solver/libsolv/database.cpp
```diff
@@ -109,7 +109,19 @@ namespace mamba::solver::libsolv
...
+        // https://github.com/openSUSE/libsolv/blob/27aa6a72c7db73d78aa711ae412231768e77c9e0/src/pool.c#L1623-L1637
+        // TODO: Make `level` configurable once the semantics and UX for verbosity are clarified.
+        // Currently, we use the behavior of `1.x` whose default value for the verbosity level was
+        // `0` in which case the logs of libsolv were not transferred. See:
+        // https://github.com/mamba-org/mamba/blob/4f269258b4237a342da3e9891045cdd51debb27c/libmamba/include/mamba/core/context.hpp#L88
+        // See: https://github.com/mamba-org/mamba/blob/1.x/libmamba/src/core/pool.cpp#L72
+        // Instead use something like:
+        // const int level = Context().output_params.verbosity - 1;
+        const int level = 0;
+        ::pool_setdebuglevel(pool().raw(), level);
```
Hind-M, line 124: In 1.x we were doing:
`pool_setdebuglevel(pool().raw(), Context::instance().output_params.verbosity - 1);`
when `verbosity` >= 3
Otherwise `pool_setdebuglevel` wasn't called. That means that it was set to something in libsolv internals not to the default value of `verbosity` in `mamba`.

jjerphan: [The default value of `libsolv`'s `pool->debugmask` is `SOLV_DEBUG_RESULT`](https://github.com/openSUSE/libsolv/blob/27aa6a72c7db73d78aa711ae412231768e77c9e0/src/pool.c#L87).

If we call it with `0` we only `&=SOLV_DEBUG_TO_STDERR` it. Not sure if it causes a problem in practice, but let's thus not call it for now at all.

Hind-M (approved):

Hind-M:
The added test was skipped here, so we should have waited for #3417 being merged before merging this...

Issue #1455, opened 2022-02-08
Implement `micromamba env update`


jonashaag:
It would be nice to have a way in Micromamba to update an env from an `environment.yml`. In Mamba this works by `mamba env update -n X -f environment.yml`.

wolfv:
I think we can use the current `micromamba update -f myenv.yml` for that.

We can add a command line option to `--prune-to-specs`.

jonashaag:
`micro mamba update -f myenv.yml` seems to ignore `myenv.yml`.

wolfv:
I am realizing that maybe that is the desired behavior and `micromamba install` is all we should ever need?

Let's picture some scenarios -- assuming we have a created environment. What should the ideal command line look like. We could quite easily add the following flags:

- `micromamba install -f env.yml` -> installs new packages to specs
- `micromamba install -f env.yml --update` -> installs & updates new packages according to `env.yml`
- `micromamba install -f env.yml --update --prune-to-specs` -> installs & updates new packages to specs, removes packages not in the environment but not in `env.yml`

jonashaag:
Yes that sounds reasonable as well. What would be the role of `micromamba update` in that case? Is it just a shorthand to `install --update`?

wolfv:
When using `install --update` we could use the `SOLVER_FLAG_INSTALL_ALSO_UPDATES` flag. It would be different from `update` because `update` doesn't install any new packages.

https://github.com/mamba-org/mamba/issues/1317

CDTR-MattConroy:
Is there any intention to add a `--prune` option to be able to remove packages that are not specified in an environment file?

shenker:
My understanding is that `mamba update -n myenv -f environment.yml --prune` will install new packages, update existing packages, and remove extraneous packages so that `myenv` ends up in the same state as it would be if you did `mamba env remove -n myenv && mamba create -n myenv -f environment.yml`.

It would be good for an analogous command to exist in micromamba. It would be especially convenient if this command created the environment if it didn't already exist. (E.g., after you run this command, you're guaranteed to end up with `myenv` matching the given `environment.yml` file regardless of the initial state or existence of `myenv`, and the command would get you there in the most efficient way.)

goyalyashpal:
> --prune-to-specs
\- @ wolfv at https://github.com/mamba-org/mamba/issues/1455#issuecomment-1032754684

that is very much needed 😃 
the issue with `env create` is that it doesn't use cache and has to download all the packages again.

[...]

PR #3419, opened 2024-08-29
fix: PyPI support for `env update`

jjerphan:
Completes #1455.

This factors the logic of the call to `pip` for the install and update path.

Note that previously the output of `pip install` was redirected up to the `libmambapy` helpers' outputs, breaking JSON outputs when `--json` is specify. The logging level aren't managing the standard output of subprocess for now, and doing this properly probably should  be done in another PR. For now, we simply call `pip` in quiet mode.

### DIFF
--- a/libmamba/CMakeLists.txt
+++ b/libmamba/CMakeLists.txt
@@ -258,6 +258,7 @@ set(
     ${LIBMAMBA_SOURCE_DIR}/api/install.cpp
     ${LIBMAMBA_SOURCE_DIR}/api/list.cpp
     ${LIBMAMBA_SOURCE_DIR}/api/pip_utils.cpp
+    ${LIBMAMBA_SOURCE_DIR}/api/pip_utils.hpp
     ${LIBMAMBA_SOURCE_DIR}/api/remove.cpp
     ${LIBMAMBA_SOURCE_DIR}/api/repoquery.cpp
     ${LIBMAMBA_SOURCE_DIR}/api/shell.cpp
--- a/libmamba/CMakeLists.txt
+++ b/libmamba/CMakeLists.txt
@@ -257,6 +257,7 @@ set(
     ${LIBMAMBA_SOURCE_DIR}/api/info.cpp
     ${LIBMAMBA_SOURCE_DIR}/api/install.cpp
     ${LIBMAMBA_SOURCE_DIR}/api/list.cpp
+    ${LIBMAMBA_SOURCE_DIR}/api/pip_utils.cpp
     ${LIBMAMBA_SOURCE_DIR}/api/remove.cpp
     ${LIBMAMBA_SOURCE_DIR}/api/repoquery.cpp
     ${LIBMAMBA_SOURCE_DIR}/api/shell.cpp
--- a/libmamba/src/api/pip_utils.cpp
+++ b/libmamba/src/api/pip_utils.cpp
@@ -118,6 +118,12 @@ namespace mamba
         const auto& deps = other_spec.deps;
         const auto& cwd = other_spec.cwd;
 
+        LOG_WARNING << fmt::format(
+            "You are using '{}' as an additional package manager.\nBe aware that packages installed with '{}' are managed independently from 'conda-forge' channel.",
+            pkg_mgr,
+            pkg_mgr
+        );
+
         TemporaryFile specs("mambaf", "", cwd);
         {
             std::ofstream specs_f = open_ofstream(specs.path());
--- a/libmamba/src/api/pip_utils.cpp
+++ b/libmamba/src/api/pip_utils.cpp
@@ -27,21 +27,20 @@ namespace mamba
             bool update
         )
         {
-            const auto get_python_path = [&]
-            { return util::which_in("python", get_path_dirs(target_prefix)).string(); };
+            const auto python = util::which_in("python", get_path_dirs(target_prefix)).string();
 
             const std::unordered_map<std::string, command_args> pip_install_command{
                 { "pip",
-                  { get_python_path(), "-m", "pip", "install", "-r", spec_file, "--no-input" } },
+                  { python, "-m", "pip", "install", "-r", spec_file, "--no-input", "--quiet" } },
                 { "pip --no-deps",
-                  { get_python_path(), "-m", "pip", "install", "--no-deps", "-r", spec_file, "--no-input" } }
+                  { python, "-m", "pip", "install", "--no-deps", "-r", spec_file, "--no-input", "--quiet" } }
             };
 
             const std::unordered_map<std::string, command_args> pip_update_command{
                 { "pip",
-                  { get_python_path(), "-m", "pip", "install", "-U", "-r", spec_file, "--no-input" } },
+                  { python, "-m", "pip", "install", "-U", "-r", spec_file, "--no-input", "--quiet" } },
                 { "pip --no-deps",
-                  { get_python_path(), "-m", "pip", "install", "-U", "--no-deps", "-r", spec_file, "--no-input" } }
+                  { python, "-m", "pip", "install", "-U", "--no-deps", "-r", spec_file, "--no-input", "--quiet" } }
             };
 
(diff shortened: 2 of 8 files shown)


libmamba/CMakeLists.txt
```diff
@@ -257,6 +257,7 @@ set(
     ${LIBMAMBA_SOURCE_DIR}/api/info.cpp
     ${LIBMAMBA_SOURCE_DIR}/api/install.cpp
     ${LIBMAMBA_SOURCE_DIR}/api/list.cpp
+    ${LIBMAMBA_SOURCE_DIR}/api/pip_utils.cpp
```
JohanMabille, line 260: I think you should add `api/pip_utils.hpp` here.

libmamba/src/api/pip_utils.cpp
```diff
@@ -0,0 +1,166 @@
...
+            };
+
+            const std::unordered_map<std::string, command_args> pip_update_command{
+                { "pip",
+                  { python, "-m", "pip", "install", "-U", "-r", spec_file, "--no-input", "--quiet" } },
+                { "pip --no-deps",
+                  { python, "-m", "pip", "install", "-U", "--no-deps", "-r", spec_file, "--no-input", "--quiet" } }
+            };
+
+            auto found_it = update ? pip_update_command.find(name) : pip_install_command.find(name);
```
JohanMabille, line 46: I wonder if something like:
```cpp
const std::unordered_map<std::string, command_arg> pip_command{
    { "pip",
      { python, "-m", "pip", "install", (update ? ("-U", "-r") : "-r"), spec_file, "--no-input", "--quiet" } }, 
      // ...
};

auto found_it = pip_command.find(name);
```

would work. I t avoids some code duplication and the creation of two maps.

libmamba/src/api/pip_utils.cpp
```diff
@@ -0,0 +1,174 @@
...
+#include <reproc/reproc.h>
+
+#include "mamba/api/install.hpp"
+#include "mamba/core/activation.hpp"
+#include "mamba/core/context.hpp"
+#include "mamba/core/util.hpp"
+#include "mamba/fs/filesystem.hpp"
+#include "mamba/util/environment.hpp"
+
+#include "tl/expected.hpp"
```
Hind-M, line 15: Should be before the mamba files includes.

micromamba/tests/test_create.py
```diff
@@ -161,6 +161,55 @@ def test_env_logging_overhead_regression(tmp_home, tmp_root_prefix, tmp_path):
...
 
+@pytest.mark.parametrize("shared_pkgs_dirs", [True], indirect=True)
+def test_env_create_pypi(tmp_home, tmp_root_prefix, tmp_path):
+    env_prefix = tmp_path / "myenv"
+
+    create_spec_file = tmp_path / "env-pypi-pkg-test.yaml"
+
+    shutil.copyfile(__this_dir__ / "env-pypi-pkg-test.yaml", create_spec_file)
+
+    res = helpers.create("-p", env_prefix, "-f", create_spec_file, "-y", "--json")
```
Hind-M, line 172: The tests should be in `test_env.py` instead because we are testing the subcommands of the env command.
That being said, I'm wondering if this is testing the right thing, I think we should use `helpers.run_env(*cmd)` instead, see `test_env_update` for reference.

micromamba/tests/test_create.py
```diff
@@ -161,6 +161,55 @@ def test_env_logging_overhead_regression(tmp_home, tmp_root_prefix, tmp_path):
...
+def test_env_create_update_pypi(tmp_home, tmp_root_prefix, tmp_path):
+    env_prefix = tmp_path / "myenv"
+
+    create_spec_file = tmp_path / "env-requires-pip-install.yaml"
+    update_spec_file = tmp_path / "env-pypi-pkg-test.yaml"
+
+    shutil.copyfile(__this_dir__ / "env-requires-pip-install.yaml", create_spec_file)
+    shutil.copyfile(__this_dir__ / "env-pypi-pkg-test.yaml", update_spec_file)
+
+    res = helpers.create("-p", env_prefix, "-f", create_spec_file, "-y", "--json")
```
Hind-M, line 192: Same here (use `run_env`).

micromamba/tests/test_create.py
```diff
@@ -161,6 +161,55 @@ def test_env_logging_overhead_regression(tmp_home, tmp_root_prefix, tmp_path):
...
+
+    # Check that pypi-pkg-test is not installed before the update
+    pip_list_output = helpers.umamba_run("-p", env_prefix, "pip", "list", "--format=json")
+    pip_packages_list = yaml.safe_load(pip_list_output)
+    assert all(pkg["name"] != "pypi-pkg-test" for pkg in pip_packages_list)
+
+    umamba_packages_list = helpers.umamba_list("-p", env_prefix, "--json")
+    assert all(pkg["name"] != "pypi-pkg-test" for pkg in umamba_packages_list)
+
+    res = helpers.update("-p", env_prefix, "-f", update_spec_file, "-y", "--json")
```
Hind-M, line 203: Use `run_env`.

Hind-M:
Can we also add tests to make sure we are behaving like `conda env update` when mixing conda-forge and pip packages in both orders?

libmamba/src/api/pip_utils.cpp
```diff
@@ -0,0 +1,174 @@
...
+    namespace
+    {
+        tl::expected<command_args, std::runtime_error> get_pip_install_command(
+            const std::string& name,
+            const std::string& target_prefix,
+            const fs::u8path& spec_file,
+            bool update
+        )
+        {
+            const auto python = util::which_in("python", get_path_dirs(target_prefix)).string();
```
Hind-M, line 30: Btw, I think we should restore the previous lambda function here, as it should take into consideration the changes in the `target_prefix` and compute the python path on demand.

micromamba/tests/test_env.py
```diff
@@ -200,3 +201,64 @@ def test_explicit_export_topologically_sorted(tmp_home, tmp_prefix):
...
+    assert res["success"]
+
+    # Check that pypi-pkg-test (that is part of the update environment) is installed after the update
+    # See: https://github.com/mamba-org/mamba/issues/2059
+    pip_list_output = helpers.umamba_run("-p", env_prefix, "pip", "list", "--format=json")
+    packages_list = yaml.safe_load(pip_list_output)
+    assert any(pkg["name"] == "pypi-pkg-test" for pkg in packages_list)
+
+    # Create a new environment with the same spec file but with conda within the shell
+    subprocess.run(f"conda update -p {update_spec_file} -f {create_spec_file} -y")
```
Hind-M, line 256: I think using `conda` (even for tests) is a very bad idea. We can just know how it should behave and check the result.
This also looks like mixing updates with `conda` and `mamba` and it shouldn't.
The error in the CI seems to be caused by setting the target prefix as a yaml file` -p {update_spec_file}` ?

libmamba/src/api/pip_utils.cpp
```diff
@@ -35,7 +35,7 @@ namespace mamba
                     "-m",
                     "pip",
                     "install",
-                    (update ? ("-U", "-r") : "-r"),
+                    (update ? "-U", "-r" : "-r"),
```
Hind-M, line 38: Why remove parentheses? They look much safer...

JohanMabille (approved):

SylvainCorlay:
🎉

azjps:
> For now, we simply call `pip` in quiet mode.

Hi, I can see why `pip`'s output is problematic here, but do you think we could switch whether to use `[uv] pip install --quiet` depending on the context? For example, passing `--quiet` if `context.output_params.json || context.output_params.quiet`. 

We check the output from `pip` (and ideally `uv` now) to see whether it has unintentionally overridden any packages installed by `mamba`, but it seems like after this change, its not possible to see this information or any other diagnostics from `pip/uv` anymore. We can still find ways to figure out this information but would be nice if we could see this a bit more easily.

Happy to open a new issue or PR if the above sounds reasonable.

PR #3434, opened 2024-09-07
fix: add warning when using defaults

wolfv:
Following all the drama about Anaconda's aggressive behavior towards users who are using the defaults channel (discussed at the NumFOCUS summit) we came up with the plan to _at least_ add a warning.

I had a quick go at adding a warning when using defaults.

We should discuss what the UX should be like:

- A warning that can be disabled either via ENV var or via `.condarc` setting
- We could also go one step further and have a `y/n` prompt, after which we would append a value to the `.condarc` file.

Which option would you all prefer?


wolfv:
cf https://github.com/conda/conda/issues/14217

SylvainCorlay:
I missed the conversation at the summit, so I cannot comment on that.

I think that it makes sense for us to help mamba users, and warning them that using certain channels may incur costs as per their TOS makes sense IMO.

However, conda is a client of this library and we should probably make sure that they can disable this warning if they decided to.

libmamba/src/api/channel_loader.cpp
```diff
@@ -63,6 +63,10 @@ namespace mamba
         {
             for (const auto& platform : channel.platforms())
             {
+                if (channel.platform_url(platform).host() == "repo.anaconda.com") {
+                    spdlog::warn("This is a commercial channel hosted by Anaconda.com. Please make sure you understand the TOS.");
```
Hind-M, line 67: I think we can use `LOG_WARNING` instead and avoid including `<iostream>`.

Hind-M:
> We should discuss what the UX should be like:
> 
>     * A warning that can be disabled either via ENV var or via `.condarc` setting
> 
>     * We could also go one step further and have a `y/n` prompt, after which we would append a value to the `.condarc` file.
> Which option would you all prefer?

I think having something via the `.condarc` would be enough.
The `y/n` prompt seems to be a good idea because that makes the user actually read the message, but I'm not sure how this could interfere with the tests and people's workflow, and how much it could be annoying...

jjerphan:
The `y/n` prompt might be blocking if called in user scripts; so I guess the first option is the best.

JohanMabille (commented):
We still need a way to deactivate this warning. I would go for a full configurable option (rc, env and public API exposed to Python) so that clients can choose how they disable it.

jjerphan:
> We still need a way to deactivate this warning. I would go for a full configurable option (rc, env and public API exposed to Python) so that clients can choose how they disable it.

How would you best do that?

libmamba/src/api/channel_loader.cpp
```diff
@@ -3,6 +3,7 @@
 // Distributed under the terms of the BSD 3-Clause License.
 //
 // The full license is in the file LICENSE, distributed with this software.
+#include <iostream>
```
JohanMabille, line 6: Can we remove this include now that we use `LOG_WARNING`?

JohanMabille:
Let's remove the test for now, there are some issues with the capture and redirection of stdout and stderr that are orthogonal to this PR.

SylvainCorlay (approved):

JohanMabille (approved):

jjerphan:
See https://github.com/mamba-org/mamba/issues/3440.

PR #3435, opened 2024-09-09
Add fallback to root prefix

Hind-M:


libmamba/src/api/remove.cpp
```diff
@@ -27,6 +27,7 @@ namespace mamba
         bool remove_all = flags & MAMBA_REMOVE_ALL;
 
         config.at("use_target_prefix_fallback").set_value(true);
+        config.at("use_root_prefix_fallback").set_value(false);
```
Hind-M, line 30: I set the flag to `false` here since it's related to environment removal and could be dangerous (so no fallback in this case).

SylvainCorlay:
Wow, all green! 😄

JohanMabille (approved):

PR #3441, opened 2024-09-12
[mamba package] Set root prefix to mamba installation prefix

### LINKED ISSUE #3426: Mamba 2.0 - installing in the base env without activation
With Conda (and Mamba 1.x), running `conda install foobar` or `mamba install foobar` will install `foobar` in the base environment when no environment is activated. However, micromamba and mamba 2.0 will complain that no prefix is specified.

This is the root cause for the current failure when [updating repo2docker to the latest rc of mamba 2.0](https://github.com/jupyterhub/repo2docker). 

I believe that we should support this behavior, as it will generally be expected from conda users who don't use environments much.

A possible approach would be to

1. Set up the mamba package so that the root prefix is set to the installation prefix of mamba.
2. Always default to the root prefix when installing a package (either for both micromamba and mamba, or just for mamba, behind a flag).

Hind-M:
Fix #3426


Hind-M:
@SylvainCorlay Not sure if this is what you had in mind, but it's a way of doing things.
I'm currently trying to see if there could be any obvious conflict with a set `conda` base.

SylvainCorlay:
This profile.d script is not run by default unfortunately.

I think this should be done in the c++.

Behind a compilation flag (only enabled in mamba, but not micromamba), bypass the root_prefix computation with a result basedbon the executable path.

Hind-M:
Hmm yeah that's another way of doing it. I figured that was more something that we want to have in the script when running it (since it's related to mamba package) which will avoid having specific cases in the codebase.
Is this supposed to be related to only mamba package or any mamba in general?

SylvainCorlay:
> Is this supposed to be related to only mamba package or any mamba in general?

I would say any mamba that is installed in a conda environment.

Hind-M:
Superseded by #3447

PR #3447, opened 2024-09-16
Compute `root prefix` as mamba install path

### LINKED ISSUE #3426: Mamba 2.0 - installing in the base env without activation

Hind-M:
Fix #3426


libmamba/src/api/configuration.cpp
```diff
@@ -718,6 +781,17 @@ namespace mamba
                     validate_existing_root_prefix(default_root_prefix_v1())
                         .or_else([](const auto& /* error */)
                                  { return validate_root_prefix(default_root_prefix_v2()); })
```
Hind-M, line 783: This is actually making sure that the root prefix will be set to `fs::u8path(util::user_data_dir()) / "mamba"` (which is in most cases something like `/home/user/.local/share/mamba`)  if it doesn't already exist.
It it exists, it checks that it's a directory containing at least `pkgs`, `conda-meta` or `envs` directories.
So mamba will only complain if that directory exists and doesn't contain one of the mentioned dirs (which I don't think happens a lot...).
In this case, we set the root prefix to the mamba installation root prefix.

libmamba/CMakeLists.txt
```diff
@@ -76,6 +76,11 @@ if(BUILD_STATIC)
     add_definitions(-DLIBMAMBA_STATIC_DEPS)
 endif()
 
+# Use mamba installation prefix to set root prefix (base)
+if(BUILD_SHARED OR BUILD_MAMBA)
+    add_definitions(-DMAMBA_USE_INSTALL_PREFIX_AS_BASE)
+endif()
```
SylvainCorlay, line 82: Thinking again about this. I think the definition should be passed to the target for the shared library, and not for the static library.

What do you think?

SylvainCorlay (approved):

PR #3448, opened 2024-09-16
Fix test in osx

Hind-M:


JohanMabille:
We may want to either be able to set one of these two variables depending on what we want to test locally, or to skip this test for micromamba, but let's do that later.

Hind-M:
> We may want to either be able to set one of these two variables depending on what we want to test locally, or to skip this test for micromamba, but let's do that later.

Hum, not sure this is doable in a clean way, since we use `TEST_MAMBA_EXE` as the executable, but that doesn't give a clear info about which one is used (`mamba` or `micromamba`) unless we check that the filename in `TEST_MAMBA_EXE`  is `mamba` or not?...

JohanMabille (approved):

Issue #3453, opened 2024-09-17
Micromamba 2.0rc5 - libmamba Error parsing version ""


bertaveira:
### Troubleshooting docs

- [X] My problem is not solved in the Troubleshooting docs

### Anaconda default channels

- [X] I do NOT use the Anaconda default channels (pkgs/* etc.)

### How did you install Mamba?

Other (please describe)

### Search tried in issue tracker

error parsing version

### Latest version of Mamba

- [X] My problem is not solved with the latest version

### Tried in Conda?

Not applicable

### Describe your issue

I tried the latest rc of micromamba 2.0rc5 as well as the latest master build #af81975 for windows. I noticed it did not work to create an environment from a yaml spec file because it would not accept versions of packages where it would request greater or lower than a certain version. If I make it greater or equal then it works.

Example of a YAML file that fails:
```
channels:
  - conda-forge
dependencies:
  - python > 3.11
  - numpy < 2.0
 ```
With this file if one would create an environment using `micromamba create -f .\environment.yaml -p .\.env` you get this fault:
```
critical libmamba Error parsing version "". Empty version.
```

One can make it not throw that error and create an environment just by changing the versions to this:
```
channels:
  - conda-forge
dependencies:
  - python >= 3.11
  - numpy <= 2.0
 ```

### environment.yml

```yaml
channels:
  - conda-forge
dependencies:
  - python > 3.11
  - numpy < 2.0
```

jjerphan:
I confirm that this is a regression for `micromamba`.

PR #3456, opened 2024-09-17
fix: Handle extra white-space in `MatchSpec`

jjerphan:
Fix https://github.com/mamba-org/mamba/issues/3453 which is just one of the existing issues with `MatchSpec`.


micromamba/tests/test_env.py
```diff
@@ -363,3 +364,44 @@ def test_env_update_pypi_with_conda_forge(tmp_home, tmp_root_prefix, tmp_path):
     pip_packages_list = yaml.safe_load(pip_list_output)
 
     assert any(pkg["name"] == "numpy" and pkg["version"] == "1.26.4" for pkg in pip_packages_list)
+
+
+@pytest.mark.parametrize("shared_pkgs_dirs", [True], indirect=True)
+def test_env_create_whitespace(tmp_home, tmp_root_prefix, tmp_path):
```
Hind-M, line 370: I don't think we need this test, tests in `test_match_spec` are enough to make sure the parsing is done correctly.
Adding this one is redundant and is not relevant in my opinion.

libmamba/tests/src/specs/test_match_spec.cpp
```diff
@@ -47,6 +47,76 @@ TEST_SUITE("specs::match_spec")
...
 
+        SUBCASE("xtensor      >=       0.12.3")
+        {
+            auto ms = MatchSpec::parse("xtensor      >=       0.12.3").value();
+            CHECK_EQ(ms.name().str(), "xtensor");
+            CHECK_EQ(ms.version().str(), ">=0.12.3");
+            CHECK(ms.build_string().is_explicitly_free());
+            CHECK(ms.build_number().is_explicitly_free());
+            CHECK_EQ(ms.str(), "xtensor>=0.12.3");
+        }
```
Hind-M, line 58: This is the exact same test as the one above.

libmamba/src/specs/match_spec.cpp
```diff
@@ -502,32 +502,58 @@ namespace mamba::specs
...
+            const std::string& bad_op = op + " ";
+            while (raw_match_spec_str.find(bad_op) != std::string::npos)
+            {
+                raw_match_spec_str = raw_match_spec_str.substr(0, raw_match_spec_str.find(bad_op)) + op
+                                     + raw_match_spec_str.substr(
+                                         raw_match_spec_str.find(bad_op) + bad_op.size()
+                                     );
+            }
+        }
+
```
Hind-M, line 530: Maybe we won't have the choice and we will need to fix this this way, but I really think that we should do this properly and stop postponing everything to later, because it will just increase the complexity (making it harder to change things afterwards), especially regarding the `MatchSpec`...
IIRC this should be rather handled [here](https://github.com/mamba-org/mamba/blob/main/libmamba/src/specs/match_spec.cpp#L441-L488), so we need to adapt the logic accordingly (and try to keep the `string_view`).

micromamba/tests/test_env.py
jjerphan: I agree with this, I just wanted to have a non-regression test for the reported issue.

libmamba/src/specs/match_spec.cpp
jjerphan: I agree with your proposal although there is some complexity with the low level parser: if we want to handle all the current cases while excluding the [PEP 508 environment markers](https://peps.python.org/pep-0508/#environment-markers), I am afraid that the amount of complexity to manage will be far more complex that this horrible yet working and short solution.

In my opinion, the long-term robust solution (as discussed in the past) is to define a grammar for `MatchSpec` and use lexers to generate parsers in applications. But this goes far beyond the scope of this PR or the time we have at hand.

micromamba/tests/env-extra-white-space.yaml
```diff
@@ -0,0 +1,8 @@
+channels:
+  - conda-forge
+dependencies:
+  - python > 3.11
+  - numpy< 2.0
```
jjerphan, line 5:
```suggestion
  - numpy < 2.0
```

micromamba/tests/test_env.py
```diff
@@ -370,3 +371,44 @@ def test_env_update_pypi_with_conda_forge(tmp_home, tmp_root_prefix, tmp_path):
...
+    res = helpers.run_env("create", "-p", env_prefix, "-f", create_spec_file, "-y", "--json")
+    assert res["success"]
+
+    # Check that the env was created
+    assert env_prefix.exists()
+    # Check that the env has the right packages
+    packages = helpers.umamba_list("-p", env_prefix, "--json")
+
+    assert any(
+        package["name"] == "python" and Version(package["version"]) >= Version("3.11")
```
Hind-M, line 395:
```suggestion
        package["name"] == "python" and Version(package["version"]) > Version("3.11")
```

libmamba/tests/src/specs/test_match_spec.cpp
```diff
@@ -47,6 +47,66 @@ TEST_SUITE("specs::match_spec")
...
+        {
+            auto ms = MatchSpec::parse("python > 3.11").value();
+            CHECK_EQ(ms.name().str(), "python");
+            CHECK_EQ(ms.version().str(), ">3.11");
+            CHECK(ms.build_string().is_explicitly_free());
+            CHECK(ms.build_number().is_explicitly_free());
+            CHECK_EQ(ms.str(), "python>3.11");
+        }
+
+        SUBCASE("numpy< 2.0")
```
Hind-M, line 60:
```suggestion
        SUBCASE("numpy < 2.0")
```

libmamba/tests/src/specs/test_match_spec.cpp
```diff
@@ -47,6 +47,66 @@ TEST_SUITE("specs::match_spec")
...
+            CHECK_EQ(ms.name().str(), "python");
+            CHECK_EQ(ms.version().str(), ">3.11");
+            CHECK(ms.build_string().is_explicitly_free());
+            CHECK(ms.build_number().is_explicitly_free());
+            CHECK_EQ(ms.str(), "python>3.11");
+        }
+
+        SUBCASE("numpy < 2.0")
+        {
+            auto ms = MatchSpec::parse("numpy< 2.0").value();
```
Hind-M, line 62: I mean the test should be modified as well:
```suggestion
            auto ms = MatchSpec::parse("numpy < 2.0").value();
```

SylvainCorlay (approved):

Issue #3472, opened 2024-09-26
Micromamba v2.0 unable to parse environment.yml


breakbotz:
### Troubleshooting docs

- [X] My problem is not solved in the Troubleshooting docs

### Anaconda default channels

- [X] I do NOT use the Anaconda default channels (pkgs/* etc.)

### How did you install Mamba?

Micromamba

### Search tried in issue tracker

Creating issue as recommended in https://github.com/mamba-org/micromamba-releases/issues/58

### Latest version of Mamba

- [X] My problem is not solved with the latest version

### Tried in Conda?

Not applicable

### Describe your issue

A unit test in my GitHub Actions pipeline is failing when attempting to create a micromamba environment using an environment.yaml which has not changed in months.

Using the latest version (curl -Ls https://micro.mamba.pm/api/micromamba/linux-64/latest), we run `micromamba env create --yes --file environment.yml` which errors with `libmamba Could not solve for environment specs`. More detailed error logs provided below and full output can be found in my [GHA pipeline run](https://github.com/BCG-X-Official/artkit/actions/runs/11044431820/job/30680379445).

### Logs

```shell
error    libmamba Could not solve for environment specs
    The following packages are incompatible
    ├─ aiohttp ~= does not exist (perhaps a typo or a missing channel);
    ├─ anthropic ~= does not exist (perhaps a typo or a missing channel);
    ├─ fluxus ~= does not exist (perhaps a typo or a missing channel);
    ├─ gamma-pytools ~= does not exist (perhaps a typo or a missing channel);
    ├─ google-generativeai ~= does not exist (perhaps a typo or a missing channel);
    ├─ graphviz ~= does not exist (perhaps a typo or a missing channel);
    ├─ groq ~= does not exist (perhaps a typo or a missing channel);
    ├─ ipywidgets ~= does not exist (perhaps a typo or a missing channel);
    ├─ jupyterlab ~= does not exist (perhaps a typo or a missing channel);
    ├─ nbsphinx ~= does not exist (perhaps a typo or a missing channel);
    ├─ notebook ~= does not exist (perhaps a typo or a missing channel);
    ├─ openai ~= does not exist (perhaps a typo or a missing channel);
    ├─ packaging ~= does not exist (perhaps a typo or a missing channel);
    ├─ pandas ~= does not exist (perhaps a typo or a missing channel);
    ├─ pre-commit ~= does not exist (perhaps a typo or a missing channel);
[...]

PR #3483, opened 2024-09-30
fix: Support for PEP 440 "Compatible Releases" (operator `~=` for `MatchSpec`)

jjerphan:
Fix https://github.com/mamba-org/mamba/issues/3472.

### DIFF
--- a/libmamba/src/specs/match_spec.cpp
+++ b/libmamba/src/specs/match_spec.cpp
@@ -4,6 +4,7 @@
 //
 // The full license is in the file LICENSE, distributed with this software.
 
+#include <iostream>
 #include <string>
 #include <string_view>
 #include <tuple>
@@ -505,6 +506,10 @@ namespace mamba::specs
         std::string raw_match_spec_str = std::string(str);
         raw_match_spec_str = util::strip(raw_match_spec_str);
 
+        // Those are temporary adaptations to handle some instances of `MatchSpec` which is not
+        // yet formally specified.
+        // For a tentative formulation of the MatchSpec see: https://github.com/conda/ceps/pull/82
+
         // Remove any with space after binary operators, such as:
         //  - `openmpi-4.1.4-ha1ae619_102`'s improperly encoded `constrains`: "cudatoolkit >= 10.2"
         //  - `pytorch-1.13.0-cpu_py310h02c325b_0.conda`'s improperly encoded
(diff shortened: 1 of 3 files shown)


libmamba/src/specs/match_spec.cpp
```diff
@@ -528,6 +533,47 @@ namespace mamba::specs
...
+        //
+        //                     ">1.8,<2|==1.7,!=1.9,~=1.7.1 py34_0"
+        //
+        // must be replaced with:
+        //
+        //                     ">1.8,<2|==1.7,!=1.9,>=1.7.1,1.7.* py34_0"
+        //
+        while (raw_match_spec_str.find("~=") != std::string::npos)
+        {
+            std::cout << "raw_match_spec_str: " << raw_match_spec_str << std::endl;
```
JohanMabille, line 557: This looks like a leftover for debugging, it should be removed (or use the LOG macros).

libmamba/src/specs/match_spec.cpp
```diff
@@ -4,6 +4,7 @@
 //
 // The full license is in the file LICENSE, distributed with this software.
 
+#include <iostream>
```
JohanMabille, line 7: This should be removed (used the logging system instead).

libmamba/tests/src/specs/test_match_spec.cpp
```diff
@@ -446,9 +446,28 @@ TEST_SUITE("specs::match_spec")
         {
             auto ms = MatchSpec::parse(R"(numpy >1.8,<2|==1.7,!=1.9,~=1.7.1 py34_0)").value();
```
JohanMabille, line 447: I don't understand why we didn't have any error here since the support for ~= operator was missing? Was it silently ignored? 

NB: this is definitely not a blocker for getting this merged

JohanMabille (changes requested):

libmamba/tests/src/specs/test_match_spec.cpp
jjerphan: It was handled in context, but not for simple cases.

I think that we also need to add a test on `micromamba`.

JohanMabille (approved):

libmamba/tests/src/specs/test_match_spec.cpp
jjerphan: Done in 7056766.

JohanMabille:
The issue is unrelated with the change, the job should be restarted when the workflow is done.

PR #3488, opened 2024-10-01
Fix `channel` and `base_url` in `list` cmd

### LINKED ISSUE #3480: micromamba2: base_url and channel are mangled in `micromamba list --json`
Note the difference below in `base_url` and `channel`. This leads to a failure in the conda-lock CI https://github.com/conda/conda-lock/pull/713

Run
```bash
micromamba list --json
```

A typical entry with micromamba 1.5.10
```json
    {
        "base_url": "https://conda.anaconda.org/conda-forge",
        "build_number": 0,
        "build_string": "ha6fb4c9_0",
        "channel": "conda-forge",
        "dist_name": "zstd-1.5.6-ha6fb4c9_0",
        "name": "zstd",
        "platform": "linux-64",
        "version": "1.5.6"
    }
```

A typical entry with micromamba 2.0.0

```json
    {
        "base_url": "https://conda.anaconda.org/conda-forge/linux-64/zstd-1.5.6-ha6fb4c9_0.conda",
        "build_number": 0,
        "build_string": "ha6fb4c9_0",
        "channel": "conda-forge/linux-64/zstd-1.5.6-ha6fb4c9_0.conda",
        "dist_name": "zstd-1.5.6-ha6fb4c9_0",
        "name": "zstd",
        "platform": "linux-64",
        "version": "1.5.6"
    }
```

I think the output from 1.5.10 of `base_url` and `channel` is correct, while the output from 2.0.0 is incorrect.

Hind-M:
Fix #3480

The channel name used in `1.x` was dropped during the "Great refactoring" (under the assumption, IIUC, that it's the same as `canonical_name`,  which was not always true...) and now that info is stored/computed nowhere.
Since this is only used in the `list` command, it's handled this way for now...

### DIFF
--- a/libmamba/src/api/list.cpp
+++ b/libmamba/src/api/list.cpp
@@ -51,7 +51,21 @@ namespace mamba
             }
         }
 
-        void list_packages(const Context& ctx, std::string regex, ChannelContext& channel_context, list_options options)
+        std::string strip_from_filename_and_platform(
+            const std::string& full_str,
+            const std::string& filename,
+            const std::string& platform
+        )
+        {
+            return rstrip(rstrip(rstrip(rstrip(full_str, filename), "/"), platform), "/");
+        }
+
+        void list_packages(
+            const Context& ctx,
+            std::string regex,
+            ChannelContext& channel_context,
+            list_options options
+        )
         {
             auto sprefix_data = PrefixData::create(ctx.prefix_params.target_prefix, channel_context);
             if (!sprefix_data)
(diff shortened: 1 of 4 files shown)


micromamba/tests/test_create.py
```diff
@@ -1231,9 +1231,8 @@ def test_create_from_oci_mirrored_channels(tmp_home, tmp_root_prefix, tmp_path,
     assert pkg["name"] == "pandoc"
     if spec == "pandoc=3.1.13":
         assert pkg["version"] == "3.1.13"
-    assert pkg["base_url"].startswith(
-        "https://pkg-containers.githubusercontent.com/ghcr1/blobs/pandoc"
-    )
+    assert pkg["base_url"] == "https://pkg-containers.githubusercontent.com/ghcr1/blobs"
+    assert pkg["channel"] == "https://pkg-containers.githubusercontent.com/ghcr1/blobs"
```
Hind-M, line 1235: We could set this to something else like `oci` or `ghcr`?
But that could be done in another PR once it's all sorted out.

libmamba/src/api/list.cpp
```diff
@@ -34,6 +35,31 @@ namespace mamba
             return a.name < b.name;
         }
 
+        // This is more or less an implementation of `util::rstrip` specific to this use case
+        // (for printing purposes), but using `std::string` instead of `std::string_view`
+        // `util::rstrip` is not used here because it leads to an UB,
+        // `using non owned/tracked strings from Channel (& co) and PackageInfo
```
JohanMabille, line 41: Not sure to understand why `util::rstrip` leads to UB. Is it still the case if you assign its result to a `std::string`?
Anyway, if we keep this implementation, we should open an issue to try to fix the UB and use a single implementation of rstrip.

micromamba/tests/test_create.py
JohanMabille: Yeah let's fix the bugs for now;, and figure out what to do for OCI in a dedicated PR.

JohanMabille (approved):

JohanMabille:
I agree that we may reintroduce the canonical_name in a future PR.

libmamba/src/api/list.cpp
JohanMabille: I meant using the funciton degined in `util::string` and assign the final result of the recursive call to a string. The function returns temporary string_view passed by copies, so there hsould not be any dangling temporary issues.

Hind-M: Well doing:
```
std::string res = std::string(util::rstrip(util::rstrip(util::rstrip(util::rstrip(full_str, filename), "/"), platform), "/"));
```
instead of using the reimplemented `rstrip` in `strip_from_filename_and_platform` definitely leads to UB, if that's what you meant.

JohanMabille: yep that's what I meant :( OK let's merge as is and open an issue to fix that UB later

PR #3501, opened 2024-10-03
debug: Show info about the pre-installed version of micromamba

jjerphan:


jjerphan:
Actually, the workflows from `main` are run so this is ineffective.

Issue #3468, opened 2024-09-26
Micromamba 2 shell hook for Powershell doesn't run in Constrained Language Mode


mleistner-bgr:
### Troubleshooting docs

- [X] My problem is not solved in the Troubleshooting docs

### Anaconda default channels

- [X] I do NOT use the Anaconda default channels (pkgs/* etc.)

### How did you install Mamba?

Micromamba

### Search tried in issue tracker

Powershell constrained language

### Latest version of Mamba

- [X] My problem is not solved with the latest version

### Tried in Conda?

Not applicable

### Describe your issue

In my organization Powershell runs by default in [Constrained Language Mode](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_language_modes?view=powershell-7.4#constrainedlanguage-mode) which limits the elements and features that can be used. micromamba 1.5 worked without problems in this language mode, but in micromamba 2.0.0 the shell hook doesn't run anymore (both in Powershell 7.4 and Windows Powershell 5.1).

As the log shows the invocation of `[System.IO.Path]::GetFileNameWithoutExtension()` is not allowed in Constrained Language Mode.
[GetFileNameWithoutExtension()](https://github.com/dotnet/runtime/blob/5535e31a712343a63f5d7d796cd874e563e5ac14/src/libraries/System.Private.CoreLib/src/System/IO/Path.cs#L268C13-L272C47) itself isn't a very sophisticated method, so it should be reasonable to replace this invocation with a different approach.

I replaced the line in question
```Powershell
$__mamba_name = [System.IO.Path]::GetFileNameWithoutExtension((Split-Path -Leaf $Env:MAMBA_EXE));
```
with a case insensitive regex replace
```Powershell
$__mamba_name = (Split-Path -Leaf $Env:MAMBA_EXE) -replace '\.exe$', ''
```
or more universal
```Powershell
$__mamba_name = (Split-Path -Leaf $Env:MAMBA_EXE) -replace '\.\w*$', ''
```
and the resulting script runs fine in constrained language mode and initializes micromamba in Powershell correctly.

### mamba info / micromamba info

```shell
libmamba version : 2.0.0
     micromamba version : 2.0.0
           curl version : libcurl/8.2.1-DEV Schannel zlib/1.2.13
     libarchive version : libarchive 3.6.2 zlib/1.2.13 liblzma/5.4.3 bz2lib/1.0.8 liblz4/1.9.4 libzstd/1.5.5
[...]

PR #3510, opened 2024-10-07
Replace `[System.IO.Path]::GetFileNameWithoutExtension` with `-replace`

mleistner-bgr:
Fix for #3468


Hind-M (approved):
Thanks @mleistner-bgr!

Issue #3505, opened 2024-10-04
`micromamba` paths


Hind-M:
From `prefix` discord:

Micromamba 2 shell activation puts non-existent dirs on Linux `$PATH`. It should check that dirs exist before adding them to `$PATH`.
One line at fault is `export PATH="${MAMBA_ROOT_PREFIX}/condabin:${PATH}` where `condabin` does not exist here.
I have `$MAMBA_ROOT_PREFIX/bin` holding `micromamba` itself, and my own login profile scripts already place it on `$PATH` and define `$MAMBA_EXE`. My base environment is otherwise empty. I also had a bothersome libmamba error message and made an empty `$MAMBA_ROOT_PREFIX/conda-meta` dir just to stop it.
Lastly, I named `$MAMBARC` as 'condarc' because `micromamba` seems to prefer `conda` names (I'm not sure).

Hind-M:
>     * Lastly, I named `$MAMBARC` as 'condarc' because `micromamba` seems to prefer `conda` names (I'm not sure).

There is actually a precedence order regarding the different configuration possibilities (cf. schema and list in https://mamba.readthedocs.io/en/latest/user_guide/configuration.html#multiple-rc-files).
So `$MAMBARC` is supposed to have the highest precedence after CLI configuration.
To make sure there is no issue or bug, can you specify your use case and what made you think that `micromamba` seems to prefer `conda` names?

PR #3511, opened 2024-10-07
Fix `micromamba` paths

Hind-M:
- Fix first point of #3505

### DIFF
--- a/libmamba/data/mamba.csh
+++ b/libmamba/data/mamba.csh
@@ -48,16 +48,12 @@ if ("$__exe_name" == "micromamba") then
 else if ("$__exe_name" == "mamba") then
     alias micromamba __mamba_wrap
 else
-    echo "Error unknow MAMBA_EXE: \"$MAMBA_EXE\", filename must be mamba or micromamba" >&2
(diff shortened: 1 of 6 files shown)


libmamba/data/mamba.csh
```diff
@@ -48,16 +48,12 @@ if ("$__exe_name" == "micromamba") then
...
+    echo "Error unknown MAMBA_EXE: \"$MAMBA_EXE\", filename must be mamba or micromamba" >&2
 endif
 
 if (! $?CONDA_SHLVL) then
     setenv CONDA_SHLVL 0
-    # In dev-mode MAMBA_EXE is python.exe and on Windows
-    # it is in a different relative location to condabin.
-    if ($?_CE_CONDA && $?WINDIR) then
-        setenv PATH "${MAMBA_ROOT_PREFIX}/condabin:${PATH}"
-    else
```
Hind-M, line 60: Removed this because I don't see any difference between the if and else...

Hind-M:
I'm not sure about the execution order here. For example, the `root_prefix/condabin` directory can be added to the path before creating it somewhere else...
Let's see what happens with the CI.

PR #3519, opened 2024-10-09
Create empty base prefix with `env update`

### LINKED ISSUE #3515: `env update` behavior
`env update` command does not work on not yet created envs (unlike `conda`).
(cf. https://github.com/mamba-org/mamba/issues/3388#issuecomment-2376581557 and https://github.com/mamba-org/mamba/issues/3354#issuecomment-2263391579).

Hind-M:
Fix #3515

### DIFF
--- a/micromamba/tests/test_env.py
+++ b/micromamba/tests/test_env.py
@@ -410,3 +410,31 @@ def test_env_create_whitespace(tmp_home, tmp_root_prefix, tmp_path):
         package["name"] == "scikit-learn" and Version(package["version"]) > Version("1.0.0")
         for package in packages
     )
+
+
+env_yaml_content_to_update_empty_base = """
+channels:
+- conda-forge
+dependencies:
+- python
+- xtensor
+"""
+
+
+@pytest.mark.parametrize("shared_pkgs_dirs", [True], indirect=True)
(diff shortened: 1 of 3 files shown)


micromamba/tests/test_env.py
```diff
@@ -307,6 +307,7 @@ def test_env_update_conda_forge_with_pypi(tmp_home, tmp_root_prefix, tmp_path):
 channels:
 - conda-forge
 dependencies:
+- python 3.10
```
Hind-M, line 310: This is to fix the CI test failure due to some known issues with `python 3.11` and `pip`.

micromamba/tests/test_install.py
```diff
@@ -741,3 +741,30 @@ def test_reinstall_with_new_version(tmp_home, tmp_root_prefix):
...
+env_yaml_content_to_install_empty_base = """
+channels:
+- conda-forge
+dependencies:
+- python
+- xtensor
+"""
+
+
+def test_install_empty_base(tmp_home, tmp_root_prefix, tmp_path):
```
Hind-M, line 755: Such a test was missing, i.e test creation of a valid environment with `install` when target prefix is the root prefix.

Issue #3393, opened 2024-08-05
2.0 regression: large overhead of `libsolv`'s `solver_unifyrules` when multichannels are used


Hind-M:
From @ndevenish in the QS lobby on gitter:
"
Is there any known issues with current `micromamba` about resource usage, possibly related to Centos/RHEL? I've had two separate people come to me this week with issues with:
a) using `micromamba` in a container build dying because it filled their entire temp disk (when installing very few packages).
b) being what looked like OOM killed after taking >60% of their memory. Both tasks which have worked before.

The out-of-disk-space instance was running:
`micromamba create -y -c conda-forge gnuplot python numpy pymca workflows>=1.7 xraylib zocalo`
and it took at least 4GB of scratch disk space (the smallest of possible locations that podman was using to do container working on their system).

The other instance didn't get past resolving (an admittedly rather large requirement) but was using >9GB of ram on a 16GB machine the last time I checked before it died.
"

Hind-M:
Used micromamba version: 2.0.0rc0

jjerphan:
I cannot reproduce the errors which you report using `conda-forge/label/micromamba_dev/linux-64/micromamba-2.0.0{rc0-1,rc1-2}`.

On my machine, installing those packages take around 1.5GiB of memory storage in the `$CONDA_PREFIX`, while using less than 1GiB of RAM.

@ndevenish: Could you provide the difference of your instances' resource usage when using `micromamba<2.0.0rc0` and `micromamba>=2.0.0rc0`?

ndevenish:
On [this](https://github.com/dials/dials/blob/main/.conda-envs/windows.txt) environment file

ndevenish:
This is exactly 700 GB btw

ndevenish:
RHEL8, 16GB memory machine:
```
curl -JLO https://raw.githubusercontent.com/dials/dials/refs/heads/main/.conda-envs/linux.txt
curl -Ls https://micro.mamba.pm/api/micromamba/linux-64/latest | tar -xvj bin/micromamba
psrecord  --plot out.png "bin/micromamba create -yp ENV/ -c conda-forge --file linux.txt"
```
![image](https://github.com/user-attachments/assets/b8340195-b22e-428a-8130-845fe90195c4)

ndevenish:
Possibly because it seems to be in a package-cache-fetching loop?
https://github.com/user-attachments/assets/cf71deec-db90-4735-93b1-b8e6365f2fe7

[...]

PR #3521, opened 2024-10-10
fix: Only register channels in the context once

jjerphan:
Partially fix https://github.com/mamba-org/mamba/issues/3393.


JohanMabille (commented):
I think we should take the opportunity of these bug fixes to do some slight refactoring. Typically here the code is duplicated for no reeason, it should be factorize out.

jjerphan:
Actually one needs to first fix [the regression observed](https://github.com/mamba-org/mamba/issues/3393#issuecomment-2405388129) between 1.x and 2.0.

JohanMabille:
I don't see how the performance regression is preventing to factorize out code.

jjerphan:
As discussed IRL, let's first merge this PR and let's tackle [this other issue](https://github.com/mamba-org/mamba/issues/3393#issuecomment-2405388129) in another one.

JohanMabille (approved):

PR #3525, opened 2024-10-14
fix: Handle space in `mamba` and `micromamba` executable absolute paths 

NewUserHa:
version: micromamba-2.0.2-0.tar.bz2

on windows cmd,
it will fail by `'C:\Program' is not recognized as an internal or external command, operable program or batch file.` when `micromamba.exe` is located at eg `C:\Program Files`. 

this PR fixed it.


jjerphan (approved):
LGTM.

The failure on the CI are unrelated and should be resolved by #3537.

Klaim (approved):
LGTM

PR #3532, opened 2024-10-14
fix `activate.bat` fail on windows 

NewUserHa:
version: micromamba-2.0.2-0.tar.bz2

curently run `activate` or `activate.bat` will end result with:
```
>> activate
>> micromamba activate
'micromamba' is not recognized as an internal or external command,
operable program or batch file.
```

the reason is DOSKEY "cannot run a macro from a command that you didn't type in manually" confirmed in docs and other sources eg https://stackoverflow.com/questions/71229141/windows-bat-file-using-start-cmd-k-with-doskey-not-work

this PR fixed it.

### DIFF
--- a/libmamba/src/core/shell_init.cpp
+++ b/libmamba/src/core/shell_init.cpp
@@ -734,7 +734,7 @@ namespace mamba
         util::replace_all(
             activate_bat_contents,
             std::string("__MAMBA_INSERT_EXE_NAME__"),
-            exe_name.string()
+            "mamba"
         );
         std::ofstream condabin_activate_bat_f = open_ofstream(
             root_prefix / "condabin" / "activate.bat"


JohanMabille:
Thanks for your PR. However, I don't understand how this fixes the issue:
- `exe_name.string()` returns a string (`mamba` or `micromamba`, depending on the executable taht runs shell init) that is inserted in the generate bat file, so in the end the doskey contains a string with the exe name (not an alias or anything else).
- except on CI where Autorun is specically disabled, micromamba activate works once the shell init has be done (and if micromamba.exe is in th path).

Am I missing something here?

NewUserHa:
the current `activate.bat` will not work 
because after .bat files got generated even there is a `DOSKEY micromamba=...` in `mamba_hook.bat` and a fresh cmd `call`ed it, 
when run `activate.bat` and the line `micromamba activate %*
` of `activate.bat` execute, it will error by `'micromamba' is not recognized as an internal or external command,
operable program or batch file.`, 
so this PR updated micromamba(from `exe_name.string()`) to use mamba(underhood mamba.bat in env) instead, because Macro set by DOSKEY only works in interactive

JohanMabille:
Got it! Actually we cannnot hardcode that name to mamba, because there are configuration where we expect a micromamba.bat file (and the mamba.bat will not exist). There is an on-going work in https://github.com/mamba-org/mamba/pull/3546 which should fix it for every use case (mamba / micromamba).

JohanMabille:
This has been fixed in #3546

PR #3545, opened 2024-10-17
Correctly rename test to be run

### LINKED ISSUE #3543: Cannot install a local `tar.bz2` package
`install_local_package` added [here](https://github.com/mamba-org/mamba/pull/3164) is not run on the CI.
So far, trying to install a local `tar.bz2` package is still not working.

Hind-M:
Fix #3543

### DIFF
--- a/libmamba/src/core/package_fetcher.cpp
+++ b/libmamba/src/core/package_fetcher.cpp
@@ -325,15 +325,11 @@ namespace mamba
 
     std::string PackageFetcher::channel() const
     {
-        // NOTE `package_type` is only set in `PackageInfo`, and its value is lost
-        // when passing through `make_package_info` from libsolv
-        // So far, the first case never happens?
-        // because `package_type` default value is `specs::PackageType::Unknown`
-        if (m_package_info.package_type == specs::PackageType::Conda)
+        if (!util::starts_with(m_package_info.package_url, "file://"))
         {
             return m_package_info.channel;
         }
-        else
+        else  // local package case
         {
             return "";
         }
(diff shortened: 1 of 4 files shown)


micromamba/tests/test_install.py
```diff
@@ -614,7 +614,11 @@ def test_no_reinstall(self, existing_cache):
         reinstall_res = helpers.install("xtensor", "--json")
         assert "actions" not in reinstall_res
 
-    def install_local_package(self):
+    @pytest.mark.skipif(
+        sys.platform == "win32",
+        reason="Set info (e.g 'name') is wrong (to be fixed)",
```
Hind-M, line 619: #3550 was opened to track this.

libmamba/src/core/package_fetcher.cpp
```diff
@@ -325,12 +325,26 @@ namespace mamba
...
     {
-        return util::concat(m_package_info.platform, '/', m_package_info.filename);
+        if (!util::starts_with(m_package_info.package_url, "file://"))
+        {
+            return util::concat(m_package_info.platform, '/', m_package_info.filename);
+        }
+        else  // local package case
+        {
+            return m_package_info.package_url;
+        }
```
Hind-M, line 347: This was added back from https://github.com/mamba-org/mamba/pull/3296/files#r1604585887, but with the right condition this time (local package or not).

Note that `m_package_info.package_type` is always set to `specs::PackageType::Unknown` here (which is why this part of code was initially removed in #3296 in the first place).
And that's because `package_type` is only set in [PackageInfo](https://github.com/mamba-org/mamba/blob/main/libmamba/src/specs/package_info.cpp#L61), and then its value is not passed when going through [make_package_info](https://github.com/mamba-org/mamba/blob/main/libmamba/src/solver/libsolv/helpers.cpp#L99-L113) from `libsolv` so it's always set to Unknown (which is the default).

JohanMabille (approved):

PR #3548, opened 2024-10-17
maint: Unpin libcurl<8.10

jjerphan:
To see if libcurl 8.10.1 solved the problem observed with libcurl 8.10.0.

### DIFF
--- a/.github/workflows/static_build.yml
+++ b/.github/workflows/static_build.yml
@@ -43,18 +43,6 @@ jobs:
         run: |
           cd micromamba-feedstock/
           sed -i '' '/conda_forge_output_validation/d' conda-forge.yml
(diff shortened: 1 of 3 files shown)


JohanMabille (approved):

jjerphan:
So 8.10.0 likely had a bug which was fixed in 8.10.1.

Do we need to do something to prevent 8.10.0 from being used? E.g. specify `!= 8.10.0` for runtime?

JohanMabille:
I don't think so, unless someone explicitely pings this broken version, the user will get the latest 8.10 version, or something < 8 if he has an upper bound. I would prefer to mark the build 8.10 as broken on ocnda-forge, even if that might be a bit extreme.

PR #3555, opened 2024-10-21
Fix test on windows

### LINKED ISSUE #3550: Info set from a locally installed package is wrong on windows
See skipped test in #3545

Hind-M:
Fix #3550


micromamba/tests/test_install.py
```diff
@@ -614,17 +614,6 @@ def test_no_reinstall(self, existing_cache):
...
 
-    @pytest.mark.skipif(
-        sys.platform == "win32",
-        reason="Set info (e.g 'name') is wrong (to be fixed)",
-    )
-    def test_install_local_package(self):
-        """Attempts to install a .tar.bz2 package from a local directory."""
-        file_path = Path(__file__).parent / "data" / "cph_test_data-0.0.1-0.tar.bz2"
-
-        res = helpers.install(f"file://{file_path}", "--json", default_channel=False)
```
Hind-M, line 625: This is the wrong way of testing, we should only provide the path and `abs_path_to_url` will take care of converting it to a valid file URL format.

micromamba/tests/test_install.py
```diff
@@ -676,6 +665,28 @@ def test_install_check_dirs(tmp_home, tmp_root_prefix):
         assert os.path.isdir(env_prefix / "lib" / "python3.8" / "site-packages")
 
 
+def test_install_local_package(tmp_home, tmp_root_prefix):
```
Hind-M, line 668: Move the test outside of the class because of `C` drive access issues on windows (I also think this is the preferable way of writing tests, because it's more isolated).

libmamba/tests/src/specs/test_match_spec.cpp
```diff
@@ -623,6 +624,54 @@ TEST_SUITE("specs::match_spec")
         }
     }
 
+    TEST_CASE("parse_url")
```
Hind-M, line 627: This was missing, and helps understanding how it's supposed to work.

JohanMabille (approved):

PR #3564, opened 2024-10-23
chore: some CMake cleanup

henryiii:
Some things I noticed when trying something locally.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,8 +5,6 @@
 # The full license is in the file LICENSE, distributed with this software.
 
 cmake_minimum_required(VERSION 3.16)
-cmake_policy(SET CMP0025 NEW) # Introduced in cmake 3.0
-cmake_policy(SET CMP0077 NEW) # Introduced in cmake 3.13
 
 project(mamba)
 
--- a/docs/source/developer_zone/dev_environment.rst
+++ b/docs/source/developer_zone/dev_environment.rst
@@ -37,7 +37,7 @@ Install development dependencies
 .. code:: bash
 
     micromamba create -n mamba -c conda-forge -f dev/environment-dev.yml
-    micromamba activate -n mamba
+    micromamba activate mamba
 
 Compile ``libmamba`` and other artifacts
 ****************************************
--- a/libmamba/CMakeLists.txt
+++ b/libmamba/CMakeLists.txt
@@ -539,7 +539,7 @@ macro(libmamba_create_target target_name linkage output_name)
             set(CMAKE_PREFIX_PATH "$ENV{VCPKG_ROOT}/installed/x64-windows-static-md/")
 
             # For Windows we have a vcpkg based build system right now.
-            find_package(LibArchive CONFIG REQUIRED)
+            find_package(LibArchive MODULE REQUIRED)
             find_package(CURL CONFIG REQUIRED)
             find_library(LIBLZMA_LIBRARIES lzma REQUIRED)
             find_library(LZ4_LIBRARY NAMES lz4)
--- a/libmamba/CMakeLists.txt
+++ b/libmamba/CMakeLists.txt
@@ -405,15 +405,15 @@ set(
 # Targets and link
 # ================
 
-find_package(fmt REQUIRED)
-find_package(spdlog REQUIRED)
-find_package(tl-expected REQUIRED)
-find_package(nlohmann_json REQUIRED)
-find_package(simdjson REQUIRED)
-find_package(yaml-cpp REQUIRED)
-find_package(reproc REQUIRED)
-find_package(reproc++ REQUIRED)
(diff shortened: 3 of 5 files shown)


CMakeLists.txt
```diff
@@ -5,8 +5,6 @@
 # The full license is in the file LICENSE, distributed with this software.
 
 cmake_minimum_required(VERSION 3.16)
-cmake_policy(SET CMP0025 NEW) # Introduced in cmake 3.0
-cmake_policy(SET CMP0077 NEW) # Introduced in cmake 3.13
```
henryiii, line 9: Setting the minimum version of CMake sets all policies to NEW that were introduced on or before that value. (Also true if you set a version range, like 3.16...3.30)

docs/source/developer_zone/dev_environment.rst
```diff
@@ -37,7 +37,7 @@ Install development dependencies
 .. code:: bash
 
     micromamba create -n mamba -c conda-forge -f dev/environment-dev.yml
-    micromamba activate -n mamba
+    micromamba activate mamba
```
henryiii, line 40: There's no `-n` here.

libmamba/CMakeLists.txt
```diff
@@ -405,15 +405,15 @@ set(
...
-find_package(fmt REQUIRED)
-find_package(spdlog REQUIRED)
-find_package(tl-expected REQUIRED)
-find_package(nlohmann_json REQUIRED)
-find_package(simdjson REQUIRED)
-find_package(yaml-cpp REQUIRED)
-find_package(reproc REQUIRED)
-find_package(reproc++ REQUIRED)
-find_package(Libsolv REQUIRED)
+find_package(fmt CONFIG REQUIRED)
```
henryiii, line 408: Using `CONFIG` explicitly makes the error message better. I didn't go through every one to figure out whether they were CONFIG or MODULE, but these are some of the first ones hit, so error message is important for these.

libmambapy/CMakeLists.txt
```diff
@@ -7,20 +7,16 @@ cmake_minimum_required(VERSION 3.18.2)
...
 if(NOT TARGET mamba::libmamba)
-    find_package(libmamba REQUIRED)
+    find_package(libmamba CONFIG REQUIRED)
     set(libmamba_target mamba::libmamba-dyn)
 else()
     set(libmamba_target mamba::libmamba)
 endif()
 
-find_package(Python COMPONENTS Interpreter Development)
+find_package(Python COMPONENTS Interpreter Development.Module)
```
henryiii, line 19: The default `Development` looks for the `Embed` target too, which isn't required and sometimes (like on manylinux) not available.

jjerphan (approved):
Hi @henryiii,

Thank you for bringing your expertise in build system and meta-build systems with this contribution!

Edit: I don't know whether the failures on Windows are related to that, and I can't debug it.

henryiii:
The PR after this one has the same windows error, and the one before this one claims to fix something on Windows.

jjerphan:
#3555 fixes an issue related to tests under Windows.

The error here is due to vcpkg which is used for the static builds of micromamba on Windows — this might be beyond our control.

JohanMabille (approved):

JohanMabille:
This looks like a network issue, libxml2 is available at [https://gitlab.gnome.org//GNOME/libxml2/-/archive/v2.11.9/libxml2-v2.11.9.tar.gz](https://gitlab.gnome.org//GNOME/libxml2/-/archive/v2.11.9/libxml2-v2.11.9.tar.gz:)

libmamba/CMakeLists.txt
```diff
@@ -539,7 +539,7 @@ macro(libmamba_create_target target_name linkage output_name)
             set(CMAKE_PREFIX_PATH "$ENV{VCPKG_ROOT}/installed/x64-windows-static-md/")
 
             # For Windows we have a vcpkg based build system right now.
-            find_package(LibArchive REQUIRED)
+            find_package(LibArchive CONFIG REQUIRED)
```
JohanMabille, line 542: I t looks like LibArhcive does not provide a cmake config file on Windows.

JohanMabille, line 542:
```suggestion
            find_package(LibArchive REQUIRED)
```

JohanMabille (changes requested):
It looks like the  LibArchive vcpkg package does not provide a cmake config file

libmamba/CMakeLists.txt
henryiii: Ah, my mistake: https://cmake.org/cmake/help/latest/module/FindLibArchive.html

```suggestion
            find_package(LibArchive MODULE REQUIRED)
```

JohanMabille (approved):

JohanMabille:
Thanks!

Issue #3495, opened 2024-10-03
Put micromamba v2 on homebrew, currently latest version there is 1.5.10


corneliusroemer:
I'd like to install micromamba v2 via homebrew, but so far only 1.5.10 is available there.

https://github.com/Homebrew/homebrew-core/blob/45656b5b8833aa5a0f26fdcf4318324c756fb994/Formula/m/micromamba.rb

Do you have plans to add it there?

corneliusroemer:
I created a PR to point this out in docs while homebrew doesn't have v2.

Once homebrew has v2, this PR should be reverted: https://github.com/mamba-org/mamba/pull/3499

henryiii:
See https://github.com/Homebrew/homebrew-core/pull/192460 and https://github.com/Homebrew/homebrew-core/pull/191869.

henryiii:
I made PR https://github.com/Homebrew/homebrew-core/pull/195371 - advice welcome on how to proceed - it seems "micromamba" now requires static linkage so the micromamba recipe can't build it, shared linkage is required on homebrew.

henryiii:
Anyone know why macOS 13 reports:

```
 /tmp/micromamba-20241025-6758-psgcn7/mamba-micromamba-2.0.1/libmamba/src/core/transaction.cpp:703:36: error: no viable constructor or deduction guide for deduction of template arguments of 'packaged_task'
                  std::packaged_task task{ [=] { return it->run(); } };
```

henryiii:
And linux:

```
Error: /usr/include/c++/11/bits/hashtable_policy.h:1004:12:   required from ‘struct std::__detail::_Insert<std::__cxx11::basic_string<char>, std::pair<const std::__cxx11::basic_string<char>, mamba::specs::Channel>, std::allocator<std::pair<const std::__cxx11::basic_string<char>, mamba::specs::Channel> >, std::__detail::_Select1st, std::equal_to<std::__cxx11::basic_string<char> >, std::hash<std::__cxx11::basic_string<char> >, std::__detail::_Mod_range_hashing, std::__detail::_Default_ranged_hash, std::__detail::_Prime_rehash_policy, std::__detail::_Hashtable_traits<true, false, true>, false>’
[...]

PR #3578, opened 2024-11-06
ci: test on macOS 13 (Intel)

henryiii:
macOS 13 is a target for homebrew, and still an important target for GHA, as it's the last Intel build they provide. I think on macOS 13 there will be a missing deduction guide error, but will have to see.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -23,7 +23,7 @@ jobs:
     name: Unix
     strategy:
       matrix:
-        os: [ubuntu-latest, macos-latest]
+        os: [ubuntu-latest, macos-latest, macos-13]
         build_type: [release, debug]
       fail-fast: false
     uses: ./.github/workflows/unix_impl.yml


jjerphan:
Thank you for this contribution, @henryiii.

Someone with a machine on macOS 13 should dive into this problem.

Note that we are only maintaining the distributions of mamba and micromamba on conda-forge, but we aren't against someone maintaining the distributions on Homebrew and upstreaming patches so that mamba and micromamba are portable or more systems.

henryiii:
Any suggestions or thoughts on https://github.com/mamba-org/mamba/issues/3495 ? The problem is that the GHA runs here use the latest LLVM compilers, and not the AppleClang compilers that come with macOS, and systems like homebrew use the AppleClang compilers. I think there are a couple of uses of C++ features that aren't supported by the older compilers. So far it looks pretty easily fixable, I think? Maybe just adding some explicit types instead of using the deduction guidelines.

There's also a linux issue there.

.github/workflows/tests.yml
```diff
@@ -23,7 +23,7 @@ jobs:
     name: Unix
     strategy:
       matrix:
-        os: [ubuntu-latest, macos-latest]
+        os: [ubuntu-latest, macos-latest, macos-13]
```
jjerphan, line 26: We could also add [`ubuntu-22.04`](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md) which is used by [Homebrew's Linux job](https://github.com/Homebrew/homebrew-core/actions/runs/11708308681/job/32609917766?pr=195371#step:1:8).

```suggestion
        os: [ubuntu-latest, ubuntu-22.04, macos-latest, macos-13]
```

henryiii: This is currently ubuntu-latest. They tried to move to ubuntu-24.04 but had to roll it back, I think it moves Jan 2025 now.

jjerphan:
Can this be closed since https://github.com/mamba-org/mamba/pull/3613 has been merged?

.github/workflows/tests.yml
```diff
@@ -24,7 +24,7 @@ jobs:
...
```
jjerphan:
```suggestion
        os: [ubuntu-latest, macos-latest, macos-15]
```

jjerphan:
@henryiii: using macos-15 resolves the problem but it seems that homebrew's CI uses macos-13, macos-14 and macos-15.

Must it use macos-13 and macos-14? I.e. would you accept those changes?

henryiii:
13 is special because it's Intel based. That's what was failing on brew. It's not the last OS to support Intel (I'm on an Intel machine mostly, and am using the latest OS), but it was the last OS they provided with Intel runners (last I checked).

I expect macOS-13 to fail, I want it to fail here, because the problem needs to be fixed. Using macOS-13 was supposed to expose the problem that needs fixing. :)

jjerphan:
We don't have a macos-13 machine to debug, so we will wait.

jjerphan:
Is Homebrew still using [`macos-13` which is being closing down](https://github.blog/changelog/2025-09-19-github-actions-macos-13-runner-image-is-closing-down/) or another similar configuration?

jjerphan:
Closing as `macos-13` is being rolled out.

Issue #3582, opened 2024-11-07
`mamba env export --json --from-history` is broken in v2


mathbunnyru:
### Troubleshooting docs

- [X] My problem is not solved in the Troubleshooting docs

### Anaconda default channels

- [X] I do NOT use the Anaconda default channels (pkgs/* etc.)

### How did you install Mamba?

Micromamba

### Search tried in issue tracker

mamba, env, export, from-history

### Latest version of Mamba

- [X] My problem is not solved with the latest version

### Tried in Conda?

I didn't try

### Describe your issue

When I run `mamba env export --json --from-history` it produces invalid json file:

```
{
  "channels": [
    "conda-forge"
  ],
  "dependencies": [
    "jupyter_core",
    "mamba",
    "python=3.12",
  ],
  "name": "base",
  "prefix": "/opt/conda"
}
```

Note: **--from-history** flag is important here, without it everything is ok, the json is invalid due to `,` in the last element of dependencies.

### mamba info / micromamba info

```shell
libmamba version : 2.0.3
          mamba version : 2.0.3
           curl version : libcurl/8.10.1 OpenSSL/3.3.2 zlib/1.3.1 zstd/1.5.6 libssh2/1.11.0 nghttp2/1.64.0
     libarchive version : libarchive 3.7.4 zlib/1.2.13 liblzma/5.2.6 bz2lib/1.0.8 liblz4/1.9.3 libzstd/1.5.6
       envs directories : /opt/conda/envs
          package cache : /opt/conda/pkgs
                          /home/jovyan/.mamba/pkgs
            environment : base
           env location : /opt/conda
      user config files : /home/jovyan/.mambarc
 populated config files : /opt/conda/.condarc
       virtual packages : __unix=0=0
                          __linux=6.10.11=0
                          __glibc=2.39=0
                          __archspec=1=aarch64
               channels : https://conda.anaconda.org/conda-forge/linux-aarch64
                          https://conda.anaconda.org/conda-forge/noarch
       base environment : /opt/conda
               platform : linux-aarch64
```

mathbunnyru:
Related to: https://github.com/mamba-org/mamba/issues/3492

The `json` output was added in https://github.com/mamba-org/mamba/pull/3559/files

@jjerphan you might want to take a look

maresb:
I'm running into a different problem of a very similar style in #3583

PR #3587, opened 2024-11-08
fix: Export `'channels'` as part of environments' export

mathbunnyru:
If I run the following:
```
mamba env create --name empty
mamba env export --name empty
```

it will give me:
```
name: empty
channels:
dependencies:

prefix: "/Users/asalikhov/test-mamba-root/envs/empty"
```

Note: `channels`, `dependencies` are always printed, even if empty.

But if I run `export` with `--json`, it will not print `channels:`:
```
{
  "dependencies": [
  ],
  "name": "empty",
  "prefix": "/Users/asalikhov/test-mamba-root/envs/empty"
}
```

I think it's always good to print `channels`, it makes parsing easier.

Related: https://github.com/mamba-org/mamba/issues/3582


micromamba/tests/test_env.py
```diff
@@ -70,31 +70,47 @@ def export_env():
...
+
+
 @pytest.mark.parametrize("channel_subdir_flag", [None, "--channel-subdir"])
 @pytest.mark.parametrize("md5_flag", [None, "--md5", "--no-md5"])
 @pytest.mark.parametrize("explicit_flag", [None, "--explicit"])
 @pytest.mark.parametrize("no_build_flag", [None, "--no-build", "--no-builds"])
 @pytest.mark.parametrize("json_flag", [None, "--json"])
 def test_env_export(
-    export_env, json_flag, no_build_flag, explicit_flag, md5_flag, channel_subdir_flag
+    channel_subdir_flag, md5_flag, explicit_flag, no_build_flag, json_flag, export_env
```
mathbunnyru, line 98: This is the order it is mentioned above, it's better to have the same order everywhere

micromamba/tests/test_env.py
```diff
@@ -70,31 +70,47 @@ def export_env():
...
         if md5_flag != "--no-md5":
             assert re.search("#[a-f0-9]{32}$", output.replace("\r", ""))
     else:
-        if json_flag:
-            # Already parsed
-            ret = output
-        else:
-            ret = yaml.safe_load(output)
+        # json is already parsed
+        ret = output if json_flag else yaml.safe_load(output)
```
mathbunnyru, line 113: I think this is quite easy to read as a one-line

mathbunnyru:
I think the CI failure is spurious and this can be merged

jjerphan (approved):
Thank you, @mathbunnyru!

PR #3595, opened 2024-11-12
Fix some warnings

mathbunnyru:


libmamba/include/mamba/core/output.hpp
```diff
@@ -170,7 +170,7 @@ namespace mamba
     {
     public:
 
-        MessageLogger(const char* file, int line, log_level level);
+        MessageLogger(const char* file, log_level level);
```
Hind-M, line 173: Can you explain why remove the line here?

libmamba/include/mamba/solver/libsolv/unsolvable.hpp
```diff
@@ -21,7 +21,7 @@ namespace solv
 
 namespace mamba
 {
-    class Palette;
+    struct Palette;
```
Hind-M, line 24: What was the warning regarding this?

libmamba/tests/src/download/test_mirror.cpp
```diff
@@ -58,7 +58,8 @@ namespace mamba::download
         TEST_CASE("PassThroughMirror")
         {
             std::unique_ptr<Mirror> mir = make_mirror("");
-            CHECK_EQ(typeid(*mir), typeid(PassThroughMirror));
+            auto& ref = *mir;
```
Hind-M, line 61: How about the warning for this change? I believe this is more related to performance but we only dereference once.

libmamba/src/api/list.cpp
```diff
@@ -71,7 +71,7 @@ namespace mamba
             }
 
             std::regex spec_pat(regex);
-            auto all_records = std::move(prefix_data.all_pkg_mgr_records());
+            auto all_records = prefix_data.all_pkg_mgr_records();
```
Hind-M, line 74: And this one?

libmamba/include/mamba/core/output.hpp
mathbunnyru: `m_line` is not used anywhere, we're not logging the line number

libmamba/include/mamba/solver/libsolv/unsolvable.hpp
mathbunnyru: `Palette` is defined as `struct`: https://github.com/mamba-org/mamba/blob/af276aeedbb0f63a40ff500832425335d88029b9/libmamba/include/mamba/core/palette.hpp#L14

There is a warning when declaration and definition are different - as far as I remember in some ABIs (MSVC, if I remember correctly) this means different mangling and could cause linking errors.

libmamba/tests/src/download/test_mirror.cpp
mathbunnyru: This is not about performance, compiler didn't like that there possibly was a call with side effects in `typeid` call (dereferencing might fail).
The easiest way to fix this is to dereference in a separate line.

Hind-M: Well in that case, it still doesn't solve the problem of dereferencing possible failure.
I suggest adding instead a `static_assert` before the `CHECK_EQ`:
`static_assert(mir != nullptr);`

libmamba/include/mamba/core/output.hpp
mathbunnyru: It doesn't give a warning (now), but it's a good point, I will remove `m_file` as well - it doesn't change any behaviour.

libmamba/tests/src/download/test_mirror.cpp
mathbunnyru: I can't do it, because `mir` is not known in a compile-time, so `static_assert` won't work here

Hind-M: And maybe also rename `ref` to something more explicit, `mir_ref` for example.

mathbunnyru: Done

Hind-M (approved):
Thanks!

PR #3596, opened 2024-11-12
Force spinx v6 in readthedocs

mathbunnyru:


mathbunnyru:
This is a duplicate, sorry. I wanted to check if I could add labels to the PR, so the CI passes. Unfortunately, it seems I can't.

PR #3600, opened 2024-11-12
maint: Update clang-format to v19

mathbunnyru:
Possible options (I didn't change any): https://clang.llvm.org/docs/ClangFormatStyleOptions.html

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -34,7 +34,7 @@ repos:
       - id: blacken-docs
         additional_dependencies: [black==22.3.0]
   - repo: https://github.com/pre-commit/mirrors-clang-format
-    rev: v18.1.2
+    rev: v19.1.3
     hooks:
       - id: clang-format
         args: [--style=file]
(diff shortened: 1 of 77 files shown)


jjerphan (commented):
LGTM if and only if the merge commit of this PR has its diff ignored since it bring no semantic changes.

For more information, see:
 - https://github.com/orgs/community/discussions/5033
 - https://www.stefanjudis.com/today-i-learned/how-to-exclude-commits-from-git-blame/

Hind-M:
To have a consistent history and since the formatting commits were always explicit, I think we should actually just leave the files diff as is (not ignored) so that we know that a change was made regarding the format.

Hind-M:
@klaim Any objections on the format result? (We may change the options if any).

Klaim:
> @Klaim Any objections on the format result? (We may change the options if any).

Looks tolerable to me. 👍🏽 

To be completely clear, there are things I'm not a fan of but that I can totally live with:
- having the return type on a second line when using the `auto ... ->` function signature syntax is nice, but I would prefer the short signatures to not have that second line, as it seems not necessary (only on short signatures) - but I can live with systematic return type separated;
- My preferred style of lambda expression is to have the `{` on the same line as the signature (like some post-java languages?), so that the content of the body looks like it's part of the expression hosting the lambda expression (`func(a, b, []{ ...`), and it avoids having too many lines in short lambdas, and also that makes the lambda expressions standout stylistically when you use a different style for most C++ expressions (`{` and `}` always on a new line for example, to mark a scope clearly); here we get a new line per `{` systematically and the signature is kind of disconnected visually - but it's really minor and I can also totally live with it;

Also I don't know the limitations of clang-format regarding these minor points, maybe it's just not possible to have the exact style I prefer. Probably not worth our time if everybody can live with this style.

So I think we're good to go. 👍🏽

jjerphan:
I let anyone of you merge this PR.

PR #3603, opened 2024-11-14
[1.x] maint: Add missing dependencies to micromamba's environment specification

jjerphan:
Fix CI failures first observed in https://github.com/mamba-org/mamba/pull/3593.

### DIFF
--- a/micromamba/tests/test_package.py
+++ b/micromamba/tests/test_package.py
@@ -6,7 +6,6 @@ import tarfile
 import zipfile
 from pathlib import Path
 
-import conda_package_handling.conda_fmt
 import pytest
 import zstandard
 from conda_package_handling import api as cph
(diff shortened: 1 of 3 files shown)


Klaim (approved):
I'm not knowledgeable the test with zstandard, but otherwise LGTM

micromamba/tests/test_package.py
```diff
@@ -148,6 +149,11 @@ def test_extract_compress(cph_test_file: Path, tmp_path: Path):
     assert_sorted(names[: len(info_files)])
 
 
+# Only run this test if zstandard is available.
+@pytest.mark.skipif(
+    not hasattr(conda_package_handling.conda_fmt, "ZSTD_COMPRESS_LEVEL"),
+    reason="zstandard not available (required by conda_package_handling)",
+)
```
Hind-M, line 156: Is `zstandard` not available in some platforms? I mean why do we need to skip the test?

Hind-M (approved):

PR #3604, opened 2024-11-14
More details in error message when failing to parse json from a python command's output

Klaim:


libmamba/src/core/prefix_data.cpp
```diff
@@ -10,8 +10,10 @@
 #include <utility>
 
 #include <reproc++/run.hpp>
+#include <fmt/ranges.h>
```
Hind-M, line 13: I think this should be included before the `reproc` header (should be in alphabetical order) to fix the linter.
FYI, running `pre-commit run --all` in the source folder will modify the files accordingly to match the linter rules.

libmamba/src/core/prefix_data.cpp
```diff
@@ -231,7 +233,15 @@ namespace mamba
         );
         if (ec)
         {
-            throw std::runtime_error(ec.message());
+            const auto message = fmt::format(
```
Hind-M, line 236: This is good to have for more details in general indeed.
But in this specific issue, I think the problem is parsing the [output](https://github.com/mamba-org/mamba/blob/main/libmamba/src/core/prefix_data.cpp#L244) which is not valid so maybe add a debug log or even a trace log (to not disturb other debugging use cases too much, if we think the log should stay permanently):
```
LOG_TRACE << "Parsing `pip inspect` output:\n" << out;
nlohmann::json j = nlohmann::json::parse(out);
```

Klaim: Ah yes wrong try/catch, thanks I'll move it.

Klaim: @Hind-M I suspect we might want these info every time there is a parsing error there without having to add trace level of logging. Or do you think it would be too much and the trace is enough?

Klaim: Ok I'll attempt with a mix of trace and throw and lets see how it looks concretely.

libmamba/src/core/prefix_data.cpp
Klaim: I'm having an issue with `pre-commit run --all` locally, sorry! I'll try to fix it but so far my attempts failed.

Hind-M (approved):

Issue #3576, opened 2024-11-06
2.0.3-0 breaks mamba-org/setup-micromamba@v2 GH Action


kmuehlbauer:
### Troubleshooting docs

- [X] My problem is not solved in the Troubleshooting docs

### Anaconda default channels

- [X] I do NOT use the Anaconda default channels (pkgs/* etc.)

### How did you install Mamba?

Other (please describe)

### Search tried in issue tracker

no other issue found

### Latest version of Mamba

- [X] My problem is not solved with the latest version

### Tried in Conda?

Not applicable

### Describe your issue

With latest micromamba 2.0.3-0 our CI mamba-org/setup-micromamba@v2 GH Action stopped working.

See https://github.com/pydata/xarray/actions/runs/11704038363/job/32595639555

```
/home/runner/micromamba-bin/micromamba list -r /home/runner/micromamba -n xarray-tests
  critical libmamba could not load prefix data
Error: The process '/home/runner/micromamba-bin/micromamba' failed with exit code 1
```

### mamba info / micromamba info

```shell
/home/runner/micromamba-bin/micromamba info -r /home/runner/micromamba -n xarray-tests
  
         libmamba version : 2.0.3
       micromamba version : 2.0.3
             curl version : libcurl/8.10.1 OpenSSL/3.3.2 zlib/1.3.1 zstd/1.5.6 libssh2/1.11.0 nghttp2/1.64.0
       libarchive version : libarchive 3.7.4 zlib/1.2.13 bz2lib/1.0.8 libzstd/1.5.6
         envs directories : /home/runner/micromamba/envs
            package cache : /home/runner/micromamba/pkgs
                            /home/runner/.mamba/pkgs
              environment : xarray-tests
             env location : /home/runner/micromamba/envs/xarray-tests
        user config files : /home/runner/.mambarc
   populated config files : /home/runner/work/_temp/setup-micromamba/.condarc
         virtual packages : __unix=0=0
                            __linux=6.5.0=0
                            __glibc=2.35=0
                            __archspec=1=x86_64_v3
                 channels : https://conda.anaconda.org/conda-forge/linux-64
                            https://conda.anaconda.org/conda-forge/noarch
         base environment : /home/runner/micromamba
                 platform : linux-64
```

### Logs

```shell
https://github.com/pydata/xarray/actions/runs/11704038363/job/32595639555
```

### environment.yml

```yaml
name: xarray-tests
channels:
  - conda-forge
  - nodefaults
dependencies:
[...]

PR #3607, opened 2024-11-15
Prevent pip "rich" ouput

Klaim:
The goal here is to help with #3576 , although there might be other issues behind it.

If someone reading this know some other flags/env-vars we could use to obtain a pure json utf-8 output, feel free to add suggestions 👍🏽


libmamba/src/core/prefix_data.cpp
```diff
@@ -261,11 +267,11 @@ namespace mamba
         {
             j = nlohmann::json::parse(out);
```
Hind-M, line 268: I think this may be useful as well:
```suggestion
            j = nlohmann::json::parse(util::strip(out));
```

wolfv:
Usually rich / pip should implement the `isatty` kind of detection to figure out wether to print colors or not. However, it could indeed be the case that someone adds an env var that forces colors for pip which would make this crash. 

You could also be more aggressive about running in a pristine environment. 

Just some thoughts :)

Klaim:
I reproduced the issue from #3576 by running the command with `FORCE_COLOR=1`. It seems that `NO_COLOR=1` is not enough to prevent the parsing failure in that context even if the color is indeed removed. So the best solution so far that works locally is to unset `FORCE_COLOR` for the time of the call.

Issue #3571, opened 2024-10-30
DOC: explain how mamba2 and micromamba2 differ


lucascolley:
Reading the mamba 2.0 changes page in the docs, I understand the general sentiment that mamba and micromamba have become more similar to each other. I also understand from other docs pages that they are still different in that they can be installed differently.

Aside from that, how do they differ? Why might I choose to use one over the other?

henryiii:
I'm assuming they are exactly identical, just one is statically linked and the other dynamically linked in https://github.com/Homebrew/homebrew-core/pull/195371. Could be wrong, though, based that on parts of the docs that looked updated!

jjerphan:
We are wording this with #3610, I would invite you to give it a review if you can. :slightly_smiling_face:

PR #3610, opened 2024-11-18
docs: Update pieces of documentation after the release of mamba 2

jjerphan:
Fixes https://github.com/mamba-org/mamba/issues/3571.


README.md
```diff
@@ -31,8 +31,17 @@
...
-The ecosystem also consists of `quetz`, an open source `conda` package server and `boa`, a fast `conda` package builder.
+`mamba` is part of the [conda-forge](https://conda-forge.org/) ecosystem, which also consists of `quetz`, an open source `conda` package server and `boa`, a fast `conda` package builder.
+
+You can read our [announcement blog post](https://medium.com/@QuantStack/open-software-packaging-for-science-61cecee7fc23).
+
+## micromamba
+
+`micromamba` is the statically linked version of `mamba`.
+
+It can be installed as a standalone executable without any dependencies, making it a perfect fit for CI/CD pipelines and containerized environments.
```
lucascolley, line 42: It may be nice to also give examples of when `mamba` would be preferable to `micromamba` (if there are any?)

lucascolley (commented):
thanks! I think it would also be useful to clarify that `mamba` and `micromamba` are functionally (almost?) the same now on https://mamba.readthedocs.io/en/latest/developer_zone/changes-2.0.html, and perhaps also https://mamba.readthedocs.io/en/latest/.

README.md
jjerphan: I cannot really think of any benefits of using `mamba` over `micromamba` for now (except retro-compatibility for end-users' application).

lucascolley: makes sense! I wonder how that can be more clearly reflected in the docs (maybe it isn't important)

JohanMabille: By being dynamically linked, mamba benefits from the updates of its dependencies, while someone using micromamba must wait for a new release of micromamba to get the updates.

I would say micromamba targets CI and "non transient", not sensible environments (like developers machine) while mamba is more for production environments. Also if you want ot use the Python API, it might be better to install mamba (otherwise it's like installing libmamba twice).

README.md
```diff
@@ -31,8 +31,17 @@
 
 At the same time, `mamba` utilizes the same command line parser, package installation and deinstallation code and transaction verification routines as `conda` to stay as compatible as possible.
 
-Mamba is part of a bigger ecosystem to make scientific packaging more sustainable. You can read our [announcement blog post](https://medium.com/@QuantStack/open-software-packaging-for-science-61cecee7fc23).
-The ecosystem also consists of `quetz`, an open source `conda` package server and `boa`, a fast `conda` package builder.
+`mamba` is part of the [conda-forge](https://conda-forge.org/) ecosystem, which also consists of `quetz`, an open source `conda` package server and `boa`, a fast `conda` package builder.
```
JohanMabille, line 34: I think we should not mention boa, we will deprecate it soon.

JohanMabille (changes requested):

README.md
lucascolley: > By being dynamically linked, mamba benefits from the updates of its dependencies, while someone using micromamba must wait for a new release of micromamba to get the updates.
> 
> I would say micromamba targets CI and "non transient", not sensible environments (like developers machine) while mamba is more for production environments. Also if you want ot use the Python API, it might be better to install mamba (otherwise it's like installing libmamba twice).

Cool! Would be great for this to make it into the docs.

jjerphan: Added with f477533, let me know if this is clear to you.

JohanMabille (approved):

README.md
jjerphan: @lucascolley: is this clear to you?

lucascolley: looks fine to me, thanks!

Issue #3542, opened 2024-10-16
Address warnings while building mamba/micromamba from source


anutosh491:
### Troubleshooting docs

- [X] My problem is not solved in the Troubleshooting docs

### Anaconda default channels

- [X] I do NOT use the Anaconda default channels (pkgs/* etc.)

### How did you install Mamba?

Micromamba

### Search tried in issue tracker

Not found

### Latest version of Mamba

- [X] My problem is not solved with the latest version

### Tried in Conda?

I do not have this problem with Conda, just with Mamba

### Describe your issue

I see lot of warnings while trying to build mamba/micromamba from source. This issue has been opened to keep track and address these.
```
/Users/anutosh491/work/mamba/libmamba/ext/solv-cpp/src/pool.cpp:96:14: warning: unused function 'is_reldep' [-Wunused-function]
   96 |         auto is_reldep(::Id id) -> bool
      |              ^~~~~~~~~
1 warning generated.

In file included from /Users/anutosh491/work/mamba/libmamba/src/util/parsers.cpp:10:
/Users/anutosh491/work/mamba/libmamba/include/mamba/util/parsers.hpp:284:56: warning: implicit conversion changes signedness: 'int' to 'std::size_t' (aka 'unsigned long') [-Wsign-conversion]
  284 |                 pos = static_cast<int>(found) * i + (1 - static_cast<int>(found)) * pos;
      |                                                      ~~^~~~~~~~~~~~~~~~~~~~~~~~~  ~
/Users/anutosh491/work/mamba/libmamba/include/mamba/util/parsers.hpp:364:61: note: in instantiation of function template specialization 'mamba::util::detail_parsers::find<char, 1UL>' requested here
  364 |                 const auto open_depth_idx = detail_parsers::find(open, text[pos]);
      |                                                             ^
/Users/anutosh491/work/mamba/libmamba/include/mamba/util/parsers.hpp:473:32: note: in instantiation of function template specialization 'mamba::util::detail_parsers::find_matching_parentheses_impl<1UL, mamba::util::detail_parsers::FindParenthesesSearcher>' requested here
  473 |         return detail_parsers::find_matching_parentheses_impl(
      |                                ^
/Users/anutosh491/work/mamba/libmamba/src/util/parsers.cpp:27:16: note: in instantiation of function template specialization 'mamba::util::find_matching_parentheses<1UL>' requested here
   27 |         return find_matching_parentheses(text, err, std::array{ open }, std::array{ close });
      |                ^
[...]

PR #3611, opened 2024-11-18
maint: Enable `-Werror` compiler flag for GCC, Clang and AppleClang

mathbunnyru:
I think we can now enable `-Werror` flag relatively easily when building with GCC, Clang and AppleClang.

I tested locally on my M2 Mac with the following compilers:
- `Apple clang version 16.0.0 (clang-1600.0.26.4)`
- `clang version 17.0.6`
- `gcc-14 (Homebrew GCC 14.2.0_1) 14.2.0`

In some places I turned errors back to warnings, in some - disabled, I will explain choices using commenting.

Fix: https://github.com/mamba-org/mamba/issues/3542


libmamba/CMakeLists.txt
```diff
@@ -264,6 +264,22 @@ set(
     ${LIBMAMBA_SOURCE_DIR}/api/shell.cpp
     ${LIBMAMBA_SOURCE_DIR}/api/update.cpp
 )
+# TODO: remove when switch to C++20
+if(CMAKE_CXX_COMPILER_ID STREQUAL "Clang" OR CMAKE_CXX_COMPILER_ID STREQUAL "AppleClang")
+    # This file uses capturing structured bindings, which was fixed in C++20
+    set_source_files_properties(
+        ${LIBMAMBA_SOURCE_DIR}/download/mirror_impl.cpp
+        PROPERTIES COMPILE_FLAGS -Wno-c++20-extensions
```
mathbunnyru, line 272: There is nothing wrong with our code and nothing to change there. So we can safely ignore the warning we're using `C++20` feature (capturing structured binding).
This can't hide any real problems, so here this warning/error will disappear.

libmamba/CMakeLists.txt
```diff
@@ -264,6 +264,22 @@ set(
...
+    )
+endif()
+
+if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+    # GCC reports dangling reference when using std::invoke with data members
+    set_source_files_properties(
+        ${LIBMAMBA_SOURCE_DIR}/solver/problems_graph.cpp
+        PROPERTIES COMPILE_FLAGS -Wno-error=dangling-reference
```
mathbunnyru, line 280: This might be useful warning/error, so I'm turning it back to being warning for this file.

libmamba/include/mamba/util/conditional.hpp
```diff
@@ -31,9 +31,7 @@ namespace mamba::util
         }
         else
         {
-            using int_t = Int;
-            return static_cast<int_t>(condition) * true_val
-                   + (int_t(1) - static_cast<int_t>(condition)) * false_val;
+            return condition ? true_val : false_val;
```
mathbunnyru, line 34: I think this is significantly easier to read and doesn't deal with arithmetics at all (which was giving warnings in some cases).

libmamba/src/core/activation.cpp
```diff
@@ -554,11 +554,11 @@ namespace mamba
         auto conda_environment_env_vars = get_environment_vars(prefix);
 
         // TODO check with conda if that's really what's supposed to happen ...
-        std::remove_if(
+        void(std::remove_if(
```
mathbunnyru, line 557: `remove_if` is marked as `[[nodiscard]]` (at least in one of my stdlibs)
There is already `TODO` above to treat this properly, so the easiest way to silence the warning is to use `void(EXPRESSION)`

libmamba/src/core/thread_utils.cpp
```diff
@@ -29,6 +29,25 @@ namespace mamba
 #ifndef _WIN32
     namespace
     {
+// `sigaddset` might be implemented as a macro calling `__sigbits(int)` function
+// At the same time `sigset_t` might be `unsigned int`
+// This causes compiler warning
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wsign-conversion"
```
mathbunnyru, line 36: I added comment explaining why this happens above
This doesn't happen in our code, but in the system code
But because `sigaddset` might be implemented as a macro, it's treated as our code.

Because, there is no bug, I completely disable this warning/error.
But I didn't want to disable it for the whole file, so I created a small function, where I applied this behaviour.

libmamba/src/solver/libsolv/matcher.cpp
```diff
@@ -193,7 +193,7 @@ namespace mamba::solver::libsolv
             return { std::cref(it->second) };
         }
 
-        return specs::Channel::resolve(std::move(uc), channel_params())
+        return specs::Channel::resolve(uc, channel_params())
```
mathbunnyru, line 196: It's `const specs::UnresolvedChannel& uc`, so no point in `std::move(uc)`

libmambapy/CMakeLists.txt
```diff
@@ -32,6 +32,14 @@ pybind11_add_module(
     src/libmambapy/bindings/solver.cpp
     src/libmambapy/bindings/solver_libsolv.cpp
 )
+# TODO: remove when `SubdirData::cache_path()` is removed
+if(CMAKE_CXX_COMPILER_ID STREQUAL "Clang" OR CMAKE_CXX_COMPILER_ID STREQUAL "AppleClang")
+    # This file uses capturing structured bindings, which was fixed in C++20
+    set_source_files_properties(
+        src/libmambapy/bindings/legacy.cpp
+        PROPERTIES COMPILE_FLAGS -Wno-error=deprecated-declarations
```
mathbunnyru, line 40: I turned this back to warning. Sometimes it's useful to see when we're using deprecated functions, so let's have it as a warning.

libmamba/ext/solv-cpp/src/pool.cpp
```diff
@@ -93,10 +93,17 @@ namespace solv
 
     namespace
     {
+// This function is only used in `assert()` expressions
+// That's why it might get reported as unused in Release builds
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wunused-function"
```
mathbunnyru, line 99: I explained above why this happens.
It's completely ok to have a function which is only used in `assert` statements.
So, I silence this warning/error

cmake/WrongCompilerOptions.cmake
```diff
@@ -0,0 +1,19 @@
+if(APPLE)
+    # It seems that there are some flags added to CXXFLAGS during `conda-build`. It might be a bug
+    # in https://github.com/conda-forge/clang-compiler-activation-feedstock. One way to fix this is
+    # to turn these errors to warnings or silence them (using `-Wno-unused-command-line-argument`
+    # and `-Wno-ignored-optimization-argument`)
+
+    # But we will go another way: we will remove wrong options from CMAKE_CXX_FLAGS
```
mathbunnyru, line 7: This is the most tricky part.

I think we have some flags, which are not part of this project and we inherit it during `conda-build`.

This is how current `main` looks like: https://github.com/mamba-org/mamba/actions/runs/11896974442/job/33150163684#step:7:1203

```
x86_64-apple-darwin13.4: warning: -framework CoreFoundation: 'linker' input unused [-Wunused-command-line-argument]
x86_64-apple-darwin13.4: warning: -framework CoreServices: 'linker' input unused [-Wunused-command-line-argument]
x86_64-apple-darwin13.4: warning: -framework Security: 'linker' input unused [-Wunused-command-line-argument]
x86_64-apple-darwin13.4: warning: -framework Kerberos: 'linker' input unused [-Wunused-command-line-argument]
x86_64-apple-darwin13.4: warning: optimization flag '-fno-merge-constants' is not supported [-Wignored-optimization-argument]
```

I think it comes from here `INFO: activate_clangxx_osx-64.sh made the following environmental changes:`:
https://github.com/mamba-org/mamba/actions/runs/11896974442/job/33150163684#step:7:1123

I'm not 100% sure about this and might be completely wrong.
If someone more experienced could take a look at this, it would be great.

mathbunnyru: Trying another approach right now

mathbunnyru: It worked; this one didn't due to some strange linking errors.

CMakeLists.txt
```diff
@@ -19,7 +19,11 @@ option(BUILD_LIBMAMBA_TESTS "Build libmamba C++ tests" OFF)
 option(BUILD_MAMBA "Build mamba" OFF)
 option(BUILD_MICROMAMBA "Build micromamba" OFF)
 option(BUILD_MAMBA_PACKAGE "Build mamba package utility" OFF)
-option(MAMBA_WARNING_AS_ERROR "Treat compiler warnings as errors" OFF)
+if(MSVC)
+    option(MAMBA_WARNING_AS_ERROR "Treat compiler warnings as errors" OFF)
+else()
+    option(MAMBA_WARNING_AS_ERROR "Treat compiler warnings as errors" ON)
```
jjerphan, line 25: I think this is a good idea even if this will probably will make some untested toolchain fail.

jjerphan (commented):
LGTM, modulo a few comments.

libmamba/src/core/activation.cpp
Klaim: This code is suspicious: either the intent was to move the elements at the end of the container, in which case the elements are not removed and the `push_back` calls that follows will add new eleemnts after the "removed" elements; or it was intended as a remove-erase, in wich case this code doesnt match the intent and hiding the warnings hides the bug.
I'm not sure which one it is here (I'm not familiar with the exact meaning of that code), @JohanMabille or @Hind-M  any idea?

libmamba/CMakeLists.txt
Klaim: Is there any way to make that change as part of the source code instead of here (like disabling warnings)?

libmambapy/CMakeLists.txt
Klaim: It makes sense for that code or even the whole library. 👍🏽

cmake/CompilerWarnings.cmake
```diff
@@ -109,6 +109,16 @@ function(mamba_target_add_compile_warnings target)
         -Wuninitialized
     )
 
+    # It seems that there are some wrong flags added to CXXFLAGS during `conda-build`. And then
+    # these flags are used to compile our source code. It might be a bug in
+    # https://github.com/conda-forge/clang-compiler-activation-feedstock.
```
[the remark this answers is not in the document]
mathbunnyru: I rewrote the comment and added more details.

**Note, this is not caused by my PR, it already happens in `main`**
But because we use warnings, not errors, no one noticed:
https://github.com/mamba-org/mamba/actions/runs/11935345040/job/33266465568#step:7:1202

Unfortunately, I couldn't understand exactly why this happens.
But I noticed `INFO: activate_clangxx_osx-64.sh made the following environmental changes:` a few lines above.
And then we have lines showing erroneous changes in `CXXFLAGS`.
That's why I think it happens in the mentioned lib.

If someone wants to dig deeper, that would be great.

libmamba/CMakeLists.txt
mathbunnyru: Yes, fortunately it happens in functions located near each other, so we can only write pragmas once.
Thanks!

libmambapy/CMakeLists.txt
mathbunnyru: For now, let's keep it for this file. If needed later, we can add this flag to the whole repo/lib I suppose.
It's better not to use deprecated functions, when possible.

libmamba/src/core/activation.cpp
Klaim: Ok now that it's merged I expect the warning to disappear

mathbunnyru: Yes, it disappeared and I merged latest `main` to my branch

mathbunnyru:
@jjerphan @Klaim I think I fixed all the comments (and a bit more), so please, take a look.

Klaim (approved):
LGTM 👍🏽

cmake/CompilerWarnings.cmake
```diff
@@ -109,6 +109,21 @@ function(mamba_target_add_compile_warnings target)
         -Wuninitialized
     )
 
+    # It seems that these flags are leaked to CXXFLAGS: `-framework CoreFoundation` `-framework
+    # CoreServices` `-framework Security` `-framework Kerberos` `-fno-merge-constants`
```
jjerphan, line 113: How is this comment related to the elements bellow? Can you precise where those are leaked? Is it when you build locally on macOS?

See also [those lines on the feedstock which make use of them](https://github.com/conda-forge/micromamba-feedstock/blob/326c9d4b3f78189f48b00771388a15ef1528e031/recipe/build.sh#L10-L14).

The use and presence of those flags must be clarified at a point, but this is orthogonal to the goal of this PR to me.

jjerphan (approved):
LGTM.

libmamba/CMakeLists.txt
JohanMabille: Technically, this construction is C++20, not C++17. It works with gcc and msvc though, and the warning is still here with Clang, even if we pass the C++20 option (that's a known bug), so we can definitely hide this one.

PR #3614, opened 2024-11-19
maint: Add dependabot config for GitHub workflows/actions

mathbunnyru:
This is quite an easy to use tool which will create PRs when there are new versions of actions used in workflows/actions


JohanMabille (approved):

JohanMabille:
Thanks ;)

Issue #2008, opened 2022-10-11
[micromamba] `micromamba env export` does not show packages installed with `pip`


zacharyburnett:
I created an environment with `micromamba`:
```shell
micromamba create -n mamba_export_test
```
and then installed a package with `pip`:
```
micromamba activate mamba_export_test
pip install requests
```
However, when running `micromamba env export`, the packages installed with `pip` are not shown:
```yaml
name: mamba_export_test
channels:
- https://repo.anaconda.com/pkgs/main
dependencies:
- python=3.10.6=hdfd78df_0
- pip=22.2.2=py310hecd8cb5_0
- xz=5.2.6=hca72f7f_0
- ncurses=6.3=hca72f7f_3
- readline=8.1.2=hca72f7f_1
- ca-certificates=2022.07.19=hecd8cb5_0
- zlib=1.2.12=h4dc903c_3
- libcxx=14.0.6=h9765a3e_0
- bzip2=1.0.8=h1de35cc_0
- wheel=0.37.1=pyhd3eb1b0_0
- tk=8.6.12=h5d9f67b_0
- libffi=3.3=hb1e8313_2
- tzdata=2022c=h04d1e81_0
- openssl=1.1.1q=hca72f7f_0
- certifi=2022.9.24=py310hecd8cb5_0
- sqlite=3.39.3=h707629a_0
- setuptools=63.4.1=py310hecd8cb5_0
```
In contrast, `conda env export` shows the `pip` packages:
```yaml
name: /Users/zburnett/micromamba/envs/mamba_export_test
channels:
  - defaults
dependencies:
  - bzip2=1.0.8=h1de35cc_0
  - ca-certificates=2022.07.19=hecd8cb5_0
  - certifi=2022.9.24=py310hecd8cb5_0
  - libcxx=14.0.6=h9765a3e_0
  - libffi=3.3=hb1e8313_2
  - ncurses=6.3=hca72f7f_3
  - openssl=1.1.1q=hca72f7f_0
  - pip=22.2.2=py310hecd8cb5_0
  - python=3.10.6=hdfd78df_0
  - readline=8.1.2=hca72f7f_1
  - setuptools=63.4.1=py310hecd8cb5_0
  - sqlite=3.39.3=h707629a_0
  - tk=8.6.12=h5d9f67b_0
  - tzdata=2022c=h04d1e81_0
  - wheel=0.37.1=pyhd3eb1b0_0
  - xz=5.2.6=hca72f7f_0
[...]

PR #3623, opened 2024-11-20
feat: List PyPI packages in environment export

jjerphan:
Fix https://github.com/mamba-org/mamba/issues/2008.


micromamba/tests/test_env.py
```diff
@@ -484,3 +484,33 @@ def test_env_update_empty_base(tmp_home, tmp_root_prefix, tmp_path):
     packages = helpers.umamba_list("-p", env_prefix, "--json")
     assert any(package["name"] == "xtensor" for package in packages)
     assert any(package["name"] == "python" for package in packages)
+
+
+@pytest.mark.parametrize("shared_pkgs_dirs", [True], indirect=True)
+@pytest.mark.parametrize("json_flag", [None, "--json"])
+def test_env_export_with_pip(tmp_home, tmp_root_prefix, tmp_path, json_flag):
```
jjerphan, line 491: Is this test sufficient, or should we test many more cases?

micromamba/tests/test_env.py
```diff
@@ -484,3 +484,33 @@ def test_env_update_empty_base(tmp_home, tmp_root_prefix, tmp_path):
...
+    env_name = "env-list_with_pip"
+    tmp_root_prefix / "envs" / env_name
```
mathbunnyru, line 493: This line does nothing, right?
I've also seen it used in other functions and other test files, so I guess we should deal with it, and probably in a separate PR

micromamba/src/env.cpp
```diff
@@ -303,10 +320,19 @@ set_env_command(CLI::App* com, Configuration& config)
                         channels.insert(chan.display_name());
                     }
                 }
+                // Add a `pip` subsection in `dependencies` listing wheels installed from PyPI
+                if (pip_versions_map.size() > 0)
```
mathbunnyru, line 324:
```suggestion
                if (!pip_versions_map.empty())
```

micromamba/tests/test_env.py
mathbunnyru: LGTM.

Just to make sure - you checked that the json/yaml looks beautiful in the output?
I think tests don't check this at all.

micromamba/tests/test_env.py
```diff
@@ -484,3 +484,33 @@ def test_env_update_empty_base(tmp_home, tmp_root_prefix, tmp_path):
...
+
+    env_file_yml = tmp_path / "test_env_yaml_content_to_install_numpy_with_pip.yaml"
+    env_file_yml.write_text(env_yaml_content_create_pip_pkg_with_version)
+
+    flags = list(filter(None, [json_flag]))
+    helpers.create("-n", env_name, *flags, no_dry_run=True)
+    helpers.install("-n", env_name, "-f", env_file_yml, *flags, no_dry_run=True)
```
mathbunnyru, line 500:
```suggestion
    helpers.create("-n", env_name, "-f", env_file_yml)
```

Couldn't we just have this simple command?
Also, I wouldn't use `*flags` here, because we care about parametrisation of `export` in this test, and it's flags are independent and may be aded later to check more cases

micromamba/tests/test_env.py
```diff
@@ -484,3 +484,33 @@ def test_env_update_empty_base(tmp_home, tmp_root_prefix, tmp_path):
...
+
+    ret = helpers.run_env("export", "-n", env_name, *flags)
+
+    # JSON is already parsed
+    ret = yaml.unsafe_load(ret) if json_flag is None else ret
```
mathbunnyru, line 505:
```suggestion
    output = helpers.run_env("export", "-n", env_name, *flags)

    # json is already parsed
    ret = output if json_flag else yaml.safe_load(output)
```

I would prefer having this the same as in other tests
And I don't understand to be honest - why you had to use `unsafe_load`?

micromamba/tests/test_env.py
```diff
@@ -484,3 +484,33 @@ def test_env_update_empty_base(tmp_home, tmp_root_prefix, tmp_path):
...
+
+    assert ret["name"] == env_name
+    assert env_name in ret["prefix"]
+    assert set(ret["channels"]) == {"conda-forge"}
+    assert any("pip" in dep for dep in ret["dependencies"])
```
mathbunnyru, line 510: This is not working 100% as you expect, because this will also be ok if there is a package with a `pip` substring in it

micromamba/tests/test_env.py
```diff
@@ -484,3 +484,33 @@ def test_env_update_empty_base(tmp_home, tmp_root_prefix, tmp_path):
...
+
+    # Check there's a dictionary with a `pip` entry in `dependencies` and that numpy is part of it.
+    pip_section_index = next(
+        i for i, dep in enumerate(ret["dependencies"]) if isinstance(dep, dict) and "pip" in dep
+    )
+    assert ret["dependencies"][pip_section_index] == {"pip": ["numpy==1.26.4"]}
```
mathbunnyru, line 516:
```suggestion
    pip_section = next(
        dep for dep in ret["dependencies"] if isinstance(dep, dict) and ["pip"] == [*dep]
    )
    assert pip_section["pip"] == ["numpy==1.26.4"]
```

And I think we can remove the assert before that

mathbunnyru (commented):
Overall, LGTM, but left a few comments

micromamba/src/env.cpp
```diff
@@ -227,6 +228,21 @@ set_env_command(CLI::App* com, Configuration& config)
                         channels.insert(chan.display_name());
                     }
                 }
+
+                // Add a `pip` subsection in `dependencies` listing wheels installed from PyPI
+                if (pip_versions_map.size() > 0)
```
JohanMabille, line 233:
```suggestion
                if (!pip_versions_map.empty())
```

micromamba/src/env.cpp
```diff
@@ -227,6 +228,21 @@ set_env_command(CLI::App* com, Configuration& config)
...
+                {
+                    dependencies << (first_dependency_printed ? ",\n" : "") << "     { \"pip\": [\n";
+                    first_dependency_printed = false;
+                    for (const auto& [k, v] : pip_versions_map)
+                    {
+                        dependencies << (first_dependency_printed ? ",\n" : "") << "      \""
+                                     << v.name << "==" << v.version << "\"";
+                        first_dependency_printed = true;
+                    }
+                    dependencies << "\n    ]\n    }";
```
JohanMabille, line 243: Consider using '\t' instead of 4 spaces.

jjerphan: conda is using spaces over tabs everywhere (even for 4 spaces), so I guess we better stick with them not to mix outputs with spaces and tabs. What do you think?

micromamba/tests/test_env.py
jjerphan: Yes, this can be simplified. Some logic of old tests have been ported to other tests but this is unnecessary.

micromamba/tests/test_env.py
jjerphan: Yes, the output is properly indented.

For now, it is hard to come up with tests assessing the proper indentation since `helpers.run_env` is returning a dictionary when `--json` is passsed.

Also, I think we should rather introduce something systematic regarding such tests.

micromamba/tests/test_env.py
jjerphan: Mostly left-overs, also because safety is not needed.

micromamba/src/env.cpp
```diff
@@ -227,6 +228,21 @@ set_env_command(CLI::App* com, Configuration& config)
                         channels.insert(chan.display_name());
                     }
                 }
+
+                // Add a `pip` subsection in `dependencies` listing wheels installed from PyPI
+                if (!pip_versions_map.empty() > 0)
```
mathbunnyru, line 233:
```suggestion
                if (!pip_versions_map.empty())
```

micromamba/tests/test_env.py
```diff
@@ -484,3 +484,38 @@ def test_env_update_empty_base(tmp_home, tmp_root_prefix, tmp_path):
...
+- pip:
+  - requests==2.32.3
+"""
+
+
+@pytest.mark.parametrize("json_flag", [None, "--json"])
+def test_env_export_with_pip(tmp_path, json_flag):
+    env_name = "env_export_with_pip"
+
+    env_file_yml = tmp_path / "test_env_yaml_content_to_install_numpy_with_pip.yaml"
```
mathbunnyru, line 503:
```suggestion
    env_file_yml = tmp_path / "test_env_yaml_content_to_install_requests_with_pip.yaml"
```

mathbunnyru (approved):

mathbunnyru (approved):

micromamba/src/env.cpp
JohanMabille: Ah indeed, I thought we were already using "\t" everywhere else. I don't think that would result in mixing tabs and spaces though, since we are not appending anything to output produced by conda, but yeah, let's keep the codebase consistent and use 4 spaces for now.

JohanMabille (approved):

zacharyburnett:
thank you so much for merging this PR! This was a sticking point in my adoption of micromamba, and now is resolved (for my use case).

Issue #3638, opened 2024-11-27
Release Process Improvements (for 2.x)


Klaim:
### Troubleshooting docs

- [X] My problem is not solved in the Troubleshooting docs

### Anaconda default channels

- [X] I do NOT use the Anaconda default channels (pkgs/* etc.)

### How did you install Mamba?

Mambaforge or latest Miniforge

### Search tried in issue tracker

N/A

### Latest version of Mamba

- [X] My problem is not solved with the latest version

### Tried in Conda?

Not applicable

### Describe your issue

We started discussing privately this subject so I'm moving the discussion here for making sure we coordinate and detect potential disagreements.

## Problems

- We need an established/official workflow to publish pre-release versions before actual releases (at least for minor versions), so that we get some user feedbacks before releasing a version. Recent experiments doing that showed how it helped growing a patch version that was actually fixing the reported issues for the related major version. Establishing at least one alpha and/or beta phase as part of the official release process would help limit publishing releases that have major regressions.
- We dont have an established version scheme to handle pre-release versions (semver allows various ways to do this).
- Our release tooling/scripts doesnt handle  correctly the pre-release part of semver.
- `micromamba/mamba --version` with a pre-released version (for example `v2.0.4alpha1`) is not 
- The current release process will generate multiple tag for 1 version, one of the tag being the date of the tagging. I believe it was historically established because of a stronger separation between mamba and libmamba, but current context implies that both behavior are very much related as mamba/micromamba is just a driver for libmamba at this point (it doesnt relay on any conda code anymore). Therefore in the current context, multiple tags for the same version feels overkill. The date tag is also apparently unnecessary (please ping us if you have a use for it).

## Solutions

These are the different points we already briefly discussed and started working on to solve these problems:

- [x] Decide on a pre-release naming scheme: we decided to go with `X.Y.Z-prerelease_type.increment` with the increment separated by a `.` so that it is easy for (future) tooling to compare/order pre-release versions. For example `1.2.3-alpha.0` or `2.1.0-beta.1`
- [ ] Make the release scripts handle (and validate) the new version schemes (I'm working on it)
- [ ] Make `mamba/micromamba --version` display the pre-release part when it is a pre-release version (I'm working on it)
- [ ] Establish a release process that starts with publishing an alpha or beta first (briefly discussed, nothing established yet). This requires some documentation of that process and maybe tweaking the release script to reflect it

### mamba info / micromamba info

```shell
N/A
```

### Logs

```shell
N/A
```

[...]

PR #3639, opened 2024-11-27
`micromamba/mamba --version` displays pre-release version names + establishes pre-release versions name scheme

Klaim:
Part of the effort for #3638 

While I started this just to allow displaying the pre-release versions in `micromamba/mamba --version`, this required also changing the release scripts to provide the relevant information to the source code. As part of that change I setup the discussed pre-release version naming scheme with validation in the  release scripts.

Also deprecates `mamba::version_arr()` and `umamba::version_arr()` functions which seem unusued. If there is a need for information about the version, I'll add another more precise function as a replacement, or we'll just remove these.


Klaim:
Note that this version accepts a minimum of 3 dot-separated fields but does not limit the number of dots over 3, so schemes with and without additional dots are considered "valid" by the validation script.

version_scheme.py
```diff
@@ -0,0 +1,65 @@
...
+
+    def __init__(self, version: str):
+        if "-" in version:
+            raise ValueError(
+                "'{}' is not a valid version name : `-` is reserved for another usage in conda packages version names".format(
+                    version
+                )
+            )
+
+        VALID_VERSION_PRERELEASE_TYPES = ("alpha", "beta")
```
Hind-M, line 16: What about `rc` and `dev`?

Hind-M:
Adding [this](https://conda-forge.org/docs/maintainer/knowledge_base/#pre-release-version-sorting) as a reference.

jjerphan:
> Adding [this](https://conda-forge.org/docs/maintainer/knowledge_base/#pre-release-version-sorting) as a reference.

[This piece of `conda.models.VersionOrder`'s docstring](https://github.com/conda/conda/blob/cc21508563912268649f207723fd5114fa21b906/conda/models/version.py#L115-L143) is also relevant (it is linked in the section Hind shared).

JohanMabille (approved):
This PR does not include the changes so that `umamba --version` shows pre-release versions, right?

Klaim:
> This PR does not include the changes so that `umamba --version` shows pre-release versions, right?

It does (if it doesnt, its a bug in this pr): `--version` is implemented [there](https://github.com/mamba-org/mamba/blob/a85eb16149caecb1d4752acb5d051c44cc5225f6/micromamba/src/umamba.cpp#L36) and [implemented in terms of the define](https://github.com/mamba-org/mamba/blob/a85eb16149caecb1d4752acb5d051c44cc5225f6/micromamba/src/version.cpp#L13) which is [generated by the scripts](https://github.com/mamba-org/mamba/pull/3639/files#diff-f3a4bb0eb70e514458bca8a0e6eae6944ec7e131827491f0a19b3638a452e06dR21) modified [here](https://github.com/mamba-org/mamba/pull/3639/files#diff-a06697bdad36b834f9a2442b621e7ff158633a4e0b7521d23d33d25f19f9b150R27).

In the previous code, the version name was parsed, the major minor and patch number were extracted, then the version string define was re-built from that. In this PR we just print the validated name that was used, so it will be whatever you passed to the script as long as it has been validated.

Note that it's not visible in this PR because I didnt use a pre-release version, but if you want I can use it with `-dev1` as we are in development versions right now, to make it very clear from the CI?

Klaim:
For clarity and because it's one of the intended usage, I added a commit that sets the version to `2.0.5.dev0`
It is important to note that to do that I had to run the release scripts and then remove the generated changelogs. I think it would be useful to modify (in another PR) `release.py` to be called standalone (currently it relies on the modified changelog from `update_changelog.py` and make it allow just changing the version in the generated headers without changing anything else, to be able to switch the version to devN just after a release.
Another future improvement would be to have a way to add the commit id to a dev version, so that we can identify which commit matches the micromamba binary, in case someone uses such binary and reports issue. Anyway that's future improvements.

Klaim:
I just realized that the cmake scripts are taking the version name from the major.minor.patch numbers only, so cmake believes this is 2.0.5. @JohanMabille @jjerphan Should I fix that in this PR or another one?

Klaim:
@JohanMabille Confirmed that version name is complete in `micromamba --version` : https://github.com/mamba-org/mamba/actions/runs/12158203081/job/33905615645?pr=3639#step:10:13

JohanMabille:
> @JohanMabille Confirmed that version name is complete in `micromamba --version` : https://github.com/mamba-org/mamba/actions/runs/12158203081/job/33905615645?pr=3639#step:10:13

Yeah my bad, I misread the code.

JohanMabille:
> I just realized that the cmake scripts are taking the version name from the major.minor.patch numbers only, so cmake believes this is 2.0.5. @JohanMabille @jjerphan Should I fix that in this PR or another one?

I think it can be done in a dedicated PR

Klaim:
Note that I've re-launched already 3 times the failing vcpkg command (which failed so far when trying to acquire a package); not sure what's going on exactly.

JohanMabille:
This was a vcpkg issue, not related to mamba.

Issue #3648, opened 2024-12-02
`mamba uninstall` not recognized as alias to `mamba remove`


fperez:
### Troubleshooting docs

- [X] My problem is not solved in the Troubleshooting docs

### Anaconda default channels

- [X] I do NOT use the Anaconda default channels (pkgs/* etc.)

### How did you install Mamba?

Mambaforge or latest Miniforge

### Search tried in issue tracker

uninstall

### Latest version of Mamba

- [X] My problem is not solved with the latest version

### Tried in Conda?

I do not have this problem with Conda, just with Mamba

### Describe your issue

`mamba uninstall` is [supposed to be an alias for `mamba remove`](https://fig.io/manual/mamba/uninstall):

![image](https://github.com/user-attachments/assets/6e54ba5c-5ba1-45ed-9114-cccddf8cc832)

But it's not recognized as such by the latest version (2.0.4):

```bash
(base) (main)longs[setup]> mamba --version
2.0.4
(base) (main)longs[setup]> mamba uninstall jupytext
The following arguments were not expected: jupytext uninstall
Run with --help for more information.
(base) (main)longs[setup]> mamba remove jupytext

[... WORKS FINE...]
```

This used to work until pretty recently, as that's always been the command I used. While `remove` is nice and maybe shorter, since many other package managers use `uninstall` as their default command, it's burned in my muscle memory. I think compatibility with this alias should be maintained, not only to avoid breaking user muscle memory, but more importantly, potential automated scripts and pipelines that rely on it.

### mamba info / micromamba info

```shell
(base) (main)longs[setup]> mamba info

       libmamba version : 2.0.4
          mamba version : 2.0.4
           curl version : libcurl/8.10.1 OpenSSL/3.4.0 (SecureTransport) zlib/1.3.1 zstd/1.5.6 libssh2/1.11.1 nghttp2/1.64.0
     libarchive version : libarchive 3.7.7 zlib/1.3.1 liblzma/5.2.6 bz2lib/1.0.8 liblz4/1.9.3 libzstd/1.5.6
       envs directories : /Users/fperez/local/conda/envs
          package cache : /Users/fperez/local/conda/pkgs
                          /Users/fperez/.mamba/pkgs
            environment : base (active)
           env location : /Users/fperez/local/conda
      user config files : /Users/fperez/.mambarc
 populated config files : /Users/fperez/.condarc
                          /Users/fperez/local/conda/.condarc
       virtual packages : __unix=0=0
[...]

PR #3650, opened 2024-12-03
fix: Reintroduce the `uninstall` command

jjerphan:
Fix https://github.com/mamba-org/mamba/issues/3648.

### DIFF
--- a/micromamba/src/umamba.cpp
+++ b/micromamba/src/umamba.cpp
@@ -62,13 +62,7 @@ set_umamba_command(CLI::App* com, mamba::Configuration& config)
 
     CLI::App* remove_subcom = com->add_subcommand("remove", "Remove packages from active environment");
     set_remove_command(remove_subcom, config);
-
-    // `uninstall` is an alias for `remove`
-    CLI::App* uninstall_subcom = com->add_subcommand(
-        "uninstall",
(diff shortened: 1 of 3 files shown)


micromamba/src/umamba.cpp
```diff
@@ -63,6 +63,13 @@ set_umamba_command(CLI::App* com, mamba::Configuration& config)
     CLI::App* remove_subcom = com->add_subcommand("remove", "Remove packages from active environment");
     set_remove_command(remove_subcom, config);
 
+    // `uninstall` is an alias for `remove`
+    CLI::App* uninstall_subcom = com->add_subcommand(
+        "uninstall",
+        "Remove packages from active environment"
+    );
+    set_remove_command(uninstall_subcom, config);
```
jjerphan, line 71: Note for reviewers: has CLI11 a feature for aliases or not?

JohanMabille: Yes, you can use `subcom->alias("uninstall");` instead of adding a new subcommand.

JohanMabille (changes requested):
I thing using an alias instead of adding a new subcommand would be better.

JohanMabille (approved):

PR #3662, opened 2024-12-06
fix: Skip empty lines in environment spec files

jjerphan:
Fix https://github.com/mamba-org/mamba/issues/3659.
Fix https://github.com/mamba-org/mamba/issues/3289.

### DIFF
--- a/libmamba/src/api/install.cpp
+++ b/libmamba/src/api/install.cpp
@@ -888,6 +888,8 @@ namespace mamba
                     {
                         throw std::runtime_error(util::concat("Got an empty file: ", file));
                     }
+
+                    // Inferring potential explicit environment specification
                     for (std::size_t i = 0; i < file_contents.size(); ++i)
                     {
                         auto& line = file_contents[i];
@@ -930,34 +932,23 @@ namespace mamba
                         }
                     }
 
-                    std::vector<std::string> f_specs;
-                    for (auto& line : file_contents)
-                    {
-                        auto lstrip_line = util::lstrip(line);
-
-                        // Skip comment lines and empty lines
-                        // Comment lines start with '#' or '@' preceded by whitespaces or tabs
-                        auto is_comment = util::starts_with(lstrip_line, "#")
-                                          || util::starts_with(lstrip_line, "@");
-                        auto is_empty = lstrip_line.empty();
-
-                        if (!is_comment && !is_empty)
-                        {
-                            f_specs.push_back(line);
-                        }
-                    }
-
+                    // If we reach here, we have a file with no explicit env, and the content of the
+                    // file just lists MatchSpecs.
                     if (specs.cli_configured())
                     {
                         auto current_specs = specs.cli_value<std::vector<std::string>>();
-                        current_specs.insert(current_specs.end(), f_specs.cbegin(), f_specs.cend());
+                        current_specs.insert(
+                            current_specs.end(),
+                            file_contents.cbegin(),
+                            file_contents.cend()
+                        );
                         specs.set_cli_value(current_specs);
                     }
                     else
                     {
-                        if (!f_specs.empty())
+                        if (!file_contents.empty())
                         {
-                            specs.set_cli_value(f_specs);
+                            specs.set_cli_value(file_contents);
                         }
                     }
                 }
--- a/libmamba/src/api/install.cpp
+++ b/libmamba/src/api/install.cpp
@@ -933,7 +933,15 @@ namespace mamba
                     std::vector<std::string> f_specs;
                     for (auto& line : file_contents)
                     {
-                        if (line[0] != '#' && line[0] != '@')
+                        auto lstrip_line = util::lstrip(line);
+
+                        // Skip comment lines and empty lines
+                        // Comment lines start with '#' or '@' preceded by whitespaces or tabs
+                        auto is_comment = util::starts_with(lstrip_line, "#")
+                                          || util::starts_with(lstrip_line, "@");
+                        auto is_empty = lstrip_line.empty();
+
+                        if (!is_comment && !is_empty)
                         {
                             f_specs.push_back(line);
                         }
--- a/libmamba/src/core/util.cpp
+++ b/libmamba/src/core/util.cpp
@@ -301,7 +301,8 @@ namespace mamba
                 line.pop_back();
             }
 
-            line = util::strip(line);
+            // Remove leading and trailing whitespace in place not to create a new string.
+            util::inplace_strip(line);
 
             // Skipping empty lines
             if (line.empty())
--- a/libmamba/src/core/util.cpp
+++ b/libmamba/src/core/util.cpp
@@ -301,24 +301,24 @@ namespace mamba
                 line.pop_back();
             }
 
-            const auto lstrip_line = util::lstrip(line);
+            line = util::strip(line);
 
             // Skipping empty lines
-            if (lstrip_line.empty())
+            if (line.empty())
             {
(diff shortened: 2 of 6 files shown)


libmamba/src/api/install.cpp
```diff
@@ -933,7 +933,13 @@ namespace mamba
                     std::vector<std::string> f_specs;
                     for (auto& line : file_contents)
                     {
-                        if (line[0] != '#' && line[0] != '@')
+                        // Skip comment lines and empty lines
+                        // Comment lines start with '#' or '@' preceded by whitespaces or tabs
+                        auto is_comment = util::starts_with(line, "#")
+                                          || util::starts_with(line, "@");
+                        auto is_empty = std::all_of(line.begin(), line.end(), ::isspace);
```
Klaim, line 940: I would prefer that we use `std::isspace` or `mamba::isspace` but not global `isspace`. It's the same as `std::isspace` but we avoid potential surprises by being explicit.

Klaim (approved):
LGTM. 

BTW did you also consider [striping](https://github.com/mamba-org/mamba/blob/main/libmamba/include/mamba/util/string.hpp#L195) every line before even processing them?

Klaim:
I wonder if filtering at [reading time](https://github.com/mamba-org/mamba/pull/3662/files#diff-ebac1cb6982f8e508ff9d8203d56f74aa3da80115d7ca84cc525a131ef23535aR886) would help even more.

micromamba/tests/test_create.py
```diff
@@ -1441,3 +1443,39 @@ def test_create_empty_or_absent_dependencies(tmp_path):
...
+env_spec_empty_lines_and_comments = """
+# The line below are empty (various number of spaces)
+"""
+
+env_spec_empty_lines_and_comments += "\n"
+env_spec_empty_lines_and_comments += "  \n"
+env_spec_empty_lines_and_comments += "    \n"
+env_spec_empty_lines_and_comments += "	\n"
+env_spec_empty_lines_and_comments += "# Un seul check vous manque, \n"
+env_spec_empty_lines_and_comments += "@ et tout est explosé\n"
```
JohanMabille, line 1457: hooked_on_a_feeling

libmamba/src/api/install.cpp
```diff
@@ -933,7 +933,15 @@ namespace mamba
...
+
+                        // Skip comment lines and empty lines
+                        // Comment lines start with '#' or '@' preceded by whitespaces or tabs
+                        auto is_comment = util::starts_with(lstrip_line, "#")
+                                          || util::starts_with(lstrip_line, "@");
+                        auto is_empty = lstrip_line.empty();
+
+                        if (!is_comment && !is_empty)
                         {
                             f_specs.push_back(line);
```
Klaim, line 946: I think it's the stripped version that should be pushed?

libmamba/src/api/install.cpp
```diff
@@ -933,7 +933,15 @@ namespace mamba
                     std::vector<std::string> f_specs;
                     for (auto& line : file_contents)
                     {
-                        if (line[0] != '#' && line[0] != '@')
+                        auto lstrip_line = util::lstrip(line);
+
+                        // Skip comment lines and empty lines
+                        // Comment lines start with '#' or '@' preceded by whitespaces or tabs
+                        auto is_comment = util::starts_with(lstrip_line, "#")
```
Klaim, line 940:
```suggestion
                        const auto is_comment = util::starts_with(lstrip_line, "#")
```

libmamba/src/api/install.cpp
```diff
@@ -933,7 +933,15 @@ namespace mamba
...
+                                          || util::starts_with(lstrip_line, "@");
+                        auto is_empty = lstrip_line.empty();
```
Klaim, line 942:
```suggestion
                        const auto is_empty = lstrip_line.empty();
```

libmamba/src/api/install.cpp
jjerphan: If we want to push the stripped version (a `std::stringview` in this context), we need to reallocate a string, which can be costly. I preferred not to for now.

What do you prefer?

Klaim: I would just re-assign the stripped version in the same string, we dont need the unstripped version and we can (maybe) reuse the memory from the initial string?
Not a big deal anyway.

libmamba/src/core/util.cpp
```diff
@@ -301,6 +301,33 @@ namespace mamba
                 line.pop_back();
             }
 
+            const auto lstrip_line = util::lstrip(line);
```
jjerphan, line 304: @Klaim: would you prefer this
```suggestion
            line = util::lstrip(line);
```

with subsequent modifications?

Klaim: Yes that 's what I meant (the move is unnecessary).

jjerphan: Done in 63348f9.

libmamba/src/util/string.cpp
```diff
@@ -447,6 +447,27 @@ namespace mamba::util
         return strip_if(input, [](Char c) { return !is_graphic(c); });
     }
 
+    void inplace_strip(std::string& input)
+    {
+        // Dedicated implementation for std::string to avoid copies
```
Klaim, line 452: That comment should be in the header.

libmamba/src/util/string.cpp
```diff
@@ -447,6 +447,27 @@ namespace mamba::util
...
+
+        if (start == std::string::npos)
+        {
+            input.clear();
+            return;
+        }
+
+        const auto end = input.find_last_not_of(" \t\n\v\f\r");
+
+        input = input.substr(start, end - start + 1);
```
Klaim, line 468: I think this is fine, but if we discover that this can create another string (not sure), then the alternative is to do 2 `.erase`, starting with the end.
Anyway that's a side-note, I think that's fine.

jjerphan: Do you have an example? Would such a processing qualifies as YAGNI?

Klaim: I meant something like
```c++
input.erase(end, input.end());
input.erase(input.begin(), start);
```
(didnt check if it's correct)

> Would such a processing qualifies as YAGNI?

Yes, I dont think we need to worry about this, it's just a side note.

I was wondering if the implementation of `string x = x.substr(b,e);` is allowed to create a string. The standard spec says this is equivalent to `x.assign(x.substr(b, e));` and `assign` which is linear complexity in that case. I would expect all implementations just overwrite without new allocations, therefore this code should be fine. 
My comment is just a side note in case we realize later that it's not doing what we expect (and then the function is not working in-place as advertised).

JohanMabille (approved):

micromamba/tests/test_create.py
```diff
@@ -1441,3 +1443,55 @@ def test_create_empty_or_absent_dependencies(tmp_path):
...
+# comment
+"""
+
+
+@pytest.mark.parametrize("env_spec", [env_spec_empty_lines_and_comments, env_repro_1, env_repro_2])
+def test_create_with_empty_lines_and_comments(tmp_path, env_spec):
+    # Non-regression test for:
+    #  - https://github.com/mamba-org/mamba/issues/3289
+    #  - https://github.com/mamba-org/mamba/issues/3659
+    memory_limit = 100  # in MB
```
AntoinePrv, line 1479: @jjerphan could we consider increasing to 150? It does not work consistently on my machine (~116mB).

PR #3664, opened 2024-12-07
maint: Remove anonymous namespace in tests

mathbunnyru:


libmamba/tests/src/solver/test_problems_graph.cpp
```diff
@@ -33,56 +33,53 @@
...
+    REQUIRE_FALSE(c.has_conflict(1));
+    REQUIRE_FALSE(c.in_conflict(1, 1));
+    REQUIRE_FALSE(c.in_conflict(1, 2));
+    REQUIRE_FALSE(c.in_conflict(3, 1));
 }
 
+using namespace mamba::specs::match_spec_literals;
+using Request = solver::Request;
+
 namespace
```
mathbunnyru, line 81: I added this one because `mpkg` function is also present in another test file (and I added anon namespace there as well)

libmamba/tests/src/util/test_string.cpp
```diff
@@ -15,581 +15,557 @@
...
-            REQUIRE(concat_dedup_splits("test/chan", "chan/foo", '/') == "test/chan/foo");
-        }
+        std::string x(strip("abc"));
+        REQUIRE(x == "abc");
+        std::string y(rstrip("abc"));
+        REQUIRE(y == "abc");
+        std::string z(lstrip("abc"));
+        REQUIRE(z == "abc");
     }
-}  // namespace mamba
```
mathbunnyru, line 595: This comment seems to be wrong, I think it was 2 nested anonymous namespaces

jjerphan:
Naive question: can you explain what is the plus-value of those changes whose diff is large an noisy?

mathbunnyru:
> Naive question: can you explain what is the plus-value of those changes whose diff is large an noisy?

Well, not much, but it makes tests look more consistent (and a bit shorter).
I'm completely OK closing this PR.

I told I would work on this when I was working on Catch2, so delivering my promise: https://github.com/mamba-org/mamba/pull/3618#discussion_r1862289092

JohanMabille:
In that case I would find it safer to replace anonymous namespaces with named namespaces instead of removing them. This could lead to duplicated symbols in the future because of tests with very gneric names (constructor, swap, etc).

Keeping anonymous namespaces is fine too.

jjerphan:
I would also keep the anonymous namespace (and thus close this PR).

We could name them, but I would not act unless we see a real benefit or unless it fixes problems.

mathbunnyru:
Ok, closing this PR

PR #3667, opened 2024-12-09
maint: Cleanup CMake files and delete not compiled files

mathbunnyru:
I did the following:
- sorted the order files are mentioned in cmake files
- added missing entries
- deleted files which are not compiled at all (I guess we don't need them?)


libmamba/CMakeLists.txt
```diff
@@ -162,8 +162,8 @@ set(
     ${LIBMAMBA_SOURCE_DIR}/util/path_manip.cpp
     ${LIBMAMBA_SOURCE_DIR}/util/random.cpp
     ${LIBMAMBA_SOURCE_DIR}/util/string.cpp
-    ${LIBMAMBA_SOURCE_DIR}/util/url.cpp
     ${LIBMAMBA_SOURCE_DIR}/util/url_manip.cpp
+    ${LIBMAMBA_SOURCE_DIR}/util/url.cpp
```
mathbunnyru, line 166: This is the correct sort order, and this is how it shows up in the tree list in VS Code, for example

libmamba/CMakeLists.txt
```diff
@@ -301,36 +301,38 @@ set(
...
     ${LIBMAMBA_INCLUDE_DIR}/mamba/util/url_manip.hpp
+    ${LIBMAMBA_INCLUDE_DIR}/mamba/util/url.hpp
     ${LIBMAMBA_INCLUDE_DIR}/mamba/util/variant_cmp.hpp
     ${LIBMAMBA_INCLUDE_DIR}/mamba/util/weakening_map.hpp
     # Implementation of version and matching specs
     ${LIBMAMBA_INCLUDE_DIR}/mamba/specs/archive.hpp
     ${LIBMAMBA_INCLUDE_DIR}/mamba/specs/authentication_info.hpp
     ${LIBMAMBA_INCLUDE_DIR}/mamba/specs/build_number_spec.hpp
     ${LIBMAMBA_INCLUDE_DIR}/mamba/specs/channel.hpp
+    ${LIBMAMBA_INCLUDE_DIR}/mamba/specs/chimera_string_spec.hpp
```
mathbunnyru, line 324: This is an example of missing line. Everything will work except that VS Code will give errors when you open this file - it annoyed me, so I decided to create this PR

libmamba/include/mamba/core/mirror.hpp
```diff
@@ -1,182 +0,0 @@
-// Copyright (c) 2023, QuantStack and Mamba Contributors
```
mathbunnyru, line 1: This file is not included by anyone, and both it and .cpp file are not mentioned in CMake, so I think it's safe to remove it.
Please, make sure I'm right

micromamba/src/microserver.cpp
```diff
@@ -1,414 +0,0 @@
-// Copyright (c) Alex Movsisyan
```
mathbunnyru, line 1: This file and `server.cpp` are not mentioned in CMake at all, so I think it's safe to remove it.
Please, make sure I'm right.

libmamba/ext/solv-cpp/tests/CMakeLists.txt
```diff
@@ -11,6 +11,7 @@ add_executable(
     src/main.cpp
     src/msvc_catch_string_view.cpp
     src/pool_data.cpp
+    src/pool_data.hpp
```
mathbunnyru, line 14: There is no separation for source and header files here, so we should mention this header file as well.

mathbunnyru: Yes.
And it also makes it consistent (all source/header files are present in CMake files).

mathbunnyru:
> LGTM.
> 
> We need to add code coverage to see any potential dead code which may exist, see #2866.
> 
> I just have one minor comment.

Thanks! Replied to your comment.

I agree that having coverage would be nice, but I don't think it would help here.
As far as I know, it only works for source/header files that are actually compiled.
So, if you have a header file, which is not included anywhere, or source file, which is not compiled, it won't not work.
(I might be wrong here)

jjerphan:
I agree with you and I meant that adding code coverage is an addiction to dead code elimination. :slightly_smiling_face:

mathbunnyru:
@jjerphan I think this is ready to be merged as well 🙂

jjerphan:
For this one, I would wait for another review of the team.

micromamba/src/server.cpp
JohanMabille, line 1: This file is still used in [micromamba's CMakeLists.txt](https://github.com/mamba-org/mamba/blob/main/micromamba/CMakeLists.txt#L48-L51), and it includes microserver.cpp, so I don't think we should remove them for now.

JohanMabille (changes requested):

micromamba/src/server.cpp
mathbunnyru: Just as a note: these files are not compiled in any way right now, and their compilation is significantly broken.

mathbunnyru: So, I think, if we want to keep these files, we need to fix the compilation and add it to the CI.
Otherwise we'll just have broken files in the repo, which no one is able to use easily.

mathbunnyru: Created an issue: https://github.com/mamba-org/mamba/issues/3677

JohanMabille: @mathbunnyru I agree; IIRC we initially wanted to remove these files and then realized we need them for another project (I think Quetz). We kept them for reference because we need to refactor them (ok, maybe it will be a full rewrite, but we still need to keep them for now), but did not have the bandwidth to do it. We'll probably tackle this in 2025.

JohanMabille (approved):

JohanMabille:
Thanks for the cleanup!

Issue #3666, opened 2024-12-08
Build regression with GCC 14 when moving to libmamba 2.0.4


dannf:
### Troubleshooting docs

- [X] My problem is not solved in the Troubleshooting docs

### Anaconda default channels

- [X] I do NOT use the Anaconda default channels (pkgs/* etc.)

### How did you install Mamba?

Other (please describe)

### Search tried in issue tracker

maybe-uninitialized

### Latest version of Mamba

- [X] My problem is not solved with the latest version

### Tried in Conda?

Not applicable

### Describe your issue

libmamba 2.0.4 no longer builds with GCC 14, as observed using wolfi-os's toolchain. See the error below. This is a regression from 2.0.3, and was introduced by commit 0a8128d23eef2d23abacd194692e112f1c25b820. Of course, we can work around this by adding `-Wno-error=maybe-uninitialized` to `CXXFLAGS`. But perhaps this should be added upstream so that each impacted downstream needn't work around this independently.

```
[...]

PR #3668, opened 2024-12-09
build: Ignore one `maybe-uninitialized` warning

jjerphan:
#3666.


JohanMabille:
We should fix the uninitialized variables instead of deactivating the check. I have a branch ready for that, I'm gonna push it.

Issue #3660, opened 2024-12-05
Environment names containing periods not shown correctly in bash command prompt


ilekka:
### Troubleshooting docs

- [X] My problem is not solved in the Troubleshooting docs

### Anaconda default channels

- [x] I do NOT use the Anaconda default channels (pkgs/* etc.)

### How did you install Mamba?

Mambaforge or latest Miniforge

### Search tried in issue tracker

prompt, PS1

### Latest version of Mamba

- [X] My problem is not solved with the latest version

### Tried in Conda?

I do not have this problem with Conda, just with Mamba

### Describe your issue

I often use (perhaps foolishly) environment names containing periods. After updating to mamba 2.0, such names are not displayed correctly in the shell command prompt: the last period and everything after it is left out. For example, activating an environment named `py3.10` or `np2.1.3` causes `(py3)` or `(np2.1)` to be prepended to the prompt.

### mamba info / micromamba info

```shell
libmamba version : 2.0.4
          mamba version : 2.0.4
           curl version : libcurl/8.10.1 OpenSSL/3.4.0 zlib/1.3.1 zstd/1.5.6 libssh2/1.11.1 nghttp2/1.64.0
     libarchive version : libarchive 3.7.7 zlib/1.3.1 liblzma/5.2.6 bz2lib/1.0.8 liblz4/1.9.3 libzstd/1.5.6
       envs directories : /home/ilekka/miniforge3/envs
          package cache : /home/ilekka/miniforge3/pkgs
                          /home/ilekka/.mamba/pkgs
            environment : py3.10 (active)
           env location : /home/ilekka/miniforge3/envs/py3.10
      user config files : /home/ilekka/.mambarc
 populated config files : /home/ilekka/.condarc
                          /home/ilekka/miniforge3/.condarc
       virtual packages : __unix=0=0
                          __linux=6.1.0=0
                          __glibc=2.36=0
                          __archspec=1=x86_64_v3
               channels : https://conda.anaconda.org/conda-forge/linux-64
                          https://conda.anaconda.org/conda-forge/noarch
       base environment : /home/ilekka/miniforge3
               platform : linux-64
```

### ~/.condarc

```yaml
auto_activate_base: false
```

ilekka:
I found something that looks like it may be the source of this behavior.

In the function below, `prefix.stem()` returns the last component of the file path stripped of the file extension, i.e. the part beginning from the last period.

[...]

Issue #2680, opened 2023-07-15
Environment names containing '.' get cut out on the shell prompt


flowln:
### Troubleshooting docs

- [X] My problem is not solved in the Troubleshooting docs

### How did you install Mamba?

Micromamba

### Search tried in issue tracker

dot,cut,prompt

### Latest version of Mamba

- [X] My problem is not solved with the latest version

### Tried in Conda?

I didn't try

### Describe your issue

When an environment is active in a shell session, it shows up in that shell's prompt. However, when that environments name contains a dot character, it gets cut of.

For instance, here I have an environment called `pyside2_5.13`, which when active, shows up like that:

Fish:
![environment name shows up as pyside5_5 instead of the correct pyside5_5.13](https://github.com/mamba-org/mamba/assets/9145768/a0e80edf-4166-4e6d-b8e3-f32aafbe15c1)

Bash:
![environment name shows up as pyside5_5 instead of the correct pyside5_5.13](https://github.com/mamba-org/mamba/assets/9145768/6d32da95-4de2-4313-b85f-446b5a8781fb)

Possibly caused by this use of `.stem()`, though it's only a guess of mine: https://github.com/mamba-org/mamba/blob/b3b6cd6d07f8badf49f7f746e624a23d024baba8/libmamba/src/core/activation.cpp#L199

### mamba info / micromamba info

```shell
libmamba version : 1.4.9
     micromamba version : 1.4.9
           curl version : libcurl/7.88.1 OpenSSL/3.1.1 zlib/1.2.13 zstd/1.5.2 libssh2/1.11.0 nghttp2/1.52.0
     libarchive version : libarchive 3.6.2 zlib/1.2.13 bz2lib/1.0.8 libzstd/1.5.2
       envs directories : /home/low/micromamba/envs
          package cache : /home/low/micromamba/pkgs
                          /home/low/.mamba/pkgs
            environment : pyside2_5.13 (active)
           env location : /home/low/micromamba/envs/pyside2_5.13
      user config files : /home/low/.mambarc
 populated config files : /home/low/.mambarc
       virtual packages : __unix=0=0
                          __linux=6.3.1=0
                          __glibc=2.37=0
                          __archspec=1=x86_64
               channels : https://conda.anaconda.org/conda-forge/linux-64
                          https://conda.anaconda.org/conda-forge/noarch
                          https://conda.anaconda.org/nodefaults/linux-64
                          https://conda.anaconda.org/nodefaults/noarch
[...]

PR #3670, opened 2024-12-09
fix: Activated environment name

jjerphan:
Fix https://github.com/mamba-org/mamba/issues/3660.
Fix https://github.com/mamba-org/mamba/issues/2680.

cc @ilekka

### DIFF
--- a/libmamba/tests/src/core/test_activation.cpp
+++ b/libmamba/tests/src/core/test_activation.cpp
@@ -16,5 +16,20 @@ namespace mamba
             // std::endl; std::cout << a.activate("/home/wolfv/miniconda3/", false) <<
             // std::endl;
         }
+
+        TEST_CASE("Activator::get_default_env")
+        {
+            ContextOptions opts;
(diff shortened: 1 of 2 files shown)


JohanMabille (approved):

libmamba/tests/src/core/test_activation.cpp
```diff
@@ -16,5 +16,24 @@ namespace mamba
             // std::endl; std::cout << a.activate("/home/wolfv/miniconda3/", false) <<
             // std::endl;
         }
+
+        TEST_CASE("Activator::get_default_env")
+        {
+            ContextOptions opts;
+            Context ctx(opts);
```
Klaim, line 23: No need to explicitly pass options when empty, [it's already the case by default](https://github.com/mamba-org/mamba/blob/main/libmamba/include/mamba/core/context.hpp#L262).

Klaim (approved):
LGTM

Issue #3047, opened 2023-12-07
Shell completion instructions do not work but shell completion works anyways.


TheSkyentist:
### Troubleshooting docs

- [X] My problem is not solved in the Troubleshooting docs

### Anaconda default channels

- [X] I do NOT use the Anaconda default channels (pkgs/* etc.)

### How did you install Mamba?

Other (please describe)

### Search tried in issue tracker

Searched but none found.

### Latest version of Mamba

- [X] My problem is not solved with the latest version

### Tried in Conda?

I didn't try

### Describe your issue

Fresh install of micromamba with a bash shell. `mamba shell completion` returns an error saying the prefix does not exist. Despite this, `micromamba activate <TAB>` fills in the environment (before and after running the reccomended setup for shell completion). 
Also tested on macOS with micromamba 1.5.4 and zsh. 

Perhaps related to a separate issue where 
`conda env remove -n <TAB>` completes but `conda env remove --name <TAB>` does not.

### mamba info / micromamba info

```shell
libmamba version : 1.5.3
     micromamba version : 1.5.3
           curl version : libcurl/8.4.0 OpenSSL/3.1.4 zlib/1.2.13 zstd/1.5.5 libssh2/1.11.0 nghttp2/1.58.0
     libarchive version : libarchive 3.7.2 zlib/1.2.13 bz2lib/1.0.8 libzstd/1.5.5
       envs directories : /home/ec2-user/micromamba/envs
          package cache : /home/ec2-user/micromamba/pkgs
                          /home/ec2-user/.mamba/pkgs
            environment : None (not found)
           env location : -
      user config files : /home/ec2-user/.mambarc
 populated config files : /home/ec2-user/.mambarc
       virtual packages : __unix=0=0
                          __linux=6.1.61=0
                          __glibc=2.34=0
                          __archspec=1=x86_64-v4
               channels : https://conda.anaconda.org/conda-forge/linux-64
                          https://conda.anaconda.org/conda-forge/noarch
       base environment : /home/ec2-user/micromamba
               platform : linux-64
```

### Logs

```shell
critical libmamba The given prefix does not exist: "/home/ec2-user/micromamba/envs/completion"
```

### ~/.condarc

```yaml
channels:
  - conda-forge
changeps1: false
channel_priority: strict
```

fabcor-maxiv:
Indeed:

```none
$ micromamba shell completion
critical libmamba The given prefix does not exist: "/home/user/.local/share/mamba/envs/completion"
```

PR #3672, opened 2024-12-09
docs: Adapt shell completion subsection

jjerphan:
Fix https://github.com/mamba-org/mamba/issues/3047.

Shell completion is available after the shell is initialized.


JohanMabille:
but mamba shell completion is still failing, right?

JohanMabille (approved):

Issue #3703, opened 2024-12-19
On certain versions of Windows, `Invoke-WebRequest -Uri https://micro.mamba.pm/install.ps1` fails without `-UseBasicParsing`


jjrob:
### Troubleshooting docs

- [X] My problem is not solved in the Troubleshooting docs

### Anaconda default channels

- [X] I do NOT use the Anaconda default channels (pkgs/* etc.)

### How did you install Mamba?

Micromamba

### Search tried in issue tracker

Did not find anything

### Latest version of Mamba

- [X] My problem is not solved with the latest version

### Tried in Conda?

Not applicable

### Describe your issue

This issue suggests an improvement to the micromamba installation documentation currently available at https://mamba.readthedocs.io/en/latest/installation/micromamba-installation.html#automatic-install. That page offers the following approach to installing micromamba:

> On Windows Powershell, use
>
> `Invoke-Expression ((Invoke-WebRequest -Uri https://micro.mamba.pm/install.ps1).Content)`

But on certain Windows configurations, or possibly PowerShell versions, this error will occur:

```
PS C:\Users\jjr8> Invoke-Expression ((Invoke-WebRequest -Uri https://micro.mamba.pm/install.ps1).Content)                                                            Invoke-WebRequest : The response content cannot be parsed because the Internet Explorer engine is not available, or
Internet Explorer's first-launch configuration is not complete. Specify the UseBasicParsing parameter and try again.
At line:1 char:21
+ ... xpression ((Invoke-WebRequest -Uri https://micro.mamba.pm/install.ps1 ...
+                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotImplemented: (:) [Invoke-WebRequest], NotSupportedException
    + FullyQualifiedErrorId : WebCmdletIEDomNotSupportedException,Microsoft.PowerShell.Commands.InvokeWebRequestCommand
```
I do not know the exact circumstances that cause this, but my best guess is that it occurs on installations of Windows Server that use the default installation options. By default, the Internet Explorer engine is not installed, being very obsolete. I think there is a Windows option to explicitly install it, but most people probably try to avoid ever doing so.

As suggested in the error message, the problem can be solved by adding the `-UseBasicParsing` option to `Invoke-WebRequest`, like this:

```
[...]

PR #3704, opened 2024-12-19
docs: Use a portable web request for Windows

jjerphan:
Fix https://github.com/mamba-org/mamba/issues/3703.

### DIFF
--- a/docs/source/installation/micromamba-installation.rst
+++ b/docs/source/installation/micromamba-installation.rst
@@ -50,7 +50,7 @@ On Windows Powershell, use
 
 .. code :: powershell
 
-   Invoke-Expression ((Invoke-WebRequest -Uri https://micro.mamba.pm/install.ps1).Content)
+   Invoke-Expression ((Invoke-WebRequest -Uri https://micro.mamba.pm/install.ps1 -UseBasicParsing).Content)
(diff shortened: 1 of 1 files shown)


docs/source/installation/micromamba-installation.rst
```diff
@@ -50,7 +50,7 @@ On Windows Powershell, use
 
 .. code :: powershell
 
-   Invoke-Expression ((Invoke-WebRequest -Uri https://micro.mamba.pm/install.ps1).Content)
+   Invoke-Expression ((Invoke-WebRequest -Uri https://micro.mamba.pm/install.ps1 -UseBasicParsing).Content)
```
Klaim, line 53: I'm not a specialist at all so I just discovered that, but `-UseBasicParsing` seems to be deprecated for Powershell >= 6
If we need to support v5 then I guess it's ok, also it's supposed to be silent with higher versions.
See: https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.utility/invoke-restmethod?view=powershell-7.4#-usebasicparsing

Klaim (approved):
LGTM, just needed to be aware of the deprecation of that flag (automatic in following powershell version)

Issue #3535, opened 2024-10-15
Missing command options in micromamba


JohanMabille:
To be filled

anutosh491:
More commands found in conda/mamba1 missing in micromamba/mamba2 

Not shown explicitly through `--help`
- [ ] compare
- [x] deactivate
  cf. https://github.com/mamba-org/mamba/pull/4350
- [ ] init
- [x] uninstall
- [ ] package (Create low-level conda packages. (EXPERIMENTAL))

Others
- [ ] doctor (Display a health report for your environment)
- [ ] export (Export a given environment)
- [ ] notices (Retrieve latest channel notifications)
- [ ] rename
- [ ] content-trust (Signing and verification tools for Conda)

Differences in flags/utilities for common commands (provided by conda but not by mamba 2)
1) Create 
- [x] --clone ENV (Create a new environment as a copy of an existing local env)
- [ ] --show-channel-urls (Show channel urls)
- [ ] -C, --use-index-cache ( Use cache of channel index files, even if it has expired)
- [ ] --dev ( Use `sys.executable -m conda` in wrapper scripts instead of  CONDA_EXE. This is mainly for use during tests where we test new conda sources against old Python versions.)
- [ ] --no-lock (Disable locking)

2) Install
- [x] --revision REVISION   (Revert to the specified REVISION)
cf. #3911
cf. #3966
- [ ] --dev
- [ ] --solver (Choose which solver backend to use)
- [ ] --no-lock 
- [ ] --show-channel-urls

3) Update (some common to create & install)
- [ ] --use-local (Use locally built packages)
- [ ] --force-reinstall (Ensure that any user-requested package is uninstalled and reinstalled)
Consider https://github.com/mamba-org/mamba/issues/3860
- [ ] --update-deps   (Update dependencies that have available update)
- [ ] --no-update-deps (Do not update or change already-installed dependencies)
- [ ] --update-specs  (Update based on provided specifications)
- [ ] --clobber (Allow clobbering of overlapping file paths within package)

4) list (some commonly listed above)
- [x] --reverse  (List installed packages in reverse order)
cf. #3705 
- [x] -c, --canonical  (Output canonical names of packages only)
cf. #3777 
- [x] --explicit (List explicitly all installed conda packages with URL)
cf. #3760 
- [x] --md5 (Add MD5 hashsum when using --explicit)
cf. #3773 
- [x] --sha256 (Add SHA256 hashsum when using --explicit)
cf. #3885 
[...]

PR #3705, opened 2024-12-20
feat: add reverse flag to list command

SandrineP:


micromamba/tests/test_list.py
```diff
@@ -28,6 +29,8 @@ def test_list(tmp_home, tmp_root_prefix, tmp_env_name, tmp_xtensor_env, env_sele
         for i in res
     )
 
+    if reverse_flag == "--reverse":
+        assert names.index("xtensor") > names.index("xtl")
```
Hind-M, line 33: Add an `else`?

libmamba/src/api/list.cpp
```diff
@@ -208,6 +231,7 @@ namespace mamba
         detail::list_options options;
         options.full_name = config.at("full_name").value<bool>();
         options.no_pip = config.at("no_pip").value<bool>();
+	    options.reverse =config.at("reverse").value<bool>();
```
Hind-M, line 234: Indentation issue here but running `pre-commit run --all` in your source folder should fix it (I guess the linter is failing in the CI as well?)

micromamba/tests/test_list.py
```diff
@@ -7,16 +7,17 @@
 from . import helpers
 
 
+@pytest.mark.parametrize("reverse_flag", ["", "--reverse"])
 @pytest.mark.parametrize("quiet_flag", ["", "-q", "--quiet"])
 @pytest.mark.parametrize("env_selector", ["", "name", "prefix"])
 @pytest.mark.parametrize("shared_pkgs_dirs", [True], indirect=True)
-def test_list(tmp_home, tmp_root_prefix, tmp_env_name, tmp_xtensor_env, env_selector, quiet_flag):
+def test_list(tmp_home, tmp_root_prefix, tmp_env_name, tmp_xtensor_env, env_selector, quiet_flag, reverse_flag):
```
Hind-M, line 14: I know this is maybe beyond this PR because it wasn't done in the first place, but we should test the non `--json` case as well (you could have missed adding the corresponding `std::sort`, and no test would have pointed that...).

jjerphan (commented):
LGTM beyond a few comments from Hind and this review.

Can you edit the description of the PR to point to the issue you mentioned then? :)

libmamba/src/api/list.cpp
```diff
@@ -168,7 +184,14 @@ namespace mamba
...
 
-            std::sort(packages.begin(), packages.end(), compare_alphabetically);
+            if (options.reverse)
+            {
+                std::sort(packages.begin(), packages.end(), compare_reverse_alphabetically);
+            }
+            else
+            {
+                std::sort(packages.begin(), packages.end(), compare_alphabetically);
+            }
```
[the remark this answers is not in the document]
Klaim: Note that the parenthesis are not necessary in `auto comparable = (` (also they dont have the same exact meaning, not that it's important here).

Klaim:
For info, my review is basically Julien's review at this point. :)

libmamba/src/api/list.cpp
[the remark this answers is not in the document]
SandrineP: If my understanding is correct, it returns a function not a boolean.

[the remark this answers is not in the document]
jjerphan: Yes, and its typing is done via `auto`. I chose `comparable` as a name based on [this C++ terminology](https://en.cppreference.com/w/cpp/concepts/equality_comparable), but it can be changed definitely.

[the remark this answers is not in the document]
Klaim: Yeah C++ concepts are essentially boolean predicates on types so their naming usually sounds like booleans, but that's a general property of the object, not the local role of that object ("what is it in this context?" that's the best name). `comparator` makes sense, thanks!

micromamba/tests/test_list.py
SandrineP: I added a test for the non --json case, but the part checking "conda-forge" is assuming that "res" always has the same format.

jjerphan: This handling of flags might be useful to boil down the two tests into a single one handling all the flags.

https://github.com/mamba-org/mamba/blob/3dcbf76a041a99c073b83630259ec81c8e44b385/micromamba/tests/test_env.py#L74-L75

jjerphan: We discussed that IRL, let's forget about my last suggestion (it would make things more complicated than necessary).

micromamba/tests/test_list.py
```diff
@@ -28,6 +31,50 @@ def test_list(tmp_home, tmp_root_prefix, tmp_env_name, tmp_xtensor_env, env_sele
...
+    # Name           Version  Build        Channel
+    # ────────────────────────────────────────────────────
+    # _libgcc_mutex  0.1      conda_forge  conda-forge
+    # _openmp_mutex  4.5      2_gnu        conda-forge
+    header_delimiter = "\u2500"
+    packages = res[res.rindex(header_delimiter) :].split("\n", 1)[1]
+    packages_list = packages.strip().split("\n")
+
+    for package in packages_list:
+        assert package[-11:] == "conda-forge"
```
jjerphan, line 71:
```suggestion
        channel = package.split(" ")[-1]
        assert channel == "conda-forge"
```

libmamba/src/api/list.cpp
```diff
@@ -75,6 +81,7 @@ namespace mamba
                 regex = '^' + regex + '$';
             }
 
+
```
jjerphan, line 84:
```suggestion
```

micromamba/tests/test_list.py
```diff
@@ -28,6 +31,51 @@ def test_list(tmp_home, tmp_root_prefix, tmp_env_name, tmp_xtensor_env, env_sele
...
+    # This is what res looks like in this case (with or without a header delimiter):
+    # List of packages in environment: "xxx"
+
+    # Name           Version  Build        Channel
+    # ────────────────────────────────────────────────────
+    # _libgcc_mutex  0.1      conda_forge  conda-forge
+    # _openmp_mutex  4.5      2_gnu        conda-forge
+    packages = res[res.rindex("Channel") :].split("\n", 1)[1]
+    packages_list = packages.strip().split("\n")
+    header_delimiter = "\u2500"
```
jjerphan, line 68: Are there problems with this:
```suggestion
    header_delimiter = "─"
```
?

SandrineP: I am not sure which of the different dashes this one is... I will have to check !

jjerphan: FYI, I copied and pasted it from the output of the failing test.

micromamba/tests/test_list.py
```diff
@@ -28,6 +31,49 @@ def test_list(tmp_home, tmp_root_prefix, tmp_env_name, tmp_xtensor_env, env_sele
...
+    assert "xtl" in res
+
+    # This is what res looks like in this case (with or without a header delimiter):
+    # List of packages in environment: "xxx"
+
+    # Name           Version  Build        Channel
+    # ────────────────────────────────────────────────────
+    # _libgcc_mutex  0.1      conda_forge  conda-forge
+    # _openmp_mutex  4.5      2_gnu        conda-forge
+    packages = res[res.rindex("Channel") :].split("\n", 1)[1]
```
jjerphan, line 66:
```suggestion
    packages = res[res.rindex("Channel") :].split("\n", 1)[1]. replace("\r", "")
```

jjerphan (commented):
Final suggestion.

micromamba/tests/test_list.py
SandrineP: I put the "replace" int he loop because I think there will be a "\r" at the end of each line.

jjerphan: OK, this also would have replaced all the instances.

PR #3709, opened 2024-12-23
Removes absolute paths from mamba.bat and  mamba_hook.bat

### LINKED ISSUE #3708: Hardcoded absolute path in mamba.bat and mamba_hook.bat
### Troubleshooting docs

### Anaconda default channels

1

### Latest version of Mamba

### Tried in Conda?

Not applicable

### Describe your issue

mamba.bat and mamba_hook.bat have hardcoded absolute paths. Additionally, mamba_hook.bat has junk code.

_No response_

### Logs

_No response_

### environment.yml

_No response_

### ~/.condarc

_No response_

pchemguy:
Replaces macros expanded to absolute paths with prefix-independent code.
Fixes #3708

### DIFF
--- a/libmamba/data/mamba.bat
+++ b/libmamba/data/mamba.bat
@@ -3,8 +3,11 @@
 
 @REM Replaced by mamba executable with the MAMBA_EXE and MAMBA_ROOT_PREFIX variable pointing
 @REM to the correct locations.
-__MAMBA_DEFINE_MAMBA_EXE__
-__MAMBA_DEFINE_ROOT_PREFIX__
+@SET "CURDIR=%CD%"
+@CD /d "%~dp0.."
+@SET "MAMBA_ROOT_PREFIX=%CD%"
+@SET "MAMBA_EXE=%CD%\Library\bin\mamba.exe"
+@CD /d "%CURDIR%"
 
 @IF [%1]==[activate]   "%~dp0__MAMBA_INSERT_ACTIVATE_BAT_NAME__" %*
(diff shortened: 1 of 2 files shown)


libmamba/data/mamba.bat
```diff
@@ -3,8 +3,11 @@
 
 @REM Replaced by mamba executable with the MAMBA_EXE and MAMBA_ROOT_PREFIX variable pointing
 @REM to the correct locations.
-__MAMBA_DEFINE_MAMBA_EXE__
-__MAMBA_DEFINE_ROOT_PREFIX__
+@SET "CURDIR=%CD%"
+@CD /d "%~dp0.."
+@SET "MAMBA_ROOT_PREFIX=%CD%"
+@SET "MAMBA_EXE=%CD%\Library\bin\mamba.exe"
+@CD /d "%CURDIR%"
```
JohanMabille, line 10: This will break micromamba installs, which install path is independent from its root prefix.

JohanMabille (commented):
What is the issue with ahving absolute paths in these scripts? They are generated when installing mamba / micromamba, so the paths should be the correct.

pchemguy:
> What is the issue with ahving absolute paths in these scripts? They are generated when installing mamba / micromamba, so the paths should be the correct.

The issue is that this is a bad practice. There is absolutely no reason for Miniforge not to be portable, except for these hardcoded absolute paths. Absolute paths should alomst never be hardcoded in scripts. This is definitely true about Python infrastructure.

PR #3722, opened 2025-01-03
Options args enhancement

Hind-M:
- [x] Fix `--remote` wrong description
- [x] Set default `TEXT` to something more explicit for options using `option_text`.
Note that options using `required()` and `excludes(some_other_option)` are not affected because modifying the `TEXT` option makes the requirement and exclusion descriptions vanish.

### DIFF
--- a/micromamba/src/common_options.cpp
+++ b/micromamba/src/common_options.cpp
@@ -478,5 +478,4 @@ init_install_options(CLI::App* subcom, Configuration& config)
             "Categories of package to install from environment lockfile"
         )
         ->option_text("CAT1 CAT2...");
-    ;
 }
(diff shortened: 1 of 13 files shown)


micromamba/src/common_options.cpp
```diff
@@ -451,17 +462,21 @@ init_install_options(CLI::App* subcom, Configuration& config)
...
-        "Categories of package to install from environment lockfile"
-    );
+    subcom
+        ->add_option(
+            "--category",
+            categories.get_cli_config<string_list>(),
+            "Categories of package to install from environment lockfile"
+        )
+        ->option_text("CAT1 CAT2...");
+    ;
```
JohanMabille, line 481: This one can be removed ;)

JohanMabille (approved):

PR #3731, opened 2025-01-07
update_changelog.py now can also take input as cli parameters

Klaim:
With this change `python update_changelog.py -d 2024-12-24 -v 2.1.0.dev0` can be used instead of interactively providing these input. If any input is missing, we still do the interactive asks.

### DIFF
--- a/update_changelog.py
+++ b/update_changelog.py
@@ -21,6 +21,7 @@ from datetime import date
 import json
 import re
 import subprocess
+import argparse
 from version_scheme import version_info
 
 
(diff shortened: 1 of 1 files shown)


JohanMabille (approved):

update_changelog.py
```diff
@@ -21,6 +21,7 @@
 import json
 import re
 import subprocess
+import argparse
```
Hind-M, line 24: Nitpick: `argparse` should be imported first (alphabetical order).

Hind-M (approved):

JohanMabille:
These are changes to the releaser which is not tested in this CI, so no need to wait for it.

PR #3732, opened 2025-01-07
Unique Release Tag 

Klaim:
This changes the release scripts so that only one tag is generated per new version, the one with the version. Part of the effort for #3638

We still generate separate changelogs for the different sub-projects for now, not sure if it will still be useful.

### DIFF
--- a/releaser.py
+++ b/releaser.py
@@ -14,6 +14,7 @@ templates = {
     "libmambapy": "libmambapy/src/libmambapy/version.py.tmpl",
 }
 
+
 def apply_changelog(name, version, changes):
     def template_substitute(contents):
         x = contents.replace("{{ version_major }}", version.major)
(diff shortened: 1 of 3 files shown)


JohanMabille:
> We still generate separate changelogs for the different sub-projects for now, not sure if it will still be useful.

I think it is. At least, it does not hurt as it is fully automated.

JohanMabille (approved):

releaser.py
jjerphan, line 1: Would it be possible to add a comment at the head of this file specifying how this script should be executed via the CLI?

jjerphan (approved):
Thank you, @Klaim.

Hind-M (approved):

releaser.py
Klaim: Actually, there is already a comment saying to refer to update_changelog.py's top comment, which describes the release process. Note however:
- I recently improved the cli of `update_changelog.py`, I'll update the comment there in this PR or another;
- `release.py` dont have any argument you're supposed to run it with not arguemnts, it reads from the changelog which should have been modified by `update_changelog.py`. I will add this clarification.

Issue #3733, opened 2025-01-08
Stochastically getting error "ZSTD decompression error: Unknown frame descriptor", persists until index cache cleaned


corneliusroemer:
<details>

### Troubleshooting docs
- [X] My problem is not solved in the Troubleshooting docs
- [X] I do NOT use the Anaconda default channels (pkgs/* etc.)

### How did you install Mamba?

Micromamba

### Search tried in issue tracker

zstd decompression error

### Latest version of Mamba

- [X] My problem is not solved with the latest version

### Tried in Conda?

I didn't try
</details>

### Describe your issue

Preface: Stochastic error, so not critical, but just opening the issue in case others face it as well. In any case there could be better error handling: cleaning of index cache when this error happens rather than make the user have to do it manually.

When doing `micromamba install mailbox` I suddenly got the error `libmamba ZSTD decompression error: Unknown frame descriptor`:

```
...snip...
━━━━━━━━━━━   0.0 B /  ??.?MB @  ??.?MB/s  0.2s
error    libmamba ZSTD decompression error: Unknown frame descriptor
warning  libmamba Download error (23) Failed writing received data to disk/application [https://conda.anaconda.org/nodefaults/noarch/repodata.json.zst]
critical libmamba Multiple errors occurred:
    Download error (23) Failed writing received data to disk/application [https://conda.anaconda.org/nodefaults/noarch/repodata.json.zst]
    Failure writing output to destination, passed 19 returned 20
    Subdir nodefaults/noarch not loaded!
```

Retrying the install didn't help. Only cleaning the index cache helped with `micromamba clean -a`

This is the first time I see the issue, but it calls for better error handling even if it's rare, so opening just in case.

Possibly related to: https://github.com/mamba-org/mamba/issues/2707

### Logs of the incident and what I did right before

<details>

```
$ micromamba install pycryptodome
conda-forge/noarch                                          Using cache
conda-forge/osx-arm64                                       Using cache
bioconda/noarch                                             Using cache
bioconda/osx-arm64                                          Using cache

Pinned packages:

  - python=3.13


Transaction

  Prefix: /Users/corneliusromer/micromamba/envs/py13

  Updating specs:

   - pycryptodome


  Package         Version  Build            Channel           Size
────────────────────────────────────────────────────────────────────
  Install:
────────────────────────────────────────────────────────────────────

  + gmp             6.3.0  h7bae524_2       conda-forge     Cached
[...]

PR #3736, opened 2025-01-09
fix: Add a hint on cache corruption

jjerphan:
Fix https://github.com/mamba-org/mamba/issues/3733.

### DIFF
--- a/libmamba/src/core/error_handling.cpp
+++ b/libmamba/src/core/error_handling.cpp
@@ -76,6 +76,12 @@ namespace mamba
                 m_aggregated_message += er.what();
                 m_aggregated_message += "\n";
             }
+
+            m_aggregated_message += "If you run into this error repeatedly, your package cache may be corrupted.\n";
+            m_aggregated_message += "Please try running `mamba clean -a` to remove this cache before retrying the operation.\n";
+            m_aggregated_message += "\n";
+            m_aggregated_message += "If you still are having issues, please report the error on `mamba-org/mamba`'s issue tracker:\n";
+            m_aggregated_message += "https://github.com/mamba-org/mamba/issues/new?assignees=&labels=&projects=&template=bug.yml";
         }
         return m_aggregated_message.c_str();
     }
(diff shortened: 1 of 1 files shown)


libmamba/src/core/error_handling.cpp
```diff
@@ -76,6 +76,12 @@ namespace mamba
                 m_aggregated_message += er.what();
                 m_aggregated_message += "\n";
             }
+
+            m_aggregated_message += "If you run into this error repeatedly, your package cache may be corrupted.\n";
```
Klaim, line 80: Note that double-quoted text chained are automaticlaly concatenated in C++, so you can remove the sequence of runtime concatenations and just inject one string:
```c++

mystring = "fniofdngofdnofdg\n"
                    "fnoidngfodnigongidognfdsgdsigonds\n"
                    "oinoinfoidsnfoidsnfoidsnfoids";
```

You could also use a [string litteral](https://en.cppreference.com/w/cpp/language/string_literal) to avoid having to use the right formatting characters:
```c++
my_message = R"(
lorem ipsum blahblah blahblah blahblah blahblah 
blahblah blahblah blahblah 


blahblah blahblah blahblah blahblah 
blahblah blahblah blahblah 
)";
```

jjerphan: I though of the string literal approach. Unfortunately, it would intend the message I believe, except if there's another way to handle it. For instance:

```cpp
// main.cpp
#include <iostream>
#include <string>


int main(int argc, char *argv[])
{
    auto my_message = R"(
    lorem ipsum blahblah blahblah blahblah blahblah blahblah blahblah


    blahblah blahblah blahblah blahblah blahblah blahblah
    blahblah blahblah
    )";

    std::cout << my_message << std::endl;

    return 0;
}
```

```
$CC main.cpp -o main ; ./main
```
```

    lorem ipsum blahblah blahblah blahblah blahblah blahblah blahblah


    blahblah blahblah blahblah blahblah blahblah blahblah
    blahblah blahblah
    
```

Let's go for the other approach, which is also less costly than the current one.

Klaim (approved):

libmamba/src/core/error_handling.cpp
Klaim: Personally I'm fine with the indentation as it's internal to a string and very visible in IDEs , but both are fine for me 👍🏽

PR #3749, opened 2025-01-15
maint: Correctly exclude json files in clang-format

mathbunnyru:


.pre-commit-config.yaml
```diff
@@ -40,7 +40,7 @@ repos:
     hooks:
       - id: clang-format
         args: [--style=file]
-        exclude: ".json"
```
mathbunnyru, line 43: I think it was regex match, that's why `libmamba/include/mamba/util/json.hpp` matched as well.
It's easier to use `exclude_types` here

jjerphan (approved):

Issue #3734, opened 2025-01-08
`micromamba self-update` unnecessarily errors with  `libmamba Could not lock non-existing path '/Users/home/micromamba/conda-meta'`


corneliusroemer:
<details>

### Troubleshooting docs

- [X] My problem is not solved in the Troubleshooting docs

### Anaconda default channels

- [X] I do NOT use the Anaconda default channels (pkgs/* etc.)

### How did you install Mamba?

Micromamba

### Search tried in issue tracker

self-update

### Latest version of Mamba

- [X] My problem is not solved with the latest version [one can't possibly self-update on the latest version ;) ]

### Tried in Conda?

I didn't try

</details>

### Describe your issue

When doing `micromamba self-update` (on version 2.0.2) I get an error, even though the update succeeds. So a false-positive error. The update code seems to assume something that isn't true in my case:

```
$ micromamba self-update
conda-forge/osx-arm64                               13.8MB @  27.1MB/s  0.4s
conda-forge/noarch                                  18.4MB @  24.6MB/s  0.7s

  Installing micromamba version: 2.0.5 (currently installed 2.0.2)
  Fetching micromamba from https://conda.anaconda.org/conda-forge/osx-arm64/micromamba-2.0.5-0.tar.bz2

error    libmamba Could not lock non-existing path '/Users/corneliusromer/micromamba/conda-meta'

Transaction starting
micromamba                                           6.1MB @   8.1MB/s  0.5s
Download only - packages are downloaded and extracted. Skipping the linking phase.
```

Actually, the update seems to have worked, despite the error. I only just found out while filing the issue. It's still a bug :)

### mamba info / micromamba info

```shell
libmamba version : 2.0.5
     micromamba version : 2.0.5
           curl version : libcurl/8.11.1 (OpenSSL/3.4.0) SecureTransport zlib/1.3.1 zstd/1.5.6 libssh2/1.11.1 nghttp2/1.64.0
     libarchive version : libarchive 3.7.7 zlib/1.3.1 bz2lib/1.0.8 libzstd/1.5.6
       envs directories : /Users/corneliusromer/micromamba/envs
          package cache : /Users/corneliusromer/micromamba/pkgs
                          /Users/corneliusromer/.mamba/pkgs
            environment : py13 (active)
           env location : /Users/corneliusromer/micromamba/envs/py13
      user config files : /Users/corneliusromer/.mambarc
 populated config files : /Users/corneliusromer/.mambarc
[...]

PR #3752, opened 2025-01-16
fix: Ensure root prefix's `conda-meta` folder presence

jjerphan:
Fix https://github.com/mamba-org/mamba/issues/3734.

### DIFF
--- a/micromamba/src/update.cpp
+++ b/micromamba/src/update.cpp
@@ -75,7 +75,7 @@ update_self(Configuration& config, const std::optional<std::string>& version)
 
     // We need to ensure that the `conda-meta` folder exists so that all further
     // operations can be performed.
-    fs::u8path conda_meta_path = ctx.prefix_params.target_prefix / "conda-meta";
+    const fs::u8path conda_meta_path = ctx.prefix_params.target_prefix / "conda-meta";
     if (!fs::exists(conda_meta_path))
     {
         fs::create_directories(conda_meta_path);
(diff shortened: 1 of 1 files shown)


micromamba/src/update.cpp
```diff
@@ -73,6 +73,14 @@ update_self(Configuration& config, const std::optional<std::string>& version)
     // the conda-meta folder of the target_prefix)
     ctx.prefix_params.target_prefix = ctx.prefix_params.root_prefix;
 
+    // We need to ensure that the `conda-meta` folder exists so that all further
+    // operations can be performed.
+    fs::u8path conda_meta_path = ctx.prefix_params.target_prefix / "conda-meta";
```
Klaim, line 78:
```suggestion
    const fs::u8path conda_meta_path = ctx.prefix_params.target_prefix / "conda-meta";
```

micromamba/src/update.cpp
```diff
@@ -73,6 +73,14 @@ update_self(Configuration& config, const std::optional<std::string>& version)
...
+    if (!fs::exists(conda_meta_path))
+    {
+        fs::create_directories(conda_meta_path);
```
Klaim, line 81: My understanding is that `create_directories` will not fail and is no-op if the path already exists, so the `if` doesnt seem necessary.

Hind-M (commented):
`conda-meta` creation is more complex than what it seems (IIRC it depends on the command we're using `create`, `update`, `install` etc and the corresponding rules/constraints are different).
So I don't think we can just create the directory here without checking the (expected) behavior more thoroughly.
Also, IIUC the issue, this is not necessary and the corresponding [fix](https://github.com/mamba-org/mamba/pull/3572) was merged after `2.0.2` version.
So having this error is normal (as it's updating from `2.0.2`).

jjerphan:
Oh, I did not see https://github.com/mamba-org/mamba/pull/3572, and the original issue is not reproducible, so I guess we can close this PR.

PR #3753, opened 2025-01-16
maint: Add missing thread and undefined sanitizers CMake options

mathbunnyru:


libmamba/CMakeLists.txt
```diff
@@ -91,6 +90,7 @@ if(ENABLE_ASAN)
     add_link_options(-fno-omit-frame-pointer -fsanitize=address)
 endif()
 
+option(ENABLE_TSAN "Enable Thread-Sanitizer (currently only supported on GCC and Clang)" OFF)
```
jjerphan, line 93: Can you group all the `option`s together bellow this comment:

```cmake
# Build options
# =============
```

?

libmamba/CMakeLists.txt
```diff
@@ -100,6 +100,7 @@ if(ENABLE_TSAN)
     add_link_options(-fno-omit-frame-pointer -fsanitize=thread)
 endif()
 
+option(ENABLE_USAN "Enable Undefined-Sanitizer (currently only supported on GCC and Clang)" OFF)
```
jjerphan, line 103: This is [Undefined Behavior Sanitizer (ubsan)](https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html), and I think this could be changed in this PR as well.

jjerphan (approved):

libmamba/CMakeLists.txt
mathbunnyru: I applied consistent naming, and changed other options as well to what is usually used.

libmamba/CMakeLists.txt
mathbunnyru: I prefer not to - if someone want to work on https://github.com/mamba-org/mamba/issues/3754, it would be easy to remove the chunk of code from here and put it to `cmake/Sanitizers.cmake`.
This way it will start working for the whole repo.

PR #3700, opened 2024-12-18
fix: Improve CUDA version detection

jjerphan:

PR #3757, opened 2025-01-16
fix: Use `LOG_DEBUG` for CUDA version detection

jjerphan:
Follow-up on #3700.

### DIFF
--- a/libmamba/src/core/virtual_packages.cpp
+++ b/libmamba/src/core/virtual_packages.cpp
@@ -103,11 +103,11 @@ namespace mamba
                             return cuda_version;
                         }
                     }
-                    LOG_WARNING << "Could not extract CUDA version from: " << cuda_version;
+                    LOG_DEBUG << "Could not extract CUDA version from: " << cuda_version;
                 }
                 else
                 {
-                    LOG_WARNING << "CUDA version not found in the JSON file (`.cuda.version` is missing)";
+                    LOG_DEBUG << "CUDA version not found in the JSON file (`.cuda.version` is missing)";
                 }
             }
             else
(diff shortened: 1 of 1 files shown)


libmamba/src/core/virtual_packages.cpp
```diff
@@ -190,13 +190,13 @@ namespace mamba
...
-            LOG_WARNING << "\n";
-            LOG_WARNING << "We recommend setting the `CONDA_OVERRIDE_CUDA` environment variable\n";
-            LOG_WARNING << "to the desired CUDA version.";
+            LOG_DEBUG << "Could not find CUDA version by, in this order:\n";
+            LOG_DEBUG << " - inspecting the `CONDA_OVERRIDE_CUDA` environment variable\n";
+            LOG_DEBUG << " - parsing the : " << cuda_version_file << "\n";
+            LOG_DEBUG << " - parsing the output of `nvidia-smi --query -u -x`\n";
+            LOG_DEBUG << "\n";
+            LOG_DEBUG << "We recommend setting the `CONDA_OVERRIDE_CUDA` environment variable\n";
+            LOG_DEBUG << "to the desired CUDA version.";
```
Hind-M, line 199: We can have these with LOG_TRACE if you think it could be useful, but previous logs could be enough and these could be removed.
But in any case, I would add back:
`LOG_DEBUG << "CUDA not found";`

jjerphan:
```suggestion
            LOG_DEBUG << "CUDA not found";
```

Hind-M (approved):
Thanks!

PR #3765, opened 2025-01-22
fix: Use CA certificates from `conda-forge::ca-certificates`

jjerphan:
See https://github.com/conda-forge/miniforge/pull/715#discussion_r1925037651.

### DIFF
--- a/libmamba/src/core/context.cpp
+++ b/libmamba/src/core/context.cpp
@@ -12,6 +12,7 @@
 #include <spdlog/sinks/stdout_color_sinks.h>
 #include <spdlog/spdlog.h>
 
+#include "mamba/api/configuration.hpp"
 #include "mamba/core/context.hpp"
 #include "mamba/core/execution.hpp"
 #include "mamba/core/output.hpp"
@@ -177,7 +178,7 @@ namespace mamba
     Context::Context(const ContextOptions& options)
     {
         on_ci = static_cast<bool>(util::get_env("CI"));
-        prefix_params.root_prefix = util::get_env("MAMBA_ROOT_PREFIX").value_or("");
+        prefix_params.root_prefix = detail::get_root_prefix();
         prefix_params.conda_prefix = prefix_params.root_prefix;
 
         envs_dirs = { prefix_params.root_prefix / "envs" };
--- a/libmamba/src/download/downloader.cpp
+++ b/libmamba/src/download/downloader.cpp
@@ -37,11 +37,6 @@ namespace mamba::download
             "/usr/local/etc/openssl/cert.pem",
             "/usr/local/share/certs/ca-root-nss.crt",
             "/usr/local/share/certs/ca-root.crt",
-            // Windows
-            "C:/Program Files/Common Files/SSL/certs/ca-bundle.crt",
-            "C:/Program Files/Common Files/SSL/certs/ca-bundle.trust.crt",
-            "C:/Program Files/Common Files/SSL/certs/ca-bundle.pem",
-            "C:/Program Files/Common Files/SSL/certs/ca-bundle.crt"
         };
 
         void init_remote_fetch_params(Context::RemoteFetchParams& remote_fetch_params)
@@ -111,6 +106,18 @@ namespace mamba::download
 
                     // Fallback on system CA certificates.
                     bool found = false;
+
+                    // TODO: find if one needs to specify a CA certificate on Windows or not
+                    // given that the location of system's CA certificates is not clear on Windows.
+                    // For now, just use `libcurl` and the SSL libraries' default.
+                    if (util::on_win)
+                    {
+                        LOG_INFO << "Using libcurl/the SSL library's default CA certification";
+                        remote_fetch_params.ssl_verify = "";
+                        found = true;
+                        remote_fetch_params.curl_initialized = true;
+                        return;
+                    }
 
                     for (const auto& loc : cert_locations)
                     {
--- a/libmamba/src/download/downloader.cpp
+++ b/libmamba/src/download/downloader.cpp
@@ -25,7 +25,7 @@ namespace mamba::download
     namespace
     {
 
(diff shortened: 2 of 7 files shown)


libmamba/src/download/downloader.cpp
```diff
@@ -76,14 +81,43 @@ namespace mamba::download
                 }
                 else if (remote_fetch_params.ssl_verify == "<system>" && util::on_linux)
                 {
+#if defined(__linux__)
+                    // Use the CA certificates from the `conda-forge::ca-certificates` (a transitive
+                    // dependency of libmamba via OpenSSL) which are installed at
+                    // `$CONDA_PREFIX/ssl/cacert.pem` if they exist.
```
jjerphan, line 87: I wonder whether OpenSSL as distributed on conda-forge are using those certificates by default.

libmamba/src/download/downloader.cpp
```diff
@@ -19,18 +19,28 @@
...
+        constexpr std::array<const char*, 9> cert_locations{
             "/etc/ssl/certs/ca-certificates.crt",                 // Debian/Ubuntu/Gentoo etc.
             "/etc/pki/tls/certs/ca-bundle.crt",                   // Fedora/RHEL 6
             "/etc/ssl/ca-bundle.pem",                             // OpenSUSE
             "/etc/pki/tls/cacert.pem",                            // OpenELEC
             "/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem",  // CentOS/RHEL 7
             "/etc/ssl/cert.pem",                                  // Alpine Linux
+            "/System/Library/Keychains/SystemRootCertificates.keychain",  // macOS
+            "/Library/Keychains/System.keychain",                         // macOS
+            "C:/Program Files/Common Files/SSL/certs/ca-bundle.crt"       // Windows
```
jjerphan, line 43: Note for reviewers: I am searching for an authoritative list of all certificates on these platforms.

libmamba/src/download/downloader.cpp
```diff
@@ -74,16 +84,62 @@ namespace mamba::download
                         LOG_INFO << "Using REQUESTS_CA_BUNDLE " << remote_fetch_params.ssl_verify;
                     }
                 }
-                else if (remote_fetch_params.ssl_verify == "<system>" && util::on_linux)
+                else if (remote_fetch_params.ssl_verify == "<system>")
```
jjerphan, line 87: The semantic of `<system>` (which is the default) is unclear for now: must it uses conda-forge's distribution of CA certificates or must it use the system's?

To me conda-forge's distribution must be used by default either by being specified explicitly or by being used implicitly as the default (in which case we need to make sure the distribution of `libcurl` and `openssl` that are linked are `conda-forge`'s; I am not even sure of this at the moment).

libmamba/src/download/downloader.cpp
JohanMabille: How does the current version find the certificates on MacOS and Windows? Not saying we should not add this, but I want to avoid redundancy.

libmamba/src/download/downloader.cpp
```diff
@@ -74,16 +84,62 @@ namespace mamba::download
...
+                    LOG_INFO << "Checking for CA certificates at: " << env_prefix_conda_cert;
+
+                    if (fs::exists(env_prefix_conda_cert))
+                    {
+                        LOG_INFO << "Using CA certificates from the environment at: "
+                                 << env_prefix_conda_cert;
+                        remote_fetch_params.ssl_verify = env_prefix_conda_cert;
+                        remote_fetch_params.curl_initialized = true;
+                        return;
+                    }
```
JohanMabille, line 130: I don't think we need all this complexity. As discussed in a meeting, checking if certificates are available under the ROOT_PREFIX sohuld be enough.

libmamba/src/download/downloader.cpp
jjerphan: I don't know. They might be found automatically on Windows and macOS.

I'll revert the change only to change the case of Linux.

libmamba/src/download/downloader.cpp
JohanMabille: I think this mostly depends on the default value of this config option:
- either it is "system"  and then we want ti to be the MAMBA_ROOT_PREFIX > system approach, and this behavior can be ovveriden by specifying another value for it.
- it is something else, and system can be used to explicitely ignore the MAMBA_ROOT_PREFIX certificates.

libmamba/src/core/context.cpp
```diff
@@ -177,7 +178,7 @@ namespace mamba
     Context::Context(const ContextOptions& options)
     {
         on_ci = static_cast<bool>(util::get_env("CI"));
-        prefix_params.root_prefix = util::get_env("MAMBA_ROOT_PREFIX").value_or("");
+        prefix_params.root_prefix = detail::get_root_prefix();
```
JohanMabille, line 181: That's the only part I'm not sure about, since `prefix_params.root_prefix` is used to store the root_prefix set by the config (see https://github.com/mamba-org/mamba/blob/main/libmamba/src/api/configuration.cpp#L1237-L1245). I don't know whether some parts of the code do something specific when this variable is empty, nor if there are use cases where it would be valid to have it empty (when using the library out of mamba / micromamba for instance).

JohanMabille (approved):
LGTM despite my doubts about the Context root_prefix member. Maybe @Klaim has
a better understanding than me regarding the initialization of these variables.

libmamba/src/download/downloader.cpp
jjerphan: I eventually performed changes for all platforms.

Klaim:
> LGTM despite my doubts about the Context root_prefix member. Maybe @Klaim has a better understanding of the initialization of these variables.

I dont think I have more understanding here, although I share your suspicion. A cursory look at how that variable is used suggests to me that most of libmamba's code relies on this variable to not be empty.

PR #3769, opened 2025-01-24
ci: Rerun pytest tests on `main` in case of failures

jjerphan:
Sporadic test failures due to reasons we cannot control (e.g. network) make an entire workflow fail.

This workflow need to be rerun manually, which takes extra time (up to 40 minutes) and distracts from work.

This proposes to automatically rerun test if they fail up to 3 times thanks to [`pytest-rerunfailures`](https://github.com/pytest-dev/pytest-rerunfailures).

### DIFF
--- a/.github/workflows/unix_impl.yml
+++ b/.github/workflows/unix_impl.yml
@@ -107,7 +107,10 @@ jobs:
           python -m pip install --no-deps --no-build-isolation ./libmambapy
       - name: Run libmamba Python bindings tests
         run: |
-          python -m pytest libmambapy/tests/ --reruns 3 ${{ runner.debug == 'true' && '-v' || '--exitfirst' }}
+          # Only rerun flaky tests on the `main` branch
+          python -m pytest libmambapy/tests/ \
+              ${{ runner.debug == 'true' && '-v' || '--exitfirst' }} \
+              ${{ github.ref == 'refs/heads/main' && '--reruns 3' || '' }}
 
   mamba_integration_tests_unix:
     name: mamba integration tests
--- a/.github/workflows/unix_impl.yml
+++ b/.github/workflows/unix_impl.yml
@@ -107,7 +107,7 @@ jobs:
           python -m pip install --no-deps --no-build-isolation ./libmambapy
       - name: Run libmamba Python bindings tests
         run: |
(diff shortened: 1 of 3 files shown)


Hind-M:
As it may be very convenient to not restart failed tests manually, I don't think it's always due to network issues in our case. I guess this PR is opened due to the multiple recent tests failures that we are encountering, but in my opinion, we should rather investigate the root cause (as it seems to abort IIRC or terminate unexpectedly) and rerunning tests automatically after failure will prevent us from doing that thoroughly.

JohanMabille:
This could be done on the main branch once the PR has been merged though (I don't know if there is an easy way to do that).

.github/workflows/unix_impl.yml
```diff
@@ -107,7 +107,10 @@ jobs:
           python -m pip install --no-deps --no-build-isolation ./libmambapy
       - name: Run libmamba Python bindings tests
         run: |
-          python -m pytest libmambapy/tests/ ${{ runner.debug == 'true' && '-v' || '--exitfirst' }}
+          # Only rerun flaky tests on the `main` branch
+          python -m pytest libmambapy/tests/ \
+              ${{ runner.debug == 'true' && '-v' || '--exitfirst' }} \
+              ${{ github.ref == 'refs/heads/main' && '--reruns 3' || '' }}
```
Hind-M, line 113: I guess we should do the same for https://github.com/mamba-org/mamba/blob/main/.github/workflows/unix_impl.yml#L165-L166

.github/workflows/windows_impl.yml
```diff
@@ -116,7 +116,10 @@ jobs:
           python -m pip install --no-deps --no-build-isolation ./libmambapy
       - name: Run libmamba Python bindings tests
         run: |
-          python -m pytest libmambapy/tests/ ${{ runner.debug == 'true' && '-v' || '--exitfirst' }}
+          # Only rerun flaky tests on the `main` branch
+          python -m pytest libmambapy/tests/ \
+              ${{ runner.debug == 'true' && '-v' || '--exitfirst' }} \
+              ${{ github.ref == 'refs/heads/main' && '--reruns 3' || '' }}
```
Hind-M, line 122: Same for `micromamba/tests`.

JohanMabille (approved):

PR #3770, opened 2025-01-27
[Windows] Fix test failure

Hind-M:


libmamba/src/core/prefix_data.cpp
```diff
@@ -227,7 +227,6 @@ namespace mamba
             { "PYTHONIOENCODING", "utf-8" },
             { "NO_COLOR", "1" },
             { "PIP_NO_COLOR", "1" },
-            { "PIP_NO_PYTHON_VERSION_WARNING", "1" },
```
Hind-M, line 230: `PIP_NO_PYTHON_VERSION_WARNING` is deprecated and is a no-op now: https://github.com/pypa/pip/issues/13154.

JohanMabille (approved):

corneliusroemer:
The PR probably should have `[Windows]` removed as the change hopefully is not windows-only, but also affects macos

PR #3776, opened 2025-01-28
fix: remove deprecated PIP_NO_PYTHON_VERSION_WARNING environment vari…

govarthenan:
The PIP_NO_PYTHON_VERSION_WARNING environment variable is deprecated
in Python 3 and no longer has any effect. This flag was originally
used to suppress Python version compatibility warnings from pip.

Since mamba primarily works with Python 3 environments now, this
flag can be safely removed from the environment variables passed
to pip inspect commands.


Hind-M:
cf. #3770

PR #3777, opened 2025-01-28
feat: add canonical flag to list command

SandrineP:

### DIFF
--- a/libmamba/src/api/list.cpp
+++ b/libmamba/src/api/list.cpp
@@ -27,11 +27,12 @@ namespace mamba
             bool reverse;
             bool explicit_;
             bool md5;
+            bool canonical;
         };
 
         struct formatted_pkg
         {
-            std::string name, version, build, channel, url, md5;
+            std::string name, version, build, channel, url, md5, build_string;
         };
 
         bool compare_alphabetically(const formatted_pkg& a, const formatted_pkg& b)
@@ -201,6 +202,7 @@ namespace mamba
                         formatted_pkgs.build = package.second.build_string;
                         formatted_pkgs.url = package.second.package_url;
                         formatted_pkgs.md5 = package.second.md5;
+                        formatted_pkgs.build_string = package.second.build_string;
                         packages.push_back(formatted_pkgs);
                     }
                 }
@@ -209,8 +211,15 @@ namespace mamba
                                                   : compare_alphabetically;
                 std::sort(packages.begin(), packages.end(), comparator);
 
-                // format and print table
-                if (options.explicit_)
+                // format and print output
+                if (options.canonical)
+                {
+                    for (auto p : packages)
+                    {
+                        std::cout << p.name << "-" << p.version << "-" << p.build_string << std::endl;
+                    }
+                }
+                else if (options.explicit_)
                 {
                     for (auto p : packages)
                     {
@@ -268,6 +277,7 @@ namespace mamba
         options.reverse = config.at("reverse").value<bool>();
         options.explicit_ = config.at("explicit").value<bool>();
         options.md5 = config.at("md5").value<bool>();
+        options.canonical = config.at("canonical").value<bool>();
 
         auto channel_context = ChannelContext::make_conda_compatible(config.context());
         detail::list_packages(config.context(), regex, channel_context, std::move(options));
--- a/libmamba/src/api/list.cpp
+++ b/libmamba/src/api/list.cpp
@@ -22,17 +22,17 @@ namespace mamba
     {
         struct list_options
         {
-            bool full_name;
-            bool no_pip;
-            bool reverse;
-            bool explicit_;
-            bool md5;
-            bool canonical;
+            bool full_name = false;
+            bool no_pip = false;
+            bool reverse = false;
+            bool explicit_ = false;
+            bool md5 = false;
+            bool canonical = false;
         };
 
         struct formatted_pkg
         {
-            std::string name, version, build, channel, url, md5, build_string;
+            std::string name, version, build, channel, url, md5, build_string, platform;
         };
 
         bool compare_alphabetically(const formatted_pkg& a, const formatted_pkg& b)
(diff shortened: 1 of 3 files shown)


libmamba/src/api/list.cpp
```diff
@@ -27,11 +27,12 @@ namespace mamba
             bool reverse;
             bool explicit_;
             bool md5;
+            bool canonical;
```
Klaim, line 30: Could you add a default value ( `= false;`) to every member of this struct? I'm not convinced that it's always initialized correctly at usage point. If something prevents you to do so, nevermind.

Klaim (approved):

libmamba/src/api/list.cpp
```diff
@@ -209,8 +211,15 @@ namespace mamba
...
-                if (options.explicit_)
+                // format and print output
+                if (options.canonical)
+                {
+                    for (auto p : packages)
+                    {
+                        std::cout << p.name << "-" << p.version << "-" << p.build_string << std::endl;
+                    }
+                }
+                else if (options.explicit_)
```
Hind-M, line 222: What should happen if we ask for `--explicit` and `--canonical`?
Here, it would consider it as `--canonical` only (because it's the first condition), but if we want it lead to an additional behavior and print both lines (`build_string` and `url`) we should use an `if` instead of `else if`.
Same for remaining cases (should the `else` be exclusive etc).

Hind-M:
A more general question, are these recently added flags supposed to be supported only with a regular output (not `json`)?

libmamba/src/api/list.cpp
SandrineP: I had this question in mind for --canonical and --export (the next one I started working on), but missed it for --canonical and --explicit. Do we want to have the same behavior as conda? If so I would have to check what appends in these cases. If not, we would have to decide on the format of the output.

SandrineP:
When the --json flag is used, we return a json with everything, no matter which other flag appears (discussed with @JohanMabille). If I add md5, all the information needed to create the outputs of the other flags would be in the json.

Hind-M:
> When the --json flag is used, we return a json with everything, no matter which other flag appears (discussed with @JohanMabille). If I add md5, all the information needed to create the outputs of the other flags would be in the json.

So these specific flags are ignored with `json` and the output never changes? (just confirming that I got it right).

SandrineP:
Yes, you're right.

libmamba/src/api/list.cpp
Hind-M: Based on the descriptions, I think this should be exclusive indeed (I think `conda` does the same but gives precedence to `--explicit`). The question is, should we error out when multiple exclusive flags are given by the user or just ignore that and have some kind of priority that we decide on (for now it's `--canonical`)?

SandrineP: I tested with conda. I will reproduce its behavior (explicit > canonical > export) and print a warning when two options are used.

libmamba/src/api/list.cpp
SandrineP: Done !

libmamba/src/api/list.cpp
SandrineP: That's done !

micromamba/tests/test_list.py
```diff
@@ -103,6 +103,27 @@ def test_list_explicit(
...
+        res = helpers.umamba_list(canonical_flag)
+
+    outputs_list = res.strip().split("\n")[2:]
+    if canonical_flag == "--canonical":
+        items = ["conda-forge/", "::"]
+        for output in outputs_list:
+            assert all(i in output for i in items)
+            assert " " not in output
+
+
```
Hind-M, line 126: We can maybe combine with `explicit` tests (if it doesn't complicate things too much) and test the case where it should warn/abort (with `excludes` option)?

micromamba/src/list.cpp
```diff
@@ -50,12 +50,12 @@ init_list_parser(CLI::App* subcom, Configuration& config)
...
-    /*auto& canonical = config.insert(Configurable("canonical", false)
-                                        .group("cli")
-                                        .description("Output canonical names of packages only."));
-    subcom->add_flag("-c,--canonical", canonical.get_cli_config<bool>(), canonical.description());*/
+    auto& canonical = config.insert(
+        Configurable("canonical", false)
+            .group("cli")
+            .description("Output canonical names of packages only. Ignored if --explicit.")
+    );
+    subcom->add_flag("-c,--canonical", canonical.get_cli_config<bool>(), canonical.description());
```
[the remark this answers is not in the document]
SandrineP: The idea was to let the user know the behavior of the commands.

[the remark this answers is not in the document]
Hind-M: I think `excludes` automatically adds a note after the flag description in the `--help` text (saying that it's excluding some flag/option). But yes you can try it out and do whatever you think is user friendly.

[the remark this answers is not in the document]
SandrineP: I tried it and I don´t think that it is the behavior we want as the command doesn´t run when using `excludes`. I will keep the initial version.

libmamba/src/api/list.cpp
```diff
@@ -209,9 +213,14 @@ namespace mamba
                                                   : compare_alphabetically;
                 std::sort(packages.begin(), packages.end(), comparator);
 
-                // format and print table
+                // format and print output
                 if (options.explicit_)
                 {
+                    if (options.canonical)
+                    {
+                        LOG_WARNING << "Warning: Option --canonical ignored because of --explicit"
```
Hind-M, line 221: No need to add `Warning` in the message here as it's redundant with `LOG_WARNING` already mentioning it.
Same for `std::endl;`

libmamba/src/api/list.cpp
```diff
@@ -209,9 +213,14 @@ namespace mamba
...
+                                    << std::endl;
```
Hind-M, line 222:
```suggestion
                        LOG_WARNING << "Option --canonical ignored because of --explicit";
```

micromamba/tests/test_list.py
```diff
@@ -101,6 +109,12 @@ def test_list_explicit(
                 assert "#" in output
             else:
                 assert "#" not in output
```
Hind-M, line 111: Add a case where we have `explicit` and `canonical`?

SandrineP: It is included in the ```if```: if there is ```--explicit``` (with or without another flag), ```explicit``` should apply.
When running the test there is a case where there are both flags ```explicit``` and ```canonical``` and it passes.

Hind-M (approved):

PR #3779, opened 2025-01-28
feat: add base flag to info command

SandrineP:


micromamba/src/info.cpp
```diff
@@ -22,6 +25,11 @@ set_info_command(CLI::App* subcom, mamba::Configuration& config)
     init_info_parser(subcom, config);
     static bool print_licenses;
 
+    auto& base = config.insert(
+        Configurable("base", false).group("cli").description("Display base environment path.")
+    );
+    subcom->add_flag("--base", base.get_cli_config<bool>(), base.description());
```
jjerphan, line 31: Can `mamba info --base` be tested?

micromamba/src/info.cpp
```diff
@@ -4,11 +4,14 @@
...
 // The full license is in the file LICENSE, distributed with this software.
 
+#include "mamba/api/configuration.hpp"
 #include "mamba/api/info.hpp"
 #include "mamba/core/context.hpp"
 
 #include "common_options.hpp"
 
+using namespace mamba;
+
```
jjerphan, line 14: What is this used for?

libmamba/src/api/info.cpp
```diff
@@ -91,102 +81,137 @@ namespace mamba
...
-            if (ctx.command_params.is_mamba_exe && !ctx.command_params.caller_version.empty())
+            if (options.base)
             {
-                items.push_back({
-                    fmt::format("{} version", get_self_exe_path().stem().string()),
-                    ctx.command_params.caller_version,
-                });
+                std::cout << ctx.prefix_params.root_prefix.string() << std::endl;
             }
+            else
```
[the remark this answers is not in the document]
SandrineP: I don't understand which parenthesis you are talking about.
If I just make this change (replacing ```std::cout``` with ```LOG_INFO``` and adding ```return;```), the output is not printed in the terminal

micromamba/src/info.cpp
SandrineP: It's for ```Configurable```, which is a ```mamba::Configurable```.

micromamba/src/info.cpp
SandrineP: I have no idea how to test it. Open to suggestions!

libmamba/src/api/info.cpp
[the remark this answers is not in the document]
jjerphan: My bad: I meant curly brackets, and let's use `std::cout`.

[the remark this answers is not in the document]
SandrineP: So I leave it as is?

micromamba/src/info.cpp
jjerphan: What do you think of testing that the usual information is not present and that an existing path is output?

micromamba/src/info.cpp
```diff
@@ -22,6 +25,11 @@ set_info_command(CLI::App* subcom, mamba::Configuration& config)
     init_info_parser(subcom, config);
     static bool print_licenses;
 
+    auto& base = config.insert(
+        Configurable("base", false).group("cli").description("Display base environment path.")
```
jjerphan, line 29: How about using this if this is possible?
```suggestion
        mamba::Configurable("base", false).group("cli").description("Display base environment path.")
```

micromamba/src/info.cpp
jjerphan: OK, in this case I propose https://github.com/mamba-org/mamba/pull/3779/files#r1936931327.

libmamba/src/api/info.cpp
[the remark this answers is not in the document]
jjerphan: Yes.

micromamba/src/info.cpp
JohanMabille: Indeed, `using namespace` directive should be avoided as much as possible.

SandrineP: It is done in /mamba/micromamba/src/list.cpp , which is why I originally went for it in /mamba/micromamba/src/info.cpp (but it used only once here, and more in list.cpp).

jjerphan: OK, we can remove (broad) `using namespace` directives in another PR.

micromamba/tests/test_info.py
```diff
@@ -78,3 +78,40 @@ def test_not_env(tmp_home, tmp_root_prefix, prefix_selection, existing_prefix):
     assert f"env location : {location}" in infos
     assert f"user config files : {tmp_home / '.mambarc'}" in infos
     assert f"base environment : {tmp_root_prefix}" in infos
+
+
+@pytest.mark.parametrize("base_flag", ["", "--base"])
+@pytest.mark.parametrize("prefix_selection", [None, "prefix", "name"])
+def test_base_subcommand(tmp_home, tmp_root_prefix, prefix_selection, base_flag):
+    os.environ["CONDA_PREFIX"] = str(tmp_root_prefix)
```
jjerphan, line 86: `pytest` exposes a `monkeypatch` fixture which can be used to [monkeypatch environment variable in the scope of a test](https://docs.pytest.org/en/stable/how-to/monkeypatch.html#monkeypatching-environment-variables). This is a bit safer and the intend is clearer.

```suggestion
def test_base_subcommand(tmp_home, tmp_root_prefix, prefix_selection, base_flag, monkey_patch):
    monkeypatch.setenv("CONDA_PREFIX", str(tmp_root_prefix))
```

Not that this suggestion might not respect linters' formatting.

micromamba/tests/test_info.py
```diff
@@ -78,3 +78,40 @@ def test_not_env(tmp_home, tmp_root_prefix, prefix_selection, existing_prefix):
     assert f"env location : {location}" in infos
     assert f"user config files : {tmp_home / '.mambarc'}" in infos
     assert f"base environment : {tmp_root_prefix}" in infos
+
+
+@pytest.mark.parametrize("base_flag", ["", "--base"])
```
Hind-M, line 83: Add `--json` case.

libmamba/src/api/info.cpp
```diff
@@ -23,25 +25,13 @@ extern "C"
...
-
-        auto channel_context = ChannelContext::make_conda_compatible(config.context());
-        detail::print_info(config.context(), channel_context, config);
-
-        config.operation_teardown();
-    }
-
     namespace detail
     {
+        struct list_options
```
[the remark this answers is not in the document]
Hind-M: Btw, I just realized that this is using snake case (as in `list_options`) but we are rather using camel case in mamba code base (`RepoqueryOptions`, `ContextOptions`, etc). So for consistency, I would say to change to `ListOptions` and `InfoOptions`.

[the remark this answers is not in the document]
SandrineP: Noted!

libmamba/src/api/info.cpp
```diff
@@ -91,9 +81,30 @@ namespace mamba
...
+            ChannelContext& channel_context,
+            const Configuration& config,
+            InfoOptions options
+        )
         {
             assert(&ctx == &config.context());
+
+            if (options.base)
+            {
+                if (ctx.output_params.json)
```
Hind-M, line 95: Hmm, after looking at the original code, we shouldn't use this condition here (and no `cout` either).
We are using `info_json_print` (for json case) and `info_pretty_print` for regular output where they internally handle the previous condition.
We should modify `items` depending on the `base` flag instead.

micromamba/tests/test_info.py
```diff
@@ -71,10 +71,59 @@ def test_not_env(tmp_home, tmp_root_prefix, prefix_selection, existing_prefix):
...
+            assert all(
+                (f"{i} :" not in infos) | (f"\n{i} :" not in infos) for i in items
+            )  # f"\n{i} :" is to handle the case of the "environment" item
+            base_environment_path = infos.replace("base environment :", "").strip()
+        assert os.path.exists(base_environment_path)
+        assert base_environment_path == str(tmp_root_prefix)
+    else:
+        items += ["base environment"]
+        if json_flag == "--json":
+            assert all(i in infos.keys() for i in items)
```
jjerphan, line 127: Nitpick: This tests that `items` is a subset of `infos.keys()` but  I guess this is sufficient 

This comment also applies elsewhere.

libmamba/src/api/list.cpp
```diff
@@ -20,7 +20,7 @@ namespace mamba
 
     namespace detail
     {
-        struct list_options
+        struct ListOptions
```
jjerphan, line 23: I agree that the current `*_options` structs should be PascalCase-named.

jjerphan (approved):

jjerphan:
I am waiting for another approval before merging.

libmamba/src/api/info.cpp
```diff
@@ -4,6 +4,8 @@
 //
 // The full license is in the file LICENSE, distributed with this software.
 
+#include <iostream>
```
Hind-M, line 7: No need for this anymore.

Issue #3601, opened 2024-11-13
Inconsistencies with globs in `MatchSpec` parsing


jaimergp:
### Troubleshooting docs

- [X] My problem is not solved in the Troubleshooting docs

### Anaconda default channels

- [X] I do NOT use the Anaconda default channels (pkgs/* etc.)

### How did you install Mamba?

Other (please describe)

### Search tried in issue tracker

matchspec glob

### Latest version of Mamba

- [X] My problem is not solved with the latest version

### Tried in Conda?

I do not have this problem with Conda, just with Mamba

### Describe your issue

I have detected some parsing / roundtrip issues with libmamba `MatchSpec` objects:

This should be valid (although it's a redundant version glob):

```python
>>> print(MatchSpec.parse("name *,*.* build*"))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
libmambapy.bindings.specs.ParseError: Found invalid version predicate in "*.*"
>>> print(MatchSpecConda("name *,*.* build*"))
name[version='*,*.*',build=build*]
```

Also, when passing this spec to the solver, the error message is lossy. Note how the trailing `*` is lost in the build string:

```python
 {
  "Install": [
    "accelerate[version=\"=*,=*\",build=\"np17py27*\"]"
  ]
}
```

```
Could not solve for environment specs
The following package could not be installed
└─ accelerate =*,=* np17py27 does not exist (perhaps a typo or a missing channel).
```

### mamba info / micromamba info

```shell
libmamba version : 2.0.3
          mamba version : 2.0.3
           curl version : libcurl/8.10.1 OpenSSL/3.4.0 zlib/1.3.1 zstd/1.5.6 libssh2/1.11.0 nghttp2/1.58.0
     libarchive version : libarchive 3.7.4 zlib/1.2.13 liblzma/5.2.6 bz2lib/1.0.8 liblz4/1.9.3 libzstd/1.5.6
       envs directories : /opt/conda/envs
          package cache : /opt/conda/pkgs
                          /root/.mamba/pkgs
            environment : base (active)
           env location : /opt/conda
      user config files : /root/.mambarc
 populated config files : /opt/conda/.condarc
       virtual packages : __unix=0=0
                          __linux=5.15.49=0
                          __glibc=2.31=0
                          __archspec=1=aarch64
               channels : https://conda.anaconda.org/conda-forge/linux-aarch64
                          https://conda.anaconda.org/conda-forge/noarch
       base environment : /opt/conda
```

[...]

PR #3783, opened 2025-01-29
docs: Document that mamba 2 only supports trailing globs in version strings

jdblischak:
Mamba 1 supported leading and internal globs in version strings (eg `*.0.0` and `0.*.0` respectively). Mamba 2 only supports trailing globs. This PR documents this breaking change.

xref: https://github.com/mamba-org/mamba/issues/3601#issuecomment-2515447329

Here is a demonstration using libmambapy 2.0.5:

```python
import libmambapy

libmambapy.version.version_info
## ('2', '0', '5')

import libmambapy.specs as specs

specs.VersionSpec.parse("0.0.*")
specs.VersionSpec.parse("0.*.0")
specs.VersionSpec.parse("*.0.0")
## Traceback (most recent call last):
##   File "<stdin>", line 1, in <module>
## libmambapy.bindings.specs.ParseError: Found invalid version predicate in "*.0.0"

# Trailing glob works as expected
specs.VersionSpec.parse("0.0.*").contains(specs.Version.parse("0.0.1"))
## True
specs.VersionSpec.parse("0.0.*").contains(specs.Version.parse("1.0.0"))
## False

# Internal glob accepted but does not work
specs.VersionSpec.parse("0.*.0").contains(specs.Version.parse("0.1.0"))
## False
specs.VersionSpec.parse("0.*.0").contains(specs.Version.parse("1.0.0"))
## False
```

Here is a comparison of micromamba 2.0.5 versus 1.5.12:

```sh
docker run --rm -it ubuntu:24.10 bash
apt-get update
apt-get install --yes curl
"${SHELL}" <(curl -L micro.mamba.pm/install.sh)
## Micromamba binary folder? [~/.local/bin]
## Init shell (bash)? [Y/n] Y
## Configure conda-forge? [Y/n] Y
## Prefix location? [~/micromamba]
source ~/.bashrc
micromamba --version
## 2.0.5

micromamba --dry-run create -p test python=3.10.*
## + python               3.10.16  he725a3c_1_cpython  conda-forge      25MB
micromamba --dry-run create -p test python=3.*.16
## error    libmamba Could not solve for environment specs
##     The following package could not be installed
##     └─ python =3.0*.16 * does not exist (perhaps a typo or a missing channel).
## critical libmamba Could not solve for environment specs
micromamba --dry-run create -p test python=*.10.16
## error    libmamba Could not solve for environment specs
##     The following package could not be installed
##     └─ python =0*.10.16 * does not exist (perhaps a typo or a missing channel).
## critical libmamba Could not solve for environment specs

micromamba self-update --version 1.5
## Installing micromamba version: 1.5.12 (currently installed 2.0.5)
micromamba --version
## 1.5.12

micromamba --dry-run create -p test python=3.10.*
## + python               3.10.16  he725a3c_1_cpython  conda-forge      25MB
micromamba --dry-run create -p test python=3.*.16
## + python               3.10.16  he725a3c_1_cpython  conda-forge      25MB
micromamba --dry-run create -p test python=*.10.16
## + python               3.10.16  he725a3c_1_cpython  conda-forge      25MB
```

### DIFF
--- a/docs/source/developer_zone/changes-2.0.rst
+++ b/docs/source/developer_zone/changes-2.0.rst
@@ -19,7 +19,7 @@ Mamba (executable)
 has been entirely replaced by the dynamically linked version of ``micromamba``,
 a statically-linked ELF based on ``libmamba``.
 
-Hence ``mamba``` now has the exact same user interface and experience as ``micromamba``.
+Hence ``mamba`` now has the exact same user interface and experience as ``micromamba``.
 
 .. warning::
 
@@ -40,6 +40,9 @@ Breaking changes include:
 - A new config ``order_solver_request`` (default true) can be used to order the dependencies passed
   to the solver, getting order independent solutions.
 - Support for complex match specs such as ``pkg[md5=0000000000000]`` and ``pkg[build='^\d*$']``.
+- Dropped support for leading and internal globs in version strings (via
+  redesigned ``VersionSpec``, which no longer handles version strings as a
+  regex). Only trailing globs continue to be supported.
 
 .. TODO OCI and mirrors
 
@@ -86,6 +89,9 @@ Changes include:
   - The redesign of ``MatchSpec``.
     The module also includes a platform enumeration, an implementation of ordered ``Version``,
     and a ``VersionSpec`` to match versions.
+    **Breaking change:** ``VersionSpec`` dropped support for leading and
+    internal globs in version strings because they are no longer handled as a
+    regex. Only trailing globs continue to be supported.
   - ``PackageInfo`` has been moved to this submodule.
     Some attributes have been given a more explicit name ``fn`` > ``filename``,
     ``url`` > ``package_url``.
@@ -127,6 +133,10 @@ The main changes are:
   - A refactoring of a purely functional ``Channel`` class,
   - Implementation of a ``UnresolvedChannel`` to describe unresolved ``Channels``,
   - A refactored and complete implementation of ``MatchSpec`` using the components above.
+  - **Breaking change:** ``VersionSpec`` dropped support for leading and
+    internal globs in version strings because they are no longer handled as a
+    regex. Only trailing globs continue to be supported. This affects version
+    strings in both the command-line interface and recipe requirements.
 
 - A cleanup of ``ChannelContext`` to be a light proxy and parameter holder wrapping the
   ``specs::Channel``.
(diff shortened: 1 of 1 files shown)


docs/source/developer_zone/changes-2.0.rst
```diff
@@ -40,6 +40,9 @@ Breaking changes include:
 - A new config ``order_solver_request`` (default true) can be used to order the dependencies passed
   to the solver, getting order independent solutions.
 - Support for complex match specs such as ``pkg[md5=0000000000000]`` and ``pkg[build='^\d*$']``.
+- Dropped support for leading and internal globs in version strings (via
+  redesigned ``VersionSpec``, which no longer handles version strings as a
+  regex). Only trailing globs continue to be supported.
```
Hind-M, line 45: I'm not sure we should advertise that leading and internal globs are definitely dropped as they may be supported later on...
An opinion @JohanMabille?

Hind-M (commented):
Thank you @jdblischak for opening this!

docs/source/developer_zone/changes-2.0.rst
JohanMabille: Indeed, this is more a regression than a voluntary drop.

jdblischak: I appreciate that the plan is to fix this regression (and I would love that!), but [mamba 2](https://github.com/mamba-org/mamba/releases/tag/2024.09.25) was released about 4 months ago. Thus I think we need to better document this regression since this change in behavior was unexpected (eg https://github.com/mamba-org/mamba/issues/3601). Some ideas:

* What if we created a new Issue (something like "bug: fix leading and internal globs in version strings") and then pin that to the top of the Issues page?
* I could update the language in this PR to clarify that this breaking change is a temporary regression. Would that work? While it is only temporary, I feel like it deserves to be documented somewhere in the list of breaking changes.

Hind-M: @jdblischak Yes the comments were only meant to say that we shouldn't document that this is dropped, only that it's a regression that we are aiming to fix. But it's definitely worth documenting!

jdblischak: @Hind-M @JohanMabille I have updated the language to reflect the fact that this breaking change was unintentional and only temporary. I'm happy to update it further based on your preferred phrasing.

Hind-M (approved):
Awesome, thanks!

PR #3787, opened 2025-01-30
Fix windows paths and add tests

### LINKED ISSUE #3781: Paths in `conda-meta/*.json` files are not normalized in Windows
### Troubleshooting docs

### Anaconda default channels

Micromamba

conda list normalized

### Latest version of Mamba

### Tried in Conda?

### Describe your issue

Environments created with `micromamba` exhibit different `conda-meta/*.json` files  than environments created with `conda`. This causes behavioral differences in `conda list`. See https://github.com/conda-forge/conda-forge-ci-setup-feedstock/issues/362 for more details.

The gist is that the `files` and `_path` fields in a `conda-meta/*.json` file is expected to be slash normalized in Windows. See this [test file](https://github.com/conda/conda/blob/83ed2a8b0326990659fa9ba0a7854267a0a8a74c/tests/data/env_metadata/envpy27win_whl/conda-meta/wincertstore-0.2-py27hf04cefb_0.json#L11) for an example.

```shell
libmamba version : 2.0.5
     micromamba version : 2.0.5
           curl version : libcurl/8.8.0-DEV Schannel zlib/1.3.1
     libarchive version : libarchive 3.7.2 zlib/1.3.1 liblzma/5.6.2 bz2lib/1.0.8 liblz4/1.9.4 libzstd/1.5.6
       envs directories : C:\Users\JaimeRodriguez-Guerr\micromamba\envs
          package cache : C:\Users\JaimeRodriguez-Guerr\micromamba\pkgs
                          C:\Users\JaimeRodriguez-Guerr\.mamba\pkgs
                          C:\Users\JaimeRodriguez-Guerr\AppData\Roaming\.mamba\pkgs
            environment : C:\Users\JaimeRodriguez-Guerr\.pixi\envs\micromamba (active)
           env location : C:\Users\JaimeRodriguez-Guerr\.pixi\envs\micromamba
      user config files : C:\Users\JaimeRodriguez-Guerr\.mambarc
 populated config files : C:\Users\JaimeRodriguez-Guerr\.condarc
       virtual packages : __win=10.0.22631=0
                          __archspec=1=x86_64
               channels :
       base environment : C:\Users\JaimeRodriguez-Guerr\micromamba
               platform : win-64
```

### Logs

```shell

```

### environment.yml

```yaml
channels:
- conda-forge
dependencies:
- conda-smithy
```

### ~/.condarc

```yaml
channels: []
show_channel_urls: true
auto_update_conda: false
add_pip_as_python_dependency: false
channel_priority: strict
report_errors: false
```

Hind-M:
- [x] Fix #3781 
- [x] Add some missing tests

### DIFF
--- a/libmamba/include/mamba/fs/filesystem.hpp
+++ b/libmamba/include/mamba/fs/filesystem.hpp
@@ -77,7 +77,7 @@ namespace mamba::fs
     std::filesystem::path normalized_separators(std::filesystem::path path);
 
     // Returns a UTF-8 string given a standard path.
-    std::string to_utf8(const std::filesystem::path& path);
+    std::string to_utf8(const std::filesystem::path& path, bool normalize_sep);
 
     // Returns standard path given a UTF-8 string.
     std::filesystem::path from_utf8(std::string_view u8string);
@@ -300,10 +300,10 @@ namespace mamba::fs
 
         //---- Conversions ----
 
-        // Returns a UTF-8 string.
+        // Returns a UTF-8 string with normalized separators.
         std::string string() const
         {
-            return to_utf8(m_path);
+            return to_utf8(m_path, true);
         }
 
         // Returns a default encoded string.
@@ -333,7 +333,7 @@ namespace mamba::fs
         // Returns a UTF-8 string using the ``/`` on all systems.
         std::string generic_string() const
         {
-            return m_path.generic_string();
+            return to_utf8(m_path.generic_string(), false);
         }
 
         // Implicit conversion to standard path.
--- a/libmamba/include/mamba/fs/filesystem.hpp
+++ b/libmamba/include/mamba/fs/filesystem.hpp
@@ -77,7 +77,7 @@ namespace mamba::fs
     std::filesystem::path normalized_separators(std::filesystem::path path);
 
     // Returns a UTF-8 string given a standard path.
-    std::string to_utf8(const std::filesystem::path& path, bool use_win_sep);
+    std::string to_utf8(const std::filesystem::path& path);
 
     // Returns standard path given a UTF-8 string.
     std::filesystem::path from_utf8(std::string_view u8string);
@@ -303,7 +303,7 @@ namespace mamba::fs
         // Returns a UTF-8 string.
         std::string string() const
         {
-            return to_utf8(m_path, true);
+            return to_utf8(m_path);
         }
 
         // Returns a default encoded string.
@@ -333,7 +333,6 @@ namespace mamba::fs
         // Returns a UTF-8 string using the ``/`` on all systems.
         std::string generic_string() const
         {
-            // return to_utf8(m_path.generic_string(), false);
             return m_path.generic_string();
         }
 
--- a/libmamba/include/mamba/fs/filesystem.hpp
+++ b/libmamba/include/mamba/fs/filesystem.hpp
@@ -77,7 +77,7 @@ namespace mamba::fs
     std::filesystem::path normalized_separators(std::filesystem::path path);
 
     // Returns a UTF-8 string given a standard path.
-    std::string to_utf8(const std::filesystem::path& path);
+    std::string to_utf8(const std::filesystem::path& path, bool use_win_sep);
 
     // Returns standard path given a UTF-8 string.
     std::filesystem::path from_utf8(std::string_view u8string);
@@ -303,7 +303,7 @@ namespace mamba::fs
         // Returns a UTF-8 string.
         std::string string() const
         {
-            return to_utf8(m_path);
+            return to_utf8(m_path, true);
         }
 
         // Returns a default encoded string.
@@ -333,7 +333,8 @@ namespace mamba::fs
         // Returns a UTF-8 string using the ``/`` on all systems.
         std::string generic_string() const
         {
-            return to_utf8(m_path.generic_string());
+            // return to_utf8(m_path.generic_string(), false);
+            return m_path.generic_string();
         }
 
         // Implicit conversion to standard path.
@@ -621,6 +622,10 @@ namespace mamba::fs
 
         std::filesystem::path m_path;
     };
+
+    // Have '/' as a separator on all platforms
+    // cf. https://github.com/mamba-org/mamba/issues/3781
+    //     u8path use_common_separator(u8path path);
 
     class directory_entry : private std::filesystem::directory_entry
     {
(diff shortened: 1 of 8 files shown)


libmamba/include/mamba/fs/filesystem.hpp
```diff
@@ -333,7 +333,7 @@ namespace mamba::fs
         // Returns a UTF-8 string using the ``/`` on all systems.
         std::string generic_string() const
         {
-            return to_utf8(m_path.generic_string());
+            return m_path.generic_string();
```
Klaim, line 336: [`std::filesystem::path::generic_string`](https://en.cppreference.com/w/cpp/filesystem/path/generic_string) will not return UTF-8 on Windows: 
> If path::value_type is char, conversion, if any, is system-dependent. This is the case on typical POSIX systems (such as Linux), where native encoding is UTF-8 and string() performs no conversion.

Typically on Windows, if the system is not set to use UTF-8 by default (feature which is still considered beta), the default assumed encoding for `char`-based paths and strings is "ANSI" which in truth can not be that depending on which version of Windows you use (Japanese one for example).

Also note that the comment describing the function clarifies that `/` will be used on all systems, this is because `std::filesystem::path::generic_string` forces this: `The / character is used as the directory separator.`

So this change doesn't seem to be appropriate as we need to ensure it's still utf-8?

I see that `to_utf8` is normalizing the separators, which basically re-set `\` as separator on windows, so maybe the fix should be there. Idea: add an optional parameter (options) in `to_utf8` to normalize or not separators, so that here it is used without normalizing the parameters?

libmamba/tests/src/core/test_filesystem.cpp
```diff
@@ -37,6 +37,43 @@ namespace mamba
             REQUIRE(y.u8string() == u8"日本語");
         }
 
+        TEST_CASE("to_utf8")
+        {
+            static constexpr auto some_path_str = u8"a/b/c";
+            std::filesystem::path some_path{ some_path_str };
```
Klaim, line 43: [`u8`](https://en.cppreference.com/w/cpp/language/string_literal) still means `char` in C++17 and changes to `char8_t` in C++20. Using it forces the compiler to encode the string into UTF-8, but `std::filesystem::path` constructor has no way to know that this is actually UTF-8 and will assume the default system encoding, which is not UTF-8 on Windows (note: it's not guaranteed on linux either, it's just "normal" these days but is not a guarantee).
In C++17 using [`std::filesystem::path::u8path()`](https://en.cppreference.com/w/cpp/language/string_literal) to make sure the path is interpreted as UTF-8 would be necessary here instead of the default constructor (that's what `mamba::fs::from_utf8` also uses as implementation). Note that in C++20 `u8` means `uchar8_t` instead of `char` but this is then interpreted automatically by `u8path()` so it wouldnt change the test if you use it.

(same issue with the following new tests)

libmamba/tests/src/core/test_filesystem.cpp
```diff
@@ -37,6 +37,43 @@ namespace mamba
             REQUIRE(y.u8string() == u8"日本語");
         }
 
+        TEST_CASE("to_utf8")
```
Klaim, line 40: These tests only check that `to_utf8`/`from_utf8` preserve the right expected separator, right? Maybe clarify that in the names of the tests?
Also adding unicode-using tests for these functions might clarify things, maybe similar tests to `"normalized_separators_unicode"` above will work well, with non-ANSI unicode values, which will reveal issues on Windows.

libmamba/include/mamba/fs/filesystem.hpp
Hind-M: > I see that `to_utf8` is normalizing the separators, which basically re-set `\` as separator on windows, so maybe the fix should be there. Idea: add an optional parameter (options) in `to_utf8` to normalize or not separators, so that here it is used without normalizing the parameters?

Yeah I was intending to add a flag there initially but then realized that `generic_string` was not doing what it claims regarding the corresponding comment. So I removed `to_utf8` call there, but if it's problematic indeed, then I will get back to the first solution.

Klaim: What do you call "the first solution"?

Hind-M: Add the optional parameter to `to_utf8` and `normalized_separators` to either keep `/` or replace with `\` in windows.

Klaim: I think the role of `normalized_separators` is explicitly to make sure it's`\` so it shouldnt need a change?

libmamba/include/mamba/fs/filesystem.hpp
```diff
@@ -77,7 +77,7 @@ namespace mamba::fs
     std::filesystem::path normalized_separators(std::filesystem::path path);
 
     // Returns a UTF-8 string given a standard path.
-    std::string to_utf8(const std::filesystem::path& path);
+    std::string to_utf8(const std::filesystem::path& path, bool normalize_sep);
```
Klaim, line 80: I think it would be better with a struct `utf8Options` or something like that, which would have this boolean as member, with a default value  and passed here. That way, 
1) it's clearer at call site: `to_uft8(path, { .normilize_sep=true })`
2) it allows future option additions without breaking the function interface

libmamba/include/mamba/fs/filesystem.hpp
```diff
@@ -73,11 +73,16 @@ namespace mamba::fs
...
+    {
+        bool normalize_sep = true;
+    };
+
     // Maintain `\` on Windows, `/` on other platforms
     std::filesystem::path normalized_separators(std::filesystem::path path);
 
     // Returns a UTF-8 string given a standard path.
-    std::string to_utf8(const std::filesystem::path& path);
+    std::string to_utf8(const std::filesystem::path& path, Utf8Options utf8_options);
```
Klaim, line 85: Consider defaulting to simplify usage
```suggestion
    std::string to_utf8(const std::filesystem::path& path, Utf8Options utf8_options = {});
```

Klaim (approved):

libmamba/include/mamba/fs/filesystem.hpp
Klaim: It's fine that way 👍🏼 (you can add the dot to the name so that it's ready to be uncommented in c++20)

PR #3793, opened 2025-02-04
ci: Fix typo in Windows workflows

jjerphan:
Follow-up of https://github.com/mamba-org/mamba/pull/3769/files.


JohanMabille (approved):

Hind-M (approved):

.github/workflows/windows_impl.yml
```diff
@@ -153,6 +151,4 @@ jobs:
           $env:MAMBA_TEST_SHELL_TYPE='powershell'
           Remove-Item -Path "env:CONDARC"
           # Only rerun flaky tests on the `main` branch
-          python -m pytest micromamba/tests/ \
-              ${{ runner.debug == 'true' && '-v --capture=tee-sys' || '--exitfirst' }} \
```
Hind-M, line 157: As the lines are quiet long, we may think of using ` in windows as an equivalent to \ on linux.

jjerphan: I agree.

I initially tried but I had an error as if `--exitfirst` was provided on a new line.

PR #3794, opened 2025-02-04
[Windows] Write command in multiple lines

Hind-M:


jjerphan (approved):
Thank you for rectifying it!

PR #3692, opened 2024-12-16
Always add `root_prefix/envs` in `envs_dirs`

Hind-M:
Fix #3655

PR #3796, opened 2025-02-04
fix: Create directories from `envs_dirs` if they do not exist

### LINKED ISSUE #3790: mamba create does not create env_dirs (and parents) if they do not exist
### Troubleshooting docs

### Anaconda default channels

Micromamba

envs, envs_dirs

### Latest version of Mamba

### Tried in Conda?

I didn't try

### Describe your issue

If a directory in `envs_dirs` does not exist, it is skipped.

This is a change in behavior from 1.x, which would do `mkdir -p`.
 

```shell
libmamba version : 2.0.6.rc2
          mamba version : 2.0.6.rc2
           curl version : libcurl/8.11.1 OpenSSL/3.4.0 zlib/1.3.1 zstd/1.5.6 libssh2/1.11.1 nghttp2/1.64.0
     libarchive version : libarchive 3.7.7 zlib/1.3.1 liblzma/5.6.3 bz2lib/1.0.8 liblz4/1.10.0 libzstd/1.5.6
       envs directories : /home/burt/test123/envs
                          /home/burt/miniforge3/envs
          package cache : /home/burt/miniforge3/pkgs
                          /home/burt/.mamba/pkgs
            environment : base
           env location : /home/burt/miniforge3
      user config files : /home/burt/.mambarc
 populated config files : /home/burt/.mambarc
                          /home/burt/.condarc
                          /home/burt/miniforge3/.condarc
       virtual packages : __unix=0=0
                          __linux=5.14.0=0
                          __glibc=2.34=0
                          __archspec=1=x86_64_v3
               channels : https://conda.anaconda.org/conda-forge/linux-64
       base environment : /home/burt/miniforge3
               platform : linux-64
```

### Logs

```shell
$ rm -rf ~/test123/envs
$ ./build/micromamba/mamba create -n testenv
Empty environment created at prefix: /home/burt/miniforge3/envs/testenv
$ ./build/micromamba/mamba env remove -n testenv -y
Nothing to do.
No packages to remove from environment.
Environment removed at prefix: /home/burt/miniforge3/envs/testenv
$ mkdir ~/test123/envs
$ ./build/micromamba/mamba create -n testenv
Empty environment created at prefix: /home/burt/test123/envs/testenv
```

### environment.yml

```yaml

```

### ~/.condarc

```yaml
channels:
  - conda-forge

auto_activate_base: false
local_repodata_ttl: 176800
channel_priority: strict
envs_dirs:
  - /home/burt/test123/envs
```

### LINKED ISSUE #3836: `micromamba create` not idempotent when `envs_dirs` does not exist
### Troubleshooting docs

### Anaconda default channels

Other (please describe)

envs_dirs

### Latest version of Mamba

### Tried in Conda?

### Describe your issue

This issue seems to overlap with #3790. I have not checked if this is solved by either of the two PRs linked there.

To reproduce,

```Dockerfile
FROM mambaorg/micromamba:2.0.5

RUN cat <<EOF > explicit.txt
# Generated by conda-lock.
# platform: linux-64
# input_hash: ab45cd56a12b42dab7d2e540f01585dd1d1f8a7b51e43009d3475f75794137ad
@EXPLICIT
https://conda.anaconda.org/conda-forge/linux-64/_libgcc_mutex-0.1-conda_forge.tar.bz2#d7c89558ba9fa0495403155b64376d81
https://conda.anaconda.org/conda-forge/linux-64/libgomp-14.2.0-h77fa898_1.conda#cc3573974587f12dda90d96e3e55a702
https://conda.anaconda.org/conda-forge/linux-64/_openmp_mutex-4.5-2_gnu.tar.bz2#73aaf86a425cc6e73fcf236a5a46396d
https://conda.anaconda.org/conda-forge/linux-64/libgcc-14.2.0-h77fa898_1.conda#3cb76c3f10d3bc7f1105b2fc9db984df
https://conda.anaconda.org/conda-forge/linux-64/tzcode-2025a-hb9d3cd8_0.conda#e18bdeebf29176790089b654cfedaaec
EOF

RUN cat <<EOF > ~/.condarc
envs_dirs:
- ~/.conda
EOF
```

```bash
docker build . -t cl777-mm
docker run --rm -it cl777-mm
micromamba create --file explicit.txt --yes --name ficc
micromamba create --file explicit.txt --yes --name ficc
```

The first create command installs into the prefix `$MAMBA_ROOT_PREFIX/envs/ficc` = `/opt/conda/envs/ficc`.

The second (and any subsequent) create command installs into the `envs_dirs` prefix `~/.conda/ficc`.

I expected this `micromamba create` command to be idempotent. After some investigation, I believe what's happening on the two successive `micromamba create` commands is:

1. 
    a. Micromamba sees that `~/.conda` from `envs_dirs` doesn't exist, so installs in `$MAMBA_ROOT_PREFIX/envs/ficc` = `/opt/conda/envs/ficc`.
    b. Micromamba creates `~/.conda/environments.txt`.
2. Micromamba sees that `~/.conda` now exists and installs in the prefix `~/.conda/ficc`.

### ~/.condarc

```yaml
envs_dirs:
- ~/.conda
```

holzman:
Fix #3790.
Fix #3836.


holzman:
Actually, I think this exposed an issue with #3692:

In 2.0.5, `-r` took precedence over `MAMBA_ROOT_PREFIX`:

```
$ export MAMBA_ROOT_PREFIX=/tmp/envroot
$ mamba create  --debug -n foo --print-config-only | yq '. | .envs_dirs'
[
  "/tmp/envroot/envs"
]

$ mamba create -r /tmp/cliroot --debug -n foo --print-config-only | yq '. | .envs_dirs'
[
  "/tmp/cliroot/envs"
]
```

but in 2.0.6:
```
$ mamba create -r /tmp/cliroot --debug -n foo --print-config-only | yq '. | .envs_dirs'
[
  "/tmp/envroot/envs",
  "/tmp/cliroot/envs"
]
```

jjerphan:
Merging `main` into this branch to get the changes of #3813 in.

micromamba/tests/test_create.py
```diff
@@ -281,8 +281,6 @@ def test_target_prefix(
     else:
         root_prefix = Path(os.environ["MAMBA_ROOT_PREFIX"])
 
-    # TODO: Remove this call to `os.makedirs` once
-    # https://github.com/mamba-org/mamba/issues/3790 is fixed
     if root_prefix_env_exists:
         os.makedirs(Path(os.environ["MAMBA_ROOT_PREFIX"]) / "envs", exist_ok=True)
```
jjerphan, line 286: I guess you meant to do this?
```suggestion
```

Can you check that `root_prefix/envs` exists on the filesystem at the end of the test if `root_prefix_env_exists`?

micromamba/tests/test_create.py
```diff
@@ -713,8 +730,6 @@ def test_root_prefix_precedence(
         if set_in_condarc:
             f.write(f"envs_dirs: [{str(condarc_envs_dirs)}]")
 
-    # TODO: Remove this call to `os.makedirs` once
-    # https://github.com/mamba-org/mamba/issues/3790 is fixed
     for envs_folder in (condarc_envs_dirs, conda_envs_dirs, cli_provided_root, mamba_root_prefix):
         os.makedirs(envs_folder, exist_ok=True)
```
jjerphan, line 735: Same comment. 
```suggestion
```

Can you check that all of those directories are present on the filesystem at the end of the test?

holzman: Good catch - whoops, I did mean to take those lines out! 

We already check for the existence of the directories here: [L750-L757](https://github.com/mamba-org/mamba/pull/3796/files#diff-63e099f2fcceb7d321e6823ad7e744dfc92cdb4dea2b5507bc9d090ab9f09e53L750-L757)

holzman:
> ```python
>     def test_create_envs_dirs(tmp_root_prefix: Path, tmp_path: Path):
>         """Create an environment when the first env dir is not writable."""
>         os.environ["CONDA_ENVS_DIRS"] = f"{Path('/noperm')},{tmp_path}"
>         env_name = "myenv"
>         helpers.create("-n", env_name, "--offline", "--no-rc", no_dry_run=True)
> >       assert (tmp_path / env_name / "conda-meta" / "history").exists()
> E       AssertionError: assert False
> E        +  where False = exists()
> E        +    where exists = (((WindowsPath('C:/Users/runneradmin/AppData/Local/Temp/pytest-of-runneradmin/pytest-0/test_create_envs_dirs_False_0') / 'myenv') / 'conda-meta') / 'history').exists
> ```
> 
> `/noperm` is not used but `C:/Users/runneradmin/AppData/Local/Temp/pytest-of-runneradmin/pytest-0/test_create_envs_dirs_False_0` (which should be the value of `tmp_root_prefix`) is.

So this test was previously successful on both Windows and Linux because `/noperm` did not exist, so it then goes to the next env_dir and creates an env in `tmp_path`.  My guess is that this test is now failing because this PR tries to `os.mkdir("/noperm")` first.  That should fail on Linux for permissions reasons, but if it succeeds on Windows, it will create the env *there* rather than in `tmp_path`.

jjerphan:
Ah yes, my bad, I misread the assertion.

jjerphan:
In the meantime, https://github.com/mamba-org/mamba/issues/3836 has been reported.

Could we see if this PR fixes it by adding a non-regression test for it?

holzman:
Ok, I’m a little stumped on this one; I rolled out a Windows dev environment and the test (as revised) passed…

jjerphan:
@holzman: I tried several alternatives, but I think it would be worth having someone with a Windows machine reproduce and understand the error locally.

Feel free to remove / revert my commits.

holzman:
Thanks @jjerphan - I *did* test in a Windows machine, unfortunately. Something must be subtly different about the environment in the actual Windows runner.  I'll see if I can figure that out in a separate project; if it turns out it's impossible to make a read-only directory for the test, we should probably just skip the test for Windows.

jjerphan (approved):
LGTM given 🟢 CI.

I am waiting for a review from someone else in the team before merging.

libmamba/src/api/configuration.cpp
```diff
@@ -490,6 +490,18 @@ namespace mamba
             {
                 for (const auto& dir : dirs)
                 {
+                    if (!fs::exists(dir))
```
Klaim, line 493: `create_directories` shouldn't report errors when attempting to create an already existing directory, so you can remove that `if` and call `create_directories` every time (keep the error handling for the other potential errors).

libmamba/src/api/configuration.cpp
```diff
@@ -490,6 +490,18 @@ namespace mamba
...
+                    {
+                        try
+                        {
+                            fs::create_directories(dir);
+                        }
+                        catch (const fs::filesystem_error& e)
+                        {
+                            LOG_WARNING << "Error creating directory " << dir << ": " << e.what()
```
Klaim, line 501: Do we really want to continue when an error occurs here?
Did you consider not catching the error so that that we stop immediately and let the caller handle the issue?

holzman: Yes, this is intentional.  I wanted to preserve the same behavior as mamba 1, which continues if it cannot create the directory.

micromamba/tests/test_create.py
```diff
@@ -662,14 +660,90 @@ def test_create_empty(tmp_home, tmp_root_prefix, tmp_path, prefix_selector, crea
...
+    os.makedirs(mamba_root_prefix, exist_ok=True)
+    monkeypatch.setenv("MAMBA_ROOT_PREFIX", str(mamba_root_prefix))
+    env_name = "foo"
+
+    for _ in range(2):
+        cmd = ["-n", env_name, "--rc-file", tmp_home / ".condarc"]
+        helpers.create(*cmd, no_rc=False)
+
+    assert not Path(mamba_root_prefix_envs / env_name).exists()
+    assert Path(condarc_envs_dirs / env_name).exists()
```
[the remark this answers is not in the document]
holzman: I disagree, we want to run both assertions after the create (the goal is to test that we fixed #3836).  We could be more thorough and check on both iterations though:

```suggestion
    for _ in range(2):
        cmd = ["-n", env_name, "--rc-file", tmp_home / ".condarc"]
        helpers.create(*cmd, no_rc=False)
        assert not Path(mamba_root_prefix_envs / env_name).exists()
        assert Path(condarc_envs_dirs / env_name).exists()
```

holzman:
This test failing isn't my fault for a change!

 I think the latest commit to dateutil has a file with CRLF which gets autoconverted during the pip install (due to dateutil's settings in .gitattributes), showing up as modified and then erroring out.

jjerphan:
Indeed.

See https://github.com/dateutil/dateutil/issues/1419.

jjerphan:
Merging `main` in to get https://github.com/mamba-org/mamba/pull/3838.

Klaim (approved):
LGTM but I'm not the most knowledgeable about the behavior tested in the python tests, it seems fine to me but I'm not sure it's correct. Maybe wait for a review from  @Hind-M before merging.

micromamba/tests/test_create.py
```diff
@@ -665,14 +663,96 @@ def test_create_empty(tmp_home, tmp_root_prefix, tmp_path, prefix_selector, crea
...
+    try:
+        helpers.create("-n", env_name, "--offline", "--no-rc", no_dry_run=True)
+    finally:
+        if platform.system() == "Windows":
+            subprocess.run(
+                ["icacls", noperm_root_dir, "/grant", r"BUILTIN\Administrators:F"], check=True
+            )
+            subprocess.run(["icacls", noperm_root_dir, "/reset"], check=True)
+        else:
+            os.chmod(noperm_root_dir, 0o755)
```
Hind-M, line 706: Can you please add comments for permissions changes to ease the reading?

micromamba/tests/test_create.py
```diff
@@ -665,14 +663,96 @@ def test_create_empty(tmp_home, tmp_root_prefix, tmp_path, prefix_selector, crea
...
+    mamba_root_prefix = tmp_path / "envroot"
+    mamba_root_prefix_envs = mamba_root_prefix / "envs"
+    condarc_envs_dirs = tmp_home / ".conda"
+
+    with open(tmp_home / ".condarc", "w+") as f:
+        f.write(f"envs_dirs: [{str(condarc_envs_dirs)}]")
+
+    os.makedirs(mamba_root_prefix, exist_ok=True)
+    monkeypatch.setenv("MAMBA_ROOT_PREFIX", str(mamba_root_prefix))
+    env_name = "foo"
```
Hind-M, line 746: I believe this can be simplified using `tmp_root_prefix` fixture:
```suggestion
    mamba_root_prefix_envs = tmp_root_prefix / "envs"
    condarc_envs_dirs = tmp_home / ".conda"

    with open(tmp_home / ".condarc", "w+") as f:
        f.write(f"envs_dirs: [{str(condarc_envs_dirs)}]")

    env_name = "foo"
```

micromamba/tests/test_create.py
```diff
@@ -665,14 +663,96 @@ def test_create_empty(tmp_home, tmp_root_prefix, tmp_path, prefix_selector, crea
...
     assert (tmp_path / env_name / "conda-meta" / "history").exists()
 
 
+@pytest.mark.parametrize("envs_dirs_source", ("condarc", "env_var"))
+def test_mkdir_envs_dirs(tmp_path, tmp_home, monkeypatch, envs_dirs_source):
+    """Test that an env dir is created if it does not exist already"""
+
+    envs_dir = tmp_path / "user_provided_envdir" / "envs"
+
+    with open(tmp_home / ".condarc", "w+") as f:
```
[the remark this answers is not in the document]
holzman: Actually this is intentional - we want a .condarc to be created in both cases since we call `helpers.create` with `no_rc=False`.

Hind-M (approved):
LGTM, thanks!

PR #3799, opened 2025-02-05
fix: Adapt `test_target_prefix`

jjerphan:


jjerphan:
I tried to reproduce the failures observed in https://github.com/mamba-org/mamba/pull/3796, but those are due to the changes made in the PR rather than to the release of 2.0.6 which is now used on the CI.

PR #3800, opened 2025-02-05
feat: add revisions flag to list command

SandrineP:


libmamba/src/core/history.cpp
```diff
@@ -203,6 +203,7 @@ namespace mamba
     std::vector<History::UserRequest> History::get_user_requests()
     {
         std::vector<UserRequest> res;
+        int i = 0;
```
JohanMabille, line 206: nitpicking: `revision_num` would be more expressive.

libmamba/src/core/history.cpp
```diff
@@ -223,6 +224,11 @@ namespace mamba
                     r.link_dists.push_back(x.substr(1));
                 }
             }
+            if ((r.link_dists.size() > 0) || (r.unlink_dists.size() > 0))
+            {
+                r.revision_num = i;
+                i += 1;
```
JohanMabille, line 230: `++i` is more idiomatic, or replace the two lines with `r.revision_num = revision_num++;`

JohanMabille (commented):
I think tihs deserves some testing (like creating an environment, install some package, add another one, uninstall, and just check from the output tht we have different revisions. No ned to check the content of the revisions, this should be done in the tests of the History itself.

libmamba/include/mamba/core/history.hpp
```diff
@@ -38,6 +38,7 @@ namespace mamba
             static UserRequest prefilled(const Context& context);
 
             std::string date;
+            int revision_num;
```
Klaim, line 41:
```suggestion
            int revision_num = 0;
```
Always (well, almost always) set a default value for `int` and `float` members as otherwise they will be undefined and will lead to (potentially invisible) undefined behavior if used before being set. (could be also another value, like `-1` if that helps catching incorrect usage)

libmamba/src/api/list.cpp
```diff
@@ -215,7 +216,42 @@ namespace mamba
...
+                        LOG_WARNING
+                            << "Option --export ignored because --revisions was also provided.";
+                    }
+                    auto user_requests = prefix_data.history().get_user_requests();
+                    for (auto r : user_requests)
+                    {
+                        if ((r.link_dists.size() > 0) || (r.unlink_dists.size() > 0))
+                        {
+                            std::cout << r.date << " (rev " << r.revision_num << ")" << std::endl;
+                            for (auto ud : r.unlink_dists)
```
Hind-M, line 242: I think that besides new linked(+) and unlinked dists(-), there is some other info, but I'm not sure what it corresponds to exactly (cf. `conda` behavior).

Example on an test env of mine:

```
2024-08-09 16:44:50  (rev 1)
     ca-certificates  {2024.6.2 (conda-forge/linux-64) -> 2024.7.4 (conda-forge/linux-64)}
     certifi  {2024.6.2 (conda-forge/noarch) -> 2024.7.4 (conda-forge/noarch)}
     openssl  {3.3.1 (conda-forge/linux-64) -> 3.3.1 (conda-forge/linux-64)}
    +bash-5.2.21 (conda-forge/linux-64)
```

libmamba/src/api/list.cpp
```diff
@@ -215,7 +216,42 @@ namespace mamba
                 std::sort(packages.begin(), packages.end(), comparator);
 
                 // format and print output
-                if (options.explicit_)
+                if (options.revisions)
```
Hind-M, line 219: `--json` case should be handled as well.

micromamba/tests/test_list.py
```diff
@@ -209,3 +210,32 @@ def test_regex(tmp_home, tmp_root_prefix, tmp_xtensor_env, quiet_flag):
...
+    env_name = "myenv"
+
+    helpers.create("-n", env_name, "python=3.8")
+    helpers.install("-n", env_name, "xeus=2.0")
+    helpers.update("-n", env_name, "xeus=4.0")
+    helpers.uninstall("-n", env_name, "xeus")
+    res = helpers.umamba_list("-n", env_name, revisions_flag)
+
+    if revisions_flag == "--revisions":
+        revisions = re.split(r"\d{4}-\d{2}-\d{2}", res)[1:]
```
Hind-M, line 226: Can you add a comment to explain the parsing and what we should expect (generally speaking)?
(in order to ease future reading of the test).

libmamba/src/api/list.cpp
Hind-M: I guess it corresponds to the unchanged packages.

libmamba/src/api/list.cpp
SandrineP: Would you have an example of the output of ```revisions --json``` with ```conda``` ?

libmamba/src/api/list.cpp
SandrineP: The ```+``` and ```-``` correspond to the added/removed packages.
The other lines in your example correspond to the updated packages ```{old_version -> new_version}```.

The output with ```conda``` doesn't always look like the one you provided, it depends on the version I guess.
We had a discussion with @JohanMabille around the output, also having in mind the future ```install --revision REVISION```, but can have another one next week before merging.

SandrineP: We just discussed it with @JohanMabille and are going to leave it as is.

Issue #3647, opened 2024-12-02
`MatchSpec` parsing error with `~=` version operator


jaimergp:
### Troubleshooting docs

- [X] My problem is not solved in the Troubleshooting docs

### Anaconda default channels

- [X] I do NOT use the Anaconda default channels (pkgs/* etc.)

### How did you install Mamba?

Mambaforge or latest Miniforge

### Search tried in issue tracker

~=

### Latest version of Mamba

- [X] My problem is not solved with the latest version

### Tried in Conda?

I do not have this problem with Conda, just with Mamba

### Describe your issue

Comes from https://github.com/conda/conda-libmamba-solver/issues/581. 

```python
>>> from libmambapy.specs import MatchSpec as LibmambaMatchSpec
>>> LibmambaMatchSpec.parse("python[version='~=3.11.0',build=*_cpython]")
libmambapy.bindings.specs.ParseError: Error parsing MatchSpec "python[version='>=3.11.0',3.11.*": Parentheses mismatch in "python[version='>=3.11.0',3.11.*"."
```

Also, the output for this workaround is weird (note the build string containing the 3.11 mark). Maybe it's just the str-conversion, but it doesn't look right:

```python
>>> spec = LibmambaMatchSpec.parse('python~=3.11.0=*_cpython')
>>> print(spec)
python[version=">=3.11.0",build="*_cpython,3.11.*"]
#                                         ^^^^^^^
```

This can be seen in the `mamba` CLI handling too:

```
$ mamba create -n py311 "python[version='~=3.11.0',build=*_cpython]"

conda-forge/noarch                                          Using cache
conda-forge/linux-aarch64                           14.8MB @  11.7MB/s  1.2s
critical libmamba Error parsing MatchSpec "python[version='>=3.11.0',3.11.*": Parentheses mismatch in "python[version='>=3.11.0',3.11.*"."
```

```
$ mamba create -n py311 "python~=3.11.0=*_cpython"
conda-forge/linux-aarch64                                   Using cache
conda-forge/noarch                                          Using cache


Transaction

  Prefix: /opt/conda/envs/py311

  Updating specs:

   - python[version=">=3.11.0",build="*_cpython,3.11.*"]

...
```

### mamba info / micromamba info

```shell
libmamba version : 2.0.4
          mamba version : 2.0.4
[...]

PR #3818, opened 2025-02-12
fix: Adaptive level for compatible Version formatting

jjerphan:
Fix https://github.com/mamba-org/mamba/issues/3647.
Fix https://github.com/mamba-org/mamba/issues/3472 properly, reverting parts of #3483.

### DIFF
--- a/libmamba/src/specs/version_spec.cpp
+++ b/libmamba/src/specs/version_spec.cpp
@@ -264,11 +264,14 @@ fmt::formatter<mamba::specs::VersionPredicate>::format(
             }
             if constexpr (std::is_same_v<Op, VersionPredicate::compatible_with>)
             {
+                // Make sure to print the version without loosing information.
+                auto version_level = pred.m_version.version().size();
+                auto format_level = std::max(op.level, version_level);
                 out = fmt::format_to(
                     out,
                     "{}{}",
                     VersionSpec::compatible_str,
-                    pred.m_version.str(op.level)
+                    pred.m_version.str(format_level)
                 );
             }
         },
--- a/libmamba/tests/src/specs/test_match_spec.cpp
+++ b/libmamba/tests/src/specs/test_match_spec.cpp
@@ -128,6 +128,20 @@ namespace
             REQUIRE(ms.str() == "abc>3");
         }
 
+        SECTION("numpy~=1.26.0")
+        {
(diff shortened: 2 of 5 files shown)


libmamba/src/specs/version_spec.cpp
```diff
@@ -264,11 +264,14 @@ fmt::formatter<mamba::specs::VersionPredicate>::format(
...
             if constexpr (std::is_same_v<Op, VersionPredicate::compatible_with>)
             {
+                // Make sure to print the version without loosing information.
+                auto version_level = pred.m_version.version().size();
+                auto format_level = std::max(op.level, version_level);
                 out = fmt::format_to(
                     out,
                     "{}{}",
                     VersionSpec::compatible_str,
-                    pred.m_version.str(op.level)
```
jjerphan, line 271: Note to reviewer: this line cause the bug as the solver relies on the valid assumption that:

```c++
std::string match_spec_as_str = "numpy~=1.26"; // for instance.
match_spec_as_str == MatchSpec::parse(match_spec_as_str).value().str()
```

here:

https://github.com/mamba-org/mamba/blob/ad9b2d6b8840825411a3c23251cfbb978ab60e89/libmamba/src/solver/libsolv/helpers.cpp#L860

which was not verified for compatible versions.

I think we need to add tests cases (in another PR) to test that this is valid for many `MatchSpec`.

libmamba/tests/src/specs/test_match_spec.cpp
```diff
@@ -128,6 +128,20 @@ namespace
...
+            auto ms = MatchSpec::parse("numpy~=1.26.0").value();
+            REQUIRE(ms.name().str() == "numpy");
+            REQUIRE(ms.version().str() == "~=1.26.0");
+            REQUIRE(ms.build_string().is_explicitly_free());
+            REQUIRE(ms.build_number().is_explicitly_free());
+            REQUIRE(ms.str() == "numpy~=1.26.0");
+
+            // TODO: test this assumption for many more cases
+            auto ms2 = MatchSpec::parse(ms.str()).value();
+            REQUIRE(ms2 == ms);
```
jjerphan, line 142: Note to reviewers: this can be implemented in another PR.

JohanMabille (approved):

PR #3834, opened 2025-02-20
maint: Add logger to `Database` at creation

jjerphan:


libmamba/src/solver/libsolv/database.cpp
```diff
@@ -39,7 +39,7 @@ namespace mamba::solver::libsolv
         Matcher matcher;
     };
 
-    Database::Database(specs::ChannelResolveParams channel_params)
+    Database::Database(specs::ChannelResolveParams channel_params, bool add_logger)
```
jjerphan, line 42: Note to reviewers: I do not think `add_loger` is the best name for this parameter.

Klaim (commented):
I dont think that's the right way to go.
First, the code/implementation doing the setup of the logger needs to be in a separate function, keep it separate from where it needs to be called.
Second, setting up logging is not the responsibility of Database and this should be done in a separate step and outside any code from Database. It shouldnt eitehr be aware of how logging happens.
A better way would be to make a Database observable and have another code plug the specific logging setup to acquire notifications from Database. There are various ways to do this, maybe we need a signal type to do this separation more easilly in the future.

PR #3837, opened 2025-02-24
feat: add envs flag to info command

SandrineP:

### DIFF
--- a/libmamba/include/mamba/api/env.hpp
+++ b/libmamba/include/mamba/api/env.hpp
@@ -18,10 +18,10 @@ namespace mamba
     class ChannelContext;
     class Configuration;
     class Context;
-    
+
     namespace details
     {
-        void set_env_list_subcommand(CLI::App* com, Configuration& config, std::string flag, std::string description);
+        void print_envs(Configuration& config);
     }
 }
 
--- a/libmamba/include/mamba/api/env.hpp
+++ b/libmamba/include/mamba/api/env.hpp
@@ -20,7 +20,7 @@ namespace mamba
     namespace detail
     {
         std::string get_env_name(const Context& ctx, const mamba::fs::u8path& px);
-        void print_envs_impl(Configuration& config);
+        void print_envs_impl(const Configuration& config);
     }
 }
 
--- a/libmamba/include/mamba/api/env.hpp
+++ b/libmamba/include/mamba/api/env.hpp
@@ -7,10 +7,6 @@
 #ifndef MAMBA_API_ENV_HPP
 #define MAMBA_API_ENV_HPP
 
-#include <string>
-
-#include <CLI/CLI.hpp>
-
 #include "mamba/api/configuration.hpp"
 
 namespace mamba
--- a/libmamba/include/mamba/api/env.hpp
+++ b/libmamba/include/mamba/api/env.hpp
@@ -4,8 +4,8 @@
 //
 // The full license is in the file LICENSE, distributed with this software.
 
-#ifndef MAMBA_API_LIST_HPP
-#define MAMBA_API_LIST_HPP
+#ifndef MAMBA_API_ENV_HPP
+#define MAMBA_API_ENV_HPP
 
 #include <string>
 
@@ -19,8 +19,9 @@ namespace mamba
     class Configuration;
     class Context;
 
-    namespace details
+    namespace detail
     {
+        std::string get_env_name(const Context& ctx, const mamba::fs::u8path& px);
         void print_envs(Configuration& config);
     }
 }
--- a/libmamba/include/mamba/api/env.hpp
+++ b/libmamba/include/mamba/api/env.hpp
@@ -15,10 +15,12 @@ namespace mamba
     class Configuration;
     class Context;
 
+    void print_envs(Configuration& config);
+
     namespace detail
     {
         std::string get_env_name(const Context& ctx, const mamba::fs::u8path& px);
-        void print_envs(Configuration& config);
+        void print_envs_impl(Configuration& config);
     }
 }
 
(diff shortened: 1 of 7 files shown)


jjerphan (commented):
`--base` and `--envs` are incompatible, so we need to adapt `test_subcommand` accordingly.

libmamba/src/api/info.cpp
```diff
@@ -79,25 +82,67 @@ namespace mamba
...
         {
             assert(&ctx == &config.context());
 
             std::vector<std::tuple<std::string, nlohmann::json>> items;
 
-            if (options.base)
+            if (options.print_licenses)
+            {
+                const std::vector<std::pair<std::string, std::string>> licenses = {
+                    { "micromamba",
```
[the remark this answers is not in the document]
SandrineP: Yes, it is the original code that I moved. Not sure what should go there...

micromamba/src/env.cpp
```diff
@@ -52,45 +32,7 @@ set_env_command(CLI::App* com, Configuration& config)
...
-
-            for (auto& env : env_manager.list_all_known_prefixes())
-            {
-                bool is_active = (env == ctx.prefix_params.target_prefix);
-                t.add_row({ get_env_name(ctx, env), is_active ? "*" : "", env.string() });
-            }
-            t.print(std::cout);
-        }
-    );
+    list_subcom->callback([&config] { mamba::print_envs(config); });
```
jjerphan, line 35:
```suggestion
    list_subcom->callback([&config] {
        // TODO: make `Configuration::load` idempotent.
        config.load();
        mamba::print_envs(config);
    });
```

jjerphan (commented):
This review hints a problem which is unrelated to the scope of this PR, which was existing before, but which I think which must be resolved nevertheless in another PR.

libmamba/src/api/env.cpp
```diff
@@ -0,0 +1,78 @@
...
+#include "mamba/core/environments_manager.hpp"
+#include "mamba/util/string.hpp"
+
+namespace mamba
+{
+    void print_envs(Configuration& config)
+    {
+        config.load();
+        mamba::detail::print_envs_impl(config);
+    }
```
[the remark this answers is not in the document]
JohanMabille: - The config cannot be loaded in its constructor, as commands may insert additional specific Configurable objects; or that means moving *all* the Configuratble object to the constructor, which is probably not something we want to do. 
- The commands in micromamba should be responsible for adding subcommands, flags and variables to store the CLI values only. They should *never* call any logic regarding the configuration loading (or any other logic specific to libmamba).
- `config.load` is a requirement of most of the command implementations in libmamba, therefore it should be called by the libmamba functions themselves. Mixing calls to config.load in micromamba and libmamba leads to the bug ecountered here: when factorizing code to avoid duplication, one needs to read the whole implementation to know whether the config has already been loaded or not. Following a systematic policy regarding the loading of the configuration avoids this kind of bugs and make the code much cleaner: micromamba inserts any Configurable it needs, API functions of libmamba starts with calling config.load, and the implementation functions never call config.load. API and implementation functions can be the same if the implementation function is not used elsewhere.

micromamba/src/env.cpp
JohanMabille: That is typically what we want to remove, see my previous cmmments.

libmamba/src/api/env.cpp
```diff
@@ -0,0 +1,78 @@
...
+            {
+                return mamba::fs::relative(px, ed).string();
+            }
+            else
+            {
+                return "";
+            }
+        }
+
+        void print_envs_impl(const Configuration& config)
```
jjerphan, line 41: Sorry, I think I should have been clearer: I meant that the signature should be changed, but other changes for the signature of this definition are necessary (_e.g._ the signature of the declaration of `print_envs_impl` at least must be changed).

(Also it seems that a `const-qualified l-value reference to the instance of `Context` could be passed.)

libmamba/include/mamba/api/env.hpp
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) 2019, QuantStack and Mamba Contributors
```
Hind-M, line 1:
```suggestion
// Copyright (c) 2025, QuantStack and Mamba Contributors
```

libmamba/src/api/env.cpp
```diff
@@ -0,0 +1,78 @@
+// Copyright (c) 2019, QuantStack and Mamba Contributors
```
Hind-M, line 1:
```suggestion
// Copyright (c) 2025, QuantStack and Mamba Contributors
```

micromamba/tests/test_info.py
```diff
@@ -80,19 +80,20 @@ def test_not_env(tmp_home, tmp_root_prefix, prefix_selection, existing_prefix):
 
 
 @pytest.mark.parametrize("base_flag", ["", "--base"])
+@pytest.mark.parametrize("envs_flag", ["", "-e", "--envs"])
 @pytest.mark.parametrize("json_flag", ["", "--json"])
 @pytest.mark.parametrize("prefix_selection", [None, "prefix", "name"])
-def test_base_subcommand(
-    tmp_home, tmp_root_prefix, prefix_selection, base_flag, json_flag, monkeypatch
+def test_subcommand(
```
Hind-M, line 86: The name of the test should stay as it is (i.e `test_base_subcommand`), because we are testing the base environment (`-n base` equivalent to `-p root_prefix`).
We should add `envs_flag` to other tests as well when relevant (`test_env` for example, maybe even all of them in the test file, I didn't check everything).

libmamba/src/api/info.cpp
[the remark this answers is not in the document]
jjerphan: Let's treat this in another PR?

jjerphan (approved):
LGTM.

Thank you, @SandrineP.

PR #3839, opened 2025-02-26
fix: Remove invalid cached tarballs

jjerphan:


libmamba/src/core/package_cache.cpp
```diff
@@ -168,7 +169,26 @@ namespace mamba
...
+                {
+                    msg << "  - Expected md5      : " << s.md5 << "\n"
+                        << "  - Effective md5     : " << validation::md5sum(tarball_path) << "\n";
+                }
+                if (!s.sha256.empty())
+                {
+                    msg << "  - Expected sha256   : " << s.sha256 << "\n"
+                        << "  - Effective sha256  : " << validation::sha256sum(tarball_path) << "\n";
+                }
+                msg << "Processing with its deletion.";
```
Klaim, line 189:
```suggestion
                msg << "Deleting " << tarball_path;
```
I think you meant "proceeding" initially.

libmamba/src/core/package_fetcher.cpp
```diff
@@ -162,7 +162,11 @@ namespace mamba
             {
                 LOG_ERROR << "Failed to download package from "
                           << error.transfer.value().effective_url << " (status "
-                          << error.transfer.value().http_status << ")";
+                          << error.transfer.value().http_status << ")\n"
+                          << "If you see this message repeatedly, the state of your installation might be corrupted,\n"
+                          << "in which case running `mamba clean --all` might fix it.\n"
+                          << "If you continue to meet this problem, please report this issue on mamba's issue tracker:\n"
+                          << "https://github.com/mamba-org/mamba/issues/new?template=bug.yml";
```
Klaim, line 169:
```suggestion
                          << "https://github.com/mamba-org/mamba/issues/";
```

So that they first check existing issues.

libmamba/src/core/package_cache.cpp
```diff
@@ -168,7 +169,26 @@ namespace mamba
...
+                LOG_TRACE << msg.str();
+                fs::remove(tarball_path);
```
Klaim, line 191: I suspect that we want to continue if the removal failed (for example because of some weird suddenly forbidden access). Consider using the `std::error_code` taking overload and reporting the error if it's set, continue in all cases.

libmamba/src/core/package_cache.cpp
```diff
@@ -168,7 +169,26 @@ namespace mamba
             }
             else
             {
-                LOG_WARNING << "Package tarball '" << tarball_path.string() << "' is invalid";
+                std::stringstream msg;
```
Klaim, line 172: Consider putting all this in a separate function for clarity. Not a big need, I just suspect it might help reading. If you end up feeling it doesnt, ignore this comment.

jjerphan: I think that inlining the message is clearer if it is only used once.

JohanMabille (approved):

libmamba/src/core/package_cache.cpp
```diff
@@ -168,7 +169,38 @@ namespace mamba
             }
             else
             {
-                LOG_WARNING << "Package tarball '" << tarball_path.string() << "' is invalid";
```
Hind-M, line 171: I think this message should stay as a WARNING as we should see it even if we don't ask for verbose mode, the other added messages would stay indeed as TRACE.

libmamba/src/core/package_cache.cpp
```diff
@@ -168,7 +169,38 @@ namespace mamba
...
+                if (!s.sha256.empty())
+                {
+                    msg << "  - Expected sha256   : " << s.sha256 << "\n"
+                        << "  - Effective sha256  : " << validation::sha256sum(tarball_path) << "\n";
+                }
+                msg << "Deleting '" << tarball_path.string() << "'";
+                LOG_TRACE << msg.str();
+
+                std::error_code ec;
+                fs::remove(tarball_path, ec);
```
Hind-M, line 193: Not sure about this:
1- The tarball may be removed somewhere else later in the codebase and we shouldn't perform the removal here (again),so we should check that.
2- We could have some use cases where the tarball is invalid, there is a warning, but we still continue and don't really care... And so removing it here would introduce a bug (this is just an assumption).

libmamba/src/core/package_cache.cpp
jjerphan: My perspective is that warnings must be shown only if this user is responsible for some unusual operations they can understand in context. I don't think this applies here (if the state of the cache is corrupted or invalid, the users are not responsible for it, but network failure generally are for instance).

IMO cache invalidations might happen but users should not be aware of them, and the caching mechanism should be responsible for updating the state of the cache (here removing the tarball) without making the overall system inconsistent.

This was my intend with changing this line from a warning to a trace.

libmamba/src/core/package_cache.cpp
jjerphan: > 1- The tarball may be removed somewhere else later in the codebase and we shouldn't perform the removal here (again),so we should check that.

I am checking this at the moment. The deletion here must not be fatal in theory, but I do not know if it might in practice.

> 2- We could have some use cases where the tarball is invalid, there is a warning, but we still continue and don't really care... And so removing it here would introduce a bug (this is just an assumption).

I think this is clearly an anti-pattern for a cache and the reason of potential subsequent failures.

libmamba/src/core/package_cache.cpp
Hind-M: IMHO even if you are not responsible for it (tarball invalidity) the warning should help the user understand what's going on. Warning logs are also here to provide a better user experience.

jjerphan: Warnings are useful if we give users some context and the ability to resolve the situation (which is currently not the case with this one IMO).

In this case, we could mention to the users that there is a cache in place which is used to avoid re-downloading packages and that one of their tarball is corruption, explaining that they can remove it manually or clear the entire caches (_e.g._ using `mamba clean --all`), but this is not something they should be doing or even think about in the first place, isn't it?

Hind-M: To me, the user is either using `mamba` at a very high level and will ignore the warning (as it is not an error) or he is familiar with (the) software (in general) and the context is clear enough for him to understand what's happening.
Cleaning the cache is a general enough concept for people to understand and use, especially with the logs you added.

jjerphan: I guess the hint given on the error is useful and sufficient in this case.

What do you think?

libmamba/src/core/package_cache.cpp
JohanMabille: > 1- The tarball may be removed somewhere else later in the codebase and we shouldn't perform the removal here (again),so we should check that.

Except in independent commands (like clean pointed out by Julien above), the tarball is not removed explicitly. An invalid tarball will trigger a new download of the package that will overwrite the invalid tarball. I would say that the removal here is harmless if the workflow succeeds (the tarball would have been overwritten), and leaves the cache in a clean state i the workflow fails for any reason.

> 2- We could have some use cases where the tarball is invalid, there is a warning, but we still continue and don't really care... And so removing it here would introduce a bug (this is just an assumption).

I don't see why that would be a bug. The cache is supposed to be transparent for the user, I don't see any usecase where we would need to keep an invalid tarball in the cache. Besides, all the information we could need to understand why the tarball was invalid are given in the logs, so I think it is safe to remove it.

libmamba/src/core/package_cache.cpp
Hind-M: In my opinion, the original line warning should stay.

JohanMabille: I agree they don't help to solve the problem, but they give a hint whjy mamba is downloading a package again while it should be in cache for instance. But honestly, I'm fine with both options, I don't have a strong opinion on this.

jjerphan: The tarball is deleted then, so the problem is fixed by mamba; that's why I think there's no reason to report the invalidity in the first place.

JohanMabille: Thinking more of this, with the latest URL change, once 2.0.7 is released, users with huge environment could be swamped with a lot of these warnings. Let's merge as is with a trace instead of a warning, we can rollback to the warning later when most of the users have upgraded.

PR #3842, opened 2025-02-27
maint: Remove unused `CURLMultiHandle::poll`

jjerphan:


JohanMabille:
This is removing an API, it is therefore backward incompatible, it should not be labelled ci_docs IMO.

jjerphan:
> it should not be labelled ci_docs IMO

Yes, I am waiting for #3843 to get it before changing the label.

> This is removing an API, it is therefore backward incompatible

So we need to wait for a major version?

jjerphan:
Closing for now, might reopen later.

Issue #3840, opened 2025-02-26
Unable to use micromamba with networked file system


chrisburr:
### Troubleshooting docs

- [x] My problem is not solved in the Troubleshooting docs

### Anaconda default channels

- [x] I do NOT use the Anaconda default channels (pkgs/* etc.)

### How did you install Mamba?

Micromamba

### Search tried in issue tracker

https://github.com/mamba-org/mamba/issues/3740

### Latest version of Mamba

- [x] My problem is not solved with the latest version

### Tried in Conda?

I do not have this problem with Conda, just with Mamba

### Describe your issue

Similar to https://github.com/mamba-org/mamba/issues/3740 but opening a new issue as that one was closed.

Creating environments fails with:

```bash
$ ./bin/micromamba create -c conda-forge --prefix $PWD/my-env python
critical libmamba filesystem error: cannot set permissions: Permission denied [/afs/cern.ch/work/c/cburr/root-prefix/pkgs/cache]
```

With strace I can see:

```log
[pid 564852] <... clock_nanosleep resumed>0x7fffff000040) = 0
[pid 564852] stat("/afs/cern.ch/work/c/cburr/root-prefix/pkgs/cache", {st_mode=S_IFDIR|0755, st_size=2048, ...}) = 0
[pid 564852] fchmodat(AT_FDCWD, "/afs/cern.ch/work/c/cburr/root-prefix/pkgs/cache", 02775) = -1 EACCES (Permission denied)
[pid 564852] openat(AT_FDCWD, "/usr/share/locale/C.UTF-8/LC_MESSAGES/libc.mo", O_RDONLY) = -1 ENOENT (No such file or directory)
[pid 564852] openat(AT_FDCWD, "/usr/share/locale/C.utf8/LC_MESSAGES/libc.mo", O_RDONLY) = -1 ENOENT (No such file or directory)
[pid 564852] openat(AT_FDCWD, "/usr/share/locale/C/LC_MESSAGES/libc.mo", O_RDONLY) = -1 ENOENT (No such file or directory)
[pid 564852] futex(0x559e5c8c7b20, FUTEX_WAKE_PRIVATE, 2147483647) = 0
[pid 564852] write(2, "\33[1m\33[41m", 9) = 9
[pid 564852] write(2, "critical libmamba", 17critical libmamba) = 17
[pid 564852] write(2, "\33[m", 3)       = 3
[...]

PR #3844, opened 2025-02-27
fix: Modify cache directory permissions in two steps

jjerphan:
Fix https://github.com/mamba-org/mamba/issues/3840.

Though this fixes the problem (as reported by @chrisburr in https://github.com/mamba-org/mamba/issues/3840#issuecomment-2684940010), I do not know how we could test this yet.

### DIFF
--- a/libmamba/tests/src/core/test_filesystem.cpp
+++ b/libmamba/tests/src/core/test_filesystem.cpp
@@ -363,9 +363,7 @@ namespace mamba
                 }
             );
 
-            // Get the information about whether the filesystem supports special bits permissions
-            // For this test, we need to check if the filesystem supports the sticky bit using try
-            // catch
+            // Get the information about whether the filesystem supports the `set_gid` bit.
             bool supports_setgid_bit = false;
 
             try
@@ -381,7 +379,7 @@ namespace mamba
 
             fs::create_directories(create_cache_dir);
 
-            // Check that the permissions of `tmp_dir` are 'rwxr-xr-x'
+            // Check that the permissions of `create_cache_dir` are 'rwxr-xr-x'
             auto create_cache_dir_permissions = fs::status(create_cache_dir).permissions();
 
             REQUIRE((create_cache_dir_permissions & fs::perms::owner_all) == fs::perms::owner_all);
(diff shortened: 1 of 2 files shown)


Klaim (commented):
As you suggested privately, a test would be welcome. Here is how I would do it:
- make the test call `create_cache_dir` with a [sub-directory path under the temporary directory provided by the system](https://en.cppreference.com/w/cpp/filesystem/temp_directory_path)
- this call should never fail in any way other than if the system is in a situation it cant even create the directory (nothing to check in the unit test here);
- check the created directory's permissions, but
   - depending on the confidence we have that `std::filesystem::perms::set_gid` is supported on the current platform/filesystem, check that one too - dont if we cant make sure it should be supported;
Alternatively we could alter the output to provide a list of errors too or something but that might not be worth the api change.

libmamba/tests/src/core/test_filesystem.cpp
```diff
@@ -349,6 +351,50 @@ namespace mamba
...
+
+            if (!ec)
+            {
+                supports_setgid_bit = (fs::status(set_gid_tmp_dir).permissions() & fs::perms::set_gid)
+                                      == fs::perms::set_gid;
+            }
+
+            create_cache_dir(cache_dir);
+
+            // Check that the permissions of `cache_dir` are _at least_ `rwxr-xr-x`
```
jjerphan, line 382: It has been observed on Linux, that `rwx` is not necessarily set (for the group) even if explicitly specified (in which case `r-x` is set instead).

libmamba/tests/src/core/test_filesystem.cpp
```diff
@@ -349,6 +351,50 @@ namespace mamba
...
+            fs::permissions(set_gid_tmp_dir, fs::perms::set_gid, fs::perm_options::add, ec);
+
+            if (!ec)
+            {
+                supports_setgid_bit = (fs::status(set_gid_tmp_dir).permissions() & fs::perms::set_gid)
+                                      == fs::perms::set_gid;
+            }
+
+            create_cache_dir(cache_dir);
+
```
Klaim, line 381:
```suggestion
            REQUIRE(fs::is_directory(cache_dir));
```
It is not clear to me if `fs::status(cache_dir).permissions()` will end up in error if the directory doesnt exist after calling this functions, so in doubt I would explicitly check the assumption that if it didnt fail the directory indeed exists.
Even better if you also check that it does not exist before calling `create_cache_dir`.

This is all an improvement suggestion.

Klaim (approved):
LGTM, with potential improvement.

Klaim (approved):

PR #3849, opened 2025-03-01
Avoid possible out of range index in MatchSpec::parse()

opoplawski:
If parse() were passed a string that started with `#` it would try to access `str[-1]` which would be an invalid reference.

### DIFF
--- a/libmamba/src/specs/match_spec.cpp
+++ b/libmamba/src/specs/match_spec.cpp
@@ -505,7 +505,7 @@ namespace mamba::specs
     auto MatchSpec::parse(std::string_view str) -> expected_parse_t<MatchSpec>
     {
         // Remove comments, i.e. everything after ` #` (space included)
-        if (const auto idx = str.find('#'); idx != std::string::npos && str[idx - 1] == ' ')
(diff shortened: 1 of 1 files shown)


libmamba/src/specs/match_spec.cpp
```diff
@@ -505,9 +505,9 @@ namespace mamba::specs
     auto MatchSpec::parse(std::string_view str) -> expected_parse_t<MatchSpec>
     {
         // Remove comments, i.e. everything after ` #` (space included)
-        if (const auto idx = str.find('#'); idx != std::string::npos && str[idx - 1] == ' ')
+        if (const auto idx = str.find(" #"); idx != std::string::npos)
         {
-            str = str.substr(0, idx);
+            str = str.substr(0, idx + 1);
```
jjerphan, line 510: The space can also be omitted:
```suggestion
            str = str.substr(0, idx);
```

jjerphan (approved):
Thank you, @opoplawski.

libmamba/src/specs/match_spec.cpp
opoplawski: Sounds good, done.

PR #3852, opened 2025-03-05
fix: Adapt root prefix' precedence for CONDA_ENVS_PATH

### LINKED ISSUE #3851: CONDA_ENVS_PATH precedence is incorrect
### Troubleshooting docs

### Anaconda default channels

CONDA_ENVS_PATH

### Latest version of Mamba

### Tried in Conda?

### Describe your issue

I don't think #3654 was completely fixed.

```
$ mkdir -p /tmp/test/envs && CONDA_ENVS_PATH=/tmp/test/envs mamba create -n test1
Empty environment created at prefix: /home/burt/micromamba/envs/test1
```

Pretty sure this is because the order in `envs_dirs_hook` needs to be switched - I'll send in a PR with a fix and test shortly.  

```shell
This isn't working on mamba 2 (tested 2.0.5, 2.0.6, and 04b2f2ca)
```

### Logs

```shell

```

### environment.yml

```yaml

```

### ~/.condarc

```yaml

```

holzman:
Fixes #3851.

```
$ mkdir -p /tmp/test/envs && CONDA_ENVS_PATH=/tmp/test/envs mamba create -n test1
Empty environment created at prefix: /tmp/test/envs/test1
```

### DIFF
--- a/libmamba/src/api/configuration.cpp
+++ b/libmamba/src/api/configuration.cpp
@@ -1007,22 +1007,17 @@ namespace mamba
         {
             // Check that "root_prefix/envs" is already in the dirs,
             // and append if not - to match `conda`
-            fs::u8path default_env_dir = context.prefix_params.root_prefix / "envs";
-            if (std::find(dirs.begin(), dirs.end(), default_env_dir) == dirs.end())
-            {
-                dirs.push_back(default_env_dir);
-            }
 
-            // Also add all the directories in the environment variable `CONDA_ENVS_PATH`.
+            // Also prepend all the directories in the environment variable `CONDA_ENVS_PATH`.
             auto conda_envs_path = util::get_env("CONDA_ENVS_PATH");
             if (conda_envs_path)
             {
                 auto paths_separator = util::pathsep();
 
                 auto paths = util::split(conda_envs_path.value(), paths_separator);
-                for (auto& p : paths)
-                {
-                    dirs.push_back(fs::u8path(p));
+                for (auto it = paths.rbegin(); it != paths.rend(); ++it)
+                {  // prepend conda_envs_path to dirs while maintaining order
+                    dirs.insert(dirs.begin(), fs::u8path(*it));
                 }
             }
 
@@ -1035,6 +1030,12 @@ namespace mamba
                     LOG_ERROR << "Env dir specified is not a directory: " << d.string();
                     throw std::runtime_error("Aborting.");
                 }
+            }
+
+            fs::u8path default_env_dir = context.prefix_params.root_prefix / "envs";
+            if (std::find(dirs.begin(), dirs.end(), default_env_dir) == dirs.end())
+            {
+                dirs.push_back(default_env_dir);
             }
         }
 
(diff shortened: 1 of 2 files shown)


micromamba/tests/test_create.py
```diff
@@ -670,6 +670,7 @@ def test_create_envs_dirs(tmp_root_prefix: Path, tmp_path: Path, monkeypatch):
     noperm_envs_dir = noperm_root_dir / "envs"
 
     monkeypatch.setenv("CONDA_ENVS_DIRS", f"{noperm_envs_dir},{tmp_path}")
+    #    monkeypatch.setenv("CONDA_ENVS_PATH", f"{noperm_envs_dir}:{tmp_path}")
```
holzman, line 673: Ah, accidentally left that commented-out line in from testing. I'll remove it.

libmamba/src/api/configuration.cpp
```diff
@@ -1007,22 +1007,17 @@ namespace mamba
         {
             // Check that "root_prefix/envs" is already in the dirs,
             // and append if not - to match `conda`
```
Hind-M, line 1009: Can you move the comments along with the moved piece of code below?

micromamba/tests/test_create.py
```diff
@@ -757,6 +758,7 @@ def test_env_dir_idempotence(tmp_path, tmp_home, tmp_root_prefix):
         assert Path(condarc_envs_dirs / env_name).exists()
 
 
+@pytest.mark.parametrize("conda_envs_x", ("CONDA_ENVS_DIRS", "CONDA_ENVS_PATH"))
```
Hind-M, line 761: We should also test having both of the environment variables set and check the precedence?

libmamba/src/api/configuration.cpp
```diff
@@ -1007,22 +1007,17 @@ namespace mamba
...
-            fs::u8path default_env_dir = context.prefix_params.root_prefix / "envs";
-            if (std::find(dirs.begin(), dirs.end(), default_env_dir) == dirs.end())
-            {
-                dirs.push_back(default_env_dir);
-            }
 
-            // Also add all the directories in the environment variable `CONDA_ENVS_PATH`.
+            // Also prepend all the directories in the environment variable `CONDA_ENVS_PATH`.
```
Hind-M, line 1011:
```suggestion
            // Prepend all the directories in the environment variable `CONDA_ENVS_PATH`.
```

micromamba/tests/test_create.py
holzman: Sure, I wasn't sure the behavior was defined, but I'll check what mamba 1 and conda do when both are set.

holzman: Makes sense.  I started looking into it and realized there's another incompatibility in the syntax. 

In conda and mamba1, both `CONDA_ENVS_DIRS` and `CONDA_ENVS_PATHS` are path-separator delimited (colon on Linux, semicolon on Windows).

In mamba2, `CONDA_ENVS_DIRS` is comma-delimited instead.

holzman: Ok: submitted #3854 to address the case if someone sets both env vars and #3855 to unify the behavior using the path separator as delimiter.

Hind-M (approved):
Thanks!

Issue #3845, opened 2025-02-27
Mamba 2.0 does not allow creating an environment prefix in place of an existing directory


bernt-matthias:
### Troubleshooting docs

- [x] My problem is not solved in the Troubleshooting docs

### Anaconda default channels

- [x] I do NOT use the Anaconda default channels (pkgs/* etc.)

### How did you install Mamba?

Mambaforge or latest Miniforge

### Search tried in issue tracker

https://github.com/mamba-org/mamba/issues/2402

### Latest version of Mamba

- [x] My problem is not solved with the latest version

### Tried in Conda?

I do not have this problem with Conda, just with Mamba

### Describe your issue

If there an existing folder `X` I get for `mamba` >= 2.0:

```
mamba create -p /usr/local/
error    libmamba Non-conda folder exists at prefix                                                                                                                   
critical libmamba Aborting.          
```

and for `Mamba < 2.0` 

```
WARNING: A directory already exists at the target location '/usr/local'
but it is not a conda environment.
```

conda behaves like `mamba` < 2.0

I'm using the latest miniforge container. Mamba >=2 is installed with conda in the container.

see also https://github.com/mamba-org/mamba/issues/2402 

### mamba info / micromamba info

```shell
active environment : base
    active env location : /opt/conda
            shell level : 1
       user config file : /root/.condarc
 populated config files : /opt/conda/.condarc
          conda version : 25.1.1
    conda-build version : not installed
         python version : 3.12.8.final.0
                 solver : libmamba (default)
       virtual packages : __archspec=1=skylake
                          __conda=25.1.1=0
                          __glibc=2.31=0
                          __linux=6.11.0=0
                          __unix=0=0
       base environment : /opt/conda  (writable)
      conda av data dir : /opt/conda/etc/conda
  conda av metadata url : None
           channel URLs : https://conda.anaconda.org/conda-forge/linux-64
                          https://conda.anaconda.org/conda-forge/noarch
          package cache : /opt/conda/pkgs
                          /root/.conda/pkgs
       envs directories : /opt/conda/envs
                          /root/.conda/envs
               platform : linux-64
             user-agent : conda/25.1.1 requests/2.32.3 CPython/3.12.8 Linux/6.11.0-17-generic ubuntu/20.04.6 glibc/2.31 solver/libmamba conda-libmamba-solver/25.1.1 libmambapy/2.0.5
                UID:GID : 0:0
             netrc file : None
           offline mode : False
```

### Logs

```shell

```

[...]

PR #3853, opened 2025-03-06
fix: Allow prefix creation in place of an existing folder

jjerphan:
Fix https://github.com/mamba-org/mamba/issues/3845.

I am thinking of a way to test this.

### DIFF
--- a/libmamba/src/api/create.cpp
+++ b/libmamba/src/api/create.cpp
@@ -5,6 +5,7 @@
 // The full license is in the file LICENSE, distributed with this software.
 
 #include <iostream>
+#include <sstream>
 
 #include "mamba/api/configuration.hpp"
 #include "mamba/api/create.hpp"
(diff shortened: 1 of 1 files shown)


libmamba/src/api/create.cpp
```diff
@@ -64,8 +65,22 @@ namespace mamba
...
+                    std::stringstream msg;
+
+                    msg << "WARNING: A directory already exists at the target location '";
+                    msg << ctx.prefix_params.target_prefix.string();
+                    msg << "'\nbut it is not a conda environment.\n";
+                    msg << "Continue creating environment?";
+
+                    if (Console::prompt(msg.str(), 'n'))
+                    {
+                        fs::remove_all(ctx.prefix_params.target_prefix);
```
Klaim, line 77: An improvement would be to do it transactionally: at the beginning of the function, use `on_scope_exit` with this code with a condition on a boolean which would tell if we need to do this delete. By default we do, except if going in the other case. That way future exceptions/errors will also remove the directory.

But that's a minor improvement here.

Klaim (approved):

jjerphan:
Closing in preference of https://github.com/mamba-org/mamba/pull/3919.

PR #3854, opened 2025-03-06
fix: Prohibit conda envs path and conda envs dirs

holzman:
`CONDA_ENVS_PATH` and `CONDA_ENVS_DIRS` are mutually exclusive in conda and mamba v1.

Based on this comment: https://github.com/mamba-org/mamba/pull/3852#discussion_r1983545819_

### DIFF
--- a/libmamba/src/api/configuration.cpp
+++ b/libmamba/src/api/configuration.cpp
@@ -1007,15 +1007,21 @@ namespace mamba
         {
             // Prepend all the directories in the environment variable `CONDA_ENVS_PATH`.
             auto conda_envs_path = util::get_env("CONDA_ENVS_PATH");
+
             if (conda_envs_path)
             {
+                if (util::get_env("CONDA_ENVS_DIRS"))
+                {
+                    const auto message = "The `CONDA_ENVS_DIRS` and `CONDA_ENVS_PATH` environment variables are both set, but only one must be declared. We recommend setting `CONDA_ENVS_DIRS` only. Aborting.";
+                    throw mamba_error(message, mamba_error_code::incorrect_usage);
+                }
+
                 auto paths_separator = util::pathsep();
 
                 auto paths = util::split(conda_envs_path.value(), paths_separator);
-                for (auto it = paths.rbegin(); it != paths.rend(); ++it)
-                {  // prepend conda_envs_path to dirs while maintaining order
-                    dirs.insert(dirs.begin(), fs::u8path(*it));
-                }
+
+                dirs.reserve(dirs.size() + paths.size());
+                dirs.insert(dirs.begin(), paths.rbegin(), paths.rend());
             }
 
             // Check that the values exist as directories
@@ -1031,7 +1037,7 @@ namespace mamba
 
             // Check that "root_prefix/envs" is already in the dirs,
             // and append if not - to match `conda`
-            fs::u8path default_env_dir = context.prefix_params.root_prefix / "envs";
+            const fs::u8path default_env_dir = context.prefix_params.root_prefix / "envs";
             if (std::find(dirs.begin(), dirs.end(), default_env_dir) == dirs.end())
             {
                 dirs.push_back(default_env_dir);
--- a/micromamba/tests/test_create.py
+++ b/micromamba/tests/test_create.py
@@ -663,6 +663,20 @@ def test_create_empty(tmp_home, tmp_root_prefix, tmp_path, prefix_selector, crea
     assert (effective_prefix / "conda-meta" / "history").exists()
 
 
+def test_create_conda_envs_dirs_and_path(tmp_root_prefix, monkeypatch):
+    """Abort when CONDA_ENVS_PATH and CONDA_ENVS_DIRS are both set"""
+    monkeypatch.setenv("CONDA_ENVS_DIRS", f"{tmp_root_prefix / 'env1'}")
+    monkeypatch.setenv("CONDA_ENVS_PATH", f"{tmp_root_prefix / 'env2'}")
+    with pytest.raises(subprocess.CalledProcessError) as info:
+        helpers.create("-n", "test", "--offline", "--no-rc", no_dry_run=True)
+
+    msg = info.value.stderr.decode()
+    assert (
+        "The `CONDA_ENVS_DIRS` and `CONDA_ENVS_PATH` environment variables are both set, but only one must be declared."
+        in msg
+    )
+
+
 def test_create_envs_dirs(tmp_root_prefix: Path, tmp_path: Path, monkeypatch):
     """Create an environment when the first env dir is not writable."""
 


libmamba/src/api/configuration.cpp
```diff
@@ -1005,24 +1005,23 @@ namespace mamba
...
+
                 auto paths_separator = util::pathsep();
 
                 auto paths = util::split(conda_envs_path.value(), paths_separator);
-                for (auto& p : paths)
-                {
-                    dirs.push_back(fs::u8path(p));
+                for (auto it = paths.rbegin(); it != paths.rend(); ++it)
+                {  // prepend conda_envs_path to dirs while maintaining order
+                    dirs.insert(dirs.begin(), fs::u8path(*it));
```
jjerphan, line 1024: This will come with `paths.size()` reallocation. Could an alternative be:

```c++
std::move(dirs.begin(), dirs.end(), std::back_inserter(paths));
dirs = paths;
```

?

libmamba/src/api/configuration.cpp
```diff
@@ -1036,6 +1035,14 @@ namespace mamba
                     throw std::runtime_error("Aborting.");
                 }
             }
+
+            // Check that "root_prefix/envs" is already in the dirs,
+            // and append if not - to match `conda`
+            fs::u8path default_env_dir = context.prefix_params.root_prefix / "envs";
```
jjerphan, line 1041:
```suggestion
            const fs::u8path default_env_dir = context.prefix_params.root_prefix / "envs";
```

jjerphan (commented):
Thank you, @holzman.

A few comments and suggestions. Apart from them, this contribution looks good to me.

libmamba/src/api/configuration.cpp
holzman: That gets a bit messy since one is a vector of u8path and one is a vector of string.  How about `reserve()` and `emplace()` instead?

jjerphan: I guess you meant `resize`, in that case yes (I have seen your comment).

holzman: My C++ is admittedly very rusty but I believe we do want `reserve` instead of `resize` (`resize` would default-construct elements at the end of the list).

jjerphan: I meant, if one is emplacing at `dirs.begin()`, all the elements need to be moved toward the end so this is $\mathcal{O}\left(n^2\right)$, where $n$ is the number of elements.

I guess we could use a `std::dequeue` but we still need to convert it at the end anyway.

holzman: Sure - the downside of a vector is the added expense for insertions.  Given that the function only gets executed once and the number of elements is probably going to be small (I would expect 0-2 in nearly all cases), I'd argue we should investigate further optimization (e.g. replacing it across the board with a `std::list` or `std::deque`) in a separate PR

jjerphan: Yes, I also agree, also sometimes operations are generally faster with `std::vector` despite a worst complexity of operations thanks to cache/memory locality.

We can merge this PR once this the conflicts are resolved. :slightly_smiling_face:

libmamba/src/api/configuration.cpp
```diff
@@ -1007,14 +1007,24 @@ namespace mamba
         {
             // Prepend all the directories in the environment variable `CONDA_ENVS_PATH`.
             auto conda_envs_path = util::get_env("CONDA_ENVS_PATH");
+
             if (conda_envs_path)
             {
+                if (util::get_env("CONDA_ENVS_DIRS"))
+                {
+                    LOG_ERROR << "The `CONDA_ENVS_DIRS` and `CONDA_ENVS_PATH` environment variables are both set, but only one must be declared.\nWe recommend setting `CONDA_ENVS_DIRS` only.)";
```
jjerphan, line 1015:
```suggestion
                    LOG_ERROR << "The `CONDA_ENVS_DIRS` and `CONDA_ENVS_PATH` environment variables are both set, but only one must be declared.\nWe recommend setting `CONDA_ENVS_DIRS` only.";
```

jjerphan (approved):

libmamba/src/api/configuration.cpp
holzman: Whoops! Code is easy, English is hard!

jjerphan: Probably due to one of my previous comment.

libmamba/src/api/configuration.cpp
```diff
@@ -1007,14 +1007,24 @@ namespace mamba
...
                 auto paths_separator = util::pathsep();
 
                 auto paths = util::split(conda_envs_path.value(), paths_separator);
+
+                dirs.reserve(dirs.size() + paths.size());
+
                 for (auto it = paths.rbegin(); it != paths.rend(); ++it)
                 {  // prepend conda_envs_path to dirs while maintaining order
-                    dirs.insert(dirs.begin(), fs::u8path(*it));
+                    dirs.emplace(dirs.begin(), fs::u8path(*it));
```
Klaim, line 1027: Note that I don't think changing `insert` to `emplace` here is an improvement: in both cases, because we pass the  value of the same type than the container's elements type, the container's new value will be move-constructed from that temporary (which might be magically optimized in both cases, but no idea if it actually is). 
If we were passing *non-copy-and-non-move constructor parameters* to `emplace` then it would have made a positive difference compared to `insert` (if my understanding is correct).
Anyway not important, just a clarification that this doesn't change anything AFAIK.

What would be an improvement (this is just a suggestion in case you want to follow the boyscout's rule, otherwise I'll note to do it in a future pr) would be to not insert elements one by one in the beginning of a vector in a loop , that moves all the elements around at each iteration (the reserve only prevents reallocations, not the moves). Instead it would be far more efficient and shorter to replace the whole loop by a call to `dirs.insert(dirs.begin(), paths.rbegin(), paths.rend())`.

libmamba/src/api/configuration.cpp
```diff
@@ -1007,14 +1007,24 @@ namespace mamba
         {
             // Prepend all the directories in the environment variable `CONDA_ENVS_PATH`.
             auto conda_envs_path = util::get_env("CONDA_ENVS_PATH");
+
             if (conda_envs_path)
             {
+                if (util::get_env("CONDA_ENVS_DIRS"))
+                {
+                    LOG_ERROR << "The `CONDA_ENVS_DIRS` and `CONDA_ENVS_PATH` environment variables are both set, but only one must be declared.\nWe recommend setting `CONDA_ENVS_DIRS` only.";
+                    throw std::runtime_error("Aborting.");
```
Klaim, line 1016: Throw [`mamba::mamba_error`](https://github.com/mamba-org/mamba/blob/main/libmamba/include/mamba/core/error_handling.hpp) instead.

Also the message should be part of both the log and the exception. You can end the message with `"... - Aborting."`

Klaim (commented):
LGTM, with just a fix on the exception.

libmamba/src/api/configuration.cpp
jjerphan: One of the threads (https://github.com/mamba-org/mamba/pull/3854#discussion_r1986946770) discussed similar aspects.

holzman: Well, that makes a lot more sense than what I was doing, thanks @Klaim !

libmamba/src/api/configuration.cpp
holzman: I can change this if you like, but just about everywhere else in `configuration.cpp` it streams the message to `LOG_ERROR` and throws a std::runtime_error with `Aborting.`

e.g.:
https://github.com/mamba-org/mamba/blob/main/libmamba/src/api/configuration.cpp#L548-L549
https://github.com/mamba-org/mamba/blob/main/libmamba/src/api/configuration.cpp#L911-L918

etc.

jjerphan: Eventually they all should be changed, but I am indifferent that they be changed now or later.

holzman: Ok, no problem.
Are you sure you want the message in both the log and the exception?  If so it shows up twice:

```
trace    libmamba Compute configurable 'envs_dirs'
error    libmamba The `CONDA_ENVS_DIRS` and `CONDA_ENVS_PATH` environment variables are both set, but only one must be declared. We recommend setting `CONDA_ENVS_DIRS` only. Aborting.
critical libmamba The `CONDA_ENVS_DIRS` and `CONDA_ENVS_PATH` environment variables are both set, but only one must be declared. We recommend setting `CONDA_ENVS_DIRS` only. Aborting.
info     libmamba ****************** Backtrace End ********************
```

I'd rather just put it in the exception:

```
trace    libmamba Compute configurable 'envs_dirs'
critical libmamba The `CONDA_ENVS_DIRS` and `CONDA_ENVS_PATH` environment variables are both set, but only one must be declared. We recommend setting `CONDA_ENVS_DIRS` only. Aborting.
info     libmamba ****************** Backtrace End ********************
```

Klaim (approved):
LGTM 👍🏽

PR #3855, opened 2025-03-06
Unify CONDA_ENVS_PATH, CONDA_ENVS_DIRS

holzman:
`CONDA_ENVS_DIRS` and `CONDA_ENVS_PATH` use the OS's path separators in conda and mamba v1.

In mamba v2, `CONDA_ENVS_DIRS` uses `,` as the delimiter.  This PR unifies the two (using the OS's path
separator for both).

Based on https://github.com/mamba-org/mamba/pull/3852#discussion_r1983967141

**This could be a breaking change (if someone's changed to using the comma delimiter).**


libmamba/src/api/configuration.cpp
```diff
@@ -1350,7 +1363,6 @@ namespace mamba
         insert(Configurable("envs_dirs", &m_context.envs_dirs)
                    .group("Basic")
                    .set_rc_configurable(RCConfigLevel::kHomeDir)
-                   .set_env_var_names({ "CONDA_ENVS_DIRS" })
```
holzman, line 1353: I know it seems weird to take this out, but this Configurable deserializes the value of the env var assuming it's YAML (e.g. comma-delimited list).  It's delimited with the os path separator instead, so we deal with that special case in the hook.

micromamba/tests/test_activation.py
```diff
@@ -880,11 +880,11 @@ def test_activate_path(tmp_empty_env, tmp_env_name, interpreter, tmp_path):
 
 
 @pytest.mark.parametrize("interpreter", get_interpreters())
-def test_activate_envs_dirs(tmp_root_prefix: Path, interpreter, tmp_path: Path):
+def test_activate_envs_dirs(tmp_root_prefix: Path, interpreter, tmp_path: Path, monkeypatch):
     """Activate an environment as the non leading entry in ``envs_dirs``."""
     env_name = "myenv"
     helpers.create("-p", tmp_path / env_name, "--offline", "--no-rc", no_dry_run=True)
-    os.environ["CONDA_ENVS_DIRS"] = f"{Path('/noperm')},{tmp_path}"
+    monkeypatch.setenv("CONDA_ENVS_DIRS", f"{Path('/noperm')}:{tmp_path}")
```
holzman, line 887: Whoops, the colon needs to be os.pathsep.  I'll fix these this afternoon when I'm near a Windows machine for testing.

jjerphan (commented):
I think that the behavior of mamba 1 must be kept (so we could add a warning if commas are used).

jjerphan (approved):
LGTM, thank you @holzman.

Waiting for another review.

JohanMabille (approved):

JohanMabille:
Thanks @holzman !

PR #3866, opened 2025-03-18
Support SHA256 hashes in @EXPLICIT files

jaimergp:
- Ports https://github.com/conda/conda/pull/14048
- Relevant CEP at https://github.com/conda/ceps/pull/79

### DIFF
--- a/libmamba/src/specs/package_info.cpp
+++ b/libmamba/src/specs/package_info.cpp
@@ -215,7 +215,13 @@ namespace mamba::specs
                 return parse_url(url).transform(
                     [&](PackageInfo&& pkg) -> PackageInfo
                     {
-                        if (is_hash(hash))
+                        if (util::starts_with(hash, "sha256:")) {
+                            hash = hash.substr(7);
+                            if (hash.size() == 64 && is_hash(hash)) {
+                                pkg.sha256 = hash;
+                            }
+                        }
+                        else if (is_hash(hash))
                         {
                             if (hash.size() == 32)
                             {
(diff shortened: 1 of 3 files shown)


Hind-M:
Thanks!

Maybe also add some tests in [test_package_info.cpp](https://github.com/mamba-org/mamba/blob/main/libmamba/tests/src/specs/test_package_info.cpp#L20-L90):

- Always check `pkg.md5` and `pkg.sha256` values
- Add a section for a `hash.size=64` (`sha256` case)

Should we support `sha256:` in urls in this PR as well? (it can be done later otherwise since it's not mandatory).

jaimergp:
> Should we support sha256: in urls in this PR as well? (it can be done later otherwise since it's not mandatory).

I'll try my best 😬

libmamba/src/specs/package_info.cpp
jjerphan, line 1: I think we might want to do something in `else` cases.

How about raising a warning when hashes are malformed?

jjerphan (commented):
Thank you for this contribution, @jaimergp.

This LGTM modulo one discussion.

libmamba/src/specs/package_info.cpp
jaimergp: Yes, that sounds sensible. I didn't add them to respect the existing logic (silent ignore). I'm not sure how to add warnings in this codebase, so feel free to commit to the branch directly :)

Hind-M (approved):

jjerphan (approved):

PR #3878, opened 2025-04-04
Use Simdjson ondemand parser instead of DOM parser

Klaim:
Running `main`'s `test_libmamba` executable currently allocates (but dont use) more than 5Gb when parsing 320Mb sized JSON files. This PR reduces this to around 2.5Gb. (observed using Visual Studio, in both Release and RelWithDebugInfo, also got some confirmation that this is observable on Linux)

- use the "on-demand" parser from simdjson instead of the "DOM" parser
- reorganize the parsing code to avoid reading json fields values more than once as this is required for proper behavior of the "on-demand" parser;
- improved `flat_set<T>` to allow `.contains(value)` where `value` has a different type than `T` but is still comparable.

### DIFF
--- a/libmamba/src/solver/libsolv/helpers.cpp
+++ b/libmamba/src/solver/libsolv/helpers.cpp
@@ -365,9 +365,6 @@ namespace mamba::solver::libsolv
             return true;
         }
 
-        template <class T>
-        class showme;
-
         template <typename JSONObject, typename Filter, typename OnParsed>
         void set_repo_solvables_impl(
             solv::ObjPool& pool,
@@ -384,7 +381,6 @@ namespace mamba::solver::libsolv
             auto packages_as_object = packages.get_object();
             for (auto pkg_field : packages_as_object)
             {
-                // showme<decltype(packages)> debug;
                 const std::string filename(pkg_field.unescaped_key().value());
                 if (filter(filename))
                 {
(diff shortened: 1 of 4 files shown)


JohanMabille (approved):

AntoinePrv (approved):

libmamba/src/solver/libsolv/helpers.cpp
```diff
@@ -530,79 +548,135 @@ namespace mamba::solver::libsolv
...
+        // BEWARE:
+        // We use below `simdjson`'s "on-demand" parser, which does not tolerate reading the same
+        // value more than once. This means we need to make sure that the objects and their fields
+        // are read and/or concretized only once and if we need to use them more than once we need
+        // to persist them in local memory. This is why the code below tries hard to pre-read the
+        // data needed in several parts of the computing in a way that prevents jumping up and down
+        // the hierarchy of json objects. When this rule is not followed, the parsing might end
+        // earlier than expected or might skip data that are read when they shouldn't be, leading to
+        // *runtime issues* that might not be visible at first. Because of these reasons, be careful
+        // when modifying the following parsing code.
```
jjerphan, line 560: Thank you for mentioning this.

jjerphan (approved):

PR #3885, opened 2025-04-09
feat: add sha256 flag to list command

SandrineP:

### DIFF
--- a/libmamba/src/api/list.cpp
+++ b/libmamba/src/api/list.cpp
@@ -27,6 +27,7 @@ namespace mamba
             bool reverse = false;
             bool explicit_ = false;
             bool md5 = false;
+            bool sha256 = false;
             bool canonical = false;
             bool export_ = false;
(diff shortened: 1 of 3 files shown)


jjerphan (approved):
LGTM modulo a few suggestions and nitpicks.

Hind-M:
I don't remember if we addressed this in the other similar previous PRs, but what should we do in case of `--json` outputs?
Print nothing, provide a json formatted output, print the same classic output as if `--json` was ignored (`conda` case it seems)?
Whatever we choose, maybe we should add a small hint in the flag description regarding the expected behavior for `--json`.

AntoinePrv (approved):

SandrineP:
> But maybe worth discussing, I just feel like this should be explicitly stated to the user as he is expecting a json format with `--json`.

`--json` is not ignored in this case (but everything else is when the `--json` flag is used).

PR #3892, opened 2025-04-10
unify channels of installed and removed packages written in history

SandrineP:


SandrineP:
This can be reviewed/commented, but not merged as the code to parse the new format is not ready yet.

libmamba/src/core/transaction.cpp
```diff
@@ -426,6 +426,25 @@ namespace mamba
...
+            auto pkg_channel = mamba::specs::Channel::resolve(
+                                   unresolved_pkg_channel,
+                                   channel_context.params()
+            )
+                                   .value();
+            auto channel_url = pkg_channel[0].platform_url(pkg.platform).str();
+            pkg.channel = channel_url;
+        };
+        for_each_to_install(m_solution.actions, fix_channel);
+        for_each_to_remove(m_solution.actions, fix_channel);
```
Hind-M, line 446: I'm wondering if, for consistency purposes, we should do this for `for_each_to_omit` as well?
Can we add a test, maybe just an assert in already existing tests, to check the new pkg channel format?

Hind-M: I didn't have specific tests in mind, but I imagine that you can check in the tests installing (could be in `test_create.py`) and uninstalling (`test_remove.py`) packages the channel string `package["channel"]` got from the `list` output for example? And that the format should be different with and without your patch?

Not sure actually if `pkg.channel` is propagated to the `list` command but we can check the output of whatever `command` allowing access to it, otherwise if it's something not really exposed, maybe a `.cpp` test would be more appropriate?

JohanMabille (approved):

PR #3897, opened 2025-04-15
Make integration tests not rely on specific organisation of packages

Klaim:
This PR changes the mamba/micromamba integration tests to not (or less) rely on knowledge of how some packages (usually `xtensor`) are organized to check that operations are producing the expected output (package files are installed correctly, as files or links etc.)
The need came because of these integration tests failing suddenly after an `xtensor` package's release where file organisation changed, for example `xtensor.hpp` wasnt anymore in the root `include/xtensor/` directory but somewhere else, breaking the tests that were looking for this specific file in the install at a location only valid on previous versions. This kind of change in file organization of a package are not only legitimate but expected for long-term maintenance and usually advertised through semantic versioning and release notes.

To make these tests less reliant upon this kind of change, this pr:
- remove attempts to locate a specific file with a specific path in the install (usually `xtensor.hpp`) and instead provides ways to search that file in the install directory, either using the package's manifest info (validating the integrity of the install, if the file is where the manifest says it should be) or through file search (when it's installed without an available manifest, which occurs when testing the cache, see `test_pkg_cache.py`)
- provides tooling for doing all that and for checking the integrity of an installed package on-demand by comparing what the manifest says files should exist and their actual location
- applies these changes for the main breaking places;

More improvements in that direction could be made but I stopped where I could run the tests after deleting `helpers.xtensor_hpp`. Most of the following improvements would be renamings of references to `xtensor` for more generic names, as these tests should be able to wokr whatever the package we are testing.

### DIFF
--- a/micromamba/tests/helpers.py
+++ b/micromamba/tests/helpers.py
@@ -542,3 +542,48 @@ def create_with_chan_pkg(env_name, channels, package):
     cmd.append(package)
 
     return create(*cmd, default_channel=False, no_rc=False)
+
+
+def check_cpp_package_install(package_name: string, install_prefix_root_dir: Path):
+    # Checks, using `assert`, that the specified package is correctly installed
+    # in the specified prefix directory.
+    #
+    # The checks are based on the following conditions:
+    # - `$install_prefix_root_dir/conda-meta/$package_name-*.json` exists
+    # - that json file has the expected package manifest format
+    # - all the files in the manifest exists in `$install_prefix_root_dir/` (link or not)
+    #
+    # We do not open the installed files.
+    #
+    # Returns f"{package_name}-{version}-{build_string}" for the info found.
+
+    assert package_name
+    assert install_prefix_root_dir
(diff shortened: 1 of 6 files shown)


Klaim:
I updated the description for clarity.

Things that might be improvable, not sure, or to discuss:
- Is it ok if `helpers.PackageChecker` is using `assert` instead of raising some error? When failing, the call stack is part of the ouput so I suspect it's ok for these tests, but diverging opinions are welcome.
- Are there other checks that should be done to verify the integrity of an installed package?

Note also that I tried at some point to refactor the tests in `test_pkg_cache.py` and `test_linking.py`, but ended up removing the refactoring as although there is a lot of common code, the refactoring was making it harder to read tests in isolation. I still think that could be improved but is not necessary for this pr.

AntoinePrv (approved):
LGTM

jjerphan (approved):

JohanMabille (approved):

PR #3907, opened 2025-04-25
Matchspec hardening

AntoinePrv:
This PR adds the choice of the `MatchSpec` parser in the API of the `Database`.
- ⚠️ This **breaks** ABI
- ✅ This **does not break** API

It also introduce a temporary configuration`experimental_matchspec_parser`, only configurable only via environment variable.
The goal is not to make our parser/matcher the default now or in the future. The way we plug it in libsolv has too many limitations (pins & repoqueries not working, data duplication...). What we want it to be able to experiment how our matchspecs handle real data to be able to use them in combination with resolvo.

### DIFF
--- a/libmamba/src/api/configuration.cpp
+++ b/libmamba/src/api/configuration.cpp
@@ -1426,7 +1426,7 @@ namespace mamba
         insert(Configurable("experimental_matchspec_parsing", &m_context.experimental_matchspec_parsing)
                    .group("Basic")
                    .description(  //
-                       "Enable experimental parsing and matching of MatchSpecs using Mamba implementation.\n"
+                       "Enable internal parsing and matching of MatchSpecs using Mamba's experimental implementation rather than Libsolv's.\n"
                        "This is not mean for production"
                    )
                    .set_env_var_names());
--- a/libmamba/src/api/configuration.cpp
+++ b/libmamba/src/api/configuration.cpp
@@ -1423,6 +1423,14 @@ namespace mamba
                    .set_env_var_names()
                    .set_post_merge_hook(detail::not_supported_option_hook));
 
+        insert(Configurable("experimental_matchspec_parsing", &m_context.experimental_matchspec_parsing)
+                   .group("Basic")
+                   .description(  //
+                       "Enable experimental parsing and matching of MatchSpecs using Mamba implementation.\n"
+                       "This is not mean for production"
+                   )
+                   .set_env_var_names());
+
         insert(Configurable("debug", &m_context.debug)
                    .group("Basic")
                    .set_env_var_names()
--- a/libmamba/src/solver/libsolv/database.cpp
+++ b/libmamba/src/solver/libsolv/database.cpp
@@ -189,7 +189,9 @@ namespace mamba::solver::libsolv
             if (settings().matchspec_parser != MatchSpecParser::Libsolv)
             {
                 return make_unexpected(
-                    "Libsolv repodata parser uses only its own MatchSpec parser",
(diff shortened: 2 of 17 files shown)


libmamba/src/solver/libsolv/database.cpp
```diff
@@ -30,17 +29,24 @@ namespace mamba::solver::libsolv
...
     {
-        explicit DatabaseImpl(specs::ChannelResolveParams p_channel_params)
-            : matcher(std::move(p_channel_params))
+        explicit DatabaseImpl(specs::ChannelResolveParams p_channel_params, Settings settings_)
+            : settings(std::move(settings_))
+            , matcher(std::move(p_channel_params))
         {
         }
 
+        Settings settings;
```
jjerphan, line 38: I understand that the addition of `settings` is the change which causes the ABI break you have mentioned in the description of this PR. Can you confirm?

libmambapy/tests/test_solver_libsolv.py
```diff
@@ -193,31 +207,43 @@ def tmp_repodata_json(tmp_path):
...
+            channel_id=channel_id,
+            add_pip_as_python_dependency=add_pip_as_python_dependency,
+            package_types=package_types,
+            repodata_parser=repodata_parser,
+        )
+
+    if (repodata_parser == "Libsolv") and (matchspec_parser != "Libsolv"):
+        with pytest.raises(libmambapy.MambaNativeException) as e:
+            add_repo_json()
+            assert "Libsolv repodata parser uses only its own MatchSpec parser" in e
```
jjerphan, line 243: How about a verbose message to explain the cause of the error to the end-user?
```suggestion
            assert "Libsolv repodata parser must only be used with its own MatchSpec parser" in e
            assert f"Currently {repodata_parser} is used for the repodata parser" in e
            assert f"while {matchspec_parser} is used for the MatchSpec parser" in e
```

libmamba/src/api/configuration.cpp
```diff
@@ -1423,6 +1423,14 @@ namespace mamba
                    .set_env_var_names()
                    .set_post_merge_hook(detail::not_supported_option_hook));
 
+        insert(Configurable("experimental_matchspec_parsing", &m_context.experimental_matchspec_parsing)
+                   .group("Basic")
+                   .description(  //
+                       "Enable experimental parsing and matching of MatchSpecs using Mamba implementation.\n"
```
jjerphan, line 1429: Nitpick to disambiguate given than `MatchSpec` are already use externally (e.g. by `conda-libmamba-solver`).
```suggestion
                       "Enable internal parsing and matching of MatchSpecs using Mamba's experimental implementation rather than Libsolv's.\n"
```

jjerphan (approved):
LGTM.

I mostly have comment to disambiguate.

Nitpick: How about changing the title of the PR for something closer to the new entry for the configuration?

libmambapy/tests/test_solver_libsolv.py
AntoinePrv: Not sure about which message. It's not a constraint that it cannot work, it just that there is nothing that allow us to plug a different parser there.

jjerphan: I think my remark was not clear.

If I were an user, I do not think I would understand the cause of this error and what I should do to resolve it.

AntoinePrv: Solved in the error message, but keeping the Python tests light since they should only test bindings are working properly.

PR #3911, opened 2025-04-29
Create packages diff between the current state and a revision

SandrineP:
This will be used to install a revision.

Adds a parsing function to handle the different formats that can be found in the history file.
Adds a function to generate de diff between the current state and a given revision. This generates maps of packages installed and removed between the two states.

### DIFF
--- a/libmamba/include/mamba/core/history.hpp
+++ b/libmamba/include/mamba/core/history.hpp
@@ -61,23 +61,37 @@ namespace mamba
         ChannelContext& m_channel_context;
     };
 
-    namespace detail
+    /** PackageDiff contains two maps of packages and their package info, one being for the
+     * installed packages, the other for the removed ones. This is used while looping on
+     * revisions to get the diff between the target revision and the current one.
+     */
+    struct PackageDiff
     {
-        /** PackageDiff contains two maps of packages and their package info, one being for the
-         * installed packages, the other for the removed ones. This is used while looping on
-         * revisions to get the diff between the target revision and the current one.
-         */
-        struct PackageDiff
-        {
-            std::unordered_map<std::string, specs::PackageInfo> removed_pkg_diff;
-            std::unordered_map<std::string, specs::PackageInfo> installed_pkg_diff;
+        std::unordered_map<std::string, specs::PackageInfo> removed_pkg_diff;
+        std::unordered_map<std::string, specs::PackageInfo> installed_pkg_diff;
 
-            [[nodiscard]] static PackageDiff
-            from_revision(std::vector<History::UserRequest> user_requests, std::size_t target_revision);
-        };
+        [[nodiscard]] static PackageDiff
+        from_revision(const std::vector<History::UserRequest>& user_requests, std::size_t target_revision);
+    };
 
-        specs::PackageInfo pkg_info_builder(std::string s);
-    }
+    /** The following function parses the different formats that can be found in the history
+     * file.
+     *
+     * conda/mamba1 format:
+     *
+     * installed: +conda-forge/linux-64::xtl-0.8.0-h84d6215_0
+     * removed: -conda-forge/linux-64::xtl-0.8.0-h84d6215_0
+     *
+     * mamba2 broken format:
+     *
+     * installed: +conda-forge::xtl-0.8.0-h84d6215_0
+     * removed: -https://conda.anaconda.org/conda-forge/linux-64::xtl-0.8.0-h84d6215_0
+     *
+     * mamba2 new format:
+     * installed: +https://conda.anaconda.org/conda-forge/linux-64::xtl-0.8.0-h84d6215_0
+     * removed: -https://conda.anaconda.org/conda-forge/linux-64::xtl-0.8.0-h84d6215_0
+     */
+    specs::PackageInfo read_history_url_entry(const std::string& s);
 }  // namespace mamba
 
 #endif
--- a/libmamba/src/core/history.cpp
+++ b/libmamba/src/core/history.cpp
@@ -327,206 +327,186 @@ namespace mamba
         }
     }
 
-    namespace detail
+    specs::PackageInfo read_history_url_entry(const std::string& s)
     {
-        // The following function parses the different formats that can be found in the history
-        // file.
-        //
-        // conda/mamba1 format:
-        //
-        // installed: +conda-forge/linux-64::xtl-0.8.0-h84d6215_0
-        // removed: -conda-forge/linux-64::xtl-0.8.0-h84d6215_0
-        //
-        // mamba2 broken format:
-        //
-        // installed: +conda-forge::xtl-0.8.0-h84d6215_0
-        // removed: -https://conda.anaconda.org/conda-forge/linux-64::xtl-0.8.0-h84d6215_0
-        //
-        // mamba2 new format:
-        // installed: +https://conda.anaconda.org/conda-forge/linux-64::xtl-0.8.0-h84d6215_0
-        // removed: -https://conda.anaconda.org/conda-forge/linux-64::xtl-0.8.0-h84d6215_0
-        specs::PackageInfo pkg_info_builder(std::string s)
+        std::string name_version_build_string;
+        std::string channel;
+        std::size_t pos_0 = s.rfind("/");
+        if (pos_0 != std::string::npos)
         {
-            std::string name_version_build_string;
-            std::string channel;
-            std::size_t pos_0 = s.rfind("/");
-            if (pos_0 != std::string::npos)
+            // s is of the form of
+            // `https://conda.anaconda.org/conda-forge/linux-64::xtl-0.8.0-h84d6215_0` or
+            // `conda-forge/linux-64::xtl-0.8.0-h84d6215_0`
+            std::string s_begin = s.substr(0, pos_0);
+            // s_begin is of the form of `https://conda.anaconda.org/conda-forge` or
+            // `conda-forge`
+            std::string s_end = s.substr(pos_0 + 1, s.size());
+            // s_end is of the form of `linux-64::xtl-0.8.0-h84d6215_0`
+            std::size_t pos = s_begin.rfind("/");
+            if (pos != std::string::npos)
             {
-                // s is of the form of
-                // `https://conda.anaconda.org/conda-forge/linux-64::xtl-0.8.0-h84d6215_0` or
-                // `conda-forge/linux-64::xtl-0.8.0-h84d6215_0`
-                std::string s_begin = s.substr(0, pos_0);
-                // s_begin is of the form of `https://conda.anaconda.org/conda-forge` or
-                // `conda-forge`
-                std::string s_end = s.substr(pos_0 + 1, s.size());
-                // s_end is of the form of `linux-64::xtl-0.8.0-h84d6215_0`
-                std::size_t pos = s_begin.rfind("/");
-                if (pos != std::string::npos)
-                {
-                    channel = s_begin.substr(pos + 1, s_begin.size());
-                }
-                else
-                {
-                    channel = s_begin;
-                }
-                name_version_build_string = std::get<1>(util::rsplit_once(s_end, "::"));
+                channel = s_begin.substr(pos + 1, s_begin.size());
             }
             else
             {
-                // s is of the form of `conda-forge::xtl-0.8.0-h84d6215_0`
-                auto double_colon_split = util::split_once(s, "::");
-                channel = std::get<0>(double_colon_split);
-                name_version_build_string = std::get<1>(double_colon_split).value_or("");
+                channel = s_begin;
             }
-
-            // `name_version_build_string` is of the form `xtl-0.8.0-h84d6215_0`
-            std::size_t pos_1 = name_version_build_string.rfind("-");
-            std::string name_version = name_version_build_string.substr(0, pos_1);
-            // `name_version` is of the form `xtl-0.8.0`
-            std::string build_string = name_version_build_string.substr(
-                pos_1 + 1,
-                name_version_build_string.size()
-            );
-
-            std::size_t pos_2 = name_version.rfind("-");
-            std::string name = name_version.substr(0, pos_2);
-            std::string version = name_version.substr(pos_2 + 1, name_version.size());
-
-            specs::PackageInfo pkg_info{ name, version, build_string, channel };
-            return pkg_info;
+            name_version_build_string = std::get<1>(util::rsplit_once(s_end, "::"));
+        }
+        else
+        {
+            // s is of the form of `conda-forge::xtl-0.8.0-h84d6215_0`
+            auto double_colon_split = util::split_once(s, "::");
+            channel = std::get<0>(double_colon_split);
+            name_version_build_string = std::get<1>(double_colon_split).value_or("");
         }
 
-        PackageDiff PackageDiff::from_revision(
-            std::vector<History::UserRequest> user_requests,
-            std::size_t target_revision
-        )
+        // `name_version_build_string` is of the form `xtl-0.8.0-h84d6215_0`
+        std::size_t pos_1 = name_version_build_string.rfind("-");
+        std::string name_version = name_version_build_string.substr(0, pos_1);
+        // `name_version` is of the form `xtl-0.8.0`
+        std::string build_string = name_version_build_string.substr(
+            pos_1 + 1,
+            name_version_build_string.size()
+        );
+
+        std::size_t pos_2 = name_version.rfind("-");
+        std::string name = name_version.substr(0, pos_2);
+        std::string version = name_version.substr(pos_2 + 1, name_version.size());
+
+        specs::PackageInfo pkg_info{ name, version, build_string, channel };
+        return pkg_info;
+    }
+
+    PackageDiff PackageDiff::from_revision(
+        const std::vector<History::UserRequest>& user_requests,
+        std::size_t target_revision
+    )
+    {
+        assert(!user_requests.empty());
+
+        struct revision
         {
-            assert(!user_requests.empty());
+            std::size_t key = 0;
+            std::unordered_map<std::string, specs::PackageInfo> removed_pkg = {};
+            std::unordered_map<std::string, specs::PackageInfo> installed_pkg = {};
+        };
 
-            struct revision
+        std::list<revision> revisions;
+        for (auto r : user_requests)
+        {
+            if ((r.link_dists.size() > 0) || (r.unlink_dists.size() > 0))
             {
-                std::size_t key = 0;
-                std::unordered_map<std::string, specs::PackageInfo> removed_pkg = {};
-                std::unordered_map<std::string, specs::PackageInfo> installed_pkg = {};
-            };
-
-            std::list<revision> revisions;
-            for (auto r : user_requests)
-            {
-                if ((r.link_dists.size() > 0) || (r.unlink_dists.size() > 0))
+                if (r.revision_num > target_revision)
                 {
-                    if (r.revision_num > target_revision)
+                    revision rev{ /*.key = */ r.revision_num };
+                    for (auto ud : r.unlink_dists)
                     {
-                        revision rev{ /*.key = */ r.revision_num };
-                        for (auto ud : r.unlink_dists)
-                        {
-                            auto pkg_info = pkg_info_builder(ud);
-                            const auto name = pkg_info.name;
-                            rev.removed_pkg[name] = std::move(pkg_info);
-                        }
-                        for (auto ld : r.link_dists)
-                        {
-                            auto pkg_info = pkg_info_builder(ld);
-                            const auto name = pkg_info.name;
-                            rev.installed_pkg[name] = std::move(pkg_info);
-                        }
-                        revisions.push_back(rev);
+                        auto pkg_info = read_history_url_entry(ud);
+                        const auto name = pkg_info.name;
+                        rev.removed_pkg[name] = std::move(pkg_info);
                     }
+                    for (auto ld : r.link_dists)
+                    {
+                        auto pkg_info = read_history_url_entry(ld);
+                        const auto name = pkg_info.name;
+                        rev.installed_pkg[name] = std::move(pkg_info);
+                    }
+                    revisions.push_back(rev);
                 }
             }
+        }
 
-            PackageDiff pkg_diff{};
+        PackageDiff pkg_diff{};
 
-            const auto handle_install = [&pkg_diff](revision& rev, const std::string& pkg_name)
+        const auto handle_install = [&pkg_diff](revision& rev, const std::string& pkg_name)
+        {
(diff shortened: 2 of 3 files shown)


JohanMabille (approved):

libmamba/src/api/install.cpp
```diff
@@ -1010,5 +1010,182 @@ namespace mamba
...
+                size_t pos = s_begin.rfind("/");
+                if (pos != std::string::npos)
+                {
+                    channel = s_begin.substr(pos + 1, s_begin.size());
+                }
+                else
+                {
+                    channel = s_begin;
+                }
+                s_pkg = util::split(s_end, "::").back();
```
AntoinePrv, line 1039: `util::split`, splits everything and creates a temporary vector, this is more efficient:
```suggestion
                s_pkg = std::get<1>(util::rsplit_once(s_end, "::"));
```

libmamba/src/api/install.cpp
```diff
@@ -1010,5 +1010,182 @@ namespace mamba
...
+            }
+            else
+            {
+                channel = util::split(s, "::").front();
+                s_pkg = util::split(s, "::").back();
```
AntoinePrv, line 1044: Same with `util::split_once` and `util::rsplit_once`

libmamba/include/mamba/api/install.hpp
```diff
@@ -111,6 +112,16 @@ namespace mamba
...
+
+        struct PackageDiff
+        {
+            std::map<std::string, specs::PackageInfo> removed_pkg_diff;
+            std::map<std::string, specs::PackageInfo> installed_pkg_diff;
+        };
+
+        specs::PackageInfo pkg_info_builder(std::string s);
+        PackageDiff
+        get_revision_pkg_diff(std::vector<History::UserRequest> user_requests, int REVISION);
```
AntoinePrv, line 124: - `int REVISION` should be lowercase if it is a function parameter
- it could have a more explicit name: `target_revision`? `revision_count`?
- It should be explained what a negative revisions implies. If there should not be negative, then make it a positive (`std::size_t` for instance). I remember we talked about it being a bit hard to use `std::optional` in the CLI11 and that `-1` as a default was reasonable, but (assuming negative have no meaning in the rest of the code), it should be checked as early as possible, cast to a positive only integer, and only used as such.

APIs are very important to explain what the code does and cannot do. Public APIs in particular cannot be changed later on without making a breaking release. To help design API, I pretend to make negotiations between the designer of the API and a user of that API (both being me and within libmamba). On the one end, if I introduce an API, it is to solve a particular problem, so as an API user I want to be able to call that new API as easily as possible with minimum effort (integer that can be negative). On the other, to improve APIs, I pretend that I am the maintainer of a library with just this API and multiple users. I should provide something that makes sense for the problem in general and should not accept data that is meaningless or too complex for me (a negative integer if I don't know what to do with it). Eventually, as a user of the new API, I make the necessary adjustments when calling the API. Maybe there is a bit of repetitive work, so I make a few helpers functions etc.

libmamba/src/api/install.cpp
```diff
@@ -1010,5 +1010,182 @@ namespace mamba
...
+                if ((r.link_dists.size() > 0) || (r.unlink_dists.size() > 0))
+                {
+                    if (r.revision_num > REVISION)
+                    {
+                        revision rev{ /*.key = */ r.revision_num };
+                        for (auto ud : r.unlink_dists)
+                        {
+                            auto pkg_info = pkg_info_builder(ud);
+                            auto name = pkg_info.name;
+                            rev.removed_pkg[name] = pkg_info;
```
AntoinePrv, line 1081: This is a whole learning topic but `move` is beneficial here 
```suggestion
                            rev.removed_pkg[name] = std::move(pkg_info);
```

libmamba/src/api/install.cpp
```diff
@@ -1010,5 +1010,182 @@ namespace mamba
...
+                        }
+                        for (auto ld : r.link_dists)
+                        {
+                            auto pkg_info = pkg_info_builder(ld);
+                            auto name = pkg_info.name;
+                            rev.installed_pkg[name] = pkg_info;
```
AntoinePrv, line 1087:
```suggestion
                            rev.installed_pkg[name] = std::move(pkg_info);
```

AntoinePrv (commented):
Thank you for the PR. It is not a very easy part of Mamba to work with and one that can be improved. Some of my suggestions may look arbitrary because things are already be done in a similar way but I think we should aim to slowly improve our APIs. 

Could you document what the `PackageDiff` and its methods does. We have the `Solution` that is also a form of "package difference" that needs to be installed in the environment, but I don't believe this is doing the same.

I am also wondering if `include/api/install.hpp` is the best header for it. My understanding is that the (extremely imperfect) `include/api/` headers are a "mirror" of things you can do with the CLI (install, remove...) and the fact that this code is tested in `test_history` suggests it may be better to put it in `history.hpp`?

As a general rule (that is not fully enforced at the moment), we should have that `include/api` headers are not used anywhere except in `include/api` & `src/api` themselves.

libmamba/src/core/history.cpp
```diff
@@ -325,4 +325,183 @@ namespace mamba
...
+                if ((r.link_dists.size() > 0) || (r.unlink_dists.size() > 0))
+                {
+                    if (r.revision_num > target_revision)
+                    {
+                        revision rev{ /*.key = */ r.revision_num };
+                        for (auto ud : r.unlink_dists)
+                        {
+                            auto pkg_info = pkg_info_builder(ud);
+                            auto name = pkg_info.name;
+                            rev.removed_pkg[name] = std::move(pkg_info);
```
Klaim, line 400: Consider using `.insert` instead of `[key]` for inserting as this would overwrite existing key-values . Not sure if that's intended or not. Also `operator[]` will first create an empty element before overwriting after `operator=` so I think most of the time `.insert` (or `.emplace` in different cases) is better.

libmamba/src/core/history.cpp
```diff
@@ -325,4 +325,183 @@ namespace mamba
...
+
+            PackageDiff pkg_diff{};
+
+            auto handle_install = [&pkg_diff](revision& rev, const std::string& pkg_name)
+            {
+                bool res = false;
+                if (auto rev_iter = rev.installed_pkg.find(pkg_name);
+                    rev_iter != rev.installed_pkg.end())
+                {
+                    auto version = rev.installed_pkg[pkg_name].version;
```
Klaim, line 421: This is searching twice for the same package.
`rev_iter` is an iterator pointing to the found name/package pair, so you should use that once obtained, for example:
```c++
const auto version = rev_iter->second.version;
```

Similar to `iter` in the following block.

libmamba/src/core/history.cpp
```diff
@@ -325,4 +325,183 @@ namespace mamba
...
+                }
+                return res;
+            };
+
+            auto handle_remove = [&pkg_diff](revision& rev, const std::string& pkg_name)
+            {
+                bool res = false;
+                if (auto rev_iter = rev.removed_pkg.find(pkg_name); rev_iter != rev.removed_pkg.end())
+                {
+                    auto version = rev.removed_pkg[pkg_name].version;
```
Klaim, line 442:
```suggestion
                    const auto version = rev_iter->second.version;
```

libmamba/src/core/history.cpp
```diff
@@ -325,4 +325,183 @@ namespace mamba
...
+            {
+                auto& revision = *(revisions.begin());
+                while (!revision.removed_pkg.empty())
+                {
+                    auto [pkg_name, pkg_info] = *(revision.removed_pkg.begin());
+                    pkg_diff.removed_pkg_diff[pkg_name] = pkg_info;
+                    revision.removed_pkg.erase(pkg_name);
+                    bool lastly_removed = true;  // whether last operation on package was a removal
+                    lastly_removed = !handle_install(revision, pkg_name);
+                    for (auto rev = ++revisions.begin(); rev != revisions.end(); ++rev)
```
Klaim, line 468:
```suggestion
                    for (auto rev = std::next(revisions.begin()); rev != revisions.end(); ++rev)
```

libmamba/src/core/history.cpp
```diff
@@ -325,4 +325,183 @@ namespace mamba
...
+    namespace detail
+    {
+        // The following function parses the different formats that can be found in the history
+        // file. conda/mamba1 format: installed: +conda-forge/linux-64::xtl-0.8.0-h84d6215_0
+        // removed: -conda-forge/linux-64::xtl-0.8.0-h84d6215_0 mamba2 broken format: installed:
+        // +conda-forge::xtl-0.8.0-h84d6215_0 removed:
+        // -https://conda.anaconda.org/conda-forge/linux-64::xtl-0.8.0-h84d6215_0 mamba2 new format:
+        // installed: +https://conda.anaconda.org/conda-forge/linux-64::xtl-0.8.0-h84d6215_0
+        // removed: -https://conda.anaconda.org/conda-forge/linux-64::xtl-0.8.0-h84d6215_0
+        specs::PackageInfo pkg_info_builder(std::string s)
```
Klaim, line 338: I'm not completely familiar with this format, but would have regex matching worked as an alternative implementation?

libmamba/src/core/history.cpp
```diff
@@ -325,4 +325,183 @@ namespace mamba
...
+                        for (auto ud : r.unlink_dists)
+                        {
+                            auto pkg_info = pkg_info_builder(ud);
+                            auto name = pkg_info.name;
+                            rev.removed_pkg[name] = std::move(pkg_info);
+                        }
+                        for (auto ld : r.link_dists)
+                        {
+                            auto pkg_info = pkg_info_builder(ld);
+                            auto name = pkg_info.name;
```
Klaim, line 405:
```suggestion
                            const auto name = pkg_info.name;
```

libmamba/src/core/history.cpp
```diff
@@ -325,4 +325,183 @@ namespace mamba
...
+            {
+                if ((r.link_dists.size() > 0) || (r.unlink_dists.size() > 0))
+                {
+                    if (r.revision_num > target_revision)
+                    {
+                        revision rev{ /*.key = */ r.revision_num };
+                        for (auto ud : r.unlink_dists)
+                        {
+                            auto pkg_info = pkg_info_builder(ud);
+                            auto name = pkg_info.name;
```
Klaim, line 399:
```suggestion
                            const auto name = pkg_info.name;
```

libmamba/src/core/history.cpp
```diff
@@ -325,4 +325,183 @@ namespace mamba
...
+            std::string name = s_pkg_.substr(0, pos_2);
+            std::string version = s_pkg_.substr(pos_2 + 1, s_pkg_.size());
+
+            specs::PackageInfo pkg_info{ name, version, build_string, channel };
+            return pkg_info;
+        }
+
+        PackageDiff PackageDiff::get_revision_pkg_diff(
+            std::vector<History::UserRequest> user_requests,
+            int target_revision
```
Klaim, line 378: Is `target_revision` allowed to be 0 or negative? Is `user_requests` allowed to be empty?
If there are requirements on these, consider adding assertions (using `assert`) at the beginnig of the function checking that these requirements are fullfilled.

libmamba/src/core/history.cpp
```diff
@@ -325,4 +325,183 @@ namespace mamba
...
+                            rev.installed_pkg[name] = std::move(pkg_info);
+                        }
+                        revisions.push_back(rev);
+                    }
+                }
+            }
+
+            PackageDiff pkg_diff{};
+
+            auto handle_install = [&pkg_diff](revision& rev, const std::string& pkg_name)
```
Klaim, line 415: minor improvment: local lambdas that act as sub-functions can be `const` as their call function is already `const` (except if marked `mutable` but it's not the case here.

libmamba/src/core/history.cpp
```diff
@@ -325,4 +325,183 @@ namespace mamba
...
+                        {
+                            lastly_removed = handle_remove(*rev, pkg_name);
+                            if (lastly_removed)
+                            {
+                                lastly_removed = !handle_install(*rev, pkg_name);
+                            }
+                        }
+                    }
+                }
+                revisions.erase(revisions.begin());
```
Klaim, line 502: `.pop_front()` would be clearer. 

Note that with `vector` this is not very performant. `.pop_back()` is cheap however. Did you consider reversing the order of this vector to be able to `.pop_back()` progressively?

libmamba/src/core/history.cpp
SandrineP: I am not sure to understand your point...

libmamba/src/core/history.cpp
SandrineP: Reversing the order wouldn't be a good solution as I iterate on the vector.
I changed it into a list, so it makes things better ( and works with ```.pop_front()```)

libmamba/src/core/history.cpp
SandrineP: If you mean using regex matching instead of ```.rfind```, I am not sure it would be more efficient.

libmamba/src/core/history.cpp
```diff
@@ -325,4 +326,186 @@ namespace mamba
...
+                else
+                {
+                    channel = s_begin;
+                }
+                s_pkg = std::get<1>(util::rsplit_once(s_end, "::"));
+            }
+            else
+            {
+                channel = std::get<0>(util::split_once(s, "::"));
+                s_pkg = std::get<1>(util::rsplit_once(s, "::"));
```
jjerphan, line 362: Is it only possible to split once?

```suggestion
                auto double_colon_split = util::split_once(s, "::");

                channel = std::get<0>(double_colon_split);
                s_pkg = std::get<1>(double_colon_split);
```

Note that [a `namespace` field between the two `:` might be included in the future](https://docs.conda.io/projects/conda/en/stable/dev-guide/api/conda/models/match_spec/index.html).

libmamba/src/core/history.cpp
```diff
@@ -325,4 +326,186 @@ namespace mamba
...
+    {
+        // The following function parses the different formats that can be found in the history
+        // file. conda/mamba1 format: installed: +conda-forge/linux-64::xtl-0.8.0-h84d6215_0
+        // removed: -conda-forge/linux-64::xtl-0.8.0-h84d6215_0 mamba2 broken format: installed:
+        // +conda-forge::xtl-0.8.0-h84d6215_0 removed:
+        // -https://conda.anaconda.org/conda-forge/linux-64::xtl-0.8.0-h84d6215_0 mamba2 new format:
+        // installed: +https://conda.anaconda.org/conda-forge/linux-64::xtl-0.8.0-h84d6215_0
+        // removed: -https://conda.anaconda.org/conda-forge/linux-64::xtl-0.8.0-h84d6215_0
+        specs::PackageInfo pkg_info_builder(std::string s)
+        {
```
jjerphan, line 340: I think it would be useful to have comments close to the code to have readers check  substrings themselves while reducing their cognitive load.

See the suggestions bellow as an example.

libmamba/src/core/history.cpp
```diff
@@ -325,4 +326,186 @@ namespace mamba
...
+                }
+                s_pkg = std::get<1>(util::rsplit_once(s_end, "::"));
+            }
+            else
+            {
+                channel = std::get<0>(util::split_once(s, "::"));
+                s_pkg = std::get<1>(util::rsplit_once(s, "::"));
+            }
+
+            std::size_t pos_1 = s_pkg.rfind("-");
```
jjerphan, line 365:
```suggestion
            // At this stage, `s_pkg` is of the form `xtl-0.8.0-h84d6215_0`
            std::size_t pos_1 = s_pkg.rfind("-");
```

libmamba/src/core/history.cpp
```diff
@@ -325,4 +326,186 @@ namespace mamba
...
+        // The following function parses the different formats that can be found in the history
+        // file. conda/mamba1 format: installed: +conda-forge/linux-64::xtl-0.8.0-h84d6215_0
+        // removed: -conda-forge/linux-64::xtl-0.8.0-h84d6215_0 mamba2 broken format: installed:
+        // +conda-forge::xtl-0.8.0-h84d6215_0 removed:
+        // -https://conda.anaconda.org/conda-forge/linux-64::xtl-0.8.0-h84d6215_0 mamba2 new format:
+        // installed: +https://conda.anaconda.org/conda-forge/linux-64::xtl-0.8.0-h84d6215_0
+        // removed: -https://conda.anaconda.org/conda-forge/linux-64::xtl-0.8.0-h84d6215_0
+        specs::PackageInfo pkg_info_builder(std::string s)
+        {
+            std::string s_pkg;
```
jjerphan, line 341: Can `s_pkg` be renamed for something more explicit like `name_version_build_string`?

libmamba/src/core/history.cpp
```diff
@@ -325,4 +326,186 @@ namespace mamba
...
+                s_pkg = std::get<1>(util::rsplit_once(s_end, "::"));
+            }
+            else
+            {
+                channel = std::get<0>(util::split_once(s, "::"));
+                s_pkg = std::get<1>(util::rsplit_once(s, "::"));
+            }
+
+            std::size_t pos_1 = s_pkg.rfind("-");
+            std::string s_pkg_ = s_pkg.substr(0, pos_1);
```
jjerphan, line 366:
```suggestion
            // and `s_pkg_` is of the form `xtl-0.8.0`
            std::string s_pkg_ = s_pkg.substr(0, pos_1);
```

libmamba/src/core/history.cpp
```diff
@@ -325,4 +326,186 @@ namespace mamba
...
+        )
+        {
+            assert(target_revision >= 0);
+            assert(!user_requests.empty());
+
+            struct revision
+            {
+                int key = -1;
+                std::map<std::string, specs::PackageInfo> removed_pkg = {};
+                std::map<std::string, specs::PackageInfo> installed_pkg = {};
```
jjerphan, line 389: If ordering is not of interest.
```suggestion
                std::unordered_map<std::string, specs::PackageInfo> removed_pkg = {};
                std::unordered_map<std::string, specs::PackageInfo> installed_pkg = {};
```

jjerphan (commented):
Thank you, @SandrineP.

libmamba/include/mamba/core/history.hpp
```diff
@@ -61,6 +62,22 @@ namespace mamba
...
+        /** PackageDiff contains two maps of packages and their package info, one being for the
+         * installed packagege, the other for the removed ones. This is used while looping on
+         * revisions to get the diff between the target revision and the current one. */
+        struct PackageDiff
+        {
+            std::unordered_map<std::string, specs::PackageInfo> removed_pkg_diff;
+            std::unordered_map<std::string, specs::PackageInfo> installed_pkg_diff;
+
+            PackageDiff
+            get_revision_pkg_diff(std::vector<History::UserRequest> user_requests, int target_revision);
```
AntoinePrv, line 76:
```suggestion
            get_revision_pkg_diff(std::vector<History::UserRequest> user_requests, std::size_t target_revision);
```

AntoinePrv: Also is this supposed to be `static`, i.e. do you access both the current `PackageDiff` *and* the one you are creating?

libmamba/src/core/history.cpp
```diff
@@ -325,4 +326,186 @@ namespace mamba
...
+            };
+
+            std::list<revision> revisions;
```
AntoinePrv, line 392: Always prefer vector over list, unless you need to keep iterators through modifications, which rarely needed. On motivation: https://www.youtube.com/watch?v=0iWb_qi2-uI&t=2680s
```suggestion
            std::vector<revision> revisions;
```

libmamba/include/mamba/core/history.hpp
```diff
@@ -61,6 +62,22 @@ namespace mamba
         ChannelContext& m_channel_context;
     };
 
+    namespace detail
+    {
```
AntoinePrv, line 66: No need to put in detail IMHO, this is a well defined abstraction.

libmamba/include/mamba/core/history.hpp
```diff
@@ -61,6 +62,22 @@ namespace mamba
...
+        /** PackageDiff contains two maps of packages and their package info, one being for the
+         * installed packagege, the other for the removed ones. This is used while looping on
```
AntoinePrv, line 68:
```suggestion
         * installed packages, the other for the removed ones. This is used while looping on
```

libmamba/include/mamba/core/history.hpp
```diff
@@ -61,6 +62,22 @@ namespace mamba
...
+         * revisions to get the diff between the target revision and the current one. */
+        struct PackageDiff
+        {
+            std::unordered_map<std::string, specs::PackageInfo> removed_pkg_diff;
+            std::unordered_map<std::string, specs::PackageInfo> installed_pkg_diff;
```
AntoinePrv, line 73: Does this need to be a map or a vector would suffice?

AntoinePrv (commented):
Getting close, just a few more nitpicks

libmamba/src/core/history.cpp
jjerphan: Yes, I don't think there can be two `::`.

libmamba/src/core/history.cpp
SandrineP: It was a vector, but following a comment from Joël about deletion of the first item, I changed it to a list. It was not a goo idea to inverse the order of the vector as I was also iterating on it.

libmamba/include/mamba/core/history.hpp
SandrineP: It needs to be a map.

libmamba/src/core/history.cpp
SandrineP: I don´t think this name would be appropriate, as there could be more information in the string.

jjerphan (approved):
LGTM.

Thank you for this really valuable contribution, @SandrineP.

libmamba/src/core/history.cpp
Klaim: Ah yes I forgot `pop_front` doesnt exist in `vector` _because_ it is indeed inefficient. My suggestion with reversed vector implied to also use the reverse iterators when you loop.

libmamba/src/core/history.cpp
Klaim: Indeed, discarding (as in `erase`) on the fly implies using a manual loop and continue iterating (usnig a reverse iterator) with the return value of `erase` , in case you didnt see that idiom yet.

libmamba/include/mamba/core/history.hpp
Klaim: Ideally it would be a flat_map as we do both `find` and iterate through these but we dont have such containers at hand unfortunately.

PR #3913, opened 2025-04-30
Explicit API and ABI stability commitments

AntoinePrv:
Explicit our commitments to stability.

Ping other stakeholders for feedback:
- @jezdez 
- @jaimergp 
- @isuruf 
- ...

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,6 +107,33 @@ While `mamba` and `micromamba` are generally a drop-in replacement for `conda` t
 
 Please refer to the instructions given by the [official documentation](https://mamba.readthedocs.io/en/latest/developer_zone/dev_environment.html).
 
+## API and ABI stability
+
+The Mamba project uses semantic versioning of the form `MAJOR.MINOR.PATCH`.
+While we try to keep things stable for users, we also need to make breaking changes to improve
+Mamba and reduce technical debt.
+Future versions of Mamba may give stronger guarantees.
+
+_libmamba_ (C++):
+We are not aware of consumers of the C++ API, so we give ourselves room for improvements.
+
+- `PATCH` releases are API and ABI backward compatible;
+- `MINOR` releases are API compatible for elements in `mamba/api`, they can break API elsewhere
+  and ABI anywhere;
+- `MAJOR` releases make no guarantees.
+
+_libmambapy_ (Python):
+
+- `PATCH` releases are API backward compatible;
+- `MINOR` releases are API backward compatible;
+- `MAJOR` releases make no guarantees.
+
+_mamba_ and _micromamba_ (executables):
+
+- `PATCH` releases are API backward compatible;
+- `MINOR` releases are API backward compatible;
+- `MAJOR` releases make no guarantees.
+
 ## Support us
 
 Only `mamba` and `micromamba` 2.0 and later are supported and are actively developed.
--- a/README.md
+++ b/README.md
@@ -109,7 +109,7 @@ Please refer to the instructions given by the [official documentation](https://m
 
 ## API and ABI stability
 
-The Mamba project uses semantic versioning of the form `MAJOR.MINOR.PATCH`.
+The Mamba project uses [semantic versioning](https://semver.org/) of the form `MAJOR.MINOR.PATCH`.
 While we try to keep things stable for users, we also need to make breaking changes to improve
 Mamba and reduce technical debt.
 Future versions of Mamba may give stronger guarantees.
--- a/README.md
+++ b/README.md
@@ -114,24 +114,59 @@ While we try to keep things stable for users, we also need to make breaking chan
 Mamba and reduce technical debt.
 Future versions of Mamba may give stronger guarantees.
 
-_libmamba_ (C++):
+### `libmamba` (C++)
+
 We are not aware of consumers of the C++ API, so we give ourselves room for improvements.
+For `libmamba`, the term _backward compatible_ is understood as follows:
+
+- _ABI backward compatible_ means that you can replace the library binary files without recompiling
+  your code with the updated headers.
+  The observed behavior will be the same, except for bugs (disappearing, hopefully) and performance.
+- _API backward compatible_ means that you must recompile your code with the new library
+  version code, but you won't need to change your code, just re-build it.
+  This applies as long as you did not use any declaration understood to be private, for instance
+  in the `detail` sub-namespaces.
+  The observed behavior will be the same, except for bugs (disappearing, hopefully) and performance.
+  When declarations are deprecated but not removed and still functional, we consider it also
+  backward compatible, as only the observed behavior during compilation changes.
+
+With this in mind, `libmamba` offers the following guarantees:
 
 - `PATCH` releases are API and ABI backward compatible;
-- `MINOR` releases are API compatible for elements in `mamba/api`, they can break API elsewhere
-  and ABI anywhere;
+- `MINOR` releases are API compatible for declarations in `mamba/api`,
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -107,6 +107,33 @@ While `mamba` and `micromamba` are generally a drop-in replacement for `conda` t
 
 Please refer to the instructions given by the [official documentation](https://mamba.readthedocs.io/en/latest/developer_zone/dev_environment.html).
 
+## API and ABI stability
+
+The Mamba project uses semantic versioning of the form `MAJOR.MINOR.PATCH`.
```
jjerphan, line 112: We should mention that mamba has two versions:
 - one for the API following semantic versioning
 - one for the ABI following [libtool's convention](https://www.gnu.org/software/libtool/manual/html_node/Updating-version-info.html#Updating-version-info)
 
And potentially point to their definitions: 

https://github.com/mamba-org/mamba/blob/1ac3174e3864194e9e971b69eb83b3a7f068e565/libmamba/CMakeLists.txt#L16-L56

jjerphan (commented):
Thank you for mentioning this.

README.md
```diff
@@ -107,6 +107,33 @@ While `mamba` and `micromamba` are generally a drop-in replacement for `conda` t
...
+  and ABI anywhere;
+- `MAJOR` releases make no guarantees.
+
+_libmambapy_ (Python):
+
+- `PATCH` releases are API backward compatible;
+- `MINOR` releases are API backward compatible;
+- `MAJOR` releases make no guarantees.
+
+_mamba_ and _micromamba_ (executables):
```
jaimergp, line 131: Is this about their CLI and behaviour? 🤔

AntoinePrv: Yes, CLI args...

jaimergp: And output? At least the JSON bits?

README.md
Klaim, line 112:
```suggestion
The Mamba project uses [semantic versioning](https://semver.org/) of the form `MAJOR.MINOR.PATCH`.
```

README.md
```diff
@@ -107,6 +107,33 @@ While `mamba` and `micromamba` are generally a drop-in replacement for `conda` t
...
+
+The Mamba project uses semantic versioning of the form `MAJOR.MINOR.PATCH`.
+While we try to keep things stable for users, we also need to make breaking changes to improve
+Mamba and reduce technical debt.
+Future versions of Mamba may give stronger guarantees.
+
+_libmamba_ (C++):
+We are not aware of consumers of the C++ API, so we give ourselves room for improvements.
+
+- `PATCH` releases are API and ABI backward compatible;
```
Klaim, line 120: We should add a definition somewhere of backward compatible as it is not the same for every situation.
Something in that direction:
```
Below, the term backward compatible means different things for different cases:
 - *ABI backward compatible* means that you can replace the library binary files without recompiling your code with the updated headers.
 - *API backward compatible* means that you might need to recompile your code with the new library version code, but you won't need to change your code, just re-build. It is possible that the recompilation is not necessary, but we can't give that guarantee so you should recompile in doubt.
 - *CLI API backward compatible* means that all the commands you used with the previous version of the executable should work with the new version of the executable.
 
 ```
 
Notice how it is not the same for cli either and the texts below should clarify that.

Another clarification that might help, something like:
```
 When features, names, functions, types etc. are deprecated but not removed and still functional, we consider it also backward compatible, although the behavior of the output might change because of the deprecation warings.
 ```
That part might be discussed, we could consider deprectation marking not backward compatible too if we wanted.

README.md
Klaim: As pointed in my other comment, clarification of behavior is required here as backward compatible means different things in cli and other cases.

README.md
AntoinePrv: Good point, though I'm not super aware of what we do for ABI versioning. Do you feel like suggesting a paragraph?

README.md
AntoinePrv: Good points! I would say all inputs/outputs *reasonably* expected to be used by a program (shell included).
So **including**:
- Executable name,
- CLI args,
- Configs, env variables,
- CLI JSON output,
- Files created: lockfiles, environment prefixes, caches...

And **excluding**:
- CLI human-readable outputs and errors

AntoinePrv:
Regarding deprecation warnings, I think we could consider them to be backward compatible in all cases:

- In C++, they show up at compile time, so only user compiling with warnings as errors will be impacted (and probably want to be informed);
- In Python deprecation warnings do not show for end users, but only in development if you enable them (which Pytest does but many other tools may not even show them);
- In the executable, if we consider my previous [suggestion](https://github.com/mamba-org/mamba/pull/3913#discussion_r2071332883) that human readable output is not subject to API, it works out as well.

For all, they might be minor disturbances for the users, but probably better than having a big breakage at MAJOR upgrade.

Klaim:
> * In C++, they show up at compile time, so only user compiling with warnings as errors will be impacted (and probably want to be informed);

If they have a CI checking that no warning appeared, that can be an issue too but at that point it's still reasonable to assume it's compatible IMO.

> * In Python deprecation warnings do not show for end users, but only in development if you enable them (which Pytest does but many other tools may not even show them);

Do we have a mechanism where deprecation warnings are emitted at runtime when C++ libmamba deprecated functions are called from python?

> * In the executable, if we consider my previous [suggestion](https://github.com/mamba-org/mamba/pull/3913#discussion_r2071332883) that human readable output is not subject to API, it works out as well.

Note that we have seen issues at some point where tests broke because they were checking output. That being said, the output in question was JSON and deprecation messages should not appear when using that, so it was a bug on our side. Just mentionning this because if we consider users that might test the output of commands, then it becomes an issue, but we could say explicitly that the JSON output will not be affected.

README.md
AntoinePrv: Additional note: anything in a `detail` namespace is considered private and not subject to API compatibility.

AntoinePrv:
> Do we have a mechanism where deprecation warnings are emitted at runtime when C++ libmamba deprecated functions are called from python?

If a function is marked deprecated in C++, it only shows at compile time in `libmambapy`. Runtime warnings are added manually in the binding code. Since **we** are the users of `libmamba` inside `libmambapy`, whatever function we use is an implementation detail. We might break API without changing the C++ function called (if we change the parameter name for instance) or we might upgrade to a newer C++ function without breaking API (by doing a wrapper).

README.md
AntoinePrv: Other useful addition: function names generated automatically by Pybind11 (`_1`, `_2`...) are also not subject to API and can be upgraded without notice to a more useful name (e.g. `solver`...) that will in turn be subject to API versioning.

JohanMabille (approved):

corneliusroemer:
>> In Python deprecation warnings do not show for end users, but only in development if you enable them (which Pytest does but many other tools may not even show them);
> Do we have a mechanism where deprecation warnings are emitted at runtime when C++ libmamba deprecated functions are called from python?

There's `FutureWarning` which by default is _not_ filtered out. You could use that. See: https://docs.python.org/3/library/exceptions.html#FutureWarning

>exception DeprecationWarning
Base class for warnings about deprecated features when those warnings are intended for other Python developers.
>Ignored by the default warning filters, except in the __main__ module ([PEP 565](https://peps.python.org/pep-0565/)). Enabling the [Python Development Mode](https://docs.python.org/3/library/devmode.html#devmode) shows this warning.

> exception FutureWarning
Base class for warnings about deprecated features when those warnings are intended for end users of applications that are written in Python.

PR #3915, opened 2025-04-30
Remove implicit zero in Version formatting

### LINKED ISSUE #3904: MatchSpec roundtrip introduces leading zeros
### Troubleshooting docs

### Anaconda default channels

matchspec zero roundtrip

### Latest version of Mamba

### Tried in Conda?

### Describe your issue

The `MatchSpec` parser introduces a leading zero `0` in version components that start with a letter when roundtripping:

```python
>>> from libmambapy.specs import MatchSpec
>>> str(MatchSpec.parse("napari=0.6.0a2.dev14+g56ffd88f"))
'napari=0.6.0a2.0dev14+0g56ffd88f'
#               ^      ^
```

This results in solver errors (package not found) when the version has a local component (anything after the `+` symbol) that starts with a letter.

```shell
libmamba version : 2.1.0
          mamba version : 2.1.0
           curl version : libcurl/8.12.1 OpenSSL/3.5.0 (SecureTransport) zlib/1.3.1 zstd/1.5.6 libssh2/1.11.1 nghttp2/1.64.0
     libarchive version : libarchive 3.7.7 zlib/1.3.1 liblzma/5.6.3 bz2lib/1.0.8 liblz4/1.10.0 libzstd/1.5.6
       envs directories : /Users/jrodriguez/.local/conda/envs
          package cache : /Users/jrodriguez/.local/conda/pkgs
                          /Users/jrodriguez/.mamba/pkgs
            environment : conda-dev (active)
           env location : /Users/jrodriguez/.local/conda/envs/conda-dev
      user config files : /Users/jrodriguez/.mambarc
 populated config files : /Users/jrodriguez/.condarc
                          /Users/jrodriguez/.local/conda/.condarc
       virtual packages : __unix=0=0
                          __osx=15.3.1=0
                          __archspec=1=arm64
               channels : https://conda.anaconda.org/conda-forge/noarch
                          https://conda.anaconda.org/conda-forge/osx-arm64
       base environment : /Users/jrodriguez/.local/conda
               platform : osx-arm64
```

### Logs

```shell
N/A
```

### environment.yml

```yaml
N/A
```

### ~/.condarc

```yaml
always_yes: true
show_channel_urls: true
solver: libmamba
console: rich
```

AntoinePrv:
Before: ``Version::parse("1.0.2.dev4").str() == "1.0.2.0dev4"``
After: ``Version::parse("1.0.2.dev4").str() == "1.0.2.dev4"``

Note that ``Version::parse(s).str() == s`` is still not true for all s.
For instance, text is converted to lowercase, ``_`` separators are converted to ``.``.

Fix #3904

### DIFF
--- a/libmamba/include/mamba/specs/version.hpp
+++ b/libmamba/include/mamba/specs/version.hpp
@@ -89,6 +89,9 @@ namespace mamba::specs
 
         VersionPart();
         VersionPart(std::initializer_list<VersionPartAtom> init);
+        VersionPart(std::vector<VersionPartAtom> atoms, bool implicit_leading_zero);
+
+        [[nodiscard]] auto str() const -> std::string;
 
         auto operator==(const VersionPart& other) const -> bool;
         auto operator!=(const VersionPart& other) const -> bool;
--- a/libmamba/include/mamba/specs/version.hpp
+++ b/libmamba/include/mamba/specs/version.hpp
@@ -62,13 +62,37 @@ namespace mamba::specs
     /**
      * A sequence of VersionPartAtom meant to represent a part of a version (e.g. major, minor).
      *
+     * In a version like ``1.3.0post1dev``, the parts are ``1``, ``3``, and ``0post1dev``.
      * Version parts can have a arbitrary number of atoms, such as {0, "post"} {1, "dev"}
-     * in 0post1dev
+     * in ``0post1dev``.
      *
      * @see  Version::parse for how this is computed from strings.
      * @todo Use a small vector of expected size 1 if performance ar not good enough.
      */
-    using VersionPart = std::vector<VersionPartAtom>;
+    struct VersionPart
+    {
+        /** The atoms of the version part */
+        std::vector<VersionPartAtom> atoms = {};
+
+        /**
+         * Whether a potential leading zero in the first atom should be considered implicit.
+         *
+         * During parsing of ``Version``, if a part starts with a literal atom, it is considered
+         * the same as if it started with a leading ``0``.
+         * For instance ``0post1dev`` is parsed in the same way as ``post1dev``.
+         * Marking it as implicit enables the possibility to remove it when reconstructing a string
+         * representation.
+         * This is desirable for compatibility with other version formats, such as Python, where
+         * a version modifier might be expressed as ``1.3.0.dev3``.
+         */
+        bool implicit_leading_zero = false;
+
+        VersionPart() = default;
+        VersionPart(std::initializer_list<VersionPartAtom> init);
+
+        auto operator==(const VersionPart& other) const -> bool;
+        auto operator!=(const VersionPart& other) const -> bool;
+    };
 
     /**
      * A sequence of VersionPart meant to represent all parts of a version.
--- a/libmamba/include/mamba/specs/version.hpp
+++ b/libmamba/include/mamba/specs/version.hpp
@@ -43,19 +43,19 @@ namespace mamba::specs
 
         [[nodiscard]] auto str() const -> std::string;
 
-        auto operator==(const VersionPartAtom& other) const -> bool;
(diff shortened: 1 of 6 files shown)


libmamba/include/mamba/specs/version.hpp
```diff
@@ -62,13 +62,44 @@ namespace mamba::specs
...
+         * This is desirable for compatibility with other version formats, such as Python, where
+         * a version modifier might be expressed as ``1.3.0.dev3``.
+         */
+        bool implicit_leading_zero = false;
+
+        VersionPart();
+        VersionPart(std::initializer_list<VersionPartAtom> init);
+        VersionPart(std::vector<VersionPartAtom> atoms, bool implicit_leading_zero);
+
+        [[nodiscard]] auto str() const -> std::string;
```
Klaim, line 94: I would prefer `to_string()`. Minor preference though.

libmamba/include/mamba/specs/version.hpp
```diff
@@ -62,13 +62,44 @@ namespace mamba::specs
...
+
+        auto operator==(const VersionPart& other) const -> bool;
+        auto operator!=(const VersionPart& other) const -> bool;
+        auto operator<(const VersionPart& other) const -> bool;
+        auto operator<=(const VersionPart& other) const -> bool;
+        auto operator>(const VersionPart& other) const -> bool;
+        auto operator>=(const VersionPart& other) const -> bool;
```
Klaim, line 101: I would prefer comparison operations to be free functions.

libmamba/include/mamba/specs/version.hpp
```diff
@@ -62,13 +62,44 @@ namespace mamba::specs
...
+         */
+        bool implicit_leading_zero = false;
+
+        VersionPart();
+        VersionPart(std::initializer_list<VersionPartAtom> init);
+        VersionPart(std::vector<VersionPartAtom> atoms, bool implicit_leading_zero);
+
+        [[nodiscard]] auto str() const -> std::string;
+
+        auto operator==(const VersionPart& other) const -> bool;
```
Klaim, line 96: If it does make sense (not sure given the implementation), add a todo comment about replacing these operators by defaulted ones in C++20

libmamba/src/specs/version.cpp
```diff
@@ -198,6 +284,124 @@ fmt::formatter<mamba::specs::VersionPartAtom>::format(
...
+
+    VersionPart::VersionPart(std::vector<VersionPartAtom> p_atoms, bool p_implicit_leading_zero)
+        : atoms(std::move(p_atoms))
+        , implicit_leading_zero(p_implicit_leading_zero)
+    {
+    }
+
+    auto VersionPart::str() const -> std::string
+    {
+        return fmt::format("{}", *this);
```
Klaim, line 313: Is this really what's expected as default string value for a version? Why not `"{}.{}.{}"` with the appropriate values?

Klaim (approved):
LGTM with minor considerations.

libmamba/include/mamba/specs/version.hpp
AntoinePrv: Absolutely not default ;)

libmamba/src/specs/version.cpp
AntoinePrv: All the logic is implemented inside the fmt formatter, so this is really just delegating.

Klaim: Ah yes, makes sense.

libmamba/include/mamba/specs/version.hpp
AntoinePrv: I like that, it shows a bit more that a new string is computed/allocated.
However it should be done all across the `specs` namespace.
Done in a follow-up #3917 because it's a big diff.

PR #3916, opened 2025-04-30
DX: libmambapy import in build tree

AntoinePrv:
This is an devloper improvement for running libmambapy tests locally.
It repoduces a valid ``libmambapy`` folder inside the build tree.
This can latter be used with ``PYTHONPATH`` to quickly import ``libmambapy`` in any environment.

The workflow of making a full ``pip install`` of the library works well for CI, but in local
devlopment, it can be error-prone as it also install ``libmamba`` in the environment, which then
takes precedences when running ``libmamba`` and ``mamba`` integration tests.


jjerphan (approved):
I never have had too much problem in the past but if it improves your workflow, I am not against making lifes easier.

JohanMabille (approved):

Issue #3332, opened 2024-06-26
Use uv package manager for installing packages from PyPI


awvwgk:
Does mamba have the possibility to install packages from PyPI via [uv](https://github.com/astral-sh/uv) instead pip?

Lumcoin:
Maybe uv has any drawbacks? It does feel odd that "mamba env create" takes forever to install pip dependencies, while uv does it within seconds...

sidmadala:
Will the integration of `uv` allow for `environment.yaml` files like this?

```yaml
name: test-env
channels:
  - conda-forge
dependencies:
  - python=3.12
  - nvidia::cuda-toolkit=12.4
  - uv
  - uv:  # Use uv for all other packages
    - torch>=2.5.0
...
```

sidmadala:
Using `uv` with `micromamba` currently doesn't export or list out pypi dependencies properly from #4013. If there is any way to add that functionality then it will be on par and a full replacement for a `pip` managed `micromamba` environment.

PR #3918, opened 2025-05-02
support installing pip dependencies with uv

iisakkirotko:
Hey! 

This PR aims to add support for using uv instead of pip to install pip dependencies into an environment in a couple ways:
- As was suggested in https://github.com/mamba-org/mamba/issues/3332#issuecomment-2528019001, use uv to install any pip dependencies if uv is present in the environment specs already.
- If the `--use-uv` command line option is present (or the corresponding config setting is set to true) and there are pip dependencies to be installed, use uv to install the dependencies instead. Furthermore, if uv is not in the environment specs already, it will be added as is the case with pip currently.

Closes https://github.com/mamba-org/mamba/issues/3332

Edit: Since one of the new tests uses a file that should be fetched from github (https://raw.githubusercontent.com/mamba-org/mamba/refs/heads/main/libmamba/tests/data/env_file/env_4.yaml vs https://raw.githubusercontent.com/iisakkirotko/mamba/refs/heads/yaml-install-uv/libmamba/tests/data/env_file/env_4.yaml), what should the link be set to pre-merge? Currently it points to my fork, but at time of merge it should probably point to main here.

### DIFF
--- a/libmamba/include/mamba/api/install.hpp
+++ b/libmamba/include/mamba/api/install.hpp
@@ -105,8 +105,12 @@ namespace mamba
 
         bool eval_selector(const std::string& selector, const std::string& platform);
 
-        yaml_file_contents
-        read_yaml_file(const Context& ctx, const std::string& yaml_file, const std::string& platform);
+        yaml_file_contents read_yaml_file(
+            const Context& ctx,
+            const std::string& yaml_file,
+            const std::string& platform,
+            bool use_uv
+        );
 
         inline void to_json(nlohmann::json&, const other_pkg_mgr_spec&)
         {
--- a/libmamba/include/mamba/core/context.hpp
+++ b/libmamba/include/mamba/core/context.hpp
@@ -130,6 +130,7 @@ namespace mamba
         bool experimental_repodata_parsing = true;
         bool experimental_matchspec_parsing = false;
         bool debug = false;
+        bool use_uv = false;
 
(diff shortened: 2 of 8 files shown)


libmamba/include/mamba/core/context.hpp
```diff
@@ -130,6 +130,7 @@ namespace mamba
         bool experimental_repodata_parsing = true;
         bool experimental_matchspec_parsing = false;
         bool debug = false;
+        bool use_uv = false;
```
JohanMabille, line 133: Note for later: this should be moved in a dedicated structure for external package managers, but let's do that later when we plug a way to support any external package manager.

JohanMabille (approved):

libmamba/include/mamba/api/install.hpp
```diff
@@ -105,8 +105,12 @@ namespace mamba
 
         bool eval_selector(const std::string& selector, const std::string& platform);
 
-        yaml_file_contents
-        read_yaml_file(const Context& ctx, const std::string& yaml_file, const std::string& platform);
+        yaml_file_contents read_yaml_file(
+            const Context& ctx,
+            const std::string& yaml_file,
+            const std::string& platform,
+            bool use_uv
```
jjerphan, line 112: Thank you for this contribution.

I think this change is not necessary since `use_uv` can be accessed from the `Context`.

PR #3919, opened 2025-05-02
Allow creating environment with empty folder as target prefix

nsoranzo:
Fix https://github.com/mamba-org/mamba/issues/3845 .

Alternative to https://github.com/mamba-org/mamba/pull/3853 .

### DIFF
--- a/libmamba/src/api/create.cpp
+++ b/libmamba/src/api/create.cpp
@@ -44,13 +44,15 @@ namespace mamba
             {
                 if (ctx.prefix_params.target_prefix == ctx.prefix_params.root_prefix)
                 {
-                    LOG_ERROR << "Overwriting root prefix is not permitted";
-                    throw std::runtime_error("Aborting.");
+                    const auto message = "Overwriting root prefix is not permitted - aborting.";
+                    LOG_ERROR << message;
+                    throw mamba_error(message, mamba_error_code::incorrect_usage);
                 }
                 else if (!fs::is_directory(ctx.prefix_params.target_prefix))
                 {
-                    LOG_ERROR << "Target prefix already exists and is not a folder";
-                    throw std::runtime_error("Aborting.");
+                    const auto message = "Target prefix already exists and is not a folder - aborting.";
+                    LOG_ERROR << message;
+                    throw mamba_error(message, mamba_error_code::incorrect_usage);
                 }
                 else if (fs::is_empty(ctx.prefix_params.target_prefix))
                 {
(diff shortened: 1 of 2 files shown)


libmamba/src/api/create.cpp
```diff
@@ -47,6 +47,15 @@ namespace mamba
                     LOG_ERROR << "Overwriting root prefix is not permitted";
                     throw std::runtime_error("Aborting.");
                 }
+                else if (!fs::is_directory(ctx.prefix_params.target_prefix))
+                {
+                    LOG_ERROR << "Target prefix already exists and is not a folder";
+                    throw std::runtime_error("Aborting.");
```
Klaim, line 53: Prefer `mamba::mamba_error` for exceptions thrown from `libmamba`. Also the exception should contain the same message as the log, so something like `"Target prefix already exists and is not a folder - aborting"`). That way the exception handling can use the message and debugging is easier.

nsoranzo: @Klaim Thanks for the review! I copied the pattern from lines 48 and 71 in this file, would you like me to update those as well?

Also, is `mamba_error_code::incorrect_usage` the proper error code to pass to `mamba_error()` in these cases?

Klaim: Yes please, not all the code of libmamba has been updated to this way indeed, and if it doesnt break tests I welcome applying the boyscout rule. Thanks! 👍🏼

I think `mamba_error_code::incorrect_usage` is correct here yes.

nsoranzo: @Klaim I think I've implemented what you suggested.

libmamba/src/api/create.cpp
```diff
@@ -45,7 +45,22 @@ namespace mamba
                 if (ctx.prefix_params.target_prefix == ctx.prefix_params.root_prefix)
                 {
                     LOG_ERROR << "Overwriting root prefix is not permitted";
-                    throw std::runtime_error("Aborting.");
+                    throw mamba_error(
+                        "Overwriting root prefix is not permitted - aborting.",
```
Klaim, line 49: I would prefer if this message was in both cases (the aborting is useful in the log too), so sometihing like:
```c++
const auto message = "Overwriting root prefix is not permitted - aborting.";
LOG_ERROR << message;
throw mamba_error(message, mamba_error_code::incorrect_usage);
```

same for the other one

nsoranzo: Of course, addressed in a rebased commit.

jjerphan (approved):
LGTM. Thank you, @nsoranzo.

AntoinePrv:
Rebasing since this has a couple days.

Issue #3920, opened 2025-05-04
Development environment guide does not explain how to build micromamba


corneliusroemer:
I'm trying to contribute and am following the "development environment" guide: https://mamba.readthedocs.io/en/latest/developer_zone/dev_environment.html

I noticed the guide says nothing about how to build micromamba from source. None of the  Cmake presets build micromamba.

AntoinePrv:
Hi @corneliusroemer, thanks for the feedback.
`mamba` and `micromamba` are the same executable, only `micromamba` is a statically linked version of `mamba`, so it's only a matter of playing with `BUILD_STATIC`... options in CMake.

I do not recommend building statically for development. I'd say there is no added benefits.
If you want to make an install from source, check out [the install doc](https://mamba.readthedocs.io/en/latest/installation/micromamba-installation.html#build-from-source), or [the feedstock](https://github.com/conda-forge/micromamba-feedstock/blob/main/recipe/build.sh).

corneliusroemer:
Thanks for your quick response @AntoinePrv 

> mamba and micromamba are the same executable, only micromamba is a statically linked version of mamba

I thought micromamba implemented a subset of mamba, so it's not quite the same? Or has this changed?

<img width="882" alt="Image" src="https://github.com/user-attachments/assets/436cf3c3-383e-43b0-8523-498d7440c733" />
https://mamba.readthedocs.io/en/latest/user_guide/micromamba.html

> I do not recommend building statically for development. I'd say there is no added benefits.

In that case one can't run the `micromamba` integration tests? Or how do you do that with mamba instead of micromamba?

Talking about this:

<img width="956" alt="Image" src="https://github.com/user-attachments/assets/599610cb-8adb-43ef-9035-74484dddf116" />
https://mamba.readthedocs.io/en/latest/developer_zone/dev_environment.html

>  so it's only a matter of playing with BUILD_STATIC... options in CMake.

My suggestion was exactly to document this so that potential contributors have an easier time getting started. The "only" can mean minutes to hours for everyone from scratch unless it's documented/shared.

May I suggest you close this issue as "no planned" as opposed too "completed" (this is a somewhat recent Github issues feature, things turn up grey then to distinguish issues based on outcome).

<img width="576" alt="Image" src="https://github.com/user-attachments/assets/b0bbf53b-f6c5-4938-8d3e-d8a73772a879" />

corneliusroemer:
Specifically, how would I debug the failing tests here without building micromamba?

[...]

PR #3921, opened 2025-05-04
feat: allow listing only direct dependencies via --direct-deps-only

### LINKED ISSUE #3344: Allow outputting of explicitly (vs implicitly) installed packages also in non-TTY output of `micromamba list`
I love that `micromamba list` outputs explicitly installed packages by making the packages bold, see screenshot:

image

I was hoping that the same information about explicit vs implicit install was also available when `micromamba list` was not output to tty, e.g. when piping into `less` or `vim -`

Of course the color gets lost then - so maybe you could optionally add the explicit vs implicit information in an extra column that can be enabled with a flag?

corneliusroemer:
resolves #3344

I can't test micromamba as I can't get it to build (see https://github.com/mamba-org/mamba/issues/3920)

It seems to work for mamba

### DIFF
--- a/libmamba/src/api/list.cpp
+++ b/libmamba/src/api/list.cpp
@@ -187,7 +187,8 @@ namespace mamba
                         auto obj = nlohmann::json();
                         const auto& pkg_info = all_records.find(key)->second;
 
-                        if (options.direct_deps_only && requested_specs.find(pkg_info.name) == requested_specs.end())
+                        if (options.direct_deps_only
+                            && requested_specs.find(pkg_info.name) == requested_specs.end())
                         {
                             continue;
                         }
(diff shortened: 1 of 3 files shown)


libmamba/src/api/list.cpp
```diff
@@ -152,6 +153,7 @@ namespace mamba
             { return (regex.empty() || std::regex_search(pkg_info.name, spec_pat)); };
 
             auto all_records = prefix_data.all_pkg_mgr_records();
+            auto requested_specs = prefix_data.history().get_requested_specs_map();
```
Klaim, line 156:
```suggestion
            const auto requested_specs = prefix_data.history().get_requested_specs_map();
```

I cant find a modification for this so I suspect `const` works here.

PR #3933, opened 2025-05-12
feat: Use ZDOTDIR if set and fallback to HOME

ickc:


AntoinePrv (approved):

libmamba/src/core/shell_init.cpp
```diff
@@ -1275,7 +1275,15 @@ namespace mamba
         }
         else if (shell == "zsh")
         {
-            fs::u8path zshrc_path = home / ".zshrc";
+            fs::u8path zshrc_path
+            // use ZDOTDIR if set and fallback to HOME
+            const char* zdotdir = std::getenv("ZDOTDIR");
```
AntoinePrv: We have `util::get_env` to properly read from the environment.

libmamba/src/core/shell_init.cpp
```diff
@@ -1349,7 +1357,15 @@ namespace mamba
...
+            if (zdotdir != nullptr && zdotdir[0] != '\0')
+            {
+                zshrc_path = fs::u8path(zdotdir) / ".zshrc";
+            } else {
+                zshrc_path = home / ".zshrc";
+            }
```
AntoinePrv: Looks like the repeated logic could be factored into a simple function.

AntoinePrv (changes requested):

ickc:
@AntoinePrv, do you have any pointers to fix the windows build and also the integration tests on UNIX platforms?

PR #3936, opened 2025-05-13
Allow users to set labels on PRs

### LINKED ISSUE #3675: Check release label / label_check workflow is annoying
### Troubleshooting docs

### Anaconda default channels

Other (please describe)

label, label_check

### Latest version of Mamba

### Tried in Conda?

Not applicable

### Describe your issue

Every time someone I create a PR, I am getting the CI failure, because of the [label_check workflow](https://github.com/mamba-org/mamba/blob/main/.github/workflows/label_check.yml).
And I also get an email and notification from GitHub that my build failed.

I think devs with some permissions don't have this issue, but all new contributors do, and that's not pleasant.
Maybe there is a better way?

_No response_

### Logs

_No response_

### environment.yml

_No response_

### ~/.condarc

_No response_

Hind-M:
# Description

Fix #3675 

## Type of Change

- [ ] Bugfix
- [x] Feature / enhancement
- [ ] CI / Documentation
- [ ] Maintenance

## Checklist

### DIFF
--- a/.github/workflows/set_pr_label.yml
+++ b/.github/workflows/set_pr_label.yml
@@ -1,0 +1,71 @@
+name: Set release label to PR based on template checks
+
+on:
+  pull_request:
+    types:
+      - opened
+      - edited
+
+jobs:
+  set_label:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Read PR description
+        id: pr_description
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const pr_body = context.payload.pull_request.body;
+            return pr_body;
+
+      - name: Parse labels from PR description with release labels mapping
+        id: parse_labels
+        run: |
+          # Define mapping from checkbox label to release label
+          declare -A label_map=(
+            ["Bugfix"]="release::bug_fixes"
+            ["Feature / enhancement"]="release::enhancements"
+            ["CI / Documentation"]="release::ci_docs"
+            ["Maintenance"]="release::maintenance"
+          )
+          # List of labels to check
+          LABELS=("Bugfix" "Feature / enhancement" "CI / Documentation" "Maintenance")
+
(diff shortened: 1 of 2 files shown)


.github/workflows/set_pr_label.yml
```diff
@@ -0,0 +1,88 @@
+name: Set release label to PR based on title and description
+
+on:
+  pull_request_target:
```
Hind-M, line 4: It seems that `GITHUB_TOKEN` has limited permissions when used on PRs (using forks), leading to a failure when trying to set labels on the PR using GHA.

One way to handle this is to use `pull_request_target` instead of `pull_request` to allow full permissions of the token.
The downside of this is that the event would run against the workflow and code from the base of the pull request (main branch) and the new code in this PR (not merged yet) would not be considered. So that would require to merge this "blindly" (so far the label to set in the script from the PR title and/or description is correctly parsed), before really checking afterwards that the workflow correctly sets the label.

Another solution would be to generate a new customized token with all the necessary permissions (considering that we "can" set permissions on PRs from forks).

Some references, for those who are interested in more details:
https://github.blog/news-insights/product-news/github-actions-improvements-for-fork-and-pull-request-workflows/
https://github.com/actions/labeler?tab=readme-ov-file#recommended-permissions

.github/workflows/set_pr_label.yml
```diff
@@ -0,0 +1,88 @@
...
+
+            // If a label was matched, add it to PR
+            if (assigned_label) {
+              await github.rest.issues.addLabels({
+                issue_number: context.issue.number,
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                labels: [assigned_label]
+              });
+            }
```
Hind-M, line 54: This fails with 403 error.

.github/workflows/set_pr_label.yml
```diff
@@ -0,0 +1,88 @@
...
```
Hind-M, line 88: This fails with 403 error.

PR #3940, opened 2025-05-13
Refactor `SubdirData` > `SubdirIndexLoader`

AntoinePrv:
This finishes the current refactoring of `SubdirData`.
- First a rename `SubdirData` > `SubdirIndexLoader`,
- Necessary work on the Python bindings to keep the current ones backward compatible,
- Adding new bindings for the new interface,
- Adding tests,
- A few more fixes here and there, and moving `ChannelContext` and `Console::stream()` (which fails when not used with external initialization).

### DIFF
--- a/libmamba/include/mamba/core/context.hpp
+++ b/libmamba/include/mamba/core/context.hpp
@@ -226,8 +226,15 @@ namespace mamba
             return {
                 /* .local_repodata_ttl */ get_local_repodata_ttl(),
                 /* .offline */ this->offline,
-                /* .repodata_check_zst */ this->repodata_use_zst,
                 /* .force_use_zst */ false  // Must override based on ChannelContext
+            };
+        }
+
+        SubdirDownloadParams subdir_download_params() const
+        {
+            return {
+                /* .offline */ this->offline,
+                /* .repodata_check_zst */ this->repodata_use_zst,
             };
         }
 
--- a/libmamba/include/mamba/core/context.hpp
+++ b/libmamba/include/mamba/core/context.hpp
@@ -227,6 +227,7 @@ namespace mamba
                 /* .local_repodata_ttl */ get_local_repodata_ttl(),
                 /* .offline */ this->offline,
                 /* .repodata_check_zst */ this->repodata_use_zst,
+                /* .force_use_zst */ false  // Must override based on ChannelContext
             };
         }
 
--- a/libmamba/include/mamba/core/context.hpp
+++ b/libmamba/include/mamba/core/context.hpp
@@ -209,7 +209,10 @@ namespace mamba
         SubdirParams subdir_params() const
         {
             return {
-                /* .local_repodata_ttl */ this->local_repodata_ttl,
+                // This is legacy where from where 1 meant to read from header
+                /* .local_repodata_ttl */ (this->local_repodata_ttl == 1)
+                    ? std::nullopt
+                    : std::optional(this->local_repodata_ttl),
                 /* .offline */ this->offline,
                 /* .use_index_cache */ this->use_index_cache,
                 /* .repodata_use_zst */ this->repodata_use_zst,
--- a/libmamba/include/mamba/core/context.hpp
+++ b/libmamba/include/mamba/core/context.hpp
@@ -208,13 +208,24 @@ namespace mamba
 
         SubdirParams subdir_params() const
         {
-            return {
+            const auto get_local_repodata_ttl = [&]() -> std::optional<std::size_t>
+            {
+                // Force the use of index cache by setting TTL to 0
+                if (this->use_index_cache)
+                {
+                    return { 0 };
+                }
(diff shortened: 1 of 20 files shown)


libmambapy/src/libmambapy/bindings/legacy.cpp
```diff
@@ -526,25 +554,123 @@ bind_submodule_impl(pybind11::module_ m)
         .def_static("whoneeds", &Query::whoneeds)
         .def_static("depends", &Query::depends);
 
-    py::class_<SubdirData>(m, "SubdirData")
```
jjerphan, line 529: How should we handle the removal of this API?

jjerphan (commented):
Still digesting this part of the code-base on my side.

libmambapy/src/libmambapy/bindings/legacy.cpp
AntoinePrv: It's handle via `SubdirDataMigrator` which keep it as it is in Python, but uses a wrapped class to avoid binding the same class twice in Pybind.

libmambapy/src/libmambapy/bindings/legacy.cpp
```diff
@@ -774,16 +917,10 @@ bind_submodule_impl(pybind11::module_ m)
...
-        .def_readwrite("retry_timeout", &download::RemoteFetchParams::retry_timeout)
-        .def_readwrite("retry_backoff", &download::RemoteFetchParams::retry_backoff)
-        .def_readwrite("user_agent", &download::RemoteFetchParams::user_agent)
-        // .def_readwrite("read_timeout_secs", &Context::RemoteFetchParams::read_timeout_secs)
-        .def_readwrite("proxy_servers", &download::RemoteFetchParams::proxy_servers)
-        .def_readwrite("connect_timeout_secs", &download::RemoteFetchParams::connect_timeout_secs);
+    ctx.def_property_readonly_static(
+        "RemoteFetchParams",
+        [](py::handle) { return py::type::of<download::RemoteFetchParams>(); }
+    );
```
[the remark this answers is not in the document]
AntoinePrv: > Is this maintaining the current import path?

Yes but note that this was not an `import` it was in `libmambapy.Context.RemoteFetchParams`, now it's a property that points to `libmambapy.RemoteFetchParams`.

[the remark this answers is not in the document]
AntoinePrv: > If so, should we set a deprecation period in place also?

No strong opinion, but I went without since there is still a `remote_fetch_params` in the Context.

libmamba/include/mamba/core/subdir_parameters.hpp
```diff
@@ -7,16 +7,28 @@
...
     struct SubdirParams
     {
-        std::size_t local_repodata_ttl = 1;
+        /**
+         * Repodata cache time to live in seconds.
+         *
+         * If not specified, then it is read from server headers.
+         */
+        std::optional<std::size_t> local_repodata_ttl_s = std::nullopt;
         bool offline = false;
```
Hind-M, line 22: Is `offline` in `SubdirParams` supposed to be the same info/value as `offline` in `SubdirDownloadParams`?
If so, can we avoid having it in multiple places?

And shouldn't we remove the redundant parameters completely from `Context` if we are defining specific subdir parameters in this file?

libmamba/include/mamba/core/context.hpp
```diff
@@ -207,12 +207,34 @@ namespace mamba
...
+        }
+
+        SubdirDownloadParams subdir_download_params() const
         {
             return {
-                /* .local_repodata_ttl */ this->local_repodata_ttl,
                 /* .offline */ this->offline,
-                /* .use_index_cache */ this->use_index_cache,
-                /* .repodata_use_zst */ this->repodata_use_zst,
+                /* .repodata_check_zst */ this->repodata_use_zst,
```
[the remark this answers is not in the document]
AntoinePrv: No, this is the Context variable which was not renamed for simplicity.

libmamba/include/mamba/core/subdir_parameters.hpp
AntoinePrv: > Is offline in SubdirParams supposed to be the same info/value as offline in SubdirDownloadParams ?

Sort of, perhaps one of them could have a different name though. But it was always use with the same value, yes.

>  If so, can we avoid having it in multiple places?

I took the direction of putting it in the two places, rather than passing the same struct to different functions where some arguments would have no effect.

> And shouldn't we remove the redundant parameters completely from Context if we are defining specific subdir parameters in this file?

I think it is not that simple, as some parameters in the Context map directly to configuration options tat we do not want to break (with also custom meaning for some values that are being handled differently). To avoid duplication, there is still only one source of truth in the Context, and these parameters are computed from the others on demand (it's just a few `bool`s and an integer).

Issue #3925, opened 2025-05-05
spdlog should be a private dependency of libmamba


JohanMabille:
Since libmamba is building with the header-only component of spdlog, and since it does not expose any spdlog object in its API, the dependency to spdlog should be private and not propagate to downstream projects trough `libmambaConfig.cmake` file.

jjerphan:
Closed by https://github.com/mamba-org/mamba/pull/4082.

PR #3945, opened 2025-05-19
maint: Make spdlog a private dependency

jjerphan:
Fix https://github.com/mamba-org/mamba/issues/3925.

:warning: The run-export on `spdlog` will need to be removed then.

### DIFF
--- a/libmamba/CMakeLists.txt
+++ b/libmamba/CMakeLists.txt
@@ -472,8 +472,9 @@ macro(libmamba_create_target target_name linkage output_name)
 
         target_link_libraries(
             ${target_name}
-            PUBLIC fmt::fmt-header-only spdlog::spdlog_header_only yaml-cpp::yaml-cpp
+            PUBLIC fmt::fmt-header-only yaml-cpp::yaml-cpp
             PRIVATE
+                spdlog::spdlog_header_only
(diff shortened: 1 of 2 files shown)


Hind-M:
Not sure if this is enough or if we are missing other things, but we should at least remove `spdlog` includes from exposed headers (e.g https://github.com/mamba-org/mamba/blob/main/libmamba/include/mamba/core/util_scope.hpp#L9 )?

JohanMabille (changes requested):
spdlog cannot be a private dependency as long as `spdlog.h` is included (even indirectly) in a publicly exposed header of libmamba.

I think it should not be problematic to make `util_scope.hpp` (the only eader including it directly) a private header of libmamba,; the issue is that the libmamba tests also use it; adding the src folder in the include directories would solve the issue for local build, but it woul not work for testing already installed libmamba (and I think we probably need this use case for some package managers). Notice that this code could be duplicated for the tests though.

JohanMabille:
Actually let's go with private headers, and deactivate the tests of private headers in the context of an already installed library.

jjerphan:
Closing in preference of https://github.com/mamba-org/mamba/pull/4082.

PR #3947, opened 2025-05-19
Introduce resolvo

jjerphan:


JohanMabille (commented):
First review, I'll add more comment later.

.github/workflows/unix_impl.yml
```diff
@@ -40,7 +40,7 @@ jobs:
             --preset mamba-unix-shared-${{ inputs.build_type }}  \
             -D CMAKE_CXX_COMPILER_LAUNCHER=sccache \
             -D CMAKE_C_COMPILER_LAUNCHER=sccache \
-            -D MAMBA_WARNING_AS_ERROR=ON \
+            -D MAMBA_WARNING_AS_ERROR=OFF \
```
AntoinePrv, line 43: We should remove this before merging.

libmamba/include/mamba/solver/resolvo/database.hpp
```diff
@@ -0,0 +1,236 @@
...
+        ::resolvo::VersionSetId alloc_version_set(std::string_view raw_match_spec);
+        ::resolvo::SolvableId alloc_solvable(specs::PackageInfo package_info);
+        std::pair<specs::Version, size_t>
+        find_highest_version(::resolvo::VersionSetId version_set_id);
+
+        // Pools for mapping between resolvo IDs and mamba types
+        bijective_map<::resolvo::NameId, ::resolvo::String> name_pool;
+        bijective_map<::resolvo::StringId, ::resolvo::String> string_pool;
+        bijective_map<::resolvo::VersionSetId, specs::MatchSpec> version_set_pool;
+        bijective_map<::resolvo::SolvableId, specs::PackageInfo> solvable_pool;
```
AntoinePrv, line 215: Are we sure we want to publicly exposed members?

libmamba/include/mamba/solver/resolvo/database.hpp
```diff
@@ -0,0 +1,236 @@
...
+
+        bool has_package(const specs::MatchSpec& spec)
+        {
+            auto candidates = get_candidates(
+                name_pool.alloc(::resolvo::String(spec.name().to_string()))
+            );
+            return !candidates.candidates.empty();
+        }
```
AntoinePrv, line 223: Nit: separate declaration from definition.

libmamba/include/mamba/solver/solver_factory.hpp
```diff
@@ -0,0 +1,45 @@
...
+#include <memory>
+#include <variant>
+
+#include "mamba/core/context.hpp"
+#include "mamba/solver/libsolv/database.hpp"
+#include "mamba/solver/libsolv/solver.hpp"
+#include "mamba/solver/resolvo/database.hpp"
+#include "mamba/solver/resolvo/solver.hpp"
+
+namespace mamba::solver
```
AntoinePrv, line 19: It would be nice to keep `mamba::solver` context-free. There could be an enum to choose the solver, and make a method in the context to compute it from `experimental_resolvo_solver`.

libmamba/include/mamba/specs/package_info.hpp
```diff
@@ -66,6 +69,13 @@ namespace mamba::specs
 
         [[nodiscard]] static auto from_url(std::string_view url) -> expected_parse_t<PackageInfo>;
 
+        [[nodiscard]] static auto from_json(
+            const std::string_view& filename,
+            simdjson::ondemand::object& pkg,
+            const CondaURL& repo_url,
+            const std::string& channel_id
+        ) -> expected_parse_t<PackageInfo>;
+
```
AntoinePrv, line 78: This ought to be a private helpers somewhere. `simdjson` is a private dependency and I think it should stay that way.

Perhaps a concept or view-like wrapper could help. On idea I had was to make a `PackageInfo` template that take a data storage class object. Then there could be `simdjson`, `libsolv`, `resolvo`, and inline `std::string`... implementations.
Let me know if you want to brainstorm it together, I'd be happy to!

AntoinePrv (commented):
This is going in the right direction! I only put a few comments on the code for now, because I think we should split the development in successive steps that would be easy to review and merge. For instance, I would

- Focus on an `mamba::solver::resolvo` implementation disconnected from the rest of mamba that passes tests similar to `libsolv/test_solver.cpp`. We can leave leave out anything `json` related to start with, or else plug a dirty implementation that uses libsolv under the hood. If some tests are generic enough, we can introduce a provisional generic Solver API to run thoses tests with all possible solvers and configuration.
- Refactor Json parsing and `PackageInfo` to avoid code-duplication, copies, and exposing private dependencies. I think something with Concept + Composition could help us drastically reduce all the code duplication we already have. Here we can plug some integration tests with data from conda-forge.
- Then we can derive a generic API, factory, and plug it into the rest of mamba.

jjerphan:
Note that this is completely work in progress, I am trying to have something which works first (while being pulled in many directions).

> Refactor Json parsing and PackageInfo to avoid code-duplication, copies, and exposing private dependencies. I think something with Concept + Composition could help us drastically reduce all the code duplication we already have. Here we can plug some integration tests with data from conda-forge.

Yes, this would be useful to also resolve https://github.com/mamba-org/mamba/issues/3954. I think this should be tackled first before all.

wolfv:
I keep wondering if it wouldn't be easier to just add C++ bindings to rattler at this point ...

jjerphan:
Closing in preference of https://github.com/mamba-org/mamba/pull/4301.

PR #3949, opened 2025-05-20
Context dependency reduction

JohanMabille:
Low level refactoring to reduce dependency on Context. The utilmate goal is to remove the Context from the Transaction class so that we can refactor its API and expose it more easily.

### DIFF
--- a/libmamba/include/mamba/core/context_params.hpp
+++ b/libmamba/include/mamba/core/context_params.hpp
@@ -8,6 +8,8 @@
 
 #include <string>
 
+#include "mamba/fs/filesystem.hpp"
+
 // TODO: having a file for this single structure is a bit of
 // overkill; this should be refactored when we have more structures
 // like this (i.e. parameters structures with no dependency on other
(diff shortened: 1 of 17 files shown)


Hind-M (approved):

libmamba/include/mamba/core/context_params.hpp
```diff
@@ -0,0 +1,35 @@
...
+//
+// The full license is in the file LICENSE, distributed with this software.
+
+#pragma once
+
+#include <string>
+
+#include "mamba/fs/filesystem.hpp"
+
+// TODO: having a file for this single structure is a bit of
```
Klaim, line 13: Side comment: the ideal setup would be each "service" or "module" (in the organisation meaning, not C++-specific) should expose in it's header the structs which serve as input to the whole "service". So structs would be located in their headers.
But here these two seems to be common for multiple parts of the code so OK.
However if there is a more general meaning for these I would prefer the name of the header to match that meaning as it's not really a context thing? Maybe `common_params.hpp`?

Klaim (approved):
LGTM

libmamba/include/mamba/core/context_params.hpp
JohanMabille: Yeah I'm not sure yet of how things will be organized in the end. Extracting the context from the transaction is a rabbit hole, and I'm pretty sure things will change again. For now, I could have put structures in dedicated headers aside their "service" but that means a lot of new headers I'm not sure to keep in the end. And defining everything in the Context header (because all of these are more or less gathering of configurables from the context) before the whole refactoring is done would lead to a lot of circular includes.

PR #3950, opened 2025-05-21
Transaction context

JohanMabille:


jjerphan (approved):
🙏

libmamba/src/core/link.cpp
```diff
@@ -164,9 +165,10 @@ namespace mamba
         std::ofstream out_file = open_ofstream(script_path);
 
         fs::u8path python_path;
-        if (m_context->has_python)
+        if (m_context->python_params().has_python)
         {
-            python_path = m_context->relocate_prefix / m_context->python_path;
+            python_path = m_context->prefix_params().relocate_prefix
```
Klaim, line 170: Does the prefix params have to be re-evaluated here or can we get them up in the function and read that? (same question for following functions, where it's relevant)

libmamba/src/core/link.cpp
```diff
@@ -164,9 +165,10 @@ namespace mamba
...
+                          / m_context->python_params().python_path;
```
Klaim, line 171: Can we get the params once and read them through the function?

Klaim (approved):
Other than minor refinements, TLDR

libmamba/src/core/link.cpp
JohanMabille: The parameter is just used here, so I did'nt capture it at the beginning of the function to avoid adding code. Further refinement will be to remove the usage of `m_transaction_context` as much as possible in the implementation of the link related classes and have the required parameters passed as function parameters, but I think that deserves itw own PR.

PR #3952, opened 2025-05-23
Out of context

JohanMabille:


JohanMabille:
Let's rework the APIs in a dedicated PR.

Issue #1026, opened 2021-06-21
Env vars in environment.yaml


quancore:
@wolfv I am trying to add env var using environment.yaml via the section of ```variables```, however my env var not in there. How can add the env var via environment.yaml , which is valid for conda.

adriendelsalle:
Hi @quancore , thank you for opening this issue!

Are you using `micromamba` or `mamba`?
It's not yet implemented in `micromamba`, and `mamba` implementation relies on `conda` and should have the exact same behavior.

quancore:
@adriendelsalle I am using micromamba, what is the best way right now to define env vars without additional config?

wolfv:
@quancore you can add them to your activation scripts or create a package that has the environment vars in the packages' activatation script https://docs.conda.io/projects/conda-build/en/latest/resources/activate-scripts.html

Besides that we don't have a good way yet.

wolfv:
Feel free to open a PR with your desired behavior, btw! :)

adriendelsalle:
It's basically achievable by adding a loop over the key/value pairs from YAML to set the env vars [here](https://github.com/mamba-org/mamba/blob/b2851c8592f015a5b4356908d9fc638a05faff25/src/api/install.cpp#L119).

The most difficult part would be to clean those env vars after the operation to not pollute following ones. Since the envs vars are defined in a file, they should maybe tied to the lifetime of the operation relying on that file.

wolfv:
We have an implementation in https://github.com/mamba-org/mamba/pull/1423 to replace environment vars in `.condarc` (yaml) files.

We could use the same code for environment.yml files.

jonashaag:
This is about setting environment variables upon activation, not replacing environment variables in the `environment.yml`, right?

wolfv:
Yes, I think so

carschandler:
Any plans to implement this? To clarify, this is the relevant section of the conda documentation:
https://conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html#setting-environment-variables

apoorvkh:
+1 this would be very helpful, thank you!

jonashaag:
It doesn't seem to be on anyone's roadmap. Mamba is free software, so if anyone wants to have this feature, they can send a PR!

jtherrmann:
From <https://github.com/mamba-org/mamba/issues/1026#issuecomment-866191868>:
[...]

PR #3955, opened 2025-05-27
Handle environment variables from `yaml` file

### LINKED ISSUE #3884: Not exactly following the standardized environment yaml file
`mamba/micromamba` are not handling everything specified in this [CEP](https://github.com/conda/ceps/pull/81).
Missing keys (to be checked more thoroughly):
`prefix`, `variables`, `platforms`, `category`

Hind-M:
Fix #3884 

As specified in the [CEP](https://github.com/conda/ceps/blob/main/cep-0024.md):

- The `prefix` is often ignored (it's the case in `conda` as well)
- `platforms` and `category` are optional and are mostly used by tools that generate lockfiles.

Therefore, those are not considered in `mamba`.

This PR handles the missing key `variables` corresponding to the environment variables to be set upon environment activation (through `state` file).

### DIFF
--- a/libmamba/include/mamba/api/install.hpp
+++ b/libmamba/include/mamba/api/install.hpp
@@ -8,6 +8,7 @@
 #define MAMBA_API_INSTALL_HPP
 
 #include <iosfwd>
+#include <map>
 #include <string>
 #include <vector>
 
@@ -80,7 +81,18 @@ namespace mamba
     {
         void create_target_directory(const Context& context, const fs::u8path prefix);
 
-        void create_empty_target(const Context& context, const fs::u8path& prefix);
+        void create_empty_target(
+            const Context& context,
+            const fs::u8path& prefix,
+            const std::map<std::string, std::string>& env_vars,
+            bool no_env
+        );
+
+        void populate_state_file(
+            const fs::u8path& prefix,
+            const std::map<std::string, std::string>& env_vars,
+            bool no_env
+        );
 
         void file_specs_hook(Configuration& config, std::vector<std::string>& file_specs);
 
@@ -102,6 +114,7 @@ namespace mamba
         {
             std::string name;
             std::vector<std::string> dependencies, channels;
+            std::map<std::string, std::string> variables;
             std::vector<other_pkg_mgr_spec> others_pkg_mgrs_specs;
         };
 
--- a/micromamba/tests/test_create.py
+++ b/micromamba/tests/test_create.py
@@ -430,9 +430,9 @@ def test_spec_file_env_vars(tmp_home, tmp_root_prefix, tmp_path, env_vars, no_en
         "dependencies: [numpy]",
     ]
     if env_vars:
-        file_content.append(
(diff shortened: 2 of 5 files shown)


micromamba/tests/test_create.py
```diff
@@ -419,6 +419,49 @@ def test_channels(tmp_home, tmp_root_prefix, tmp_path, cli, yaml, env_var, rc_fi
...
+    env_name = "env-check-env-vars"
+
+    spec_file = tmp_path / "env-check-env-vars.yaml"
+    file_content = [
+        "dependencies: [numpy]",
+    ]
+    if env_vars:
+        file_content.append(
+            "variables: {'MY_ENV_VAR': 'My Value', 'MY_OTHER_ENV_VAR': 'Another Value'}",
+        )
```
jjerphan, line 435: Looking at [this section from conda's documentation](https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html#setting-environment-variables), it seems that is also a valid option which we must test:

```yaml
variables:
  MY_ENV_VAR: 'My Value'
  MY_OTHER_ENV_VAR: 'Another Value'
```

jjerphan (commented):
Thank you for this PR which also tackles the long-standing https://github.com/mamba-org/mamba/issues/1026.

libmamba/include/mamba/api/install.hpp
```diff
@@ -78,7 +79,18 @@ namespace mamba
     {
         void create_target_directory(const Context& context, const fs::u8path prefix);
 
-        void create_empty_target(const Context& context, const fs::u8path& prefix);
+        void create_empty_target(
+            const Context& context,
+            const fs::u8path& prefix,
+            const std::map<std::string, std::string>& env_vars,
+            bool no_env
+        );
```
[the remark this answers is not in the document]
JohanMabille: Indeed, although this is in a detail namespace, this is visible in the ABI and this change is backward incompatible? We'ell have to update the ABI version accordingly (we can consider this change as API backward compatible since this method is not supposed to be used by clients).

On the long run, we should move all these detail functions in private headers (but this deserves a dedicated PR)

JohanMabille (approved):

micromamba/tests/test_create.py
```diff
@@ -419,6 +419,49 @@ def test_channels(tmp_home, tmp_root_prefix, tmp_path, cli, yaml, env_var, rc_fi
         assert res["channels"] == ["conda-forge"]
 
 
+@pytest.mark.parametrize("shared_pkgs_dirs", [True], indirect=True)
+@pytest.mark.parametrize("env_vars", (False, True))
+@pytest.mark.parametrize("no_env", (False, True))
+def test_spec_file_env_vars(tmp_home, tmp_root_prefix, tmp_path, env_vars, no_env):
```
jjerphan, line 425: I tested locally, and activating the environment successfully sets environment variables.

(I wonder whether there's a test checking that environment variables are set when the environment is activated.)

jjerphan (approved):

PR #3956, opened 2025-05-27
Fix fmt headers includes

Hind-M:
I don't know how this is building on the CI without failures but it's not the case on my machine.
`#include <fmt/format.h>` should have been moved alongside `fmt::format_error` in https://github.com/mamba-org/mamba/pull/3942

### DIFF
--- a/libmamba/include/mamba/specs/chimera_string_spec.hpp
+++ b/libmamba/include/mamba/specs/chimera_string_spec.hpp
@@ -11,6 +11,7 @@
 #include <variant>
 
 #include <fmt/core.h>
+#include <fmt/format.h>
 
 #include "mamba/specs/error.hpp"
 #include "mamba/specs/glob_spec.hpp"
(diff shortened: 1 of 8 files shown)


JohanMabille:
I don't even understand why we need(ed) this include because "fmt/core.h" already includes "fmt/format/h", see [core.h](https://github.com/fmtlib/fmt/blob/11.1.4/include/fmt/core.h).

jjerphan:
Probably this depends on some versions of `fmt`?

JohanMabille:
Thta's my guess, but `core.h` has included `format.h`  since version 11.0, @Hind-M is it possible that you locally build with fmt 10? fmt 11 is supposed ot be backward incompatible with fmt 10, so current main should not build with such an old version.

JohanMabille:
OK so that's it, building with 10.2 my assumption about backward incompatible changes was wrong.

PR #3965, opened 2025-06-02
Compile with C++20

AntoinePrv:
# Description

## Type of Change

- [ ] Bugfix
- [ ] Feature / enhancement
- [ ] CI / Documentation
- [x] Maintenance

## Checklist


libmamba/include/mamba/util/flat_set.hpp
```diff
@@ -31,11 +32,9 @@ namespace mamba::util
      * Like, ``std::set``, uniqueness is determined by using the equivalence relation.
      * In imprecise terms, two objects ``a`` and ``b`` are considered equivalent if neither
      * compares less than the other: ``!comp(a, b) && !comp(b, a)``
-     *
-     * @todo C++23 This is implemented in <flat_set>
      */
     template <typename Key, typename Compare = std::less<Key>, typename Allocator = std::allocator<Key>>
-    class flat_set : private std::vector<Key, Allocator>
+    class MAMBA_DEPRECATED_CXX23 flat_set : private std::vector<Key, Allocator>
```
Klaim, line 37: Note that the `flat_set` in c++23 is a bit different from this one (and assumed mored performant): it's actually 2 vectors, one with keys one with values, with still the map-like interface. It might not change a thing here, I'm not sure.

libmamba/tests/src/core/test_filesystem.cpp
```diff
@@ -42,74 +46,119 @@ namespace mamba
...
-#if defined(_WIN32)
-            REQUIRE(fs::to_utf8(some_path, { /*normalize_sep=*/true }) == u8"a\\b\\c");
-#else
-            REQUIRE(fs::to_utf8(some_path, { /*normalize_sep=*/true }) == some_path_str);
-#endif
+            std::filesystem::path some_path = some_path_str;
+
+            REQUIRE(
+                fs::to_utf8(some_path, { .normalize_sep = false })
+                == util::to_utf8_std_string(some_path_str)
```
Klaim, line 53: As `util::to_utf8_std_string(some_path_str)` is used multiple time, maybe make a named constant from it?

libmamba/src/util/encoding.cpp
```diff
@@ -296,4 +296,26 @@ namespace mamba::util
...
+
+    auto to_utf8_std_string(std::u8string_view text) -> std::string
+    {
+        std::string result;
+        result.reserve(text.size());
+        for (char8_t c : text)
+        {
+            result.push_back(static_cast<char>(c));
+        }
+        return result;
```
JohanMabille: Nitpicking: result could be initialized with std::ranges::views::transform

libmamba/src/util/encoding.cpp
```diff
@@ -296,4 +296,26 @@ namespace mamba::util
...
+    auto to_u8string(std::string_view text) -> std::u8string
+    {
+        std::u8string result;
+        result.reserve(text.size());
+        for (char c : text)
+        {
+            result.push_back(static_cast<char8_t>(c));
+        }
+        return result;
+    }
```
JohanMabille, line 315: Same remark here

libmamba/src/fs/filesystem.cpp
```diff
@@ -44,26 +45,27 @@ namespace mamba::fs
...
-        }
+            if (utf8_options.normalize_sep)
+            {
+                return normalized_separators(path).u8string();
+            }
+            else
+            {
+                return path.u8string();
+            }
+        }();
```
JohanMabille, line 60: Nitpicking: maybe a simple ternary operator would be easier to read?

Klaim: Note that the whole `if` bloc is the previous code, but yes it could be reduced by a ternary operator. Not sure it would be easier to read, it would just take less space. Or do you mean also removing the called-in-place lambda?

libmamba/include/mamba/util/flat_set.hpp
AntoinePrv: Mostly a reminder to not let it go away. I think the only requirement we use is that it's a set-like thing.

Klaim: Oh wait, I was thinking about `flat_map`, not `flat_set`, ignore my remark 👍🏽

libmamba/tests/src/core/test_filesystem.cpp
AntoinePrv: There are a few subtleties of what is given a `string` and what is give a `u8string` so I went methodic and conservative.

AntoinePrv:
There are still many things that can be improved with C++20 (`rg 'TODO.*C\+\+20'`), including more general use of ranges where possible.
This is just the PR making the switch. All the changes will come gradually as we touch different part of the code.

Klaim:
Agreed, a C++ version change pr should focus on just making sure the project still builds after changing the compiler flags. The other improvements definitely has to be built over it, not in it.

PR #3968, opened 2025-06-04
Use range in Solution

AntoinePrv:
# Description
- **Add ranges to solution**
- **Add Solution::packages**
- **Move libmamba to Solution range methods**
- **Remove Solution for_each functions**

## Type of Change

- [ ] Bugfix
- [ ] Feature / enhancement
- [ ] CI / Documentation
- [x] Maintenance

## Checklist

### DIFF
--- a/.github/workflows/brew.yml
+++ b/.github/workflows/brew.yml
@@ -29,7 +29,7 @@ jobs:
 
   build_homebrew:
     name: Build on homebrew
-    runs-on: macos-13
+    runs-on: macos-15
 
     steps:
       - name: Checkout mamba repository
--- a/cmake/CompilerWarnings.cmake
+++ b/cmake/CompilerWarnings.cmake
@@ -97,8 +97,8 @@ function(mamba_target_add_compile_warnings target)
         -Wconversion
         # Warn on sign conversions
         -Wsign-conversion
-        # Warn if a null dereference is detected
-        -Wnull-dereference
+        # Warn if a null dereference is detected Deactivated because it produced too many false
+        # positive with ranges while we are not doing many pointer operations. -Wnull-dereference
         # Warn if float is implicit promoted to double
         -Wdouble-promotion
         # Warn on security issues around functions that format output (ie printf)
--- a/libmamba/include/mamba/solver/solution.hpp
+++ b/libmamba/include/mamba/solver/solution.hpp
@@ -114,7 +114,7 @@ namespace mamba::solver
             );
         }
 
-        template <typename Range>
+        template <std::ranges::range Range>
         auto packages_to_remove_impl(Range& actions)
         {
             namespace views = std::ranges::views;
@@ -161,7 +161,7 @@ namespace mamba::solver
             );
         }
 
-        template <typename Range>
+        template <std::ranges::range Range>
         auto packages_to_install_impl(Range& actions)
         {
             namespace views = std::ranges::views;
@@ -204,7 +204,7 @@ namespace mamba::solver
             );
         }
 
-        template <typename Range>
+        template <std::ranges::range Range>
         auto packages_to_omit_impl(Range& actions)
         {
             namespace views = std::ranges::views;
@@ -249,7 +249,7 @@ namespace mamba::solver
             return out;
         }
 
-        template <typename Range>
+        template <std::ranges::range Range>
         auto packages_impl(Range& actions)
         {
             namespace views = std::ranges::views;
--- a/libmamba/include/mamba/solver/solution.hpp
+++ b/libmamba/include/mamba/solver/solution.hpp
@@ -13,7 +13,6 @@
 #include <vector>
 
 #include "mamba/specs/package_info.hpp"
-#include "mamba/util/loop_control.hpp"
 #include "mamba/util/type_traits.hpp"
 
 namespace mamba::solver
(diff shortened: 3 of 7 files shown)


.github/workflows/brew.yml
```diff
@@ -29,7 +29,7 @@ jobs:
 
   build_homebrew:
     name: Build on homebrew
-    runs-on: macos-13
+    runs-on: macos-15
```
AntoinePrv, line 32: 👀

cmake/CompilerWarnings.cmake
```diff
@@ -97,8 +97,8 @@ function(mamba_target_add_compile_warnings target)
         -Wconversion
         # Warn on sign conversions
         -Wsign-conversion
-        # Warn if a null dereference is detected
-        -Wnull-dereference
+        # Warn if a null dereference is detected Deactivated because it produced too many false
+        # positive with ranges while we are not doing many pointer operations. -Wnull-dereference
```
AntoinePrv, line 101: 👀

libmamba/include/mamba/solver/solution.hpp
```diff
@@ -67,22 +67,23 @@ namespace mamba::solver
         using action_list = std::vector<Action>;
 
         action_list actions = {};
-    };
 
-    template <typename Iter, typename UnaryFunc>
-    void for_each_to_remove(Iter first, Iter last, UnaryFunc&& func);
-    template <typename Range, typename UnaryFunc>
-    void for_each_to_remove(Range&& actions, UnaryFunc&& func);
+        /** Return a view of all unique packages involved in the solution. */
```
Klaim: I think it would be helpful to clarify in these documentation the validity of these views when modifying member functions are called. Are they invalid? Depends on what is called? When it's called? etc. (kind of same way we specify the validity of iterators for a container)

libmamba/include/mamba/solver/solution.hpp
```diff
@@ -112,35 +113,27 @@ namespace mamba::solver
...
             );
         }
-    }
 
-    // TODO(C++20): Poor man's replacement to range filter transform
-    template <typename Iter, typename UnaryFunc>
-    void for_each_to_remove(Iter first, Iter last, UnaryFunc&& func)
-    {
-        for (; first != last; ++first)
+        template <typename Range>
```
Klaim: As we are now in C++20, I would prefer if any new/changed template type parameter be constrained by C++ concepts, in particular in situations where we already have the concept in the STL. Here it would be one of the [range concepts](https://en.cppreference.com/w/cpp/ranges.html) (see "range concepts" section. Not sure which one is expected here.

Same remark to other functions below.

libmamba/include/mamba/solver/solution.hpp
```diff
@@ -218,35 +203,72 @@ namespace mamba::solver
...
-                }
             }
+            return out;
         }
+
+        template <typename Range>
+        auto packages_impl(Range& actions)
+        {
+            namespace views = std::ranges::views;
+            return actions                                                             //
```
Klaim, line 272: What are the comments for?

libmamba/include/mamba/solver/solution.hpp
```diff
@@ -218,35 +203,72 @@ namespace mamba::solver
...
+                   | views::transform([](auto& a) { return package_unique_ptrs(a); })  //
```
Klaim, line 273: Potential very minor improvement, can be ignored: the lambda wouldnt be necessary if such algorithm (`package_unique_ptrs`) were implemented as a callble object (with a template call operator member) instead of a template free function.

libmamba/src/core/transaction.cpp
```diff
@@ -242,16 +243,13 @@ namespace mamba
         else
         {
             // The specs to install become all the dependencies of the non intstalled specs
-            for_each_to_omit(
-                m_solution.actions,
-                [&](const specs::PackageInfo& pkg)
+            for (const specs::PackageInfo& pkg : m_solution.packages_to_omit())
```
Klaim, line 246: I'm not sure if this case is always valid in C++20, I remember some issues with temporary ranges in for-range loop, but I might be mixing up with another case.

libmambapy/src/libmambapy/bindings/solver.cpp
```diff
@@ -302,36 +302,44 @@ namespace mambapy
...
                     return out;
                 }
             )
+            .def(
+                "to_install",
+                [](const Solution& solution) -> std::vector<specs::PackageInfo>
+                {
+                    // TODO(C++23): std::ranges::to
+                    auto range = solution.packages_to_install();
+                    return { range.begin(), range.end() };
```
Klaim, line 324: Doesnt `{` calls the consrtuctor which takes only values? 
Or is this callilng the range constructor because the type doesnt match the values?
I assumed that you would get compilation errors in that case, not overload resolution.

Klaim: OK I checked and that works as expected. TIL

libmamba/include/mamba/solver/solution.hpp
AntoinePrv: Forcing a nice formatting to clang-format 😅

libmamba/include/mamba/solver/solution.hpp
AntoinePrv: Good call! I'm just getting started with Concepts.

libmamba/include/mamba/solver/solution.hpp
Klaim: Ah yes... sad but necessary for this kind of code. 👍🏽

.github/workflows/brew.yml
jjerphan: Is this required? If so `micromamba` might not be distributable for the all builds of homebrew which use macos-13 and macos-14.

Slightly related: https://github.com/mamba-org/mamba/pull/3578#issuecomment-2948615175.

cmake/CompilerWarnings.cmake
jjerphan:
```suggestion
        # Warn if a null dereference is.
        # Currently ceactivated because it produced too many false
        # positive with ranges while we are not doing many pointer operations.
        # -Wnull-dereference
```

AntoinePrv:
@jjerphan about macos, yes it is the minimum version I was able to bump to in order to make `<range>` fully work.
This is why I drew attention to it. On the one hand this PR is not striclty necessary, on the other, the point of moving to C++20 is to use it, so that bump will come sooner or later.

Also note that this is a Homebrew requirement, since conda-forge ships it's own libcxx, it should not be the same for `miniconda`/`micromamba`.

Here are the hardware requirements for macos-15 ([Wiki](https://en.wikipedia.org/wiki/MacOS_Sequoia)):
- iMac (2019 and later)
- iMac Pro (2017)
- MacBook Air (2020 and later)
- MacBook Pro (2018 and later)
- Mac Mini (2018 and later)
- Mac Pro (2019 and later)
- Mac Studio (all models)

PR #3981, opened 2025-06-09
Fix env vars substitution from env yaml file

Hind-M:
# Description

## Type of Change

- [x] Bugfix
- [ ] Feature / enhancement
- [ ] CI / Documentation
- [ ] Maintenance

## Checklist


libmamba/src/core/util.cpp
```diff
@@ -86,6 +86,42 @@ namespace mamba
...
+                // strip $
+                var = var.substr(1);
+            }
+            auto val = util::get_env(var);
+            if (val)
+            {
+                s.replace(match[0].first, match[0].second, val.value());
+                // It turns out to be unsafe to modify the string during
+                // sregex_iterator iteration. Start a new search by recursing.
+                return expandvars(s);
```
jjerphan, line 119: So should we only have one iteration of the `std::sregex_iterator` above?

jjerphan (approved):

libmamba/src/core/util.cpp
jjerphan: For instance, instead of recursing, we can instead create new `sregex_iterator` iterators until no replacements are performed.

Hind-M: I haven't looked into the details of these comments and if it's improvable. So far left it as it is. I think any improvements attempts should/may be done in a dedicated PR.

PR #3984, opened 2025-06-12
`synchronized_value`

Klaim:
# Description

Adds an implementation of `synchronized_value` inspired by `boost::thread::synchronized_value` and the Concurrency TS 2's specifications. It doesnt follow these exactly for the advanced features but do so for the basic ones.

As there is a reticence to depend upon Boost, we decided it would be best to have an implementation at hand for libmamba's internals as we have many cases of tricky mutex handling. The logging split which is WIP will depon on this tool.

Supports shared mutex by using shared locking instead of normal/exclusive locking when the access is const.

## Type of Change

- [ ] Bugfix
- [ ] Feature / enhancement
- [ ] CI / Documentation
- [x] Maintenance
- [x] Internal tooling

## Checklist

### DIFF
--- a/libmamba/include/mamba/util/synchronized_value.hpp
+++ b/libmamba/include/mamba/util/synchronized_value.hpp
@@ -9,6 +9,7 @@
 
 #include <concepts>
 #include <mutex>
+#include <shared_mutex>
 #include <tuple>
 
 namespace mamba::util


Klaim:
A note about the documentation: doxygen's autobrief generates the brief from the first line (until the first dot) of the doxygen comments, so there is no need to add `@brief`, no need to add an empty first line etc. 
I also used the less noisy doxygen comments form (`*` vertical is really not helping reading) and didnt use `@param` when the brief already describe enough of the behavior and the parameters are constrained enough.

Klaim:
Looks like my local VS2022 (preview) has fixes that the CI doesnt have yet. I'll do some woarkaounds.

libmamba/include/mamba/util/synchronized_value.hpp
```diff
@@ -0,0 +1,607 @@
...
+    {
+        return std::scoped_lock{ mutex... };
+    }
+
+    namespace details
+    {
+        template <Mutex M>
+        M& mutex_ref()
+        {
+            static M m;
```
AntoinePrv: Why static?

Klaim: that function is only used in non-evaluated context so I just wrote that to avoid having to construct a mutex. In theory we could remove the definition I think.
But tu clarify: I needed a reference to be returned, not a r-value reference, not a moved-from, not a value. Returning a temporary would have not compiled.

Klaim: I can confirm I can just remove the definition, I'll push the change. I suspect there might be a standard tool to get what I wnat but I didnt find it , `std::declval` isnt it.

PR #4001, opened 2025-07-01
maint: handle `fmt>=11.2`

Klaim:
# Description

## Type of Change

- [ ] Bugfix
- [ ] Feature / enhancement
- [ ] CI / Documentation
- [x] Maintenance

## Checklist

### DIFF
--- a/libmambapy/src/libmambapy/bindings/utils.cpp
+++ b/libmambapy/src/libmambapy/bindings/utils.cpp
@@ -126,9 +126,9 @@ namespace mambapy
                     const auto fg = style.get_foreground();
                     if (fmt_is_rgb(fg))
                     {
-                        return { { fmt::rgb(fg.value.rgb_color) } };
+                        return { { fmt::rgb(fg.value()) } };
                     }
-                    return { { static_cast<fmt::terminal_color>(fg.value.term_color) } };
+                    return { { static_cast<fmt::terminal_color>(fg.value()) } };
                 }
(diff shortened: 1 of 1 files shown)


.github/workflows/static_build.yml
```diff
@@ -159,7 +159,7 @@ jobs:
             nlohmann_json
             simdjson-static>=3.3.0
             spdlog
-            fmt>=11.1.0
+            fmt>=11.2.0
```
Hind-M: If we decide to not support previous versions, we should remove [this function](https://github.com/mamba-org/mamba/blob/main/libmambapy/src/libmambapy/bindings/utils.cpp#L25-L32) and use the correct api directly where needed (`!is_terminal_color()`).

Hind-M:
Just FYI in case you didn't notice, when opening a PR with the template, you can either use the `Type of change` section and check the corresponding box or use conventional commit keywords in the title.
If both are used, it will just be redundant (checking `Maintenance` and use `chore` in the title for example) but won't be problematic or would lead to the `Check release label` workflow failure if different => 2 release labels.

There is this comment in the template to remind people of that behavior:
`<!-- Please skip this part if you are already using conventional commit keywords in the PR title. -->`

Note that `maint` is not considered as a conventional commit keyword and therefore is ignored but `fix` is 😉.

Hind-M (approved):
LGTM, thanks!

Issue #2288, opened 2023-02-13
Libraries logging callbacks


AntoinePrv:
For `libmamba` to be free of side effects (such as printing), it is best to provide dependency-free callback functions for logging purposes.
Users, including micromamba (or a verbose `libmambacli`), can then set these callbacks to forward the output to a log, a stdout/stderr...

## Yes-goals
It would useful if the callback functions handle the following inputs:
- Source file and line of call (this will be mostly useful for debugging/development)
- Some levels (Info, Debug, Warning, Error)
- Some namespace (e.g. libmamba, curl, libsolv) so that we can use these as callbacks in `libmamba` private dependencies as well.

## No-goals
Re-writing a logging library is not a goal here. Everything that can be done by the user (eventually though a 3rd party) should be excluded, for example:
- Any IO
- Time and date (they can be computed in the user callback)
- Dynamic behavior such as silencing through an environment variable

## Maybe-goals
- Thread safety: should it be the responsibility of the user callback to be thread safe, or should libmamba only call the user callbacks in a thread safe fashion? Customizable?
- `std::ostream` (or similar way to provide a message by parts) support for avoiding allocating temporary strings
- Lazy `fmt` formatting and concatenating on lib mamba side, for instance
  ```cpp
  template<typename Args
  log_error_fmt(const char* fmt, Args const& args...){
    if(user_log_error){
        user_log_error(fmt::format(fmt, args));
    }
  }
  ```

### Source
In C++20, we'de have `std::source_location`. Otherwise we can use macros with `__FILE__` and the like (no portable function name though).

### Enums vs mulitple callbacks 
Should users provide one callback per log level
```cpp
log_info(std::string_view msg, /* etc */);
log_debug(std::string_view msg, /* etc */);
...
```
Or a single function with a log level
```cpp
log(mamba::LogLevel lvl, std::string_view msg)
```
The former is mostly useful for if there is no way to pass the message by parts (e.g. `std::ostream`) to the user so that temporary allocations can be deactivated as needed.

Klaim:
Here is my current opinion summarized:
- there should be one unique callback stored for all logging handling. No need to handle thread-safety on our side, just let the callback handle it;
- the callback should take an object, a log record, and that's it - all the informations are in there;
- informations coming from our library should be minimal, so message, "channel" and some kind of level  and location seems adequate;
- No formatting should happen in this system, if the callback wants to format they can, or not, that happens there.
[...]

PR #4016, opened 2025-07-16
Logging impl separation

Klaim:
# Description

Introduces a new logging API in `libmamba` and replaces current `spdlog`-related code by usage of that new API, with `spdlog` being one of the possible implementations. Future PRs will move that implementation out of `libmamba`.

- [x] new logging system for users to plug-in a log handler which will receive
  log records from `libmamba` and do whatever appropriate logging impl
  they want with it (this is separate from the "console" output, only
  concerns logging);
- [x] provides a `spdlog`-based log handler, which should do what previous
  `libmamba` versions did, with less user-side fluff;
- [x] `Context` can be initialized with a provided log handler;
- [x] By default, if no provided log handler and logging is enabled,
  `Context` will setup a `spdlog`-based log handler;
- [x] `libmamba`'s logging macros uses the new system;
- [x] `libsolv` and `libcurl` loggers uses the new system;
- [x] various cleanup resulting from these changes;
- [x] logging API specific tests, including test/utility log handler implementations;

Note that this is a stepping stone towards removing `spdlog` as dependency
of `libmamba` (see also https://github.com/mamba-org/mamba/pull/3945)

Also related to #2288 and #3415.

Note that this affects only internal logging, not "normal" library output (which goes through a different circuit, either `std::cout` directly or through `mamba::Console` which should be unaffected here), but at least isolates the logging mechanism from the other library-purpose-specific details.

## Type of Change

- [ ] Bugfix
- [x] Feature / enhancement
- [ ] CI / Documentation
- [ ] Maintenance

## Checklist

### DIFF
--- a/libmamba/tests/libmamba_logging/test_logging_common.hpp
+++ b/libmamba/tests/libmamba_logging/test_logging_common.hpp
@@ -382,9 +382,11 @@ namespace mamba::logging::testing
 
             for (std::size_t i = start_log_idx; i < options.log_count; ++i)
             {
-                log_impl_func({ .message = fmt::format(fmt::runtime(message_format), i),
-                                .level = options.level,
-                                .source = options.log_sources.front() });
+                log_impl_func(
+                    { .message = fmt::format(fmt::runtime(message_format), i),
+                      .level = options.level,
+                      .source = options.log_sources.front() }
+                );
             }
         };
 
--- a/libmamba/src/core/logging_spdlog.cpp
+++ b/libmamba/src/core/logging_spdlog.cpp
@@ -83,7 +83,8 @@ namespace mamba::logging::spdlogimpl
         assert(pimpl);
         if (sources.empty())
         {
-            throw std::invalid_argument("LogHandler_spdlog must be started with at least one log source"
+            throw std::invalid_argument(
+                "LogHandler_spdlog must be started with at least one log source"
             );
         }
 
--- a/libmamba/src/core/package_database_loader.cpp
+++ b/libmamba/src/core/package_database_loader.cpp
@@ -55,12 +55,14 @@ namespace mamba
         database.set_logger(
             [](solver::libsolv::LogLevel level, std::string_view msg)
             {
-                logging::log({
-                    .message = std::string{ msg },
-                    .level = to_mamba(level),
-                    .source = log_source::libsolv,
-                    // THINK: add a location? this line?
-                });
+                logging::log(
+                    {
+                        .message = std::string{ msg },
+                        .level = to_mamba(level),
+                        .source = log_source::libsolv,
+                        // THINK: add a location? this line?
+                    }
+                );
             }
         );
     }
--- a/libmamba/src/download/downloader.cpp
+++ b/libmamba/src/download/downloader.cpp
@@ -318,9 +318,11 @@ namespace mamba::download
                         symbol_for(type),
                         Console::hide_secrets(std::string_view(data, size))
                     );
-                    logging::log({ .message = std::move(message),
-                                   .level = log_level::info,
-                                   .source = log_source::libcurl });
+                    logging::log(
+                        { .message = std::move(message),
+                          .level = log_level::info,
+                          .source = log_source::libcurl }
+                    );
                     break;
                 }
                 default:


Klaim:
As this is now documented (first pass at least...) and mainly lacks tests, I think it's time to review the design.

I will note here a short guide to help navigating the code when reviewing this:

1. The main change is adding a separate logging system which takes a logging implementation, which is also provided. Logging macros didn't change, behavior of the logging system shouldn't have changed, existing tests should not have required any change, and integration tests should pass (hopefully).

2. The important files to check first are listed below, the other files are all side-effects of these files, or additional tests. Please read and review them first as they describe the new logging system.
    - `libmamba/core/logging.hpp` : the new logging system, also see the cpp file.
    - `libmamba/core/logging_spdlog.hpp` : the current default logging implementation, using spdlog (as before this pr), also see the cpp file.
    - `libmamba/core/logging_tools.hpp` : some simple log handler implementations usefule for testing or to replace the default implementation by simpler ones.
  
 3. The content of `logging_spdlog.*` files is isolated so that we can move these into an optional library, which would then be used by (micro)mamba but not necessarilly by other direct users of libmamba. We could then also set the standard output log handler as a default.
 
 4. I added notes in comments (TODO, FIXME and REVIEW) for which you might have opinion or useful information, please feeel free to answer these before we merge this.
 
 5. Inlining choices can be disputable, same for location of the logging system files. I did the simplest thing but we can move things around easilly so if you have reasons to prefer other location and moving implementations into cpp files, I'm very open to suggestions and reasoning.
 
 6. The cases where `exit(0);` are called lead to execution paths which can be tricky to get right. I tried to get things right here but it lead to some weird code in logging.cpp. No idea if it could be made better, without just forbidding calls to `exit(0)` in mamba and libmamba.
 
 7. The concurrency policy in the logging system is that only the current logging parameters are protected. Registering (or replacing) a log handler in the system and stopping the whole logging system are not thread-safe operations. Once a log handler is registered, calling the other operations provided by the logging system (like `mamba::logging::log(...)`) are as thread-safe as the log handler implementation.
 This might be weird but it avoids adding our own thread-safe calls protection over one that could be already implemented in the log handler. For example `spdlog` already has that kind of protection through mutex locks. If we were to make sure the logging system operations are all thread-safe, we would add execution costs to every logging operations. So the current approach is that we require (through documentation) that the user of the logging system do whatever is needed to never call `set_log_handler` and `stop_logging` concurrently, and to always provide `LogHandler` implementations which have member functions thread-safe.
The alternative, adding our thread-safety over the implentation, is still a possible choice if we wanted to remove completely the possibility of data races whatever the log handler used, but at the cost of performance. We can easily decide  to go in that direction instead. I'm not super at ease with allowing data-races by design and only have documentation as a warning, personally. But I'm also not at ease with the performance cost in this context as logging happens a lot in this library (and it is not at all well optimized).

8. In `mamba::logging::AnyLogHandler` there is a possibility to have an SBO optimization. I intended to do it as part of this pr but I think it'll be easier once this design is validated.

What is missing at this point - ALL DONE:
- DONE tests, tests, tests!
- FIXED: there is an issue I need to fix: `LogHandler` currently requires the type to be move-able, but when we take a pointer to such type (in `AnyLogHandler` constructor), we end up requiring the pointer to point on a type for a move-able type, which is not really a requirement we need in that pointer case. I'll be fixing that soon.
- FIXED: there might still be random crashes with quick-exit scenarios, I think I solved them all but not sure yet.

libmamba/include/mamba/core/logging.hpp
```diff
@@ -0,0 +1,1183 @@
...
+        /// Formatting pattern to use in formatted logs.
+        std::string_view log_pattern{ "%^%-9!l%-8n%$ %v" };  // FIXME: IS THIS SPECIFIC TO spdlog???
+    };
+
+    /** Specifies the source a `LogRecord` is originating from.
+        This is mainly useful for debugging issues coming from
+        dependencies that have logging callbacks.
+    */
+    enum class log_source  // THINK: "source" isn't the best way to put it, maybe channel?
+                           // component? sink?
```
JohanMabille: I think `source` is fine:

- `channel` would be confusing since there is already a notion of channel in mamba, which is orthgonal to linking.
- `sink` sounds more like the destination than the source (a concrete logger in the case of an adaptive layer for instance).
- `component` might sounds as a subpart of the logging system.

Maybe `emitter` or `origin` or `sender` could be good alternatives to `source`?

libmamba/include/mamba/core/logging.hpp
```diff
@@ -0,0 +1,1183 @@
...
+        switch (source)
+        {
+            case log_source::libmamba:
+                return "libmamba";
+            case log_source::libcurl:
+                return "libcurl";
+            case log_source::libsolv:
+                return "libsolv";
+            case log_source::tests:
+                return "tests";
```
JohanMabille: Using a `constexpr array` + indexed access (like [here](https://github.com/mamba-org/mamba/pull/4016/files#diff-dd84df6e04bc28417e1a7bf2889a8e063cdb3fc362e3bc057a54d7aaaf8827b2R78-R82)) would make it easier to add new log sources  / emitters in the future.

libmamba/include/mamba/core/logging.hpp
```diff
@@ -0,0 +1,1183 @@
...
+
+            /// Source location of this log if available, otherwise empty.
+            std::source_location location = {};  // assigned explicitly to please apple-clang
+
+            // comparisons are mainly used for testing
+            auto operator==(const LogRecord& other) const noexcept -> bool = default;
+        };
+
+        // TODO REVIEW: it might make more sense to talk about sinks than sources when it comes to
+        // the implementation
```
JohanMabille: See [this comment](https://github.com/mamba-org/mamba/pull/4016/files#r2329484136)

libmamba/include/mamba/core/logging.hpp
```diff
@@ -0,0 +1,1183 @@
...
+
+                @see `mamba::logging::set_log_handler`
+                @see `mamba::logging::AnyLogHandler::start_log_handling`
+
+                @param params  Logging system parameters values currently set in the logging system.
+                @param sources List of possible `LogRecords` origins used by libmamba. This is
+                               mostly used in implementations that need to setup separate log sinks depending
+                               on the source.
+            */
+            handler.start_log_handling(params, sources); // TODO: REVIEW consider not providing `sources` as it is always `all_sources()`?
```
JohanMabille: I see two use cases in favor of keeping the `sources`  parameter:
- allow to register many log handlers that will handle different subset of sources
- be able to filter out some log sources when debugging.

libmamba/include/mamba/core/logging.hpp
```diff
@@ -0,0 +1,1183 @@
...
+                resources or proceeds to cleanup or prepare whatever it needs to before
+                starting receiving log records.
+
+                The implementation must keep track of the logging parameters and take them into account as much as possible.
+
+                No other specified operations will be used by the logging system before this one.
+                After this operation ends, all other specified operations might be called
+                concurrently.
+
+                This operation must be thread-safe. REVIEW: is this necessary?
```
JohanMabille: I think it depends on when the user is "authorized" to register a log handler:
- either it must be set before starting any multithreaded operation (downloading, extracting etc), and then we can relax this constraint, but we should add some mechanism to forbid later registration
- the user can register a log handler whenever he wants, and we need to keep this constraint.

If we don't want to make a choice now, keeping the constraint and relaxing it later is easier than safer than the opposite.

libmamba/include/mamba/core/logging.hpp
```diff
@@ -0,0 +1,1183 @@
...
+
+                It could also be called once while the program exits without calling `mmamba::logging::stop_logging`, for example when `::exit(0);` is called.
+                Naturally this will always be called before the implementation's destructor.
+
+                The intent is to allow the implementation to cleanup or release resources that wont be necessary anymore after this call,
+                but dont destroy the object at this point.
+
+                No other specified operations will be used by the logging system after this one.
+
+                This operation must be thread-safe. REVIEW: is this necessary?
```
JohanMabille: The requirement should be consistent with that of [start_log_handling](https://github.com/mamba-org/mamba/pull/4016/files#r2329582632)

libmamba/include/mamba/core/logging.hpp
```diff
@@ -0,0 +1,1183 @@
...
+
+                This operation must be thread-safe.
+
+                @see `mamba::logging::enable_backtrace`
+                @see `mamba::logging::log_backtrace`
+                @see `mamba::logging::AnyLogHandler::enable_backtrace`
+                @see `mamba::logging::AnyLogHandler::log_backtrace`
+
+            */
+            handler.enable_backtrace(size_t(42));
```
JohanMabille, line 392: I haven't found the question to ask yet 😅

libmamba/include/mamba/core/logging.hpp
```diff
@@ -0,0 +1,1183 @@
...
+        return { log_source::libmamba, log_source::libcurl, log_source::libsolv };
+    }
+
+    namespace logging
+    {
+        // FIXME: this is a placeholder, replace it by the real thing once available.
+        // The intent is to have a type doing SBO when possible, but act as unique_ptr otherwise.
+        // Might require allowing to use shared_ptr too, or some kind of value_ptr.
+        template <typename Interface, std::size_t local_storage_size = sizeof(std::shared_ptr<Interface>)>
+        using SBO_storage = std::unique_ptr<Interface>;
```
JohanMabille, line 193: What real thing are you thinking about? A `std::variant`?

libmamba/include/mamba/core/logging.hpp
```diff
@@ -0,0 +1,1183 @@
...
+                However if the log record has a level greater or equal than the current flush threshold,
+                the implementation musth flush the sink related to the log record's source.
+
+                This operation must be thread-safe.
+
+                @see `mamba::logging::log`
+                @see `mamba::logging::AnyLogHandler::log`
+                @see The `LOG_...` macros.
+            */
+            handler.log(log_record);
```
JohanMabille, line 362: Why not `std::as_const` here?

libmamba/include/mamba/core/logging.hpp
```diff
@@ -0,0 +1,1183 @@
...
+/////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+namespace mamba::logging
+{
+    // NOTE: the following definitions are inline for performance reasons.
+
+    // TODO: find a better name?
+    template <typename Func, typename... Args>
+        requires std::invocable<Func, AnyLogHandler&, Args...>
+    auto call_log_handler_if_existing(Func&& func, Args&&... args) -> void
```
JohanMabille, line 1009: safely_call_log_handler?

libmamba/include/mamba/core/logging.hpp
```diff
@@ -0,0 +1,1183 @@
...
+    {
+        return object;
+    }
+
+    template <LogHandler T>
+    T& as_ref(T* object)
+    {
+        assert(object);
+        return *object;
+    }
```
JohanMabille, line 1090: Is it worth moving these definition to a utils file and relax the constraint on the template parameter? This could probably be used for other types.

libmamba/include/mamba/core/logging.hpp
```diff
@@ -0,0 +1,1183 @@
...
+    inline auto AnyLogHandler::set_log_level(log_level new_level) -> void
+    {
+        assert(m_storage);
+        m_storage->set_log_level(new_level);
+    }
+
+    inline auto AnyLogHandler::set_params(LoggingParams new_params) -> void
+    {
+        assert(m_storage);
+        m_storage->set_params(new_params);
```
JohanMabille:
```suggestion
        m_storage->set_params(std::move(new_params));
```

libmamba/include/mamba/core/logging_tools.hpp
```diff
@@ -0,0 +1,533 @@
...
+            size_t backtrace_max = 0;  // 0 means disabled
+
+        public:
+
+            auto is_enabled() const
+            {
+                return backtrace_max > 0;
+            }
+
+            auto push_if_enabled(LogRecord& record) -> bool
```
JohanMabille, line 68: Consider accepting it by value instead

libmamba/include/mamba/core/logging_tools.hpp
```diff
@@ -0,0 +1,533 @@
...
+                record.message
+            );
+
+            return out;
+        }
+    }
+
+    /** `LogHandler` that retains `LogRecord`s in order of being logged.
+        Can hold any number of records or just the specified number of last records.
+        BEWARE: If not the max number of records is not specified, memory will be consumed at each
```
JohanMabille:
```suggestion
        BEWARE: If the max number of records is not specified, memory will be consumed at each
```

libmamba/include/mamba/core/logging.hpp
```diff
@@ -0,0 +1,1183 @@
...
+        namespace details
+        {
+            // NOTE: this looks complicated because it's a workaround for `std::vector`
+            // implementations which are not `constexpr` (required by c++20), we defer the vector
+            // creation to the moment it's needed. Constexpr constructor is required for a type
+            // which is usable in a `constinit` definition, which is required to avoid the
+            // static-initialization-fiasco (at least for initialization).
+            // TODO: once homebrew stl impl has `constexpr` vector, replace all this by just `using
+            // MessageLoggerBuffer = vector<LogRecord>;`
+            struct MessageLoggerBuffer
```
JohanMabille, line 978: I guess this will disappear with MessageLogger?

Klaim:
> Regarding the requirements on the thread-safety of the underlying logging system, a solution could be to keep the requirements in the documentation, and provide a LoggingWrapper that adds synchronization to every logging method before forwarding the call to the "concrete" logging system.

It's possible to do that but it requires either to add another type-erasing type, or make that type templated on the real type, in which case `synchronized_value<MyLogHandler>` would be basically what you describe. If we support that type or use a concept which would be `LogHandler` except all `x.y()` would be replaced by `x->y()` but the object is not a pointer, then we could support any wrapping type too, including `synchronized_value<MyLogHandler>`

Anyway I agree with the idea. Not sure if it's worth having that in this PR, except if we add support explicitly for `synchronized_value<MyLogHandler>` but I would prefer a more generic solution.

libmamba/include/mamba/core/logging.hpp
Klaim: `emitter` can be confused with a signal/observer system, `origin` would work I think, `sender` would add confusion in c++26.

I guess we can continue with `source` unless someone has objections at review. I'll remove the comments.

libmamba/include/mamba/core/logging.hpp
Klaim: Yes so you're thinking about when the user want to add more sources basically, in which case we have an issue because currently only `log_sources` values are allowed, but we could have an extension mechanism. Or did I misunderstand?

libmamba/include/mamba/core/logging.hpp
Klaim: - the current `set_log_handler` isn't thread-safe, that is, the assignation of the new log handler isn't protected. We cannot protect it without protecting all the other operations, unfortunately (or by trying a hack I'm not convinced we should consider). So basically we are in the case of your first point;
- but I disagree with forbiding later registration, it's problematic for testing at minimum;

I added this note mainly because as `set_log_handler` is not thread-safe, we could assume that the implemntation of the loghanlder `start_logging` will always be called without protection, therefore the whole `set_log_handler` operation will always be called in a situation where the user knows there is no concurrent logging happening. However I suspect that implementations will still need the thread-safety protection implemented even for this function to make the whole api consistent and coherent in behavior...

So ok let's keep the constraint for now, if someone find it unnecessary and gives a good example we can remove the constraint by tweaking the documentation 👍🏽

libmamba/include/mamba/core/logging.hpp
Klaim: Note that the whole "backtrace" feature is required here only because we use the feature originally from spdlog, and I dont want to break the current behavior. Otherwise I find this kind of feature at this layer of abstraction overkill.

libmamba/include/mamba/core/logging.hpp
Klaim: The implementation using SBO would be essentially a wrapper around `std::variant< std::array<std::byte, local_storage_size>, some_ptr_type<Interface> >` , yes. The wrapping would select the appropriate type depending on what's passed to it as value to store: either it's already a pointer moved into `some_ptr_type`, or it's non-pointer in which case either it's small enough to be moved in the array or it needs to moved in heap-allocated memory managed by the `some_ptr_type`. That pointer type could be pre-defined to `unique_ptr` if we force move-only on `AnyLogHandler`, or we need a `value_ptr` to enable copyable types too. Choices choices.

But if it's ok with you i prefer focusing on that in a separate PR (contrary to my initial plan).

libmamba/include/mamba/core/logging.hpp
Klaim: We want to allow copies and moves of log record, most of the time they will be moved. I'm not 100% sure but if here we used `as_const` we would imply that there will be no move?

libmamba/include/mamba/core/logging.hpp
Klaim: I would avoid `safely` as it conveys either thread-safety or something in that direction, and the name then would mean "we'll call the function safely", but here we mean to either call or not call the function depending on if the handler exists or not.

Not sure how to shorctut that "only if log handler exist, otherwise do nothing" thing.

libmamba/include/mamba/core/logging.hpp
Klaim: I could, but no idea if it's worth doing this. Might be a bit too generic without the concept? Or maybe other constraints would work generically.

Anyway yes we can do that quickly, tell me if you prefer me to do it in this pr.

libmamba/include/mamba/core/logging_tools.hpp
Klaim: That would either force a copy of the log record that might not be used,
or force a move that is not acceptable because in the case where we do nothing we want to move the log record somewhere else, see https://github.com/mamba-org/mamba/pull/4016/files/e36d3d6e64f45cf93794c3d1cb6bc73e7c5a964c#diff-42be0a002f3733e0a1995f8c588b9100c62776d0b339541a666f4120b6ee1b88

But maybe it's a sign that this function is weird XD It's kind of a juggling between the caller and the callee here.

libmamba/include/mamba/core/logging.hpp
Klaim: Yes but right now I cannot remove it without breaking something (I suspect) so I won't change it in this PR.
But yeah, if we remove MessageLogger, this whole goes away with it.

libmamba/include/mamba/core/logging.hpp
JohanMabille: Ok

libmamba/include/mamba/core/logging.hpp
JohanMabille: Yes, that was just to have an idea of what you have in mind.

libmamba/include/mamba/core/logging.hpp
JohanMabille: > but I disagree with forbiding later registration, it's problematic for testing at minimum;

In that case, let's just not advertise of this possibility :D

libmamba/include/mamba/core/logging.hpp
JohanMabille: I meant we could want to register a log handler per library (one for libmamba, one for libcurl, etc), but not sure it would make sense. And for the second point, I meant we may want to filter the log of some libraries for debugging purpose (like we want the log_level to be debug, but we don't want toremove the noise from some libraries we know (or suspect) to be unrelated to the investigated bug.

libmamba/include/mamba/core/logging.hpp
JohanMabille: I was asking because you use it [in theprevious requirement](https://github.com/mamba-org/mamba/pull/4016/files#diff-dd84df6e04bc28417e1a7bf2889a8e063cdb3fc362e3bc057a54d7aaaf8827b2R303), and it seems ok to accept the argument by value in [the implementation](https://github.com/mamba-org/mamba/pull/4016/files#diff-dd84df6e04bc28417e1a7bf2889a8e063cdb3fc362e3bc057a54d7aaaf8827b2R709). But I agree that `std::as_const` might be confusing regarding move operations, so maybe a better alternative would be to remove it from the previous constraint? In anycase, I think we should keep things homogeneous here.

libmamba/include/mamba/core/logging.hpp
JohanMabille: I agree this should be done at a higher level. This can be refactored in a second time.

libmamba/include/mamba/core/logging.hpp
JohanMabille: Indeed `safely` is not a good one; I don't have a better name in mind for now, but since it's not really a public API, we can change it later.

libmamba/include/mamba/core/logging.hpp
JohanMabille: That can be done later if we need it, I just wanted to avoid reinventing the whell in the future.

libmamba/include/mamba/core/logging_tools.hpp
JohanMabille: I just found it weird to have a function accepting a reference, and possibly moving its parameter in the implementation. Maybe an additional warning / comment about the bool return parameter could avoid this feeling.

JohanMabille:
> Anyway I agree with the idea. Not sure if it's worth having that in this PR, except if we add support explicitly for synchronized_value<MyLogHandler> but I would prefer a more generic solution.

I agree it can (and probably should) be implemented in a dedicated PR where we can discuss the implementation.

libmamba/include/mamba/core/logging.hpp
Klaim: > I meant we could want to register a log handler per library (one for libmamba, one for libcurl, etc), but not sure it would make sense.

I think it's best if there is only one log handler and the log record tells the source yeah (as currently), to avoid any complexity with whatever the user of the library provides.

> And for the second point, I meant we may want to filter the log of some libraries for debugging purpose (like we want the log_level to be debug, but we don't want toremove the noise from some libraries we know (or suspect) to be unrelated to the investigated bug.

Yes this is covered by having the source in the `LogRecord`.

For `start_log_handling` , I was wondering if it's important to pass explicitly the potential sources to the implementation knowing that that parameter will always be `all_sources()`. But let's keep it that way for now 👍🏽 .

libmamba/include/mamba/core/logging.hpp
Klaim: Indeed, the expectation of normal usage is that the beginning of the program will start the logging ,either explicitly or through Context's constructor (see my changes in this pr), and then disable logging in main either explciitly or through Context's destructor, so that it's correctly scoped (no code executing after main()). But yeah for example libmambapy needs to be able to also register a log handler (in a future pr) that logs using python stuffs, and that means you need to be able to set a log handler at any time, assuming you made sure it's not changed concurrently. The current api reflect that, mainly through `mamba::logging::set_log_handler` (also by returning the previous handler).

libmamba/include/mamba/core/logging.hpp
Klaim: Ah yes I didnt plan to move parameters, that's why I used as_const  there. LoggingParams is a completly different case in my mind.

> But I agree that std::as_const might be confusing regarding move operations, so maybe a better alternative would be to remove it from the previous constraint?

Works for me, will do.

libmamba/include/mamba/core/logging_tools.hpp
Klaim: I'll try to improve this, at a minimum on the docummentation side yes.

Klaim:
I completed the tests so this is now ready for review. 

I recommend reading this first to navigate the changes: https://github.com/mamba-org/mamba/pull/4016#issuecomment-3249830003

It might look a lot but most of the changes are happening in a few files, all the others are referring to these ones.

libmamba/include/mamba/core/logging.hpp
```diff
@@ -0,0 +1,1265 @@
...
+        critical,
+
+        // Special values:
+        off,
+        all
+    };
+
+    inline constexpr auto operator<=>(log_level left, log_level right) noexcept
+    {
+        return static_cast<int>(left) <=> static_cast<int>(right);
```
AntoinePrv, line 116: Should we have `all = 0` `trace = 1, ...` then?

libmamba/include/mamba/core/logging.hpp
```diff
@@ -0,0 +1,1265 @@
...
+        warn,
+        err,
+        critical,
+
+        // Special values:
+        off,
+        all
+    };
+
+    inline constexpr auto operator<=>(log_level left, log_level right) noexcept
```
AntoinePrv: Nit: doesn't `constexpr` imply `inline`?

libmamba/include/mamba/core/logging.hpp
```diff
@@ -0,0 +1,1265 @@
...
+#include <optional>
+#include <source_location>
+#include <sstream>
+#include <string>
+#include <typeindex>
+#include <utility>
+#include <vector>
+
+
+// TODO: rename these macros in a more friendly/conflict-averse manner, with a mamba prefix for
```
AntoinePrv, line 73: Agreed!

libmamba/include/mamba/core/logging.hpp
```diff
@@ -0,0 +1,1265 @@
...
+    {
+        return static_cast<int>(left) <=> static_cast<int>(right);
+    }
+
+    inline constexpr auto operator<=>(log_level left, unsigned long right) noexcept
+    {
+        return static_cast<unsigned long>(left) <=> right;
+    }
+
+    /// @returns The name of the specified log level as an UTF-8 null-terminated string.
```
AntoinePrv: We should stick to on documentation style.
```suggestion
    /** @returns The name of the specified log level as an UTF-8 null-terminated string. */
```

libmamba/include/mamba/core/logging.hpp
```diff
@@ -0,0 +1,1265 @@
...
+#define LOG_CRITICAL    LOG(mamba::log_level::critical)
+// clang-format on
+
+namespace mamba
+{
+    /** Level of logging, used to filter out logs which are at a lower level than the current one.
+        @see `mamba::LoggingParams`
+        @see `mamba::logging::LogRecord`
+        @see `mamba::logging::set_log_level`
+     */
```
AntoinePrv, line 99: NIT: doc style
```suggestion
    /** Level of logging, used to filter out logs which are at a lower level than the current one.
     * @see `mamba::LoggingParams`
     * @see `mamba::logging::LogRecord`
     * @see `mamba::logging::set_log_level`
     */
```

libmamba/include/mamba/core/logging.hpp
```diff
@@ -0,0 +1,1265 @@
...
+    inline constexpr auto name_of(log_source source) noexcept -> const char*
+    {
+        constexpr std::array names{ "libmamba", "libcurl", "libsolv", "tests" };
+
+        assert(source < sizeof(names));
+        return names.at(static_cast<std::size_t>(source));
+    }
+
+    /// @returns All `log_source` values as a range.
+    // FIXME: should be constexpr but some compilers don't implement vector's constexpr destructor
```
AntoinePrv, line 180: Also `std::array` is fine because all sources are known at compile time.

libmamba/src/core/logging.cpp
```diff
@@ -0,0 +1,263 @@
...
+                [[maybe_unused]] auto result = safe_invoke([this] { this->stop_log_handling(stop_reason::program_exit); })
+                    .map_error([](const mamba_error& error){
+                            // Here with report the error in the standard output to avoid any logging
+                            // implementation.
+                            const auto message = fmt::format(
+                                "mamba::logging termination failure: call to `stop_log_handling()` ended with an error (caught, logged, skipped): {}",
+                                error.what()
+                            );
+                            std::cerr << message << std::endl;
+                            std::cout << message << std::endl;
```
AntoinePrv: Both?

libmamba/src/core/logging.cpp
```diff
@@ -0,0 +1,263 @@
...
+    namespace
+    {
+        auto
+        make_safe_log_record(std::string_view message, log_level level, std::source_location location)
+        {
+            // THINK: maybe remove as much locals as possible to enable optimizations with
+            // temporaries
+            // TODO: use fmt or fmt::format to do the space prepend
+            const auto secured_message = Console::hide_secrets(message);
+            auto formatted_message = prepend(secured_message, "", std::string(4, ' ').c_str());
```
AntoinePrv: Nit ?
```suggestion
            auto formatted_message = prepend(secured_message, "", "    ");
```

libmamba/include/mamba/core/common_types.hpp
AntoinePrv, line 1: Remove the file?

libmamba/CMakeLists.txt
```diff
@@ -377,6 +379,9 @@ set(
     ${LIBMAMBA_INCLUDE_DIR}/mamba/core/fsutil.hpp
     ${LIBMAMBA_INCLUDE_DIR}/mamba/core/history.hpp
     ${LIBMAMBA_INCLUDE_DIR}/mamba/core/invoke.hpp
+    ${LIBMAMBA_INCLUDE_DIR}/mamba/core/logging.hpp
+    ${LIBMAMBA_INCLUDE_DIR}/mamba/core/logging_spdlog.hpp
+    ${LIBMAMBA_INCLUDE_DIR}/mamba/core/logging_tools.hpp
```
Hind-M, line 383: Should we put these alongside `.cpp` files to not have them as public headers? (at least the `spdlog` one).
Not sure this is possible with the current state of this PR but could be done later.

Hind-M:
Looking into the logs of the new tests, I see a lot of output (which may not be relevant and make it cumbersome/ hard to read or to interpret if we want to debug with theses tests for example). Not sure how we could handle this, maybe redirect the output while testing?

libmamba/include/mamba/core/logging.hpp
Klaim: That specific enum originally has values set to match the equivalent values of `spdlog::level`, see for example [that function](https://github.com/mamba-org/mamba/blob/2ee50e0eb30cfd66a3ec8b977b6452156741e45f/libmamba/src/core/context.cpp#L119) which now has an equivalent [there](https://github.com/mamba-org/mamba/pull/4016/files#diff-260b64ebfbe0218451c2ed618c68d2954725aae07bd30e47f901199667f5e431R23-R30).

`log_level::all` now matches `spdlog::level::n_levels` so that [when using `spdlog::flush_on` we don't have to convert the value further](https://github.com/mamba-org/mamba/pull/4016/files#diff-2f609aada4ad4e67ba2e6b0c6d3763e4fc43b02a0435642bc6e6cd4fb9b24cd3R220).

If it wasnt to keep the behavior as close as possible as before this PR, I would have indeed tried to have more "logical" values.

libmamba/include/mamba/core/logging.hpp
Klaim: Yes for functions, I'll remove the inline where unnecessary. 👍🏽

libmamba/include/mamba/core/logging.hpp
Klaim: However, that will have to wait another PR :)

libmamba/include/mamba/core/logging.hpp
Klaim: I'm not a fan of that style for one-liners, I dont see it as consistency personally when it comes to doxygen-style comments.

But if you insist ok 👍🏽

libmamba/include/mamba/core/logging.hpp
Klaim: Sorry but I disagree:
- these additional stars adds noise, I find that harder to read personally;
- it's completely unnecessary, they don't add anything;
- you lost the indentation which is designed to keep paragraphs easy to read;
- we have other parts of the code where the style I used is used.

libmamba/include/mamba/core/logging.hpp
Klaim: I actually started with `std::array` but it caused issues (ignoring standard library implementations which still didnt have `constexpr` `std::array`) because the functions that need to get the values dont necessarilly know how many values there is. So I tried to take a span or a range but that lead to other complications (lifetime because we return a temporary or a reference to a static object, etc.).

If you find an alternative that works here and at call points, I'm interested.

libmamba/src/core/logging.cpp
Klaim: Maybe only one of these make sense but I'm not sure which. This is a worst case and need to be reported in some ways and not be invisible.

What do you think?

libmamba/src/core/logging.cpp
Klaim: That's factored from [this code](https://github.com/mamba-org/mamba/blob/2ee50e0eb30cfd66a3ec8b977b6452156741e45f/libmamba/src/core/output.cpp#L583-L613).
I agree. I suspect it was that way to make visible that there is 4 chars.

libmamba/include/mamba/core/common_types.hpp
Klaim: I think I kept it because I was not sure if `LoggingParams` shouldnt be there , but yeah in the end it was not necessary.
@JohanMabille did you have plans to move stuffs in that file?

libmamba/CMakeLists.txt
Klaim: No they need to be in public headers, these 3 headers are public API of the library (hence the extensive documentation 😅). This is not an internal API, to be clear, it's supposed to be used by library users. 

As a side note: after this PR we'll move the `logging_spdlog.hpp/cpp` into a separate library.

Klaim:
> Looking into the logs of the new tests, I see a lot of output (which may not be relevant and make it cumbersome/ hard to read or to interpret if we want to debug with theses tests for example). Not sure how we could handle this, maybe redirect the output while testing?

We have several cases:
- tests using `LogHandler_History` dont output, so these are already invisible.
- tests using `LogHandler_StdOut` by default output to `std::cout` but some tests uses a `std::stringstream` so they dont have visible output. I think I didnt do that for all tests, so I can apply that. However for the concurrency stress test I would need to add a `std::ostream` implementation that's thread-safe but doesnt output visibly.
- tests using `LogHandler_spdlog` will just output because `spdlog` will do that by default. For these not sure what to do, I'll look into if there is a way to tell `spdlog` to not output without having to change how `LogHandler_spdlog` works. Ideally if there was a way to redirect to a null-sink yes that would help.

I would prefer that we dont hide the output of the whole program as if it fails we need to see the catch report. But probably that can be worked at the tests level.

libmamba/include/mamba/core/logging.hpp
Klaim: I did the change to not use `as_const` in this concept.

Klaim:
Looks like the only way to test `LogHandler_spdlog` without output is to use "null sink" in spdlog, but they are also part of the loggers. I can do that as an option in `LogHandler_spdlog` but that implies that the test doesnt reflect reality.

Not sure yet how to do better for  this one.

Klaim:
I went with a way to replace spdlog sinks to a (thread-safe) null sinks while keeping the loggers created, so I think it's good enough for our tests.

libmamba/include/mamba/core/logging.hpp
AntoinePrv: I think I'd be good to switch all to the `///` style eventually so that it's easier to be consistent.

Klaim: Yep it's done in current version (with `/**`), I agree we could switch to `///` at some point.

libmamba/include/mamba/core/logging.hpp
Klaim: I think I'll re-try a quick attempt with `std::span` on a static array, I'm not sure why it was not working but maybe it was because of the compiler for homebrew not being up to date enough.

Klaim: Nope wrong tool, it leads to many potential issues because of ownership.
I wish we had access to something like `std::inplace_vector` (C++26) for this kind of things.

libmamba/src/core/logging.cpp
JohanMabille: I think `std::cerr` is the stream made for reporting errors here. Which workflow could make this invisible if we report on `std::cerr` only?

libmamba/include/mamba/core/common_types.hpp
JohanMabille: Yes, although I don't remember which ones to be honest. We can keep it as a placeholder, or remove it and add it again later when we need it, I don't have any strong opinion about it.

libmamba/src/core/logging.cpp
Klaim: Not certain, I remember issues with only using `std::cerr` but cant find what it was about.
I'll use `std::cerr` only, we can easily change that later if necessary.

libmamba/include/mamba/core/common_types.hpp
Klaim: Ok I'll remove it then.

libmamba/include/mamba/core/invoke.hpp
```diff
@@ -3,6 +3,8 @@
 
 #include <functional>
 
+#include <fmt/core.h>  // TODO: replace by `<format>` once available on all ci compilers
```
jjerphan, line 6: I guess one issue could be open for this to identify what is left to do.

libmamba/include/mamba/core/invoke.hpp
```diff
@@ -3,6 +3,8 @@
...
+
```
jjerphan, line 7: I guess one issue could be open for this to identify what is left to do.

libmamba/include/mamba/core/tasksync.hpp
jjerphan, line 1: Nit: should the changes be reverted here?

libmamba/src/api/configuration.cpp
```diff
@@ -2234,19 +2234,20 @@ namespace mamba
             exit(0);
         }
 
-        m_context.set_log_level(m_context.output_params.logging_level);
+        m_context.set_log_level(m_context.output_params.logging_level);  // TODO remove that
+                                                                         // function from Context
```
jjerphan: Is it possible to add a comment near the declaration or definition of `Context::set_log_level`?

libmamba/src/core/context.cpp
```diff
@@ -157,25 +52,24 @@ namespace mamba
         }
     }
 
-    void Context::enable_logging()
+    void Context::enable_logging(logging::AnyLogHandler log_handler)  // THINK: change name?
+                                                                      // start_logging?
```
jjerphan: Having `start_logging` and `stop_logging` is slightly better IMO, but I do not have strong feelings about it.

libmamba/src/core/singletons.cpp
```diff
@@ -32,6 +34,46 @@ namespace mamba
...
+    //--- Logging System
+    //------------------------------------------------------------------
+
+    namespace logging::details
+    {
+        // FIXME:
+        // This should really be using `std::shared_mutex` but on some compilers they
+        // dont accept `constinit` here, so until this is fixed we'll lose performance using
+        // `std::mutex` when we mostly are reading data, not modifying,
+        // but at least we maintain correctness and avoid static-init fiasco.
```
jjerphan, line 46: Is it just some version of Apple Clang or are there other compilers which are not supporting this correctly? If you remember them, could you list them here?

libmamba/src/core/singletons.cpp
```diff
@@ -32,6 +34,46 @@ namespace mamba
...
+        // The handler MUST NOT be protected from concurrent calls at this level
+        // as that would add high performance cost to logging from multiple threads.
+        // Instead, we expect the implementation to handle concurrent calls correctly
+        // in a thread-safe manner which is appropriate for this implementation.
+        //
+        // There are many ways to implement such handlers, including with a mutex,
+        // various mutex types, through a concurrent lock-free queue of logging record feeding
+        // a specific thread responsible for the output and file writing, etc.
+        // There are too many ways that we can't predict and each impl will have it's own
+        // best strategy.
```
jjerphan, line 64:
```suggestion
        // There are too many ways that we can't predict and each impl will have its own
        // best strategy.
```

libmamba/src/core/singletons.cpp
```diff
@@ -101,6 +143,7 @@ namespace mamba
     bool MessageLoggerData::use_buffer(false);
     std::vector<std::pair<std::string, log_level>> MessageLoggerData::m_buffer({});
 
+
```
jjerphan, line 146:
```suggestion
```

libmamba/include/mamba/core/logging_tools.hpp
```diff
@@ -0,0 +1,674 @@
...
+            -> void
+        {
+            queue.push_back(std::move(record));
+            if (max_elements > 0 and queue.size() > max_elements)
+            {
+                queue.pop_front();
+            }
+        }
+
+        /** Backtrace feature implementation in it's most basic form.
```
jjerphan, line 40:
```suggestion
        /** Backtrace feature implementation in its most basic form.
```

libmamba/include/mamba/core/logging.hpp
```diff
@@ -0,0 +1,1266 @@
...
+                               mostly used in implementations that need to setup separate log sinks depending
+                               on the source.
+            */
+            handler.start_log_handling(params, sources);
+
+            /** When this log handler was registered in the logging system and later
+                another implementation object is registered to replace it or while stopping the logging system,
+                this operation will be called once after unregistering the object from the logging system.
+
+                It could also be called once while the program exits without calling `mamba::logging::stop_logging`, for example when `::exit(0);` is called.
```
jjerphan, line 298:
```suggestion
                It could also be called once while the program exits without calling `mamba::logging::stop_logging`,
                for example when `::exit(0);` is called.
```

libmamba/include/mamba/core/logging.hpp
```diff
@@ -0,0 +1,1266 @@
...
+                Naturally this will always be called before the implementation's destructor.
+
+                The intent is to allow the implementation to cleanup or release resources that wont be necessary anymore after this call,
```
jjerphan, line 301:
```suggestion
                The intent is to allow the implementation to cleanup or release resources that wont be necessary
                anymore after this call, but not destroy the object at this point.
```

Klaim:
> I am not sure the purpose of this PR is entirely clear.

Is there some missing clarifications in the first post?
(btw I dont know if you saw [this message)](https://github.com/mamba-org/mamba/pull/4016#issuecomment-3249830003)

The goal of the logging api is for libmamba to emit log records without having to depend on a specific logging system implementation, and that implementation to use is decided by the program, not the library. But we still use spdlog by default for now until next pr.

> Can you mention that this entire system would allow different dependencies to share the same logging system (spdlog being one candidate)?

The idea is that it's the program (the end-user of the build-graph) that decides such logging system, not the libraries. I can add that the consequence is  that multiple libraries using the same logic of not imposing a logging library (if they are well designed) can indeed share the same logging system setup by the program.

libmamba/include/mamba/core/invoke.hpp
Klaim: I think I wrote this with a previous version of the dependencies that didnt work, `<format>` might now work in the CI (it didnt when I began). I'll try to use `<format>` see if there is still some issues 👍🏽

libmamba/include/mamba/core/tasksync.hpp
Klaim: It's proper formatting of the doc. Why would I revert? Do you see an issue? You would prefer it as a separate pr?

libmamba/src/api/configuration.cpp
Klaim: Will do 👍🏽

libmamba/src/core/context.cpp
Klaim: I agree. Renaming now breaks the `Context` api though, is that ok for you? @JohanMabille I'm fine with that but just want to be sure.

libmamba/src/core/singletons.cpp
Klaim: I purposefully didnt specify which compilers because it might depend on compiler versions (even msvc and gcc and normal clang) too. But yes I can list the ones from our CI, it was apple clang from homebrew if my memory is correct. I'll check and update 👍🏽

libmamba/include/mamba/core/logging.hpp
Klaim: It's weird that the formatting tool didnt catch that...

libmamba/include/mamba/core/logging_tools.hpp
Klaim: For info,
I tried to do something like that (pseudocode):
```c++
auto push_if_enabled(LogRecord&& record, Func on_disabled)
{
     if(is_enabled())
     {
          history.push_back(move(record));
     }
     else
     {
         on_disabled(std::move(record));
     }
}
```

But it led to other issues in the calling side, so I ended up getting back to the same api but with more documentation, as we discussed.

libmamba/src/core/context.cpp
Klaim: I think I'll just do the deprecation dance (deprecate + add the better alternative)

libmamba/include/mamba/core/logging.hpp
jjerphan: (Could documentation style be enforced by a linter?)

libmamba/src/core/context.cpp
JohanMabille: We should keep this API backward compatible as we want to release soon without breaking downstream packages. The deprecation dance is definitely the thing to do here.

Klaim: I just realized that this function is actually private! I'll just rename it then.

libmamba/include/mamba/core/invoke.hpp
Klaim: I'll make a PR to track the issue.

libmamba/include/mamba/core/logging.hpp
```diff
@@ -0,0 +1,1315 @@
...
+    to decide (in the end-user target/program, ideally) which implementation of a logging
+    system to use by this library when it emits log records.
+    This prevents libmamba from imposing a specific implementation and dependency
+    for logging.
+    It also allows the end-user program to use only one logging library for all it's
+    dependencies that emit log records and have a similar way to handle these records.
+
+    Historically libmamba was using `spdlog` internally and used a lot of it's advanced
+    features. To not break the implementation of libmamba but still enable different
+    logging logging libraries to be used, the log-handler implementations that can be used
```
jjerphan: Thank you for this comprehensive comment.
```suggestion
    logging libraries to be used, the log-handler implementations that can be used
```

libmamba/tests/libmamba_logging/test_logging_common.hpp
```diff
@@ -0,0 +1,572 @@
...
+            auto set_params(LoggingParams) -> void {}
+            auto log(LogRecord) -> void {}
+            auto enable_backtrace(size_t) -> void {}
+            auto disable_backtrace() -> void {}
+            auto log_backtrace() -> void {}
+            auto log_backtrace_no_guards() -> void {}
+            auto flush(std::optional<log_source> = {}) -> void {}
+            auto set_flush_threshold(log_level) -> void {}
+
+        // clang-format on
```
jjerphan, line 183: Side-note: why is this formatting different?

libmamba/tests/libmamba_logging/test_logging_common.hpp
```diff
@@ -0,0 +1,572 @@
...
+            const auto random_bactrace = [&] {
+                return random->roll_dice(0,1) ? 0 : random->roll_dice(1, 50);
+            };
+
+            const auto random_params = [&] {
+                return LoggingParams{
+                    .logging_level = random_log_level(),
+                    .log_backtrace = random_bactrace(),
+                };
+            };
```
jjerphan: Nit.

```suggestion
            const auto random_backtrace = [&] {
                return random->roll_dice(0,1) ? 0 : random->roll_dice(1, 50);
            };

            const auto random_params = [&] {
                return LoggingParams{
                    .logging_level = random_log_level(),
                    .log_backtrace = random_backtrace(),
                };
            };
```

libmamba/tests/libmamba_logging/test_logging_common.hpp
```diff
@@ -0,0 +1,572 @@
...
+                    REQUIRE(level != log_level::off);
+                    REQUIRE(level != log_level::all);
+                },
+
+                [&]{
+                    set_logging_params(random_params());
+                },
+
+                [&]{
+                    enable_backtrace(random_bactrace());
```
jjerphan:
```suggestion
                    enable_backtrace(random_backtrace());
```

libmamba/tests/libmamba_logging/test_logging_common.hpp
Klaim: The current linter setup will make these empty functions take 3 lines each plus space between them, it's noisy. We dont need it well formatted, it's just a tool to check that compilation will happen as expected and we need the type to be a valid `LogHandler`, which is why we need these declarations but dont need the implementations.

libmamba/include/mamba/core/logging.hpp
Klaim: > (Could documentation style be enforced by a linter?)

I dont think it formats doxygen comments? Anyway if it does we could do a pass another time, although I would prefer if the doxygen documnetation is not noisy when reading from source. I think that requires some work to get right but yeah in another pr we could do that.

libmamba/include/mamba/core/logging.hpp
Klaim: I dont think it formats doxygen comments? Anyway if it does we could do a pass another time, although I would prefer if the doxygen documnetation is not noisy when reading from source. I think that requires some work to get right but yeah in another pr we could do that.

(sorry, github interface is messing with me XD)

libmamba/src/api/configuration.cpp
Klaim: I marked that function as deprecated (incoming), although I also noted that we'll need a way to notify that logging parameters have changed so that the output params in Context is updated, either that or removing the output params but I dont think that will be possible without a rewrite of the cli parameters processing.
I'll make separate pr for the notifications.

Klaim: Correction: I need to do it in this pr to avoid adding new deprecation warning in the ci.

Klaim: Re-correction: it's too long to do that right now, I'll just move the comment and make a separate pr for that feature.

JohanMabille:
2.3.3 is out, let's merge.

PR #4021, opened 2025-07-21
encoding: fix missing cassert include

javacruft:


Klaim:
There are several issues from different dependencies changes on the CI, we're trying to solve them in #4023 (including this change)

PR #4023, opened 2025-07-23
fix CI issues related to moving dependencies

Klaim:
# Description

API changes in pybind11 v3.0.0 breaks the CI. We didnt plan that change so we restrict mamba's build to use the previous versions.

## Type of Change

- [ ] Bugfix
- [ ] Feature / enhancement
- [ ] CI / Documentation
- [x] Maintenance

## Checklist

### DIFF
--- a/dev/CMakePresetsMamba.json
+++ b/dev/CMakePresetsMamba.json
@@ -23,7 +23,8 @@
     {
       "cacheVariables": {
         "BUILD_MICROMAMBA": "ON",
-        "BUILD_STATIC": "ON"
+        "BUILD_STATIC": "ON",
+        "ENABLE_WIN32_XMLLITE": "ON"
       },
       "hidden": true,
       "name": "mamba-static"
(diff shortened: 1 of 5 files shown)


dev/CMakePresetsMamba.json
```diff
@@ -23,7 +23,8 @@
     {
       "cacheVariables": {
         "BUILD_MICROMAMBA": "ON",
-        "BUILD_STATIC": "ON"
+        "BUILD_STATIC": "ON",
+        "ENABLE_WIN32_XMLLITE": "ON"
```
JohanMabille, line 27: Do we still need it in the preset if it is always set in the CMakeLists.txt file?

Klaim: The CI was passing withtout that change, I didnt try it yet after removing the explicit link to `xmllite.lib` which I find weird. It makes sense as a system-provided library, but libarchive already has it's own way of using it ([see there](https://github.com/libarchive/libarchive/blob/cc97338239c5b1c0e0ffcb323d8cc4fd5929bf54/CMakeLists.txt#L1267)) but  enabling it in our mamba cmake file is not enough to make this work apparently.
Not sure what's best.

Klaim: I added some clarification of the context, with a `TODO` note.

PR #4042, opened 2025-08-07
`synchronized_value` move and copy implementation

Klaim:
# Description

Implemented move operations on `synchronized_value` which were explicitly `= delete;` previously, intentionally until needed. Now I need them for #4016 so here we go.
Also added tests for copy and move. I also added tests for when T is convertible to U or vice-versa and we want to copy or move one into the other when these are into `synchronized_value` (previous code didnt pass all the tests I just added so I fixed these cases). I also switched from using `std::assignable_from` to a custom `weakly_assignable_from` which doesnt require `std::common_reference<T, U>`, same previous issue than checking if `operator==` can be used (see `weakly_equality_comparable` comments for details).

## Type of Change

- [ ] Bugfix
- [ ] Feature / enhancement
- [ ] CI / Documentation
- [x] Maintenance

## Checklist

### DIFF
--- a/libmamba/tests/src/util/test_synchronized_value.cpp
+++ b/libmamba/tests/src/util/test_synchronized_value.cpp
@@ -293,7 +293,6 @@ namespace
 
             synched_convertible_value c{ { a->x + 1 } };
             REQUIRE(a != c);
-
         }
 
         SECTION("move constructible")


Klaim:
Looks like there are connection issues with the repository making integration tests fail.

libmamba/include/mamba/util/synchronized_value.hpp
```diff
@@ -519,6 +570,9 @@ namespace mamba::util
 
         T m_value;
         mutable M m_mutex;
+
+        template <std::default_initializable, Mutex>
+        friend class synchronized_value;
```
Hind-M, line 575: Is there a specific reason for this?

libmamba/include/mamba/util/synchronized_value.hpp
```diff
@@ -560,9 +614,55 @@ namespace mamba::util
         m_value = other.m_value;
     }
 
+    template <std::default_initializable T, Mutex M>
+    synchronized_value<T, M>::synchronized_value(synchronized_value&& other) noexcept
+    {
+        auto _ = lock_as_exclusive(other.m_mutex);
```
Hind-M, line 620: I'm wondering about the differences between:
```
template <Mutex M>
[[nodiscard]]
auto lock_as_readonly(M& mutex)
{
    return std::unique_lock{ mutex };
}
```
and
```
template <Mutex M>
[[nodiscard]]
auto lock_as_exclusive(M& mutex)
{
    return std::unique_lock{ mutex };
}
```

libmamba/include/mamba/util/synchronized_value.hpp
Klaim: Yes, compilers will refuse to compile the implementations of constructors and `operator=` for convertible types because `synchronized_value<T,M>` and `synchronized_value<U,M2>` are not the same type so `private:` prevents to implement them as we need to access the private members of both types and fail to do so as it is forbidden.

Th lines says essentially enables access to `synchronized_vallue<X,Y>` private members, whatever `X` or `Y` is, from the implementation of this template's functions.

libmamba/include/mamba/util/synchronized_value.hpp
Klaim: They do the same thing, but `lock_as_readonly` has overloads that would be better matches for `SharedMutex`. If these matches best, for example when using `std::shared_mutex`, then the lock is "shared" which essentially means it's read-only, with multiple readers allowed. This is best when you need the value to be readable for cheap compared to an exclusive lock.

These `lock_as_...` functions essentially are meant to select the best lock depending on the situation. `lock_as_exclusive` in particular is there for symetry and making it clear what was the intent (exclusive access vs shared access)

Note that if `scoped_lock{ ...}`  had a way to automatically do shared-locking when instructed to and possible, I wouldh ave used that and not bothered with these functions, but nope. At least it makes it clear what's the intent here.

Hind-M (approved):

PR #4057, opened 2025-09-02
Fix install of /etc/profile.d/mamba.sh when `--prefix` is set

### LINKED ISSUE #4055: Don’t use CMAKE_INSTALL_PREFIX for `/etc`
https://github.com/mamba-org/mamba/blob/9872bed0f5a4f22f64e263ed0e6ee8e71f8e6810/micromamba/CMakeLists.txt#L128

It breaks setting `--prefix` dynamically, see https://github.com/mamba-org/mamba/issues/3554

flying-sheep:
Fixes #4055

TODO: Is updating `cmake_minimum_required` correct or is there a way for people to get that policy active with lower versions?

### DIFF
--- a/micromamba/CMakeLists.txt
+++ b/micromamba/CMakeLists.txt
@@ -5,9 +5,6 @@
 # The full license is in the file LICENSE, distributed with this software.
 
 cmake_minimum_required(VERSION 3.16...4.1)
-cmake_policy(SET CMP0025 NEW) # Introduced in cmake 3.0
-cmake_policy(SET CMP0077 NEW) # Introduced in cmake 3.13
-cmake_policy(SET CMP0192 NEW) # Introduced in cmake 4.1
 project(micromamba)
 
 include(GNUInstallDirs)


henryiii:
You can do a range, then the highest one in the range is used for policy defaults.

micromamba/CMakeLists.txt
```diff
@@ -4,9 +4,10 @@
 #
 # The full license is in the file LICENSE, distributed with this software.
 
-cmake_minimum_required(VERSION 3.16)
+cmake_minimum_required(VERSION 4.1)
 cmake_policy(SET CMP0025 NEW) # Introduced in cmake 3.0
```
henryiii: These two are not needed since the minimum is higher than the introduction version. That’s what setting the minimum does.

flying-sheep: sorry, GitHub must have missed up the line range: what does “these two” refer to?

Are you saying I should remove two `cmake_policy(SET ...)` lines? If so, which ones?

I’m not super eager to understand CMake with all its intricacies, so please just tell me what to change or do a ```` ```suggestion ```` :smile:

henryiii: I couldn't figure out how to select a range on my phone. :)

micromamba/CMakeLists.txt
```diff
@@ -4,9 +4,10 @@
 #
 # The full license is in the file LICENSE, distributed with this software.
 
-cmake_minimum_required(VERSION 3.16)
+cmake_minimum_required(VERSION 4.1)
```
henryiii:
```suggestion
cmake_minimum_required(VERSION 3.16...4.1)
```

micromamba/CMakeLists.txt
```diff
@@ -4,9 +4,10 @@
...
 cmake_policy(SET CMP0025 NEW) # Introduced in cmake 3.0
 cmake_policy(SET CMP0077 NEW) # Introduced in cmake 3.13
+cmake_policy(SET CMP0192 NEW) # Introduced in cmake 4.1
```
henryiii:
```suggestion
```

micromamba/CMakeLists.txt
flying-sheep: I also feel like the selection process is wonky in the beta of the new “Files Changed” UI: https://github.com/orgs/community/discussions/163932

flying-sheep:
@henryiii is this good to go?

PR #4058, opened 2025-09-02
Fix nodiscard errors

AntoinePrv:
# Description

- Some warnings/error were properly handled, the compiler just needed a little help;
- Some are converted from ignored to throwing (behaviour change);
- Some are silenced (same behaviour) with a comment indicating that we don't know why.
  Far from perfect, but if it got this far, we can assume it's okayish until we work on that code again.

## Type of Change

- [x] Bugfix
- [ ] Feature / enhancement
- [ ] CI / Documentation
- [ ] Maintenance

## Checklist

### DIFF
--- a/libmamba/src/api/install.cpp
+++ b/libmamba/src/api/install.cpp
@@ -695,7 +695,7 @@ namespace mamba
                                                .value<std::vector<detail::other_pkg_mgr_spec>>())
                     {
                         auto result = install_for_other_pkgmgr(ctx, other_spec, pip::Update::No);
-                        if (!result.has_value())
+                        if (!result)
                         {
                             static_assert(std::is_base_of_v<std::exception, decltype(result)::error_type>);
                             throw std::move(result).error();
--- a/libmamba/src/api/install.cpp
+++ b/libmamba/src/api/install.cpp
@@ -694,7 +694,12 @@ namespace mamba
                     for (auto other_spec : config.at("others_pkg_mgrs_specs")
                                                .value<std::vector<detail::other_pkg_mgr_spec>>())
                     {
-                        install_for_other_pkgmgr(ctx, other_spec, pip::Update::No);
+                        auto result = install_for_other_pkgmgr(ctx, other_spec, pip::Update::No);
+                        if (!result.has_value())
+                        {
+                            static_assert(std::is_base_of_v<std::exception, decltype(result)::error_type>);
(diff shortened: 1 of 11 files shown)


libmamba/src/api/install.cpp
```diff
@@ -694,7 +694,12 @@ namespace mamba
                     for (auto other_spec : config.at("others_pkg_mgrs_specs")
                                                .value<std::vector<detail::other_pkg_mgr_spec>>())
                     {
-                        install_for_other_pkgmgr(ctx, other_spec, pip::Update::No);
+                        auto result = install_for_other_pkgmgr(ctx, other_spec, pip::Update::No);
+                        if (!result.has_value())
```
Klaim: .has_error(); ?

libmamba/src/core/package_fetcher.cpp
```diff
@@ -468,7 +468,8 @@ namespace mamba
     {
         if (cb)
         {
-            safe_invoke(*cb, event);
+            // FIXME why is this unused?
```
Klaim: That's probably because it's a callback call and we don't want exceptions to escape but also can't do anything if it fails. Maybe logging would be useful, maybe not but the error must not be propagated. `safe_invoke` is often used in these case where we call user code and don't want its failure to impact the system calling.

Klaim (approved):
It is unfortunate that usage of expected + warnings makes everything using it so noisy...
But this PR is acceptable as a quick fix.

Note however that every time you ask why there is no handling of a `safe_invoke` call, it is always (in this pr at least) because we dont want to propagate exceptions when calling callbacks coming from code external to the local system (provided by the user of the system). So the only possible improvements for these are either:
- logging then ignoring (most of the these calls dont do that, from what I read from this pr) - the potential risk is adding verbosity in output where there is none currently, and maybe break tests that relies on that ouput (might require several rounds of testing to get right);
- keep ignoring and not propagating, but replace the TODO comment by a clarification that "we dont want to propagate errors coming from user's callbacks";

I'll let you decide if these potential improvements should be done here or in a following PR, for now this is good enough for me.

libmamba/src/api/install.cpp
AntoinePrv: Does not exist in `tl::expected` but I replaced with `if(!result)`.

Klaim (approved):

PR #4083, opened 2025-10-21
Removed deprecated libcurl backend

JohanMabille:
`CURLSSLBACKEND_SECURETRANSPORT` has been deprecated in curl 8.15, see https://daniel.haxx.se/blog/2025/01/14/secure-transport-support-in-curl-is-on-its-way-out/ for rationale.

Latest libcurl available on coda-forge is 8.16, and this prevents the feedstock from building (since we treat warnings as errors). This PR removes the deprecated backend and fixes the build.

### DIFF
--- a/.github/workflows/static_build.yml
+++ b/.github/workflows/static_build.yml
@@ -45,11 +45,17 @@ jobs:
         run: |
           cd micromamba-feedstock/
           sed -i '' '/conda_forge_output_validation/d' conda-forge.yml
-      - name: Replaces python 3.9 with 3.12
+      - name: Replaces python 3.9 with 3.13
         if: ${{ matrix.platform == 'linux' }}
         run: |
           cd micromamba-feedstock/recipe/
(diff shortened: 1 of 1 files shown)


JohanMabille:
> There's another usage here:
> 
> https://github.com/mamba-org/mamba/blob/e082e9b3abacbc5373eb48397cc7595719cf49c9/libmamba/src/core/singletons.cpp#L95-L101
> 
> Which back-end should we use instead?

The answer is not obvious. So instead of removing it, I'm gonna locally ignore the deprecation warning to fix the build, and open an issue to track it.

jjerphan:
I would try to use `CURLSSLBACKEND_OPENSSL`.

JohanMabille:
> I would try to use `CURLSSLBACKEND_OPENSSL`.

Yeah let's actually do that, we cannot simply ignore the warning without setting an upper bound since the lattest libcurl [does not provide this backend anymore](https://github.com/conda-forge/curl-feedstock/commit/2db04fc5f30b88777617de8aa3f1a94332ec76b0)

jjerphan (approved):

JohanMabille:
Let's merge to fix the CI, and try to upgrade curl for emulated architecture in a dedicated PR.

PR #4085, opened 2025-10-23
Support for `mambajs`'s environment lockfile format

Klaim:
# Description

Add support for mambajs environment lockfile format in addition to the currently supported conda/mamba environment lockfile format. 

Both formats don't provide the exact same set of information so we try to at least provide and use what's mandatory and fill in the additional data depending on the format.

The code implementing the parsing of each format has been put in separate cpp files, but the public api is unchanged, so one could just call the public api with either kind of file and we dont need to change anything on the cli front.

See:
 - https://github.com/emscripten-forge/mambajs/blob/main/packages/mambajs-core/schema/lock.v1.0.0.json
 - https://github.com/emscripten-forge/mambajs/blob/main/packages/mambajs-core/schema/lock.v1.0.1.json
 - example: 
[nblink-lock.json](https://github.com/user-attachments/files/23102108/nblink-lock.json)

NOTE: this also fixes some logic bugs in the processing of lockfile information. For example, previously we didn't install "noarch" packages because we assumed only packages which platform/subdir matches the current platform need to be installed. This is incorrect and has been fixed as part of this changeset.

## Type of Change

- [ ] Bugfix
- [x] Feature / enhancement
- [ ] CI / Documentation
- [ ] Maintenance

## Checklist


Klaim:
The integration tests are wired but not working yet because the test files are not correct yet (incoming).

Klaim:
This is reviewable, however some of the test lockfiles used in (micro)mamba integration are not correct yet (or there migth still be a few bugs, but I think it's just the lockfiles), which is why the CI is still red. The python tests wont change anyway and I think the overall set of c++ changes are reviewable.

Note that I moved all the integration tests lockfiles into a separate directory in the same way we did with `libmamba/tests/env_lockfiles`.

Klaim:
Added to description:
> NOTE: this also fixes some logic bugs in the processing of lockfile information. For example, previously we didn't install "noarch" packages because we assumed only packages which platform/subdir matches the current platform need to be installed. This is incorrect and has been fixed as part of this changeset.

libmamba/src/core/transaction.cpp
```diff
@@ -591,18 +598,34 @@ namespace mamba
...
+                        const auto& channel = channels.front();
+
+                        l_pkg.channel = channel.id();
+
+                        if (l_pkg.package_url.empty())
+                        {
+                            l_pkg.package_url = l_pkg.url_for_channel(
+                                channel.url().str(Credentials::Show)
+                            );
+                        }
```
JohanMabille, line 625: I know this was already here, but this seems very weird to me. There should be a single "source" for the channel. (This should not be addressed in this PR though).

Also why  do we need to do this instead of `l_pkg.channel = pkg.channel` ? Again, this can be investigated / addressed in a dedicated PR. We should probably open a issue to track this.

PR #4098, opened 2025-11-20
Add support for repodata v3 conditionals

### LINKED ISSUE #4089: Support for repodata and MatchSpec for conditional / extra / flags
Hi All,

### Background

There is a draft [CEP](https://github.com/conda/ceps/pull/111) to add support for the next evolution of Repodata  and MatchSpec. This proposal makes the following changes:

1. Extra (optional) dependency sets that can be activated by the user or other packages
2. Conditional dependencies to be activated on a specific platform, or when other package versions are also pulled in (for example requiring importlib_metadata; if python <3.10)
3. Flags to make it easier to select variants confidently (instead of globbing the build string)

### Why this is needed

Although this would be a nice update for the conda ecosystem in general, we are also interested in this for the conda CLI for better integration with the PyPI ecosystem. Having support for at least conditional dependencies is a prerequisite to adding wheels to repodata. We discussed more about that in our recent [roadmap post](https://conda.org/blog/conda-roadmap-q4-2025#better-integration-with-the-pypi-ecosystem).

While we work towards getting the CEP across the finish line, it would be great to start to support this in mamba as well. Is mamba interested in supporting these features so that we could make use of them in the conda-libmamba-solver?

### Scope of Changes

From my naive review of the codebase, the main changes needed would be in:

- `libmamba/include/mamba/specs/repo_data.hpp` - Add `extras` and `flags` fields
- `libmamba/src/solver/libsolv/helpers.cpp` - Parse conditional syntax and evaluate conditions
- `libmamba/src/specs/match_spec.cpp` - Support extras and flag syntax in MatchSpec
- Solver integration - Handle conditional deps and extras during resolution

danyeaw:
# Description

Partially closes #4089 by adding conditional support based on the draft CEP. This will need to be updated once we agree on the syntax, if we use the bracket style of passing in a conditional key it will require less parsing for this implementation.

I do not have experience with the mamba code base - this was my first time contributing and setting up a development environment. I also only limited experience with C++ - so all feedback is greatly appreciated.

This PR was assisted by a LLM in order to help gauge the amount of work this would be. The initial prompt was "What changes would be needed to support repodata v3, similar to how Rattler supports the implementation?" Follow-up context using the following steps:

1. Create `MatchSpecCondition` data structure (header and implementation) with tests
2. Update `MatchSpec` to include condition field and parsing
3. Update repodata parsing to handle conditional dependencies (parse only, no evaluation)
4. Integrate conditional dependencies into solver with condition evaluation against pool

A follow-up change could also add support for extras and flags.

## Type of Change

- [ ] Bugfix
- [X] Feature / enhancement
- [ ] CI / Documentation
- [ ] Maintenance

## Checklist

- [X] I have commented my code, particularly in hard-to-understand areas
- [X] My changes generate no new warnings
- [X] I have run `pre-commit run --all` locally in the source folder and confirmed that there are no linter errors.
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] New and existing tests pass locally with my changes

### DIFF
--- a/libmamba/tests/src/specs/test_match_spec.cpp
+++ b/libmamba/tests/src/specs/test_match_spec.cpp
@@ -5,6 +5,7 @@
 // The full license is in the file LICENSE, distributed with this software.
 
 #include <fstream>
+#include <unordered_set>
 
 #include <catch2/catch_all.hpp>
 #include <nlohmann/json.hpp>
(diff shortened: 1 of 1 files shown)


danyeaw:
Hey @baszalmstra, since you are very familiar with the Rattler implementation, would you have a few minutes to review this?

libmamba/include/mamba/specs/match_spec_condition.hpp
```diff
@@ -0,0 +1,173 @@
...
+         */
+        struct Or
+        {
+            std::unique_ptr<MatchSpecCondition> left;
+            std::unique_ptr<MatchSpecCondition> right;
+
+            [[nodiscard]] auto operator==(const Or& other) const -> bool;
+        };
+
+        using variant_type = std::variant<MatchSpecCondition_, std::unique_ptr<And>, std::unique_ptr<Or>>;
```
baszalmstra: Dont think these need to be unique_ptrs.

baszalmstra:
Wow, this is quite the PR! A thorough review is definitely required!

I checked the codebase a little bit, and I think this method is fundamentally flawed. Although it will handle cases with virtual packages because they are known upfront, it will not work with version checks. E.g. `numpy >=2; if python >=3.12` & `numpy <2; if python <3.12`.  Or even just `numpy; if foobar`, the solution might not even contain `foobar` even though it is available in the pool. These cases can only be solved at solve time. And they are used plenty in the Python ecosystem. But I'm sure you are aware of this. 

I think the code around checking these "static" virtual packages could also be improved a lot. There is a custom check in there now but you could also just match the condition against prebuild virtual packages. 

On the C++ front there are also definitely some things to improve. Although my C++ is very Rusty (pun intended), I think the use of raw pointers and unique_ptrs, could use some tweaking. There were multiple `= default` implementations added in places that werent really touched, and in some functions it was removed where I think it isnt necessary. 

Regardless, libsolv actually _does_ already support conditional dependencies, and I think it's just a matter of wiring them up properly. Rattler also has a libsolv implementation where I am experimenting with this, see [this pr](https://github.com/conda/rattler/pull/1836).

If I were a maintainer of mamba (which, to be clear, I am not), I would first try to split this PR up into more individualized components before merging. Start with the parsing, potentially make conditionals optional initially, and then work on integrating the solver. I would definitely not merge this PR with the current functionality because it contains a lot of unhandled edge cases.

danyeaw:
Hi @baszalmstra, thanks so much for taking a look at this and the feedback, I really appreciate it!

I didn't realize that libsolv had conditional dependency support through REL_COND, REL_AND, and REL_OR operators, that's amazing and does simplify the work at parse-time we have to do which is amazing.

I was mostly using this PR for discovery, and I'll document what I learned here: https://github.com/conda/conda-planning/issues/106. I really like the 3 step breakdown of the work that you proposed. It looks like a 4th step is also needed to set `ENABLE_COMPLEX_DEPS=ON` in the libsolv-feedstock recipe.

PR #4101, opened 2025-11-24
fix: Resolve ca-certificates installed in the local environment

jjerphan:
# Description

Fix https://github.com/conda-forge/mamba-feedstock/issues/362.

## Type of Change

- [x] Bugfix
- [ ] Feature / enhancement
- [ ] CI / Documentation
- [ ] Maintenance

## Checklist

### DIFF
--- a/libmamba/src/download/downloader.cpp
+++ b/libmamba/src/download/downloader.cpp
@@ -85,32 +85,47 @@ namespace mamba::download
                 // from `conda-forge::ca-certificates` and the system CA certificates.
                 else if (remote_fetch_params.ssl_verify == "<system>")
                 {
+                    // See the location of the CA certificates as distributed by
+                    // `conda-forge::ca-certificates`:
+                    // https://github.com/conda-forge/ca-certificates-feedstock/blob/main/recipe/meta.yaml#L25-L29
+                    const fs::u8path prefix_relative_conda_cert_path = (util::on_win ? fs::u8path("Library") / "ssl" / "cacert.pem" : fs::u8path("ssl") / "cert.pem");
+
                     const fs::u8path executable_path = get_self_exe_path();
-                    // Find the supposed environment prefix.
+
+                    // Find the environment prefix using the path of the `mamba` or `micromamba`
+                    // executable (we cannot assume the existence of an environment variable or
+                    // etc.).
+                    //
                     // `mamba` or `micromamba` is installed at:
+                    //
                     //    - `${PREFIX}/bin/{mamba,micromamba}` on Unix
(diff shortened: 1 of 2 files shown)


libmamba/src/download/downloader.cpp
```diff
@@ -84,19 +85,41 @@ namespace mamba::download
...
-                    fs::u8path root_prefix = detail::get_root_prefix();
-                    fs::u8path env_prefix_conda_cert = root_prefix / "ssl" / "cacert.pem";
-
-                    LOG_INFO << "Checking for CA certificates at the root prefix: "
+                    const fs::u8path executable_path = get_self_exe_path();
+                    // Find the supposed environment prefix.
+                    // `mamba` or `micromamba` is installed at:
+                    //    - `${PREFIX}/bin/{mamba,micromamba}` on Unix
+                    //    - `${PREFIX}/Library/bin/{mamba,micromamba}.exe` on Windows
+                    const fs::u8path env_prefix
```
Klaim, line 104: This whole block is extracted from `get_root_prefix` right? https://github.com/mamba-org/mamba/blob/6f11ca27b4cf4987ee86ffd0174954784c114022/libmamba/src/api/configuration.cpp#L699-L706

I would prefer if you factorize that in a function called here and in `get_root_prefix`.

Also, I think this is a bit fragile: is that supposed to work with micromamba which can be installed in any place (my micromamba never has `../bin` as an example) and doesnt have a libmamba install?

Isnt there an env variable already for the env prefix? Or we really want the one from the executable here?

I see we later check the existence of this path so maybe it's not important and it's Ok with micromamba, just asking to be sure. :)

jjerphan: > Also, I think this is a bit fragile: is that supposed to work with micromamba which can be installed in any place (my micromamba never has `../bin` as an example) and doesnt have a libmamba install?

This is only meant to support a new use-case (see https://github.com/conda-forge/mamba-feedstock/issues/362#issuecomment-3528420719).

> Isnt there an env variable already for the env prefix? Or we really want the one from the executable here?

I do not think we can expect that an environment variable is set for the use-case given above.

jjerphan: I thought of factorizing the few lines, which we can do yes, but I am not sure that it is really worth it given there's a few of them and given contextual comments are different.

Klaim: I didnt spot the difference at first, basically the starting point being dll or exe, even if it's the same logic.

I think I would put these in clearly named functions, personally, but it's not exactly "necessary", just helps me reduce the noise. So  ok let's keep it that way for now👍🏽

Klaim (approved):

PR #4071, opened 2025-10-01
Fix empty depends/constrains when installing explicit spec files

benmoss:
# Description

This PR fixes a bug where empty dependencies and constraints were not properly handled when installing packages from explicit spec files. This fixes #4070 and #4052 

## Type of Change

- [x] Bugfix
- [ ] Feature / enhancement
- [ ] CI / Documentation
- [ ] Maintenance

## Checklist

- [x] My code follows the general style and conventions of the codebase, ensuring consistency
- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [x] I have run `pre-commit run --all` locally in the source folder and confirmed that there are no linter errors.
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing tests pass locally with my changes

PR #4110, opened 2025-11-29
fix: Correct `repodata_record.json` metadata for installations using explicit specifications

### LINKED ISSUE #4095: Incorrect `repodata_record.json` precedence for explicit installs
This follows up on https://github.com/mamba-org/mamba/issues/4052 to indicate some problems that still remain.

### Summary

Since v2.1.1, libmamba switched in https://github.com/mamba-org/mamba/pull/3901 to "repodata-first" when constructing `info/repodata_record.json` (intended to honor channel repodata patches over `info/index.json`). However, in "explicit installs", the repodata object is a [URL-derived stub](https://github.com/mamba-org/mamba/blob/cd0e161f18c494d7b462deb9bf56901da187acb0/libmamba/src/specs/package_info.cpp#L193), not real channel repodata. Making that stub authoritative caused `repodata_record.json` to be written with incomplete/zeroed fields in v2.1.1 through v2.3.2.

In the v2.3.3 release, https://github.com/mamba-org/mamba/pull/4071 attempted to mitigate this by removing empty `depends`/`constrains` from the repodata object so they can be back-filled from `index.json`. However, this is applied unconditionally, based only on emptiness, with no check that the source actually came from an explicit install. As a result:

1. Channel repodata patches that intentionally set `depends`/`constrains` to the empty list cannot be represented in the local `repodata_record.json`: the code erases empty lists and refills from `index.json`, silently undoing the patch.
2. Some fields in explicit installs remain wrong in v2.3.3 (`build_number`, `license`, `timestamp`, and `track_features`), keeping the cache divergent from the best-known info.

This behavior corrupts the local cache metadata and propagates to tools that rely on `repodata_record.json` (e.g., conda-lock). Theoretically, this has the risk of leading to incomplete or incorrect lockfiles as in https://github.com/conda/conda-lock/pull/862 ([summary](https://github.com/maresb/conda-lock/blob/handle-bad-repodata2/tests/test-corrupt-repodata/README.md)). Thankfully, experimentation shows that https://github.com/mamba-org/mamba/pull/4071 was sufficient to resolve a serious bug involving disappearing dependencies, so this issue is far less critical than https://github.com/mamba-org/mamba/issues/4052.

### Expected behavior

* **Normal channel installs:** `repodata_record.json` should mirror the channel record actually used by the solver, including patches, even when a patch intentionally sets `depends`/`constrains` to `[]`. No back-fill from `index.json` should override an explicit channel decision.
* **Explicit installs:** writing `repodata_record.json` should not persist skeletal/placeholder values that zero out fields relative to either channel or artifact metadata.

---

### Concrete example (PyYAML)

As a good baseline, v2.1.0 wrote a more complete `repodata_record.json` for this package:

```json
{
    "arch": null,
    "build": "pyh7db6752_0",
    "build_number": 0,
    "build_string": "pyh7db6752_0",
    "channel": "conda-forge",
    "constrains": [],
    "depends": [],
    "fn": "pyyaml-6.0.3-pyh7db6752_0.conda",
    "license": "",
    "license_family": "MIT",
    "md5": "b12f41c0d7fb5ab81709fcc86579688f",
    "name": "pyyaml",
    "noarch": "python",
    "platform": null,
    "size": 45223,
    "subdir": "noarch",
    "timestamp": 0,
    "track_features": "",
    "url": "https://conda.anaconda.org/conda-forge/noarch/pyyaml-6.0.3-pyh7db6752_0.conda",
    "version": "6.0.3"
}
```

Since v2.3.3, here is the result for the same artifact after an explicit install:

```json
{
    "arch": null,
    "build": "pyh7db6752_0",
    "build_number": 0,
    "build_string": "pyh7db6752_0",
    "channel": "conda-forge",
    "depends": [
        "python >=3.10.*",
        "yaml"
    ],
    "fn": "pyyaml-6.0.3-pyh7db6752_0.conda",
    "license": "",
    "license_family": "MIT",
    "md5": "b12f41c0d7fb5ab81709fcc86579688f",
    "name": "pyyaml",
    "noarch": "python",
    "platform": null,
    "size": 45223,
    "subdir": "noarch",
    "timestamp": 0,
    "track_features": "",
    "url": "https://conda.anaconda.org/conda-forge/noarch/pyyaml-6.0.3-pyh7db6752_0.conda",
    "version": "6.0.3"
}
```

Diff:

```diff
@@ -4,13 +4,12 @@
     "build_number": 0,
     "build_string": "pyh7db6752_0",
     "channel": "conda-forge",
-    "constrains": [],
     "depends": [
         "python >=3.10.*",
         "yaml"
     ],
     "fn": "pyyaml-6.0.3-pyh7db6752_0.conda",
-    "license": "MIT",
+    "license": "",
     "license_family": "MIT",
     "md5": "b12f41c0d7fb5ab81709fcc86579688f",
     "name": "pyyaml",
@@ -18,8 +17,8 @@
     "platform": null,
     "size": 45223,
     "subdir": "noarch",
-    "timestamp": 1758891992558,
-    "track_features": "pyyaml_no_compile",
+    "timestamp": 0,
+    "track_features": "",
     "url": "https://conda.anaconda.org/conda-forge/noarch/pyyaml-6.0.3-pyh7db6752_0.conda",
     "version": "6.0.3"
 }
```

Even after https://github.com/mamba-org/mamba/pull/4071, `license`, `timestamp`, and `track_features` are still clobbered in explicit installs, and `constrains` disappears. While not demonstrated by this diff, `build_number` is also set to zero.

maresb:
## ⚠️ Important Disclosure

**I do not know C++.** This PR was extremely heavily LLM-assisted (Claude in Cursor). As a result, **please review this PR skeptically and critically**. I have spent significantly more time than typical attempting to verify correctness to the extent I'm able.

# Description

This PR fixes [#4095](https://github.com/mamba-org/mamba/issues/4095): incorrect `repodata_record.json` precedence for explicit installs.

## Background

Since v2.1.1, libmamba switched to "repodata-first" when constructing `info/repodata_record.json` (intended to honor channel repodata patches over `info/index.json`). However, for **explicit installs** (URL-based), the "repodata" object is a URL-derived stub containing placeholder/zero values—not real channel repodata. Making that stub authoritative caused `repodata_record.json` to be written with incomplete/zeroed fields.

**Affected versions:**
- **v2.1.1–v2.3.2**: Full corruption (all stub fields written)
- **v2.3.3+**: Partial fix via #4071 (`depends`/`constrains` fixed, but `license`, `timestamp`, `build_number`, `track_features` still corrupted)

The v2.3.3 partial fix (#4071) removed empty `depends`/`constrains` arrays so they could be backfilled from `index.json`. However, this approach has two problems:

1. **Channel patches undone**: Patches that intentionally set `depends`/`constrains` to `[]` are silently overwritten by `index.json` values
2. **Other fields still wrong**: `build_number`, `license`, `timestamp`, and `track_features` remained corrupted

## Solution

This PR leverages the existing `defaulted_keys` field (originally introduced in PR #1120 for signature verification, but no longer populated since the 2023 libsolv wrapper refactor) to distinguish URL-derived stub values from authoritative solver/channel values:

- **URL-derived packages** (`from_url()`): `defaulted_keys` lists stub field names (e.g., `["_initialized", "license", "timestamp", ...]`). These fields are erased before merging with `index.json`, allowing correct values to be filled in.

- **Solver-derived packages** (`make_package_info()`): `defaulted_keys = ["_initialized"]` only. The sentinel signals "trust ALL fields"—nothing is erased, preserving channel patches including intentionally empty arrays.

The `_initialized` sentinel enables **fail-hard verification**: `write_repodata_record()` throws `std::logic_error` if it's missing, catching any code path that fails to properly initialize `defaulted_keys`.

### All PackageInfo creation paths covered

Every code path that creates a `PackageInfo` now sets `defaulted_keys`:

| Code path | `defaulted_keys` value | Rationale |
|-----------|----------------------|-----------|
| `from_url()` | `["_initialized", "license", "timestamp", ...]` | URL-derived stubs; replace with `index.json` |
| `make_package_info()` (solver) | `["_initialized"]` | Trust all solver/channel-patched fields |
| Channel repodata (JSON) | `["_initialized"]` | Trust all channel repodata fields |
| Lockfile conda format | Inherited from `from_url()`, with `depends`/`constrains` removed when sha256 present | Trust lockfile deps when metadata is trustworthy |
| Lockfile mambajs format | `["_initialized", "build_number", "license", ...]` | Stub fields not in mambajs format |
| History URL entries | `["_initialized", "license", "timestamp", ...]` | Skeletal metadata from history |

### libsolv round-trip preservation

`defaulted_keys` is serialized as a comma-separated string in the `SOLVABLE_KEYWORDS` libsolv field, allowing it to survive the round-trip through the solver database. This is critical because URL-derived packages (e.g., URLs specified on the command line) pass through the solver before installation. Without this, `defaulted_keys` would be lost and stub values would be preserved instead of being replaced by `index.json` data.

### Consistency improvements

- `depends` and `constrains` are always present as arrays in `repodata_record.json` (matching conda behavior)
- `track_features` is omitted when empty (matching conda behavior)
- `md5` and `sha256` are computed from the tarball when missing

### Note on cache healing

An earlier version of this PR included cache healing to detect and repair corrupted caches from v2.1.1+. This has been removed because #4163 (merged into `main`) redesigns the package cache layout, effectively busting all old caches. Users upgrading will re-download packages into the new cache layout, so all metadata in the new locations will be correct from the start.

## Review Recommendation

**Please review commit-by-commit.** Each commit has an extended commit message with:
- Detailed explanation of the changes
- Test status (assertions passed/failed)
- Rationale for design decisions

The commits follow TDD (Test-Driven Development) with alternating RED (failing tests) and GREEN (fix) commits:

| # | Type | Description |
|---|------|-------------|
| 01 | RED | Tests for `defaulted_keys` population in `from_url()` |
| 02 | GREEN | Populate `defaulted_keys` in `from_url()` and `make_package_info()` |
| 03 | RED | Tests for URL-derived metadata and channel patch preservation |
| 04 | GREEN | Use `defaulted_keys` in `write_repodata_record()` and `constructor.cpp` |
| 05 | RED | Tests for consistent field presence and checksums |
| 06 | GREEN | Ensure consistent field presence and compute missing checksums |
| 07 | RED | Tests for lockfile packages having `_initialized` in `defaulted_keys` |
| 08 | GREEN | Set `defaulted_keys` in lockfile parsing for both conda and mambajs formats |
| 09 | RED | Tests for `defaulted_keys` storage in libsolv solvable wrapper |
| 10 | GREEN | Add `defaulted_keys` storage to solvable wrapper via `SOLVABLE_KEYWORDS` |
| 11 | RED | Tests for `defaulted_keys` preservation through solver round-trip |
| 12 | GREEN | Preserve `defaulted_keys` through libsolv Database add/retrieve cycle |
| 13 | RED | Test for `read_history_url_entry()` `defaulted_keys` population |
| 14 | GREEN | Populate `defaulted_keys` in `read_history_url_entry()` |
| 15 | RED | Test for trusted conda lockfile `depends` not in `defaulted_keys` |
| 16 | GREEN | Remove `depends`/`constrains` from `defaulted_keys` for trusted conda lockfiles |

Commits 01–04 implement the core fix. Commits 05–06 add consistency improvements. Commits 07–16 ensure all `PackageInfo` creation paths are covered, including lockfile parsing, libsolv solver round-trip, history entries, and trusted lockfile handling.

## Type of Change

- [x] Bugfix

## Checklist


maresb:
238e7ea is green up to a flaky connection error

Still todo: more thoroughly vet the last 6 commits, and revise the commit description accordingly.

Brief summary: I'm enabling libsolv to propagate `defaulted_keys` because there are situations where the solver receives explicit-style URLs, e.g. specifying a URL on the command line, and so without this (or some other mechanism) we still risk writing bad stub values.

maresb:
@jjerphan, I've been working on fixing the subtle cache corruption issue in https://github.com/mamba-org/mamba/issues/4095. This has been plaguing conda-lock, and leads to a really complex chain of problems described in https://github.com/conda/conda-lock/issues/896.

In the meantime, I notice that in your recent changes, namely #4163, the caching changes for the next release are expected to bust the whole cache:

> The cache layout has changed. Existing caches under the old layout are no longer used. Users will have re-download packages after upgrading.

This is a great opportunity to achieve a definitive fix for all the mamba-related corruption issues in conda-lock: if we manage to merge the correctness fix of the current PR before the next release, then we will know that that all the metadata in the new locations is consistent! :tada:

I've had this PR in draft for quite a while because I want to understand all the individual changes myself before requesting review. I expect to get to this point later today or tomorrow.

In order to take advantage of this opportunity, I was hoping you could answer these questions:

1. Do you have a timeline for the upcoming release?
2. Would you be willing to consider including this fix with the upcoming release?
3. If so, how can I be most helpful to ensure this gets merged and doesn't create any undue burdens for you?

jjerphan:
Thank you for this contribution, @maresb.

> Do you have a timeline for the upcoming release?

We don't have a timeline, although the next version to be released is likely going to be 2.6.0 since `main` now support sharded repodata.

> Would you be willing to consider including this fix with the upcoming release?

I am not against having it part of the next release, and I think it would be better if it is part of it.

> If so, how can I be most helpful to ensure this gets merged and doesn't create any undue burdens for you?

I do not think we can prevent any undue burden, but having a properly documented PR like what you came up with using atomic commits (assuming comments are actually exact) seems to be the best pathway to me.

libmamba/src/specs/package_info.cpp
jjerphan, line 1: Can you store those values for `defaulted_keys` as `constexpr` to avoid duplications?

maresb: Good call @jjerphan, done.

maresb:
Failing test is timeout on `test_parsable_env_history_with_metadata[True]`, and looks to me like a flake.

maresb:
Hi @jjerphan, I see that I missed 2.6.0rc0. Is there anything I can still do to get this into the upcoming release, so that the new sharded repodata remains consistent?

jjerphan:
Hello, we can stil have it as part of 2.6.0. I'll do a review.

libmamba/ext/solv-cpp/src/solvable.cpp
```diff
@@ -518,4 +518,61 @@ namespace solv
         // (Ab)using meaningless key
         ::solvable_set_num(raw(), SOLVABLE_INSTALLSTATUS, static_cast<Num>(val));
     }
+
+    auto ObjSolvableViewConst::defaulted_keys() const -> std::vector<std::string>
+    {
+        // Retrieve the comma-separated string from SOLVABLE_KEYWORDS
+        // (repurposed for conda-specific metadata since it's not used in the conda ecosystem)
```
jjerphan: Does this encoding allow extending usages of `SOLVABLE_KEYWORDS` for other applications than transferring `default_keys` without impacting this use-case?

libmamba/ext/solv-cpp/tests/src/test_solvable.cpp
```diff
@@ -283,5 +283,77 @@ namespace
...
+            {
+                // PURPOSE: Verify that a single-element list works correctly
+                // This is the common case for channel-derived packages
+                solv.set_defaulted_keys({ "_initialized" });
+                repo.internalize();
+
+                auto retrieved = solv.defaulted_keys();
+                REQUIRE(retrieved.size() == 1);
+                REQUIRE(retrieved[0] == "_initialized");
+            }
```
jjerphan, line 355: I think having another test with at least another values for keys but `"_initialized"` would be useful to specify the behavior (e.g. would `"_initialized"` be retrieved or not in this case?).

```cpp
            SECTION("Missing `_initialized` in `defaulted_keys`")
            {
                solv.set_defaulted_keys({ "other_value" });
                repo.internalize();

                auto retrieved = solv.defaulted_keys();
                REQUIRE(retrieved.size() == 1);
                REQUIRE_FALSE(retrieved[0] == "other_value");
            }
```

libmamba/include/mamba/specs/package_info.hpp
```diff
@@ -19,6 +19,28 @@
...
+        inline constexpr const char* fn = "fn";
+        inline constexpr const char* license = "license";
+        inline constexpr const char* timestamp = "timestamp";
+        inline constexpr const char* track_features = "track_features";
+        inline constexpr const char* depends = "depends";
+        inline constexpr const char* constrains = "constrains";
+        inline constexpr const char* md5 = "md5";
+        inline constexpr const char* sha256 = "sha256";
+        inline constexpr const char* size = "size";
+        inline constexpr const char* url = "url";
```
jjerphan: `inline constexpr std::string_view` would be more C++ idiomatic and also convenient for manipulating values.

libmamba/include/mamba/specs/package_info.hpp
```diff
@@ -52,8 +74,10 @@ namespace mamba::specs
         std::vector<std::string> track_features = {};
         std::vector<std::string> dependencies = {};
         std::vector<std::string> constrains = {};
-        // WARNING Be aware that `defaulted_keys` value, if set later,
-        // is not passed when going through `make_package_info` from libsolv
+        // Tracks which fields have stub/default values that should be replaced by index.json.
+        // Set by from_url() with field names, and by make_package_info() with {"_initialized"}.
+        // The "_initialized" sentinel indicates the PackageInfo was properly constructed.
+        // See GitHub issue #4095.
```
jjerphan: Completing this suggestion, could you also mention the different cases which exists (from explicit specification in lock files / "URL-derived packages" (?), from the conda-package's `index.json` and from the `repodata` records) and to refer to them consistently?

```suggestion
        // `defaulted_keys` tracks fields whose stub/default should be replaced
        // by values in `index.json` (metadata file within the conda-package).
        //
        // It is set by `PackageInfo::from_url` with field names, and by
        // `make_package_info` with `{"_initialized"}`.
        // The `"_initialized"` sentinel indicates the `PackageInfo`
        // was properly constructed.
        //
        // See GitHub issue #4095.
```

libmamba/src/core/env_lockfile_conda.cpp
```diff
@@ -73,6 +75,10 @@ namespace mamba
                 package.info.channel = maybe_parsed_info->channel;
                 package.info.build_string = maybe_parsed_info->build_string;
                 package.info.platform = maybe_parsed_info->platform;
+                // Copy defaulted_keys for fail-hard verification in write_repodata_record().
+                // The "_initialized" sentinel proves this PackageInfo was properly constructed.
+                // See issue #4095.
```
jjerphan:
```suggestion
```

libmamba/src/specs/package_info.cpp
jjerphan: I meant that some lists for initializing `defaulted_keys` are identical (_e.g._ the ones at l.160, l.187, and l.214).

Could we avoid this duplication?

micromamba/tests/test_constructor.py
```diff
@@ -83,3 +94,381 @@ def test_extract_pkgs(self):
...
+        os.makedirs(cls.pkgs_dir, exist_ok=True)
+
+        # Save original env vars
+        cls.orig_root_prefix = os.environ.get("MAMBA_ROOT_PREFIX")
+        cls.orig_prefix = os.environ.get("CONDA_PREFIX")
+
+        # Set test env vars
+        os.environ["MAMBA_ROOT_PREFIX"] = cls.root_prefix
+        os.environ["CONDA_PREFIX"] = cls.root_prefix
+
```
jjerphan: Nit: Could [`monkeypatching` environment variables with `pytest`](https://docs.pytest.org/en/stable/how-to/monkeypatch.html) be used here?

libmambapy/bindings/specs.cpp
```diff
@@ -726,7 +726,13 @@ namespace mambapy
                 py::arg("track_features") = decltype(PackageInfo::track_features)(),
                 py::arg("depends") = decltype(PackageInfo::dependencies)(),
                 py::arg("constrains") = decltype(PackageInfo::constrains)(),
-                py::arg("defaulted_keys") = decltype(PackageInfo::defaulted_keys)(),
+                // Default to ["_initialized"] for backward compatibility.
+                // Before issue #4095 fix, make_package_info() always set ["_initialized"],
+                // so all PackageInfo from the solver had it. Users creating PackageInfo
+                // manually should get the same default behavior: "trust all fields".
+                // Users who want index.json to fill in gaps should explicitly set
+                // defaulted_keys to include the stub field names.
```
jjerphan: I am not sure that this would be a use-case (arguably `index.json`'s authority lacks precedence over rapodata records'?), also we would need to make it possible (a YAGNI?). I think this part could be omitted.

```suggestion
```

libmambapy/bindings/specs.cpp
```diff
@@ -726,7 +726,13 @@ namespace mambapy
                 py::arg("track_features") = decltype(PackageInfo::track_features)(),
                 py::arg("depends") = decltype(PackageInfo::dependencies)(),
                 py::arg("constrains") = decltype(PackageInfo::constrains)(),
-                py::arg("defaulted_keys") = decltype(PackageInfo::defaulted_keys)(),
+                // Default to ["_initialized"] for backward compatibility.
+                // Before issue #4095 fix, make_package_info() always set ["_initialized"],
+                // so all PackageInfo from the solver had it. Users creating PackageInfo
+                // manually should get the same default behavior: "trust all fields".
```
jjerphan: This should be moved from the bindings to the definition of `PackageInfo::default_keys` if it contains any new kind of information (I would not duplicate information, but only mention to report to existing comments, such as the getter and setter of `PackageInfo::default_keys`).

libmamba/src/core/package_fetcher.cpp
```diff
@@ -452,29 +458,110 @@ namespace mamba
...
+                       m_package_info.defaulted_keys.end(),
+                       specs::defaulted_key::initialized
+                   )
+                   != m_package_info.defaulted_keys.end();
+        };
+        if (!contains_initialized())
         {
-            repodata_record.erase("depends");
+            throw std::logic_error(
+                "PackageInfo missing _initialized sentinel in defaulted_keys. "
```
jjerphan: Nit: Code formatting brings information, in error message, but more generally in all prose, including comments'.
```suggestion
                "`PackageInfo` missing `_initialized` sentinel in `defaulted_keys`. "
```

jjerphan (changes requested):
Thank you, @maresb!

A few request, comments, and suggestions.

Once they are addressed, I think we could introduce this fix in 2.6.0.rc1, and adapt it further if required.

I do not have faulty installatio I could test this fix on though. Do you?

libmamba/ext/solv-cpp/src/solvable.cpp
maresb: Good question. The current encoding assumes exclusive ownership of `SOLVABLE_KEYWORDS` — it stores `defaulted_keys` as a plain comma-separated list with no namespace prefix. This means other data couldn't coexist in the same field without a format change.

In practice this is fine: `SOLVABLE_KEYWORDS` is unused in the conda ecosystem (libsolv maps it from RPM's `Keywords` tag, which has no conda equivalent), so there's no collision risk. The encoding is also purely internal — it only lives in the in-memory libsolv pool and `.solv` cache files, never in user-facing artifacts — so we can evolve the format later if a new use case arises.

I've added a comment documenting this assumption at the encoding site in 22976a93.

libmamba/ext/solv-cpp/tests/src/test_solvable.cpp
maresb: Good idea to test this case — added in 6a2dca02.

One note on the assertion: the solvable layer is a dumb key-value store; it faithfully round-trips whatever strings it receives, so `{ "other_value" }` comes back as `{ "other_value" }`. The `_initialized` sentinel is not enforced at this layer — validation happens in `write_repodata_record()`, which checks for `_initialized` presence and throws `std::logic_error` if it's missing.

So the test asserts round-trip fidelity (`REQUIRE(retrieved[0] == "other_value")`) rather than rejection, with a comment explaining the layered design. I also added a multi-value test matching the `from_url()` pattern and a comment noting the comma-encoding limitation.

libmamba/include/mamba/specs/package_info.hpp
maresb: Done — all 18 constants changed to `inline constexpr std::string_view` in 40e76fb3.

libmamba/include/mamba/specs/package_info.hpp
maresb: Good point — rewrote the doc comment in 72f8ea31 to cover all four creation contexts (URL-derived, repodata-derived, lockfile-derived, history-derived) with consistent terminology and the three state semantics (empty = invalid, `{"_initialized"}` = trust all fields, `{"_initialized", ...}` = listed fields are stubs).

Also consolidated the duplicated design explanations in `helpers.cpp`, `package_fetcher.cpp`, `solvable.hpp`, and `bindings/specs.cpp` to reference this canonical doc instead (dc20b0c6).

libmamba/src/core/env_lockfile_conda.cpp
maresb: Done — removed the 3-line narrating comment. Also did a sweep across all files in this PR to remove similar narrating comments (15 total) and add backtick formatting to code identifiers in surviving comments and error messages (ba000c68).

libmamba/src/specs/package_info.cpp
maresb: Done — defined shared `constexpr std::array` constants for the conda URL and wheel/tar.gz patterns in `package_info.cpp`'s anonymous namespace. The three identical wheel/tar.gz lists now reference a single `wheel_targz_defaulted_keys` array (00ec5361).

micromamba/tests/test_constructor.py
maresb: Done — refactored `TestURLDerivedMetadata` and `TestChannelPatchPreservation` to use `pytest.MonkeyPatch.context()` + `tmp_path_factory` instead of manual `os.environ` save/restore and `tempfile.mkdtemp` + `shutil.rmtree` (b0455e56). Left `TestInstall` untouched since it's pre-existing code.

libmambapy/bindings/specs.cpp
maresb: Done — removed the speculative comment. The remaining block is replaced with a brief 2-line reference to `PackageInfo::defaulted_keys` (261d55ed).

libmambapy/bindings/specs.cpp
maresb: Done — the canonical doc is now on `PackageInfo::defaulted_keys` in `package_info.hpp` (72f8ea31). The bindings comment, `helpers.cpp`, `solvable.hpp`, `package_fetcher.cpp`, and `package_info.cpp` all now reference that canonical doc instead of duplicating the design explanation (261d55ed, dc20b0c6).

libmamba/src/core/package_fetcher.cpp
maresb: Done — added backtick formatting to code identifiers (`PackageInfo`, `_initialized`, `defaulted_keys`, etc.) in the error message and in surviving comments across all files touched by this PR (ba000c68).

maresb:
Addressed all review feedback. Individual replies are on each comment thread. Additional commits beyond what was requested:

- `refactor: use pathlib in test_constructor.py` (4ada0da0) — converted `os.path`/`os.makedirs` to `pathlib.Path` in the two new test classes (code introduced in this PR).
- `docs: remove narrating comments in test_constructor.py` (43b37f59) — removed comments that restate what the next line does in the two new test classes (code introduced in this PR).
- `fix: add size to defaulted_keys for URL-derived and lockfile paths` (5bd523c0) — improves internal consistency; final correctness unaffected (post-processing already handled this).
- `fix: add url to defaulted_keys for history-derived packages` (d1905f64) — improves internal consistency; final correctness unaffected (history-derived packages don't reach `write_repodata_record()`).
- `fix: add subdir to defaulted_keys when URL lacks platform` (f349bacf) — actual bugfix: for `file://` URLs and wheel/tar.gz, `repodata_record.json` contained `subdir: ""` instead of the correct value from `index.json`.
- `test: comprehensive repodata_record.json check for file:// URL installs` (7ef5d0b2) — C++ test that verifies every field in the resulting `repodata_record.json` for a `file://` URL install (would have caught the `subdir` bug).

libmamba/include/mamba/specs/package_info.hpp
maresb: Good point — rewrote the doc comment in 72f8ea31 with all four creation contexts and consistent terminology. Here's the full picture:

**URL-derived** (`from_url()`): The URL and filename provide name, version, build_string (conda only), channel, subdir (conda only), filename, and package_url. Everything else is a struct default (0, "", {}) and listed in `defaulted_keys`:

- Conda URL: `_initialized`, `build_number`, `license`, `size`, `timestamp`, `track_features`, `depends`, `constrains`.
- Wheel/tar.gz URL: same + `build`, `build_string` (no build info in filename).
- Git URL: same + `version`, `channel`, `subdir`, `fn`, and optionally `name` (if no `#egg=`).

Fields NOT listed but handled correctly:

- `md5`/`sha256`: `to_json` only writes them when non-empty, so they're absent from the JSON and `index.json` fills them via `insert()`. Post-processing computes from tarball if still missing.
- `noarch`: default `No` → `to_json` skips it → `index.json` provides if applicable.
- When URL has a `#hash` suffix, the hash is set from the fragment and correctly NOT in defaulted_keys (trusted).

**Repodata-derived** (channel `repodata.json`, via the solver / `make_package_info()`): all fields are authoritative — including intentional channel patches that may differ from `index.json`. `defaulted_keys = {"_initialized"}` only (no stubs). Backward compat: old `.solv` cache files without stored `defaulted_keys` are treated as `{"_initialized"}` since they came from authoritative channel repodata.

**Lockfile-derived**:

- Conda v1 lockfile: starts with `from_url()` defaults, then the lockfile populates name, version, hashes, deps, and constrains. Trust logic: when `sha256` is present and `dependencies`/`constrains` are non-empty, those are removed from `defaulted_keys` (trusted as potentially reflecting repodata patches). When `sha256` is absent, they stay as stubs (lockfile may have been generated during v2.1.1–v2.3.2 bug window).
- Mambajs lockfile: `_initialized`, `build_number`, `license`, `size`, `timestamp`, `track_features`, `depends`, `constrains` — lockfile provides name, version, build, subdir, channel, hashes; the rest are stubs.

**History-derived** (`read_history_url_entry()`): only name, version, build_string, and channel are parsed from the history entry. Everything else is a stub: `_initialized`, `build_number`, `license`, `timestamp`, `track_features`, `depends`, `constrains`, `subdir`, `md5`, `sha256`, `size`, `fn`, `url`.

The doc also explains how `write_repodata_record()` uses `defaulted_keys` at extraction time: it erases the listed stub fields, then merges with the tarball's `index.json` (which only fills missing keys). This preserves repodata-derived values (channel patches) while replacing URL-derived stubs with the package's own metadata.

Also consolidated the duplicated design explanations in `helpers.cpp`, `package_fetcher.cpp`, `solvable.hpp`, and `bindings/specs.cpp` to reference this canonical doc instead (dc20b0c6).

maresb:
> I do not have faulty installation I could test this fix on though. Do you?

Just realized that I missed this. I will work on this now...

libmamba/include/mamba/specs/package_info.hpp
maresb: While doing a field-by-field audit of the `defaulted_keys` lists, I found that `subdir` was missing. For `file://` URLs the platform can't be parsed from the path, so `subdir` stays `""`. Without being listed in `defaulted_keys`, that empty string persists in `repodata_record.json` instead of being backfilled from `index.json`.

Fixed in f349bacf: conda URLs now conditionally add `subdir` when `platform` is empty, and wheel/tar.gz URLs always include it (they never parse a platform).

I also added a reproduction script and posted the results on [issue #4095](https://github.com/mamba-org/mamba/issues/4095#issuecomment-4209731816).

maresb:
Reproduced issue #4095 on 2.6.0.rc0 and verified the fix. Posted a self-contained reproduction script and results on [the issue](https://github.com/mamba-org/mamba/issues/4095#issuecomment-4209731816).

During the field-by-field audit prompted by https://github.com/mamba-org/mamba/pull/4110#discussion_r3053494048, I also found and fixed a `subdir` bug: for `file://` URLs and wheel/tar.gz packages, `repodata_record.json` contained `subdir: ""` instead of the correct value from `index.json` (e.g. `"noarch"`). Fixed in f349bacf.

micromamba/tests/test_constructor.py
maresb: The GitHub conversation threading is a bit weird:

> I meant that some lists for initializing defaulted_keys are identical (e.g. the ones at l.160, l.187, and l.214).
>
>Could we avoid this duplication?


I answered this above in https://github.com/mamba-org/mamba/pull/4110#discussion_r3053405856

maresb:
@jjerphan, thanks for the excellent review! I believe I've now thoroughly addressed everything. Since there are several commits, I've tied each commit either to one of your review comments or to https://github.com/mamba-org/mamba/pull/4110#issuecomment-4208629168 in hopes that it's easy to review one commit at a time. Ready for the next review round when you are.

jjerphan:
Please do not force push after review, this makes subsequent reviews much harder, especially for PRs with a lot of changes like this one.

maresb:
Apologies @jjerphan, that force-push from earlier just cleaned up a few commits that I made strictly post-review. The underlying reviewed history with SHAs remains completely unchanged. But on principle I'll avoid all force-pushes going forward.

jjerphan (approved):

libmamba/ext/solv-cpp/src/solvable.cpp
```diff
@@ -518,4 +518,63 @@ namespace solv
...
+        // `SOLVABLE_KEYWORDS` repurposed for conda-specific `defaulted_keys` storage.
+        const char* str = ::solvable_lookup_str(const_cast<::Solvable*>(raw()), SOLVABLE_KEYWORDS);
+        if (str == nullptr || str[0] == '\0')
+        {
+            return {};
+        }
+
+        std::vector<std::string> result;
+        std::string current;
+        for (const char* p = str; *p != '\0'; ++p)
```
Klaim: I would prefer an explicit [string split](https://cppreference.com/cpp/ranges/split_view) instead of a raw loop here. Both for clarity, less code and avoid potential bugs that are hard to spot at a glance in such code.

libmamba/ext/solv-cpp/src/solvable.cpp
```diff
@@ -518,4 +518,63 @@ namespace solv
...
+        // RPM's Keywords tag, which has no conda equivalent), so collision risk is nil.
+        if (keys.empty())
+        {
+            // Store empty string for empty list (libsolv's unset behavior is unreliable)
+            ::solvable_set_str(raw(), SOLVABLE_KEYWORDS, "");
+            return;
+        }
+
+        std::string serialized;
+        for (std::size_t i = 0; i < keys.size(); ++i)
```
Klaim: I would prefer that the strings be explicitly joined instead of a raw loop here. Maybe [using this](https://github.com/mamba-org/mamba/blob/f68450c18af142437c640cf25f3b43c553e7edc5/libmamba/include/mamba/util/string.hpp#L631) or [maybe this](https://en.cppreference.com/cpp/ranges/join_view).

Klaim (commented):
Thanks for the PR, I could only focus on the structure of it as I'm less familiar with how that feature should effectively work. My main issue is with raw loops that should be replaced by explicit, tested, and shorter algorithms (where it makes sense).

Note that I'm not 100% sure it's fine to have the name constants values defined in the header, but I guess as long as we dont offer an ABI compatibility it should be somewhat OK.

libmamba/src/core/package_fetcher.cpp
```diff
@@ -452,29 +458,89 @@ namespace mamba
...
+        // See `PackageInfo::defaulted_keys`. Issue #4095.
+        auto contains_initialized = [&]()
+        {
+            return std::find(
+                       m_package_info.defaulted_keys.begin(),
+                       m_package_info.defaulted_keys.end(),
+                       specs::defaulted_key::initialized
+                   )
+                   != m_package_info.defaulted_keys.end();
+        };
```
JohanMabille:
```suggestion
        const bool contains_initialized = std::ranges::find(m_package_info.defaulted_keys, specs::defaulted_key::initialized)
            != m_package_info.defaulted_keys.end();
```

libmamba/src/core/package_fetcher.cpp
```diff
@@ -452,29 +458,89 @@ namespace mamba
...
+        auto needs_md5 = !repodata_record.contains("md5") || !repodata_record["md5"].is_string()
+                         || repodata_record["md5"].get<std::string>().empty();
+        if (needs_md5)
+        {
+            repodata_record["md5"] = validation::md5sum(m_tarball_path);
+        }
+
+        auto needs_sha256 = !repodata_record.contains("sha256")
+                            || !repodata_record["sha256"].is_string()
+                            || repodata_record["sha256"].get<std::string>().empty();
```
JohanMabille: This is quite similar to the md5 related code and could be factorized in a lambda:

```cpp
auto needs_checksum = [&repodata_record](std::string_view arg)
{
    return !repodata_record.contains(arg) || !repodata_record[arg].is_string()
                         || repodata_record[arg].get<std::string>().empty();
}
```

And then use it for both md5 and sh256 directly in the `if` statement.

maresb:
Thanks everyone for the feedback!!!

I'm working on this right now...

JohanMabille:
Thanks for fixing this issue! And apologies for the long delayed review, we'll try to ship this fix in the next minor release (2.6.0)

libmamba/ext/solv-cpp/src/solvable.cpp
maresb: Replaced the hand-rolled split with `std::views::split`. Couldn't reach for `mamba::util::split` here because `solv-cpp` is a standalone subproject that doesn't depend on libmamba (dependency goes the other way). One subtlety: `std::views::split` keeps empty segments where the hand-rolled parser dropped them; the round-trip is still safe because no caller ever stores empty keys, and the existing `Empty defaulted_keys` test in `test_solvable.cpp` locks down the empty-list round-trip. See 876d33f7.

libmamba/ext/solv-cpp/src/solvable.cpp
maresb: Same subproject constraint as the last comment (`solv-cpp` doesn't depend on libmamba, so `mamba::util::join` is off the table), and `std::ranges::join_with_view` is C++23 only. I extracted a small `join_csv` helper into solvable.cpp's anonymous namespace so the call site is now a single expression — the explicit join you asked for, even if the helper's body remains a loop. See 876d33f7.

libmamba/src/core/package_fetcher.cpp
maresb: Applied exactly as suggested. See 45bab80f.

libmamba/src/core/package_fetcher.cpp
maresb: Factored out as `needs_checksum`, used inline in both `if` statements. See fcb11f60.

libmamba/ext/solv-cpp/src/solvable.cpp
maresb: Small follow-up: the first version of this change used `std::views::split`, but linuxbrew CI exposed a GCC 11 portability issue there, so I replaced it with a local `split_csv` helper in 950624a2. The call site is still a single explicit split rather than a raw loop; this just avoids the toolchain-specific failure.

JohanMabille (approved):

maresb:
Thanks for the review. The `solvable.cpp` raw loops are addressed in 876d33f7, with a GCC 11/linuxbrew portability follow-up in 950624a2, and I applied the same principle to an erase-remove idiom in `env_lockfile_conda.cpp` and a few test-file contains lambdas in 11a91956 and d413fa9d.

On the header-defined name constants: this follows the existing `inline constexpr` pattern in libmamba. A shift to ABI-friendly `extern const` + `.cpp` definitions would be broader than the current PR.

I believe that addresses the review points currently raised, but I’m happy to do another round if there’s anything I missed or anything you’d like tightened further.

maresb:
Regarding the CI failures, it looks like there's an issue with `repo.mamba.pm`:

```bash
$ curl -I https://repo.mamba.pm/conda-forge/noarch/repodata.json
curl: (7) Failed to connect to repo.mamba.pm port 443 after 16 ms: Could not connect to server
```

jjerphan:
Resolved thanks to @atrawog.

jjerphan:
Thank you, @maresb!

jjerphan:
This fix is present as part of 2.6.0.rc4 which is being released as:

- `conda-forge/label/micromamba_prerelease::micromamba==2.6.0.rc4`
- `conda-forge/label/mamba_prerelease::mamba==2.6.0.rc4`

micromamba can be updated to this RC with:

```
micromamba self-update -c conda-forge/label/micromamba_prerelease "micromamba==2.6.0.rc4"
```

Issue #342, opened 2020-06-21
Mamba does not handle pip interoperability


rtgdk:
Conda supports pip interoperability - https://docs.conda.io/projects/conda/en/latest/user-guide/configuration/pip-interoperability.html - to satisfy dependencies with pip packages and remove pip-installed packages when upgrading them with conda. This is set by `conda config --set pip_interop_enabled True`.

Steps to reproduce : 
1. Download miniconda and install mamba (I have 0.3.7 installed)
2. Install boto3==1.14.4 through pip - `pip install boto3==1.14.4`
3. At this time both mamba list and pip list shows boto3 as 1.14.4 version
4. Set the conda config for pip interoperability - `conda config --set pip_interop_enabled True` (mamba does not have config command till now)
5. Then downgrade boto3 through mamba to 1.13.21 version - `mamba install -y boto3==1.13.21`
6. After this, mamba list and pip list still shows 1.14.4 but importing from python imports 1.13.21 version.
7. Site-packages shows both versions. 
```
> ls lib/python3.7/site-packages/ | grep boto3
boto3
boto3-1.13.21.dist-info
boto3-1.14.4.dist-info
```

If step 5 was done by conda - `conda install -y boto3==1.13.21`, it would remove the 1.14.4 version from site-packages and conda list/pip list will point to the right version.

Question is, does mamba respect all conda configs? If not, is there any other way to handle the pip interoperability in mamba. Basically, mamba should remove the pip-installed packages when upgrading/downgrading.

wolfv:
Indeed, we currently haven't yet taken much care about pip interoperability. 
Are you interested in contributing this?

rtgdk:
@wolfv Can you give some pointers from where should I start?

wolfv:
We need to utilize the functions from conda that figure out which pip packages are currently installed, and somehow add them to the installed packages.


For example, this is how they fetch the packages installed from pip:
https://github.com/conda/conda/blob/b7bd38660104973da7900d0eb09fb326fb998c0d/conda/core/prefix_data.py#L216-L299

Howevre I am also not sure how to properly handle the interop. Maybe we should first figure out how it's done in detail, e.g. document the desired behavior here and then start implementing it.

What should happen with pip installed packages: should they get overwritten with conda packages? Should they be updated? etc.

DManowitz:
[...]

PR #4123, opened 2025-12-15
feat: Support Prefix Data Interoperability

jjerphan:
# Description

Fix https://github.com/mamba-org/mamba/issues/342.

## Type of Change

- [ ] Bugfix
- [x] Feature / enhancement
- [ ] CI / Documentation
- [ ] Maintenance

## Checklist

### DIFF
--- a/libmamba/src/api/install.cpp
+++ b/libmamba/src/api/install.cpp
@@ -439,26 +439,15 @@ namespace mamba
                                 /* .clean_dependencies= */ false,
                             }
                         );
-                    }
-                    else
-                    {
-                        // No pip package conflict, use normal Install
-                        request.jobs.emplace_back(
-                            Request::Install{
-                                std::move(spec),
-                            }
-                        );
+                        continue;
                     }
                 }
-                else
-                {
-                    // No pip package conflict, use normal Install
-                    request.jobs.emplace_back(
-                        Request::Install{
-                            std::move(spec),
-                        }
-                    );
-                }
+                // No pip package conflict, use normal Install
+                request.jobs.emplace_back(
+                    Request::Install{
+                        std::move(spec),
+                    }
+                );
             }
         }
         else


libmamba/src/core/transaction.cpp
```diff
@@ -448,6 +612,28 @@ namespace mamba
...
+                        is_requested = true;
+                        break;
+                    }
+                }
+                // Skip if not explicitly requested (preserve pip package)
+                if (!is_requested)
+                {
+                    continue;
+                }
+            }
```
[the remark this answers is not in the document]
jjerphan: I do not think that it should be present; removed with 042d05bda057075d6a2c742916ac3e0df5a0c418.

libmamba/src/core/link.cpp
```diff
@@ -553,7 +584,11 @@ namespace mamba
 
         json_file.close();
 
-        fs::remove(json);
+        // Only remove the JSON file if it exists
```
Klaim: This check should be unnecessary as `std::filesystem::remove` will just return false if the file doesnt exist.

libmamba/src/api/install.cpp
```diff
@@ -377,9 +377,12 @@ namespace mamba
         }
     }
 
-    auto
-    create_install_request(PrefixData& prefix_data, std::vector<std::string> specs, bool freeze_installed)
-        -> solver::Request
+    auto create_install_request(
+        PrefixData& prefix_data,
+        std::vector<std::string> specs,
+        bool freeze_installed,
```
Klaim: Minor improvement: move the bool parameters in a (local) options struct and use that here.

libmamba/src/core/transaction.cpp
```diff
@@ -428,17 +488,121 @@ namespace mamba
...
+        // Helper function to uninstall a pip package
+        const auto uninstall_pip_package = [&](const std::string& name)
+        {
+            const auto get_python_path = [&]
+            {
+                return util::which_in("python", util::get_path_dirs(ctx.prefix_params.target_prefix))
+                    .string();
+            };
+
+            const auto args = std::array<std::string, 5>{ get_python_path(),
```
Klaim: I think you can just fill `full_args` directly at initialization:
```c++
const std::vector<std::string> full_args{
   get_python_path(),
   "-m",
   "pip",
   "uninstall",
   "-y",
   name
   };
```

libmamba/src/core/transaction.cpp
```diff
@@ -428,17 +488,121 @@ namespace mamba
...
+            }
+        };
+
+        // Uninstall all pip packages that conflict with packages being installed BEFORE installing
+        // This ensures pip packages are removed before conda packages are installed, avoiding
+        // duplicate uninstall attempts and warnings
+        if (ctx.prefix_data_interoperability)
+        {
+            const auto& pip_records = prefix.pip_records();
+            std::set<std::string> pip_packages_to_uninstall;
```
Klaim, line 548: Note that `flat_set` (in our utilities) would be an interesting alternative here because we then go through the elements sequentially afterwards. Not a big deal though, I suspect it's not performance sensitive.

libmamba/src/core/transaction.cpp
```diff
@@ -428,17 +488,121 @@ namespace mamba
...
             }
-            Console::stream() << "Unlinking " << pkg.str();
-            const fs::u8path cache_path(m_multi_cache.get_extracted_dir_path(pkg));
-            UnlinkPackage up(pkg, cache_path, &transaction_context);
-            up.execute();
-            rollback.record(up);
+
+            // Check if this is a pip package
+            // Pip packages are already uninstalled before this loop, so we skip them here
+            bool is_pip_package
```
Klaim:
```suggestion
            const bool is_pip_package
```

libmamba/src/core/transaction.cpp
```diff
@@ -448,6 +612,28 @@ namespace mamba
...
                 break;
             }
+
+            // When interoperability is disabled, skip installing conda packages that conflict with
+            // pip packages UNLESS the package is explicitly requested (to allow updates to work)
+            if (!ctx.prefix_data_interoperability && prefix.pip_records().contains(pkg.name))
+            {
+                // Check if this package is explicitly requested
+                bool is_requested = false;
+                for (const auto& spec : m_requested_specs)
```
Klaim: This looks like `std::any_of`?
```c++
// Skip if not explicitly requested (preserve pip package)
const auto is_explicitly_requested = [&](const auto& spec){ 
    return spec.name().is_exact() && spec.name().to_string() == pkg.name; 
};
if( std::ranges::any_of(m_requested_specs, is_explicitly_requested) )
{
    continue;
}
```
something like that?

libmamba/tests/src/core/test_prefix_interoperability.cpp
```diff
@@ -0,0 +1,467 @@
...
+                return util::LoopControl::Break;
+            }
+        );
+
+        REQUIRE_FALSE(found_pip_pkg);
+    }
+
+    SECTION("Pip packages ARE included when prefix interoperability is enabled")
+    {
+        ctx.prefix_data_interoperability = true;
```
Klaim: Consider restoring the previous value once out of the scope of this section or at least of the whole test?

libmamba/src/api/configuration.cpp
```diff
@@ -1668,6 +1668,20 @@ namespace mamba
...
+                   .group("Solver")
+                   .set_rc_configurable()
+                   .set_env_var_names(
+                       { "CONDA_PREFIX_DATA_INTEROPERABILITY", "MAMBA_PREFIX_DATA_INTEROPERABILITY" }
+                   )
+                   .description("Enable prefix interoperability")
+                   .long_description(
+                       "Enable interoperability between conda packages and pip packages.\n"
+                       "When enabled, mamba will use pip-installed packages to satisfy dependencies,"
+                       "pip-installed packages to satisfy dependencies, and will remove pip-installed "
```
Klaim: > pip-installed packages to satisfy dependencies,

is repeated here.

libmamba/src/core/link.cpp
```diff
@@ -609,9 +644,30 @@ namespace mamba
...
+                return std::make_tuple(
+                    std::string(validation::sha256sum(dst)),
+                    rel_dst.generic_string()
+                );
+            }
+            catch (const std::exception& e)
+            {
+                LOG_DEBUG << "Could not compute SHA256 of existing file " << dst
+                          << ", will remove it: " << e.what();
+                fs::remove(dst);
```
Klaim: Note that if remove throws (because of some fs/os issues, not because the file doesnt exist), this will crash the whole process. Consider moving the remove outside the catch so that it's not executed while an exception handling is in flight.

libmamba/src/core/package_database_loader.cpp
```diff
@@ -158,6 +159,34 @@ namespace mamba
...
         // TODO(C++20): We could do a PrefixData range that returns packages without storing them.
         auto pkgs = prefix.sorted_records();
+
+        // When prefix_data_interoperability is enabled, include pip packages in the solver database
+        // so they can be removed when replaced by conda packages. The solver will handle conflicts
+        // and remove pip packages when conda packages with the same name are installed.
+        // This is part of the prefix interoperability feature.
+        if (ctx.prefix_data_interoperability)
+        {
+            for (const auto& [name, pip_pkg] : prefix.pip_records())
```
Klaim: Side note: This loop is fine as it is 👍🏽 I was wondering if that might be rewritten using `std::ranges::filter` for clarity, to get a vew of all  `pip_pkg` existing in `pkgs` (maybe using `std::ranges::contains` too), but it doesnt feel worth the effort at this point.

libmamba/src/core/transaction.cpp
```diff
@@ -428,17 +488,121 @@ namespace mamba
...
+        {
+            const auto& pip_records = prefix.pip_records();
+            std::set<std::string> pip_packages_to_uninstall;
+
+            // Collect all pip packages that need to be uninstalled:
+            // 1. Pip packages marked for removal by the solver
+            // 2. Pip packages that conflict with packages being installed or requested
+            for (const auto& [name, pip_pkg] : pip_records)
+            {
+                bool in_removal = pip_package_names.contains(name);
```
Klaim: `const` all the `bool`s 🤡

libmamba/tests/src/core/test_prefix_interoperability.cpp
```diff
@@ -0,0 +1,467 @@
...
+#include "mamba/core/package_database_loader.hpp"
+#include "mamba/core/prefix_data.hpp"
+#include "mamba/core/util.hpp"
+#include "mamba/solver/libsolv/database.hpp"
+#include "mamba/specs/package_info.hpp"
+#include "mamba/util/environment.hpp"
+
+#include "mambatests.hpp"
+
+using namespace mamba;
```
Klaim: To prevent potential ADL-related nasty surprises, I would prefer if you put the code of the test in the `mamba` namespace instead of `using namespace` (as in just wrapping all the code in `namespace mamba{ ...`).

libmamba/src/core/transaction.cpp
jjerphan: But the insertions will be more costly because the elements being inserted aren't sorted, isn't it?

Klaim: Indeed, except if you can `.reserve` first and the element type is well optimized for moves - you'll then exchange moving elements around (sorted) vs allocating nodes. Hard to say which one is the most expensive, it also depends on the number of elements, that's why I dont think this is really important in this specific case (I dont really know). Personally for not-huge items sets, if I know I will have a loop going element by elemnts, I prefer a flat container as it avoids at least costs in there, but indeed it has to be balanced against with the insertion costs.
Feel free to ignore this note👍🏽

jjerphan: I also mind data locality for small containers, but reallocation are also quite costly _a priori_.

I do not think this is the important thing to optimize for now (I think that I would spend time on sharded repodata support), but I am definitely not against reconsidering using it (we need to compare how they perform in practice for this case). :)

libmamba/src/core/package_database_loader.cpp
```diff
@@ -158,6 +161,36 @@ namespace mamba
...
+        // This is part of the prefix interoperability feature.
+        if (ctx.prefix_data_interoperability)
+        {
+            // Filter pip packages to only those that don't have a conda equivalent
+            auto pip_packages_to_add = prefix.pip_records()
+                                       | std::ranges::views::filter(
+                                           [&pkgs](const auto& pair)
+                                           {
+                                               const auto& name = pair.first;
+                                               return !std::ranges::any_of(
```
Klaim, line 177: Minor thing: not sure we have it available on all standard library implems, but `std::ranges::contains(` with a projection argument could be shorter and clearer:
```c++
return std::ranges::contains(pkgs, name, &PackageInfo::name)
```

jjerphan: This interface of [`std::ranges::contains` (from C++26, it seems?)](https://en.cppreference.com/w/cpp/algorithm/ranges/contains.html) is not yet implemented in the version of `libstdcxx` currently used to distribute projects on conda-forge.

Klaim: Ah yes I forgot, I added a implementation in a recent pr but it's not merged yet. Nevermind.

libmamba/src/api/install.cpp
```diff
@@ -377,19 +377,27 @@ namespace mamba
...
+        const InstallRequestOptions options = {}
+    ) -> solver::Request
     {
         using Request = solver::Request;
 
         const auto& prefix_pkgs = prefix_data.records();
 
         auto request = Request();
-        request.jobs.reserve(specs.size() + freeze_installed * prefix_pkgs.size());
+        request.jobs.reserve(specs.size() + options.freeze_installed * prefix_pkgs.size());
```
JohanMabille: Please explicitly cast `freeze_installed` into `size_t`

jjerphan: For my defense, I am not responsible for this _fourberie_ originally.

jjerphan:
Note that we need to `else` to handle both cases (hence the `continue` originally).

dfc4f85 removes the `continue` (but also duplicates the statement). Should we keep this commit?

jjerphan (commented):
⇑ A response to one of the threads of the last review.

JohanMabille:
> Note that we need to `else` to handle both cases (hence the `continue` originally).
> 
> [dfc4f85](https://github.com/mamba-org/mamba/commit/dfc4f85f7d5975d6022497450374405a61cbcbd5) removes the `continue` (but also duplicates the statement). Should we keep this commit?

Nope, I noticed that I published the review and deleted the comment. I should have edited it. Let's revert to previous state.

PR #4126, opened 2025-12-19
feat: Inject channels+mirrors coming from mambajs env-lockfiles

Klaim:
# Description

So far the handling of environment lockfile read package specifications from them and used that to get the packages but the channels urls specified in the lockfile were read but then ignored.

This change makes libmamba use the urls specified by lockfiles, only for mambajs for now, and put them at the top of the priority of urls to use for each channel name (could be easily changed later if necessary).

The core of the feature is located in `MTransaction create_explicit_transaction_from_lockfile()` (`transaction.cpp`), reviewers should start reading from there.

As part of that implementation:
- inject env-lockfile channels urls in `Context` and `ContextChannel` after reading the lockfile -- this unfortunately happens after the channels has been created with other sources of information so it is a bit contrieved by the execution flow;
- added `contains` standard algorithm implementation that will be replaced by the standard once available;
- replaced the channel selection resolution by a simpler shortcut based on the channel cache in `ChannelContext` (see the change in `trancation.cpp` in `MTransaction::execute`. This change has the potential to be breaking although we didn't find so far an explicit issue.
- changed the position of the publishing of json output in `MTransaction::execute` so that it relies on the deduced information instead of approximations pre-execution.
- `ChannelContext::make_channel` now allows adding new unique mirror urls to existing channels.
- Improved error information when failing to find package cache info.

## Type of Change

- [ ] Bugfix
- [x] Feature / enhancement
- [ ] CI / Documentation
- [ ] Maintenance

## Checklist


libmamba/include/mamba/specs/channel.hpp
```diff
@@ -74,6 +80,13 @@ namespace mamba::specs
         auto clear_display_name() -> std::string;
         void set_display_name(std::string display_name);
 
+        // Adds mirror urls if not already recorded, by default added at the end of the mirrors
+        // list.
+        void add_mirror_urls(
+            const std::vector<CondaURL>& additional_mirrors,
+            UrlPriority priority = UrlPriority::low
+        );
+
```
AntoinePrv, line 89: I don't have the full context here, but if possible it would be best if these could be part of `ChannelParams`/`resolve` workflow.

The general idea of `specs::Channel` is that with `resolve` you always obtain what you want, (pure functional no mutable changes). This is because currently with `libsolv` hacks *etc*, `specs::Channel` are rather short-lived, converted back and forth to strings. I believe functions like `set_platforms` are use for only local reasoning.

libmamba/src/core/channel_context.cpp
```diff
@@ -274,19 +273,32 @@ namespace mamba
...
         {
+            // inject additional mirrors if not already existing in current channels
+            for (auto& channel : it->second)
+            {
+                // TODO C++23: replace all this by  std::vector(from_range_t, ...)
+                auto urls_view = specs::as_conda_urls(mirrors);
+                std::vector<specs::CondaURL> urls(urls_view.begin(), urls_view.end());
+
+                channel.add_mirror_urls(urls, new_mirrors_priority);
+            }
```
AntoinePrv: If this logic could be made part of `ChannelParams`/`Channel::resolve` that would avoid adding extra complexity to the channel creation logic.

Klaim:
I keep seeing issues with the tests locally but I suspect it's actually an unrelated problem, I might have to spam the ci a bit, sorry for the noise!

jjerphan:
FYI, I also rebased this branch onto `main`.

I kept the previous state of the branch on [`backup/env-lockfile-channels`](https://github.com/mamba-org/mamba/tree/backup/env-lockfile-channels).

libmamba/include/mamba/specs/channel.hpp
Klaim: Noting here for future readers: we discussed this IRL and ended up with the conclusion (if my memory is correct, correct me @AntoinePrv if I'm misremembering) that this function is still necessary except if we add a new category of urls for channels coming from the lockfiles , whcih would probably be overcomplicated.
The other comments emitted at the time were addressed the same day with proper changes.

djangoliv:
Any chance this could move forward? 
We're hitting this in our project: we create environments in a pod from a mambajs lockfile, and every package from a channel that isn't on conda.anaconda.org fails with a 404, because the URL from channelInfo is ignored.

Klaim:
> Any chance this could move forward? We're hitting this in our project: we create environments in a pod from a mambajs lockfile, and every package from a channel that isn't on conda.anaconda.org fails with a 404, because the URL from channelInfo is ignored.

Thanks for checking here! This has been only made ready for review yesterday so we need to go through that and once merged we'll be able to make a beta build for you to test. 👍🏼

djangoliv:
Great, thanks for the quick answer! Happy to test the beta build when it's ready.
We have a real lockfile that reproduces the issue (147 packages, one of them from a channel that only exists on prefix.dev), so it should be a good check. Just ping me here.

libmamba/src/core/package_cache.cpp
```diff
@@ -432,6 +434,28 @@ namespace mamba
                         // for explicit env, we have no size, nor sha256 so we need to
                         // set valid true here
                         valid = true;
+
+                        const auto repodata_url = repodata_record["url"].get<std::string>();
+                        if (!repodata_url.empty())
+                        {
+                            const auto pkg_url = repodata_record["url"].get<std::string>();
```
JohanMabille: This variable is redundant with `repodata_url` defined above.

libmamba/include/mamba/core/env_lockfile.hpp
```diff
@@ -77,6 +77,8 @@ namespace mamba
             std::vector<Channel> channels;
             std::vector<std::string> platforms;
             std::vector<std::string> sources;
+            bool enable_channels = true;  ///< `true` if we need to take into account the specified
```
JohanMabille: It could be nice (for future us) to document whe and why we should ignore the specified channels.

libmamba/src/download/mirror_map.cpp
```diff
@@ -36,29 +37,40 @@ namespace mamba::download
 
     namespace
     {
-        template <class It>
-        bool contains(It first, It last, MirrorID id)
+        template <std::ranges::input_range R>
+        bool contains_mirror_with_id(const R& r, const MirrorID& id)
```
JohanMabille: Nitpicking: maybe `contains_mirror` is enough?

Klaim: I'll tweak it.
I added the clarification in the name because it is not looking for a specific mirror but only one with the right id, it's not for example comparing other values or the memory addresses of the mirrors.
But I guess the parameters and being local to that file are enough.

JohanMabille (approved):

AntoinePrv (approved):

PR #4135, opened 2026-01-12
fix: Only call `get_all_running_processes_info` when needed

mihaitodor:
Do not call `get_all_running_processes_info()` unless the log level requires printing debug logs.

When running micromamba in a specific production environment, we keep getting `Segmentation fault (core dumped)` and, by adding more logs in the code, I was able to narrow it down to this particular log entry, where it tries to print the currently running processes. I can confirm that the crash doesn't occur in `get_all_running_processes_info()` and it is something in the `MessageLogger` implementation, but I have a hard time narrowing it down since I don't have access to run gdb in that environment or collect core dumps. I believe it is something to do with the length of the resulting string, but I've not been able to get it to reproduce reliably in a unit test. However, we don't use debug logs and, in this case, I think the code should skip this log clause completely, which should unblock us.

# Description

## Type of Change

- [x] Bugfix
- [ ] Feature / enhancement
- [ ] CI / Documentation
- [ ] Maintenance

## Checklist

### DIFF
--- a/libmamba/src/core/run.cpp
+++ b/libmamba/src/core/run.cpp
@@ -320,7 +320,10 @@ namespace mamba
             LOG_WARNING << "Could not create proc dir: " << proc_dir() << " (" << ec.message() << ")";
         }
 
-        LOG_DEBUG << "Currently running processes: " << get_all_running_processes_info();
+        if (logging::get_log_level() < log_level::info)
+        {
+            LOG_DEBUG << "Currently running processes: " << get_all_running_processes_info();
+        }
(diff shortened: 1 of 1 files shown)


jjerphan (commented):
Thank you, @mihaitodor.

Indeed `get_all_running_processes_info` only must be called when needed.

mihaitodor:
@jjerphan Thank you for confirming! I fixed the linter error (should've checked again before pushing).

jjerphan:
I have rebased your PR to have #4137 in.

PR #4141, opened 2026-01-16
feat: Shard Parsing

jjerphan:
# Description

A first slice of #4142, a solution for #3431.

## Type of Change

- [ ] Bugfix
- [x] Feature / enhancement
- [ ] CI / Documentation
- [ ] Maintenance

## Checklist

- [x] My code follows the general style and conventions of the codebase, ensuring consistency
- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [x] I have run `pre-commit run --all` locally in the source folder and confirmed that there are no linter errors.
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing tests pass locally with my changes

PR #4145, opened 2026-01-20
feat: Shard types and utils

jjerphan:
# Description

Slice of #4141.

## Type of Change

- [ ] Bugfix
- [x] Feature / enhancement
- [ ] CI / Documentation
- [ ] Maintenance

## Checklist

### DIFF
--- a/libmamba/include/mamba/core/shard_types.hpp
+++ b/libmamba/include/mamba/core/shard_types.hpp
@@ -30,7 +30,7 @@ namespace mamba
      * - Contains only fields needed for dependency traversal, reducing memory usage
      *   when processing many shards.
      * - Conversion to RepoDataPackage happens lazily when building repodata for
-     *   the solver, deferring parsing costs until needed.
+     *   the solver, deferring parsing costs until actually needed.
      *
      * **Comparison with specs::PackageInfo:**
      * - PackageInfo is the runtime representation used for installed packages,
@@ -59,10 +59,14 @@ namespace mamba
      * 4. **Flexible msgpack handling**: Custom parsing handles various msgpack types
      *    for sha256/md5 (strings, bytes, EXT types), easier with a dedicated structure.
      *
+     * This structure supports all fields defined in the shard format specification.
+     * See https://conda.org/learn/ceps/cep-0016 for the complete shard format specification.
+     *
      * @see specs::RepoDataPackage The canonical package record type used for repodata.json
      * @see specs::PackageInfo The runtime package representation used throughout the codebase
      * @see to_repo_data_package() Conversion function to RepoDataPackage
      * @see from_repo_data_package() Conversion function from RepoDataPackage
+     * @see https://conda.org/learn/ceps/cep-0016 CEP 16 - Sharded Repodata specification
      */
     struct ShardPackageRecord
     {
@@ -76,6 +80,10 @@ namespace mamba
         std::vector<std::string> constrains;
         std::optional<std::string> noarch;
         std::size_t size = 0;
+        std::optional<std::string> license;
+        std::optional<std::string> license_family;
+        std::optional<std::string> subdir;
+        std::optional<std::size_t> timestamp;
     };
 
     /**
--- a/libmamba/include/mamba/core/shard_types.hpp
+++ b/libmamba/include/mamba/core/shard_types.hpp
@@ -1,0 +1,200 @@
+// Copyright (c) 2024, QuantStack and Mamba Contributors
+//
+// Distributed under the terms of the BSD 3-Clause License.
+//
+// The full license is in the file LICENSE, distributed with this software.
+
+#ifndef MAMBA_CORE_SHARD_TYPES_HPP
+#define MAMBA_CORE_SHARD_TYPES_HPP
+
+#include <map>
+#include <optional>
+#include <string>
+#include <vector>
+
+#include "mamba/specs/package_info.hpp"
+#include "mamba/specs/repo_data.hpp"
+
+namespace mamba
+{
+    /**
+     * Package record dictionary for shard data.
+     *
+     * This is a simplified representation of package metadata used in shards.
+     * It exists separately from other package types for several reasons:
+     *
+     * **Comparison with specs::RepoDataPackage:**
+     * - Uses primitive types (string for version, string for noarch) instead of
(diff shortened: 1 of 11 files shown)


libmamba/src/core/shard_types.cpp
```diff
@@ -0,0 +1,151 @@
...
+        if (auto version = specs::Version::parse(record.version))
+        {
+            pkg.version = version.value();
+        }
+        pkg.build_string = record.build;
+        pkg.build_number = record.build_number;
+        pkg.sha256 = record.sha256;
+        pkg.md5 = record.md5;
+        pkg.depends = record.depends;
+        pkg.constrains = record.constrains;
```
jjerphan: This respects the style of the code-base, but I really do not like it.

I think that we should rather used designated initializers (named parameters for constructors) now that we use C++20.

What do reviewers think?

This comment also applies elsewhere.

libmamba/include/mamba/core/shard_types.hpp
```diff
@@ -0,0 +1,200 @@
...
+     * Maps to the Python ShardDict type. Contains all versions of a package
+     * in both .tar.bz2 and .conda formats.
+     */
+    struct ShardDict
+    {
+        /** Packages in .tar.bz2 format, keyed by filename. */
+        std::map<std::string, ShardPackageRecord> packages;
+
+        /** Packages in .conda format, keyed by filename. */
+        std::map<std::string, ShardPackageRecord> conda_packages;
```
JohanMabille, line 101: Why do we need to distinguish the two kinds of packages?

libmamba/include/mamba/core/shard_types.hpp
```diff
@@ -0,0 +1,200 @@
...
+        RepodataInfoDict info;
+
+        /** Repodata version. */
+        std::size_t repodata_version = 2;
+
+        /** Packages in .tar.bz2 format, keyed by filename. */
+        std::map<std::string, ShardPackageRecord> packages;
+
+        /** Packages in .conda format, keyed by filename. */
+        std::map<std::string, ShardPackageRecord> conda_packages;
```
JohanMabille: Maybe these two maps could be replaced with an instance of ShardDict?

libmamba/include/mamba/core/shard_types.hpp
```diff
@@ -0,0 +1,200 @@
...
+    };
+
+    /**
+     * Information dictionary from repodata.
+     *
+     * Contains channel metadata including base URLs and subdir information.
+     */
+    struct RepodataInfoDict
```
JohanMabille: THe name is a bit ocnfusing with RepodataDict. Besides, it's not really a dict. RepoMetadata? RepoBaseURLs?

libmamba/include/mamba/core/shard_types.hpp
```diff
@@ -0,0 +1,200 @@
...
+     * This conversion is used when building repodata for the solver.
+     */
+    specs::RepoDataPackage to_repo_data_package(const ShardPackageRecord& record);
+
+    /**
+     * Convert specs::RepoDataPackage to ShardPackageRecord.
+     *
+     * This conversion is used when treating monolithic repodata as shards.
+     */
+    ShardPackageRecord from_repo_data_package(const specs::RepoDataPackage& record);
```
JohanMabille, line 172: Not sure why we need this; or more precisely, not sure why we need conversions in both ways. I guess that we'll have different loaders, one for the Shards and one for the repodata. And then we'll have a single representation (either ShardPackageRecord or RepoDataPackage), so theoretically we should need only one conversion. Or am I missing somehting?

libmambapy/CMakeLists.txt
```diff
@@ -53,7 +54,7 @@ target_include_directories(bindings PRIVATE bindings)
 mamba_target_add_compile_warnings(bindings WARNING_AS_ERROR ${MAMBA_WARNING_AS_ERROR})
 
 target_link_libraries(
-    bindings PRIVATE pybind11::pybind11 ${libmamba_target} mamba::libmamba-dyn-spdlog
+    bindings PRIVATE pybind11::pybind11 ${libmamba_target} mamba::libmamba-dyn-spdlog msgpack-c
```
jjerphan, line 57: This is necessary for the build to pass on Windows; I do not understand why this is required there.

libmamba/include/mamba/core/shard_types.hpp
jjerphan: IIRC, we mostly need it if we need to report problems which are specific to one of the group. I need to recheck.

libmamba/include/mamba/core/shard_types.hpp
jjerphan: Indeed. 👀

libmamba/include/mamba/core/shard_types.hpp
jjerphan: Yes, I have added it for testing roundtrips between structures, but we should only have one of them _a priori_.

jjerphan: Ah, and it was used for treating the flat repodata as a shard (using a `ShardLike` class inheriting from the base class like `Shard`, which I removed because this was an overkill or YAGNI).

libmamba/include/mamba/core/shard_types.hpp
jjerphan: It basically follows `specs::Repodata`'s members:

https://github.com/mamba-org/mamba/blob/16ca83c584cb94206af947c9f767e8f4616400d1/libmamba/include/mamba/specs/repo_data.hpp#L175-L189

JohanMabille (approved):

libmambapy/CMakeLists.txt
JohanMabille: Let's track it in an issue after merging this one.

opoplawski:
Are you *really* wanting the C version of msgpack instead of the C++ one?

jjerphan:
Yes because the dependency chain is much more minimalistic (on conda-forge `libmsgpack-cxx` also depends on `libboost`).

jjerphan:
Is it a problem for you?

opoplawski:
Not really.  It isn't currently packaged for Fedora so I had to submit a new package review.  I also see that msgpack-c hasn't seen an update in a while, but perhaps it's just mature.

Issue #4096, opened 2025-11-12
Segmentation fault with ctrl-c while downloading


davidbrochart:
### Troubleshooting docs

- [x] My problem is not solved in the Troubleshooting docs

### Anaconda default channels

- [x] I do NOT use the Anaconda default channels (pkgs/* etc.)

### How did you install Mamba?

Micromamba

### Search tried in issue tracker

segmentation

### Latest version of Mamba

- [x] My problem is not solved with the latest version

### Tried in Conda?

Not applicable

### Describe your issue

With:
```bash
micromamba clean -i
micromamba install pip
```
And hitting ctrl-c while downloading the repodata, I get a `Segmentation fault (core dumped)` (on Ubuntu 24.04).

### mamba info / micromamba info

```shell
libmamba version : 2.3.3
     micromamba version : 2.3.3
           curl version : libcurl/8.14.1 OpenSSL/3.5.4 zlib/1.3.1 zstd/1.5.7 libssh2/1.11.1 nghttp2/1.67.0
     libarchive version : libarchive 3.8.2 zlib/1.3.1 bz2lib/1.0.8 libzstd/1.5.7 libb2/bundled
       envs directories : /home/david/micromamba/envs
          package cache : /home/david/micromamba/pkgs
                          /home/david/.mamba/pkgs
            environment : base
           env location : /home/david/micromamba
      user config files : /home/david/.mambarc
 populated config files : /home/david/.condarc
       virtual packages : __unix=0=0
                          __linux=6.14.0=0
                          __glibc=2.39=0
                          __archspec=1=x86_64_v3
               channels : https://conda.anaconda.org/conda-forge/linux-64
                          https://conda.anaconda.org/conda-forge/noarch
                          https://conda.anaconda.org/nodefaults/linux-64
                          https://conda.anaconda.org/nodefaults/noarch
       base environment : /home/david/micromamba
               platform : linux-64
```

### Logs

```shell

```

### environment.yml

```yaml

```

### ~/.condarc

```yaml

```

jjerphan:
I confirm that I can reproduce with 2.4.0.

Klaim:
@jjerphan if you didnt start investigating this one I'll take a look very soon

jjerphan:
I did not, feel free to investigate.

Klaim:
Reproduced on Windows 👍🏽

PR #4146, opened 2026-01-21
Graceful handling of download cancel/interruption

Klaim:
# Description

- Fixes https://github.com/mamba-org/mamba/issues/4096
- Improves output clarity when doing ctrl-c (although the output will depend on which stage it has been triggered).
- Removed or fixed some misleading output.
- Modified various log handling to reduce noise specifically when handling user-interruptions.

## Type of Change

- [x] Bugfix
- [ ] Feature / enhancement
- [ ] CI / Documentation
- [ ] Maintenance

## Checklist

### DIFF
--- a/libmamba/src/core/subdir_index.cpp
+++ b/libmamba/src/core/subdir_index.cpp
@@ -557,7 +557,7 @@ namespace mamba
             // missing subdirs (e.g. local path as a `noarch` but no `linux-64`).
             for (auto& result : results)
             {
-                // Don't log if it's a user interuption.
+                // Don't log if it's a user interruption.
                 if (!result.has_value() and not result.error().is_stop)
                 {
                     LOG_WARNING << "Failed to load subdir: " << result.error().message;


Klaim:
@jjerphan Nice, thanks for checking! I didnt have time to test on linux yet, that's good news as it's exactly what I see on Windows.

There are still some issues for now with this pr:
- when I tested interrupting the download of packages, which happens after downloading the repo indices and confirming the download (I used `--yes` for shortcut) , I ended up in some kind of infinite loop. @jjerphan could you try see if you can reproduce that one too? I need to investigate what's going on;
- we have an issue with the logging being overwritten by the progress bars, it's not a big deal but does impact clarity, which is why there is this weird hack of injecting multiple logs so that at least one copy of that log ends up visible, a line after the progress bars. We discussed some ways to fix it with @JohanMabille , including maybe stopping the progress bars altogether.

jjerphan:
> when I tested interrupting the download of packages, which happens after downloading the repo indices and confirming the download (I used `--yes` for shortcut) , I ended up in some kind of infinite loop. @jjerphan could you try see if you can reproduce that one too? I need to investigate what's going on;

I also can reproduce it locally.

Klaim:
This version works for me mechanically, but the output differs depending on when you trigger ctrl-c:
- while the repositories indexes are being downloaded -> returns non-zero with an explicit error about being interrupted, maybe warnings too;
- while the repositories indexes are being parsed (before starting downloads) -> returns zero, no error;
- while waiting for confirmation from user (when not using `--yes`) -> shows that it starts downloading but then return immediately zero, no error, maybe warnings;
- while downloading packages -> returns zero, maybe warnings.

It is still an improvement though, fixing these outputs would require normalizing the error handling of libmamba's code, which we already know is near impossible without a better separation of the different stages of processing.

Klaim:
I also dont have a good way to automatically test this unfortunately, it's too time-sensitive.

I'll cleanup and call it a day.

jjerphan (approved):
LGTM.

---

One UX nitpick: users on Linux obtain on `SIGINT`:
```
^Cwarning  libmamba !!!!
warning  libmamba Interruption requested by user - stopping download... 
warning  libmamba !!!!
warning  libmamba Failed to load subdir: Stopped by user request
warning  libmamba Failed to load subdir: Stopped by user request
critical libmamba Multiple errors occurred:
    Interrupted by user
```

Which is quite verbose, unexpected or even potentially scary to some users I would say.

I think something lighter like:

```
    Interrupted by user
```

would be more appropriate.

Klaim:
@jjerphan I agree in principle. The one you're showing is repo indexes download has been cancelled, with the error (it doesnt displayt he same later). The line of error can be overwritten by the console's progress bars and not even be visible even if logged as expected.
The warnings are there mostly to "push" the logging below the console output, as a temporary hack.
I can remove them, but it will look weirder.

I didn't find a simple way to remove the error in this case.

However, in last push, I did manage to make the output clearer when interrupting at any stage of the package download or before it (except at repo indexes download).

I'll try to see if I can improve the case you point but it seems difficult as long as the console overwrites the output.

Klaim:
@jjerphan In changes I just pushed:
- I removed the intempestive warning log about interruption of download;
- I did some changes to how errors are handled in `main()` IFF we detect that an exception is actually due to a user interruption, and only in that case.
- In that case we return zero instead of 1.

The result (when canceling dnl of repo):
```
warning  libmamba Failed to load subdir: Stopped by user request
warning  libmamba Failed to load subdir: Stopped by user request
warning  libmamba Multiple errors occurred:
    Interrupted by user
```

Notes:
- this is quite a potentially big change, because that might note handle well all cases of exceptions, but I tried my best to cover what I think is correct;
- the first 2 warnings are generated in the original code, not sure if I should remove them;
- the last warning is because of the error handling in main. I kept a warning instead of info log because info logs gets either overwritten or dumped for some reason (maybe our mamba-specific spdlog setup) even with forcing a log flush.

It's less scary than before but not perfect.

Klaim:
@jjerphan I managed to reduce the noise for that case to
```
warning  libmamba Interrupted by user
```

jjerphan:
Thank you!

I think we can have someone else's review and merge it.

PR #4148, opened 2026-01-23
feat: Shards

jjerphan:
# Description

Slice of #4141.

## Type of Change

- [ ] Bugfix
- [x] Feature / enhancement
- [ ] CI / Documentation
- [ ] Maintenance

## Checklist

### DIFF
--- a/libmamba/src/core/shards.cpp
+++ b/libmamba/src/core/shards.cpp
@@ -881,58 +881,44 @@ namespace mamba
                 mamba_error_code::unknown
             );
         }
-
+        std::vector<std::uint8_t> compressed_data;
         if (std::holds_alternative<download::Filename>(success.content))
         {
             std::string filename_value = std::get<download::Filename>(success.content).value;
-            LOG_DEBUG << "Download returned Filename: " << filename_value;
+            LOG_DEBUG << "Shard download returned filename: " << filename_value;
 
             shard_file = filename_value;
             LOG_DEBUG << "Using filename from download result: " << shard_file.string();
 
-            if (!fs::exists(shard_file))
+            std::ifstream file(shard_file.string(), std::ios::binary);
+            if (!file.is_open())
             {
                 return make_unexpected(
-                    "Shard file not found at path: " + shard_file.string(),
+                    "Failed to open downloaded shard file: " + shard_file.string(),
                     mamba_error_code::unknown
                 );
             }
+            LOG_DEBUG << "Reading shard file for package '" << package
+                      << "': " << shard_file.string();
+            compressed_data = std::vector<std::uint8_t>(
+                std::istreambuf_iterator<char>(file),
+                std::istreambuf_iterator<char>()
+            );
+            file.close();
+            LOG_DEBUG << "Done reading shard file for package '" << package
+                      << "': " << compressed_data.size() << " bytes";
         }
         else  // std::holds_alternative<download::Buffer>(success.content)
         {
-            LOG_DEBUG << "Shard download returned buffer, writing to artifact path";
             const auto& buffer = std::get<download::Buffer>(success.content);
-            std::ofstream out_file(shard_file.string(), std::ios::binary);
-            if (!out_file.is_open())
-            {
-                return make_unexpected(
-                    "Failed to create shard file for buffer: " + shard_file.string(),
-                    mamba_error_code::unknown
-                );
-            }
-            out_file.write(buffer.value.data(), static_cast<std::streamsize>(buffer.value.size()));
-            out_file.close();
-            LOG_DEBUG << "Wrote buffer to file: " << shard_file.string() << " ("
-                      << buffer.value.size() << " bytes)";
+            LOG_DEBUG << "Shard download returned buffer";
+            compressed_data.assign(buffer.value.begin(), buffer.value.end());
+            LOG_DEBUG << "Done reading shard buffer for package '" << package
+                      << "': " << compressed_data.size() << " bytes";
         }
 
-        std::ifstream file(shard_file.string(), std::ios::binary);
-        if (!file.is_open())
-        {
-            return make_unexpected(
-                "Failed to open downloaded shard file: " + shard_file.string(),
-                mamba_error_code::unknown
-            );
-        }
-
-        std::vector<std::uint8_t> compressed_data(
-            (std::istreambuf_iterator<char>(file)),
-            std::istreambuf_iterator<char>()
-        );
-        file.close();
-
-        LOG_DEBUG << "Reading shard file for package '" << package << "': " << shard_file.string()
-                  << " (" << compressed_data.size() << " bytes compressed)";
+        LOG_DEBUG << "Processing shard for package '" << package << "': " << compressed_data.size()
+                  << " bytes compressed";
 
         auto decompressed_result = decompress_zstd_shard(compressed_data);
         if (!decompressed_result.has_value())


libmamba/src/core/shards.cpp
```diff
@@ -0,0 +1,1219 @@
...
+     * Shards         *
+     ******************/
+
+    Shards::Shards(
+        ShardsIndexDict shards_index,
+        std::string url,
+        const specs::Channel& channel,
+        const specs::AuthenticationDataBase& auth_info,
+        const download::mirror_map& mirrors,
+        const download::RemoteFetchParams& remote_fetch_params,
```
JohanMabille: Since these members will be copied in fine, they should be passed by value (and moved in the data member), so the user can choose whether to copy or move at the call site.

libmamba/src/core/shards.cpp
```diff
@@ -0,0 +1,1219 @@
...
+                return static_cast<std::uint64_t>(obj.via.i64);
+            }
+            throw std::runtime_error("Expected INTEGER type");
+        }
+
+        auto msgpack_object_to_string_array(const msgpack_object& obj) -> std::vector<std::string>
+        {
+            std::vector<std::string> result;
+            if (obj.type == MSGPACK_OBJECT_ARRAY)
+            {
```
JohanMabille, line 73:
```suggestion
            {
                result.reserve(obj.via.array.size);
```

libmamba/src/core/shards.cpp
```diff
@@ -0,0 +1,1219 @@
...
+                    {
+                        // Only skip for optional fields, required fields should not be nil
+                        if (key == "sha256" || key == "md5" || key == "constrains" || key == "noarch")
+                        {
+                            continue;
+                        }
+                        // For required fields, log and continue
+                        LOG_DEBUG << "Field '" << key << "' is nil in ShardPackageRecord, skipping";
+                        continue;
+                    }
```
JohanMabille, line 169: Nothing is parsed in this block, I think it should be moved before the try block.

Besides, it can be simplified with something like:

```cpp
                        if (key != "sha256" && key != "md5" && key != "constrains" && key != "noarch")
                        {
                            LOG_DEBUG << "Field '" << key << "' is nil in ShardPackageRecord, skipping";
                        }
                        continue;
```

libmamba/src/core/shards.cpp
```diff
@@ -0,0 +1,1219 @@
...
+                            try
+                            {
+                                record.sha256 = msgpack_object_to_string(val_obj);
+                            }
+                            catch (...)
+                            {
+                                // Skip if conversion fails
+                            }
+                        }
+                    }
```
JohanMabille, line 195: Maybe it would make sense to move this block into a dedicated function for parsing sha256 field?

libmamba/src/core/shards.cpp
```diff
@@ -0,0 +1,1219 @@
...
+                            try
+                            {
+                                record.md5 = msgpack_object_to_string(val_obj);
+                            }
+                            catch (...)
+                            {
+                                // Skip if conversion fails
+                            }
+                        }
+                    }
```
JohanMabille, line 199: Samre remark here for `md5`

libmamba/src/core/shards.cpp
```diff
@@ -0,0 +1,1219 @@
...
+        std::string url = shards_base_url() + shard_name;
+        LOG_DEBUG << "Constructed shard URL for package '" << package << "': " << url;
+        return url;
+    }
+
+    /**
+     * Get the relative path for a shard (for use with download::Request).
+     * Returns path relative to channel base, e.g., "linux-64/shards/<hash>.msgpack.zst"
+     */
+    auto Shards::shard_path(const std::string& package) const -> std::string
```
JohanMabille: It would be nice to have the "relative" word in the name of the method.

libmamba/src/core/shards.cpp
```diff
@@ -0,0 +1,1219 @@
...
+        {
+            // For relative URLs, join with repodata URL
+            result = util::url_concat(m_url, "/", shards_base_url_str);
+        }
+
+        if (!util::ends_with(result, "/"))
+        {
+            result += "/";
+        }
+
```
JohanMabille, line 337: Can't we reuse functions from [url_manip.hpp](https://github.com/mamba-org/mamba/blob/main/libmamba/include/mamba/util/url_manip.hpp) here, especially the one for building / concatenating URLs?

libmamba/src/core/shards.cpp
```diff
@@ -0,0 +1,1219 @@
...
+                normalized_shards = normalized_shards.substr(1);
+            }
+
+            // Construct path: platform/shards/<hash>.msgpack.zst
+            std::string path = util::url_concat(platform, "/", normalized_shards);
+            if (!util::ends_with(path, "/"))
+            {
+                path += "/";
+            }
+            return path + shard_name;
```
JohanMabille: Similar question here, can't we reuse functions from [url_manip](https://github.com/mamba-org/mamba/blob/main/libmamba/include/mamba/util/url_manip.hpp) for extracting the parts of the URL we need?

libmamba/include/mamba/core/shards.hpp
```diff
@@ -0,0 +1,140 @@
...
+
+        /** Store new shard data. */
+        void visit_shard(const std::string& package, const ShardDict& shard);
+
+        /** Fetch an individual shard for the given package. */
+        auto fetch_shard(const std::string& package) -> expected_t<ShardDict>;
+
+        /** Fetch multiple shards in one go. */
+        auto fetch_shards(const std::vector<std::string>& packages)
+            -> expected_t<std::map<std::string, ShardDict>>;
```
JohanMabille: I find these APIs a bit confusing; from the implementation, `visit_shard` and `process_shard` do almost the same thing, the only different is that `process_shard` takes an additional parameter it does not use, and logs;
`shard_loaded` checks a condition set by `visit_shaed`. 

If the Shards class is responsible for downloading the shards files, it should expose a simple API to the user (only fetch_shards); the caching and checking if the shards of the dependencies have been fetched should be hidden.

Also the `visit`  word is a bit confusing, since in C++ it is usually related to the visitor pattern.

libmamba/src/core/shards.cpp
```diff
@@ -0,0 +1,1219 @@
...
+
+    void Shards::visit_shard(const std::string& package, const ShardDict& shard)
+    {
+        LOG_DEBUG << "Visiting shard for package '" << package << "': " << shard.packages.size()
+                  << " .tar.bz2 packages, " << shard.conda_packages.size() << " .conda packages";
+        m_visited[package] = shard;
+    }
+
+    void Shards::process_fetched_shard(
+        const std::string& /* url */,
```
JohanMabille: Why this argument? This method is private and does not use it.

libmamba/src/core/shards.cpp
```diff
@@ -0,0 +1,1219 @@
...
+                            unpacked.zone = nullptr;
+                        }
+                        LOG_WARNING << "Failed to parse msgpack for shard " << url << ": "
+                                    << e.what();
+                    }
+                }
+            }
+        }
+
+        return results;
```
JohanMabille, line 1032: This function is definitely too long and should be split, with correct locality for temporary variables.

libmamba/src/core/shards.cpp
```diff
@@ -0,0 +1,1219 @@
...
+                // Create cache directory using user cache dir, not URL path
+                fs::u8path cache_dir_path;
+                if (auto cache_home = util::get_env("XDG_CACHE_HOME"))
+                {
+                    cache_dir_path = fs::u8path(*cache_home) / "conda" / "pkgs";
+                }
+                else
+                {
+                    cache_dir_path = fs::u8path(util::user_cache_dir()) / "conda" / "pkgs";
+                }
```
JohanMabille: This is independent from the couple `[url, package]` and should be out of the loop.

libmamba/src/core/shards.cpp
jjerphan: Do you refer to all of them, even `mirrors`?

JohanMabille: Ah I missed it was a const reference. In that case, it would be better to
- make the member a pointer, and potentially provide a private function returning a const reference (so you don't have to dereference it each time you need it).
- OR declare the assignment operators (copy and move) as deleted; they are by default, but since this class is publicly exposed, it would be better to state it explicitly rather than ending with a compilation error about a private data.

I am in favor of the first solution since it offers more flexibility to the user, but I haven't seen how this class will be used.

EDIT: it should also be stated in the constructor doc that the mirror_map should remain alive as long as the Shard is alive. I don't know if avoiding the copy is really worth the cost here.

jjerphan: Actually, we can omit storing the `mirrors` for now.

See 40805cd.

libmamba/src/core/shards.cpp
```diff
@@ -0,0 +1,1197 @@
...
+            download::Request request(
+                package + "-shard",
+                download::MirrorName(mirror_name),
+                url_path,
+                artifact->path().string(),
+                /*head_only*/ false,
+                /*ignore_failure*/ false
+            );
+
+            request.on_success = [artifact = std::move(artifact), url, package](
```
JohanMabille: No need to capture artifact since it is not used in the lambda.

libmamba/src/core/shards.cpp
```diff
@@ -0,0 +1,1197 @@
...
+                                     const download::Success& /* success */
+                                 )
+            {
+                (void) artifact;
```
JohanMabille: This should be removed.

libmamba/src/core/shards.cpp
```diff
@@ -0,0 +1,1197 @@
...
+            ShardDict shard;
+
+            auto parse_package_records = [&package](
+                                             const msgpack_object& map_obj,
+                                             std::map<std::string, ShardPackageRecord>& target_map,
+                                             const std::string& log_prefix,
+                                             const std::string& map_name
+                                         )
+            {
+                if (map_obj.type == MSGPACK_OBJECT_MAP)
```
JohanMabille: Since this lambda is local and the condition is checked at the callsite, it should be removed from here.

libmamba/src/core/shards.cpp
```diff
@@ -0,0 +1,1197 @@
...
+                        catch (const std::exception& e)
+                        {
+                            LOG_WARNING << "Failed to parse package record in '" << map_name
+                                        << "': " << e.what();
+                        }
+                    }
+                }
+            };
+
+            if (obj.type == MSGPACK_OBJECT_MAP)
```
JohanMabille: This should be checked as an early return condition instead of wrapping the loop.

libmamba/src/core/shards.cpp
```diff
@@ -0,0 +1,1197 @@
...
+                mamba_error_code::unknown
+            );
+        }
+
+        if (std::holds_alternative<download::Filename>(success.content))
+        {
+            std::string filename_value = std::get<download::Filename>(success.content).value;
+            LOG_DEBUG << "Download returned Filename: " << filename_value;
+
+            if (filename_value.find("://") == std::string::npos)
```
JohanMabille: This is totally useless. Since you pass a filename when building the request (see https://github.com/mamba-org/mamba/pull/4148/changes#diff-c59caeb7ad4161eebdecbf6a5bf548b1b151cb2393afae7b22a98029dd048561R663 and how artifact is built), and since the result was successful, success.contente.value *is* a filename.

libmamba/src/core/shards.cpp
```diff
@@ -0,0 +1,1197 @@
...
+            {
+                shard_file = filename_value;
+                LOG_DEBUG << "Using filename from download result: " << shard_file.string();
+
+                int retries = 5;
+                while (!fs::exists(shard_file) && retries > 0)
+                {
+                    std::this_thread::sleep_for(std::chrono::milliseconds(100));
+                    retries--;
+                }
```
JohanMabille: Why? The download is finished and successful, so shard_file must exist. If not, then there's a bug that we must fix in the downloader.

libmamba/src/core/shards.cpp
```diff
@@ -0,0 +1,1197 @@
...
+                    "Failed to create shard file for buffer: " + shard_file.string(),
+                    mamba_error_code::unknown
+                );
+            }
+            out_file.write(buffer.value.data(), static_cast<std::streamsize>(buffer.value.size()));
+            out_file.close();
+            LOG_DEBUG << "Wrote buffer to file: " << shard_file.string() << " ("
+                      << buffer.value.size() << " bytes)";
+        }
+        else
```
JohanMabille: Again, not possible if the download was successful, by construction `success.content` necessarily contains one of its alternative.

libmamba/src/core/shards.cpp
```diff
@@ -0,0 +1,1197 @@
...
+            );
+        }
+
+        if (!fs::exists(shard_file))
+        {
+            return make_unexpected(
+                "Shard file does not exist: " + shard_file.string(),
+                mamba_error_code::unknown
+            );
+        }
```
JohanMabille: This check is unecessary here, since it will be performed below when opening the file.

libmamba/src/core/shards.cpp
```diff
@@ -0,0 +1,1197 @@
...
+            cache_miss_urls,
+            cache_miss_packages,
+            package_to_artifact_path,
+            artifacts
+        );
+
+        if (requests.empty())
+        {
+            return results;
+        }
```
JohanMabille: This one can be removed, by construction if `url_to_package` is not empty, then requests won't be empty.

libmamba/src/core/shards.cpp
jjerphan: Probably some debugging left-overs.

libmamba/src/core/shards.cpp
```diff
@@ -0,0 +1,1149 @@
...
+                    "Shard file not found at path: " + shard_file.string(),
+                    mamba_error_code::unknown
+                );
+            }
+        }
+        else  // std::holds_alternative<download::Buffer>(success.content)
+        {
+            LOG_DEBUG << "Shard download returned buffer, writing to artifact path";
+            const auto& buffer = std::get<download::Buffer>(success.content);
+            std::ofstream out_file(shard_file.string(), std::ios::binary);
```
JohanMabille: I don't think it is necessary to write the buffer to a file, you could initialize `compressed_data` directly from the buffer.

jjerphan: Indeed, I reworked the logic to directly populate `compressed_data`.

I wonder whether we should change the initialization (using a lambda) so that it is `const`-qualified. That's not too much a logic in `libmamba`, and probably this will overcomplexify it.

JohanMabille: Yeah I think your change is enough.

PR #4153, opened 2026-01-27
fix: Clone pip packages with environment

jjerphan:
# Description

Follow up of https://github.com/mamba-org/mamba/pull/4102#issuecomment-3802943383.

## Type of Change

- [x] Bugfix
- [ ] Feature / enhancement
- [ ] CI / Documentation
- [ ] Maintenance

## Checklist

- [x] My code follows the general style and conventions of the codebase, ensuring consistency
- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [x] I have run `pre-commit run --all` locally in the source folder and confirmed that there are no linter errors.
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing tests pass locally with my changes

PR #4158, opened 2026-02-05
maint: YAML environment specification utilities

jjerphan:
# Description

Per-requisite for https://github.com/mamba-org/mamba/pull/4153.

## Type of Change

- [ ] Bugfix
- [ ] Feature / enhancement
- [ ] CI / Documentation
- [x] Maintenance

## Checklist

### DIFF
--- a/libmamba/include/mamba/api/install.hpp
+++ b/libmamba/include/mamba/api/install.hpp
@@ -115,22 +115,10 @@ namespace mamba
 
         bool operator==(const other_pkg_mgr_spec& s1, const other_pkg_mgr_spec& s2);
 
-        struct yaml_file_contents
-        {
-            std::string name;
-            std::vector<std::string> dependencies, channels;
-            std::map<std::string, std::string> variables;
-            std::vector<other_pkg_mgr_spec> others_pkg_mgrs_specs;
-        };
+        // yaml_file_contents moved to environment_yaml.hpp
+        struct yaml_file_contents;  // Forward declaration
 
         bool eval_selector(const std::string& selector, const std::string& platform);
-
-        yaml_file_contents read_yaml_file(
-            const Context& ctx,
-            const std::string& yaml_file,
-            const std::string& platform,
-            bool use_uv
-        );
 
         inline void to_json(nlohmann::json&, const other_pkg_mgr_spec&)
         {
--- a/libmamba/include/mamba/api/install.hpp
+++ b/libmamba/include/mamba/api/install.hpp
@@ -115,10 +115,23 @@ namespace mamba
 
         bool operator==(const other_pkg_mgr_spec& s1, const other_pkg_mgr_spec& s2);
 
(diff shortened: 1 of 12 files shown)


libmamba/src/api/environment_yaml.cpp
```diff
@@ -0,0 +1,571 @@
...
+                // e.g., "https://conda.anaconda.org/conda-forge" -> "conda-forge"
+                if (channel_str.find("conda.anaconda.org/") != std::string::npos)
+                {
+                    auto parts = util::rsplit(channel_str, '/');
+                    if (parts.size() >= 2)
+                    {
+                        return parts[parts.size() - 1];
+                    }
+                }
+            }
```
jjerphan, line 111: We could benefit from using `expected` here, but this would be yet another refactor which would impact other files.

What is the best trade-off? Should we open another PR?

libmamba/src/api/environment_yaml.cpp
```diff
@@ -0,0 +1,571 @@
...
+        }
+
+        // Helper function: Convert PackageInfo to MatchSpec string
+        std::string package_to_spec_string(
+            const specs::PackageInfo& pkg,
+            ChannelContext& channel_context,
+            bool no_builds = false,
+            bool ignore_channels = false,
+            bool include_md5 = false
+        )
```
jjerphan, line 156: This could be a method on `PackageInfo`, but it is only used here and there already are similar free functions; so I am not sure that this is worth it.

libmamba/include/mamba/api/install.hpp
jjerphan, line 1: Are changes made to this file API breakages?

micromamba/src/env.cpp
```diff
@@ -215,79 +223,161 @@ set_env_command(CLI::App* com, mamba::Configuration& config)
...
+                                    std::string key = it.key();
+                                    std::string lower_key = mamba::util::to_lower(key);
+                                    yaml_contents.variables[lower_key] = it.value().get<std::string>();
+                                }
+                            }
+                        }
+                        catch (nlohmann::json::exception&)
+                        {
+                            // If parsing fails, leave variables empty
+                        }
```
jjerphan: Should we warn instead here?

jjerphan (commented):
Some points of discussions.

libmamba/src/api/environment_yaml.cpp
Klaim: I don't think that would help with `channel_context.make_channel` as it is used everywhere as a getter for channel data and it do need to throw as any failure with that call would be a failure.
I prefer to use `expected` when the error is part of the algorithm, for example as return type for parsing functions (there are other examples).

What I'm not sure of is why is there a try-catch here? "resolution" isnt supposed to happen in `make_channel`, so I dont get the meaning of the comment in the `catch`.
Is that why you were thinking of using `expected`?

PR #4161, opened 2026-02-12
feat: Shard Traversal

jjerphan:
# Description

Slice of https://github.com/mamba-org/mamba/pull/4142.

## Type of Change

- [ ] Bugfix
- [x] Feature / enhancement
- [ ] CI / Documentation
- [ ] Maintenance

## Checklist

- [x] My code follows the general style and conventions of the codebase, ensuring consistency
- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [x] I have run `pre-commit run --all` locally in the source folder and confirmed that there are no linter errors.
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing tests pass locally with my changes

PR #4142, opened 2026-01-16
feat: Support Sharded Repodata

jjerphan:
# Description

A solution towards #3431.

Note that this PR was opened for discoverability of and discussions around this working solution.

This PR is not meant to be merged, and the CI runs are likely going to fail: parts are progressively going to be extracted an review independently in different PRs.

## Type of Change

- [ ] Bugfix
- [x] Feature / enhancement
- [ ] CI / Documentation
- [ ] Maintenance

## Checklist

- [x] My code follows the general style and conventions of the codebase, ensuring consistency
- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [x] I have run `pre-commit run --all` locally in the source folder and confirmed that there are no linter errors.
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing tests pass locally with my changes

PR #4162, opened 2026-02-12
feat: Sharded Repodata

jjerphan:
# Description

Slice of https://github.com/mamba-org/mamba/pull/4142.

Comes after https://github.com/mamba-org/mamba/pull/4161 and https://github.com/mamba-org/mamba/pull/4167.

## Type of Change

- [ ] Bugfix
- [x] Feature / enhancement
- [ ] CI / Documentation
- [ ] Maintenance

## Checklist

### DIFF
--- a/libmamba/src/api/utils.cpp
+++ b/libmamba/src/api/utils.cpp
@@ -17,6 +17,7 @@
 #include "mamba/core/output.hpp"
 #include "mamba/core/util.hpp"
 #include "mamba/fs/filesystem.hpp"
+#include "mamba/specs/match_spec.hpp"
 #include "mamba/util/environment.hpp"
 
 #include "utils.hpp"
@@ -237,6 +238,24 @@ namespace mamba
                 }
             }
         }
+    }
+
+    std::vector<std::string> extract_package_names_from_specs(const std::vector<std::string>& specs)
+    {
+        std::vector<std::string> names;
+        for (const auto& s : specs)
+        {
+            auto parsed = specs::MatchSpec::parse(s);
+            if (parsed && !parsed->name().is_free())
+            {
+                auto name = parsed->name().to_string();
+                if (!name.empty())
+                {
+                    names.push_back(std::move(name));
+                }
+            }
+        }
+        return names;
     }
 
     void add_pip_if_python(std::vector<std::string>& root_packages)
--- a/libmamba/src/core/util.cpp
+++ b/libmamba/src/core/util.cpp
@@ -1725,22 +1725,5 @@ namespace mamba
         return copy;
     }
 
-    std::vector<std::string> extract_package_names_from_specs(const std::vector<std::string>& specs)
-    {
-        std::vector<std::string> names;
-        for (const auto& s : specs)
-        {
-            auto parsed = specs::MatchSpec::parse(s);
-            if (parsed && !parsed->name().is_free())
-            {
-                auto name = parsed->name().to_string();
-                if (!name.empty())
-                {
-                    names.push_back(std::move(name));
-                }
-            }
-        }
-        return names;
-    }
 
 }  // namespace mamba
--- a/libmamba/tests/src/core/test_sharded_repodata_integration.cpp
+++ b/libmamba/tests/src/core/test_sharded_repodata_integration.cpp
@@ -32,7 +32,8 @@
 
 #include "mambatests.hpp"
 
-// Forward declaration for utility implemented in core/util.cpp
+// extract_package_names_from_specs is implemented in api/utils.cpp (header is not in public include
+// path)
 namespace mamba
 {
     std::vector<std::string> extract_package_names_from_specs(const std::vector<std::string>& specs);


libmamba/src/api/channel_loader.cpp
```diff
@@ -128,11 +139,174 @@ namespace mamba
...
+                    for (const auto& [filename, record] : shard.conda_packages)
+                    {
+                        packages_by_url[node_id.channel].push_back(
+                            to_package_info(record, filename, channel_id, platform, base_url)
+                        );
+                    }
+                }
+                catch (const std::exception&)
+                {
+                }
```
jjerphan: I am not sure what would be the best way to handle failures (mamba's exception handling is not yet unified.), I guess we could warn?

libmamba/src/api/update.cpp
```diff
@@ -173,7 +197,17 @@ namespace mamba
...
-        auto exp_loaded = load_channels(ctx, channel_context, db, package_caches);
+        auto root_packages = extract_package_names_from_specs(raw_update_specs);
+
+        // When updating python with sharded repodata, also include pip in root packages.
+        // This also avoids pulling old versions of python (1.x) which do not depend on
+        // other packages, which is a choice the solver can make.
+        if (ctx.repodata_use_shards)
+        {
+            add_pip_to_root_packages_if_python_present(root_packages);
+        }
```
jjerphan, line 187: This is usually handled by `add_pip_as_python_dependency` in the `Database`, I need to understand why we need now.

libmamba/src/api/install.cpp
```diff
@@ -624,7 +651,31 @@ namespace mamba
...
+                }
+            }
+
+            // When installing python with sharded repodata, also include pip in root packages.
+            // This also avoids pulling old versions of python (1.x) which do not depend on
+            // other packages, which is a choice the solver can make.
+            if (ctx.repodata_use_shards)
+            {
+                add_pip_to_root_packages_if_python_present(root_packages);
+            }
```
jjerphan, line 656: See https://github.com/mamba-org/mamba/pull/4162/changes#r2832378251.

libmamba/src/api/channel_loader.cpp
jjerphan, line 1: I think we would benefit from first refactoring and documenting implementations of the `load_*` methods.

jjerphan: See https://github.com/mamba-org/mamba/pull/4169.

libmamba/include/mamba/core/shard_traversal.hpp
```diff
@@ -79,7 +79,7 @@ namespace mamba
          */
         void reachable(
             const std::vector<std::string>& root_packages,
-            const std::string& strategy = "pipelined",
+            const std::string& strategy = "bfs",
```
jjerphan, line 82: Two strategies exist for computing the `RepodataSubset` (both are implemented in `conda-libmamba-solver`):
 - `"bfs"`: for exploring the RepodataSubset in breadth
 - `"pipelined"`: a depth-first search alternative

I observed that `"bfs"` runs faster than the `"pipelined"`, but I do not think that we need to maintain this `"pipelined"` strategy necessarily.

libmamba/src/api/install.cpp
```diff
@@ -25,11 +27,36 @@
...
+                has_pip = true;
+            }
+        }
+        if (has_python && !has_pip)
+        {
+            root_packages.push_back("pip");
+        }
+    }
+}
+
```
jjerphan: Is it correct to include this function in an anonymous namespace to avoid having it in the `mamba` namespace?

libmamba/src/api/update.cpp
jjerphan: This might be because the packages' shards are unordered and `python-1.2-0` comes at the top of the package shard for python (currently [`d803a1392fb5c9b09d39a87646cf0abbca0ae13b03d161f6c8ab89a4d0a846b0.msgpack`](https://shards.prefix.dev/conda-forge/d803a1392fb5c9b09d39a87646cf0abbca0ae13b03d161f6c8ab89a4d0a846b0.msgpack))

```json
  "packages": {
    "python-1.2-0.tar.bz2": {
```

So it might be picked first.

jjerphan: I think it would help having a canonical representation with the latest builds (of last version and last build number) being shown first.

libmamba/src/download/mirror_impl.cpp
```diff
@@ -77,9 +77,16 @@ namespace mamba::download
...
+            // If the file corresponds to repodata or repodata shards index, the tag is `latest`,
+            // and there is no need for splitting parts.
+            // Examples:
+            //   linux-64/repodata.json
+            //   linux-64/repodata.json.zst
+            //   linux-64/repodata_shards.msgpack.zst
+            //   noarch/repodata_shards.msgpack.zst
+            if (util::ends_with(path, ".json") || util::ends_with(path, ".json.zst")
+                || util::ends_with(path, "repodata_shards.msgpack")
+                || util::ends_with(path, "repodata_shards.msgpack.zst"))
```
jjerphan: Should we keep hardcoded values in place like this?

libmamba/src/api/update.cpp
jjerphan: Gentle ping @baszalmstra: do you think having sorted shards would make sense and would help? :)

baszalmstra: currently the order of the records is in lexical order. For repodata.json this was useful because it allowed binary searching for specific records of a particular package. For shards this is less useful because its all about one package anyway. 

the lexical ordering does have the benefit that it is "roughly" ordered by version. So the latest are at the bottom.

Im not sure though what enforcing newer versions first gives you here though?

jjerphan: > the lexical ordering does have the benefit that it is "roughly" ordered by version. So the latest are at the bottom.

Yes, currently they are not sorted by version (for instance builds of 3.x comes between the only build of 1.2 and other builds of 2.7).

Sorting (increasingly or decreasingly) would allow implementations to do less work to have ordered collection of builds then.

This is not something we strictly need. I just was curious about whether there was a particular design choice for it or not.

baszalmstra: I see! Yeah I dont think the CEP text says anything about order.

libmamba/src/api/install.cpp
JohanMabille: Putting a function into an anonymous namespace is a way to avoid symbol leaking: your function will be private to the current compilation unit, and nothing else outside could call it, even if someone adds a `extern add_pip_to_root_packages_if_python_present(std::vector<std::string>& root_packages)` declaration and links with this file. So it's not just a matter of not having the function in the mamba namespace. (For the record, the "old" C way of doing this is to define the function as static in the cpp).

jjerphan: Yes (I knew that), I just wanted to know whether it is appropriate or whether it must be done for all utility functions which we want to be only local to their translation unit.

jjerphan: This function is duplicated in `update.cpp`, so we likely need to have it once in the namespace in an appropriate translation unit IMO.

libmamba/include/mamba/core/context.hpp
```diff
@@ -218,6 +220,13 @@ namespace mamba
             };
         }
 
+        /**
+         * Check if status messages can be reported to stdout.
+         * Returns true if Console is available and JSON output is not enabled.
+         * Use this before printing status messages to ensure they don't interfere with JSON output.
+         */
+        [[nodiscard]] static bool can_report_status();
```
JohanMabille: This reintroduce kind of global behavior of the context, while the implementation explicitely relies on the Console. THis method should be implemented in Console directly (this way, it would be easier to remove this latest singleton when we refactor it).

libmamba/include/mamba/core/shard_traversal.hpp
JohanMabille: The additional code for the `"pipelined"` strategy is not that much, so I think it is safe to keep it for now.

libmamba/include/mamba/core/shards.hpp
```diff
@@ -45,14 +45,18 @@ namespace mamba
...
          */
         Shards(
             ShardsIndexDict shards_index,
             std::string url,
             specs::Channel channel,
             specs::AuthenticationDataBase auth_info,
             download::RemoteFetchParams remote_fetch_params,
-            std::size_t download_threads = 10
+            std::size_t download_threads = 10,
+            const download::mirror_map* mirrors = nullptr
```
JohanMabille: I think we should prefer `std::optional` over pointers in APIs (and yeah I know, we haven't been systematic about this in the past, we should fix it too).

libmamba/include/mamba/solver/solution.hpp
```diff
@@ -22,39 +22,46 @@ namespace mamba::solver
         struct Omit
         {
             specs::PackageInfo what;
+            auto operator==(const Omit& other) const -> bool = default;
```
JohanMabille:
```suggestion
            bool operator==(const Omit& other) const = default;
```

JohanMabille: And same for the other `operator==` below, the trailing return syntax make the code heavier for no benefit here.

libmamba/include/mamba/specs/package_info.hpp
```diff
@@ -93,6 +93,13 @@ namespace mamba::specs
     void to_json(nlohmann::json& j, const PackageInfo& pkg);
 
     void from_json(const nlohmann::json& j, PackageInfo& pkg);
+
+    /**
+     * Compare two packages by version first, then by build number.
+     * Returns true if lhs should come before rhs (lhs < rhs).
+     */
+    auto compare_packages_by_version_and_build(const PackageInfo& lhs, const PackageInfo& rhs)
+        -> bool;
```
JohanMabille: No need for trailing return syntax here.

libmamba/src/api/channel_loader.cpp
```diff
@@ -133,11 +143,223 @@ namespace mamba
...
+            ));
+        }
+
+        LOG_DEBUG << "Shard index fetched for " << subdir.name();
+        const auto& channel = subdir.channel();
+        std::string current_repodata_url = subdir.repodata_url().str();
+
+        // For all subdirs sharing the same channel URL, fetch their shard indices and build
+        //    a Shards instance per subdir; collect them into a RepodataSubset.
+        std::vector<std::shared_ptr<Shards>> all_shards;
```
JohanMabille, line 285: I realize I missed that when reviewing #4161 ; why do we need shared pointers here? IIUC, this does not seem to be a shared resource whose lifetime is not deterministic. This does not have to be changed in that PR since it would involde some refactoring, but we should track that in an issue.

libmamba/src/api/channel_loader.cpp
```diff
@@ -133,11 +143,223 @@ namespace mamba
...
+            {
+                LOG_WARNING << "Failed to load package " << node_id.package << " from "
+                            << node_id.channel << ": " << e.what();
+            }
+            catch (const std::exception& e)
+            {
+                LOG_WARNING << "Failed to load package " << node_id.package << " from "
+                            << node_id.channel << ": " << e.what();
+            }
+        }
```
JohanMabille: I think this deserves a dedicated function.

libmamba/src/api/channel_loader.cpp
```diff
@@ -133,11 +143,223 @@ namespace mamba
...
+                repo_name,
+                solver::libsolv::PipAsPythonDependency(ctx.add_pip_as_python_dependency)
+            );
+            std::size_t idx = url_to_subdir_idx[channel_url];
+            database.set_repo_priority(repo, priorities[idx]);
+            if (channel_url == current_repodata_url)
+            {
+                result_repo = std::move(repo);
+            }
+        }
```
JohanMabille: This block also deserves a dedicated function.

libmamba/src/api/channel_loader.cpp
```diff
@@ -133,11 +143,223 @@ namespace mamba
...
+
+        // For each channel URL with packages, add a repo to database (unless already in
+        //    loaded_subdirs_with_shards), sort packages by version/build, and set repo
+        //    priority. The repo for the requested subdir's repodata URL is returned on success.
+        std::optional<solver::libsolv::RepoInfo> result_repo;
+        for (const auto& [channel_url, pkgs] : packages_by_url)
+        {
+            std::string repo_name = subdirs[url_to_subdir_idx[channel_url]].name();
+            if (loaded_subdirs_with_shards.count(repo_name))
```
JohanMabille: Nitpicking: prefer `contains` over `count`

libmamba/src/api/channel_loader.cpp
```diff
@@ -133,11 +143,223 @@ namespace mamba
...
+            catch (const std::runtime_error& e)
+            {
+                LOG_WARNING << "Failed to load package " << node_id.package << " from "
+                            << node_id.channel << ": " << e.what();
+            }
+            catch (const std::exception& e)
+            {
+                LOG_WARNING << "Failed to load package " << node_id.package << " from "
+                            << node_id.channel << ": " << e.what();
+            }
```
JohanMabille: I don't think we need to distinguish all the exception types here since we are logging almost the same thing. The `e.what()`  message should be explicit enough so that we know which kind of exception was thrown.

libmamba/src/api/channel_loader.cpp
```diff
@@ -133,11 +143,223 @@ namespace mamba
...
+        const std::vector<solver::libsolv::Priorities>& priorities
+    ) -> expected_t<solver::libsolv::RepoInfo>
+    {
+        auto& subdir = subdirs[subdir_idx];
+        LOG_DEBUG << "Attempting to load subdir with shards: " << subdir.name();
+
+        // Return error if shards are not applicable (disabled, stale metadata, or no roots).
+        if (!ctx.repodata_use_shards
+            || !subdir.metadata().has_up_to_date_shards(ctx.repodata_shards_ttl)
+            || root_packages.empty())
```
JohanMabille: This is already checked before calling this method, we should avoid redundant defensive code in the implementation.

libmamba/src/api/channel_loader.cpp
```diff
@@ -220,26 +442,92 @@ namespace mamba
...
+                            subdirs_needing_index.end(),
+                            subdir_params
+                        ),
+                        ctx.authentication_info(),
+                        ctx.mirrors,
+                        ctx.download_options(),
+                        ctx.remote_fetch_params,
+                        nullptr
+                    );
+                }
```
JohanMabille: This could be reduced with a ternary operator for the latest argument. LAter we can take the opportunity to replace the pointer by an optional and the the SubdirIndexMonitor instantiated only when needed.

libmamba/src/api/channel_loader.cpp
```diff
@@ -257,41 +545,101 @@ namespace mamba
...
             for (std::size_t i = 0; i < subdirs.size(); ++i)
             {
-                SubdirIndexLoader& subdir_index_loader = subdirs[i];
+                auto& subdir = subdirs[i];
+                bool use_shards = ctx.repodata_use_shards
+                                  && subdir.metadata().has_up_to_date_shards(ctx.repodata_shards_ttl)
+                                  && !root_packages.empty();
+
+                // Skip if this subdir was already loaded as part of a sharded same-channel load.
+                if (loaded_subdirs_with_shards.count(subdir.name()))
```
JohanMabille: Nitpicking: prefer `contains`  over `count`

libmamba/src/api/channel_loader.cpp
JohanMabille, line 664: This funtion is definitely too long, but let's refactor it in a dedicated PR so it is clear which changes related to shards are made here.

libmamba/src/api/utils.cpp
```diff
@@ -238,4 +238,25 @@ namespace mamba
             }
         }
     }
+
+    void add_pip_to_root_packages_if_python_present(std::vector<std::string>& root_packages)
```
JohanMabille: I think `add_pip_if_python` would be enough to understand.

libmamba/src/download/mirror_impl.cpp
JohanMabille: We should probably assign them to constexpr global variables.

JohanMabille (changes requested):

libmamba/include/mamba/core/context.hpp
jjerphan: My bad, I wanted it to be a static class method on `Console`, not `Context`.

libmamba/src/api/channel_loader.cpp
jjerphan: `std::shared_pointer`s are used because `Shards` are also referenced in the `RepodataSubset::m_shards_by_url`.

We can simply store `Shards` in a `std::vector` for `RepodataSubset::m_shards` (i.e. do not define `ShardsPtrVector` i.e. `std::vector<std::shared_ptr<Shards>>`) and simply reference `Shards` in `RepodataSubset::m_shards_by_url` by their indices in `RepodataSubset::m_shards`.

libmamba/src/specs/package_info.cpp
```diff
@@ -590,4 +592,21 @@ namespace mamba::specs
         // TODO: add checks about members that need to have a valid value
         return fmt::format("{}/{}", channel_mirror_platform_url, filename);
     }
+
+    auto compare_packages_by_version_and_build(const PackageInfo& lhs, const PackageInfo& rhs) -> bool
```
jjerphan: We could turn this in `PackageInfo::operator<`.

libmamba/src/api/channel_loader.cpp
JohanMabille: Yes, if we can have a clear ownership model, I think it is better and easier to maintain than relying on shared pointers. Let's rework this in a dedicated PR.

libmamba/src/specs/package_info.cpp
JohanMabille: In that case we should probably implement this through `operator<=>`  to get other comparison operators for free. But I'm not sure of the benefits, so let's do that in a dedicated PR if we decide to do so. Feel free to open issues to track this to.

jjerphan: https://github.com/mamba-org/mamba/issues/4176 was opened accordingly.

I will document comparison of `PackageInfo` then.

PR #4163, opened 2026-02-13
feat: Redesign packages caches hierarchically on channels and platforms

jjerphan:
# Description

## Problem

Packages were cached globally by filename, without separating them by channel. That caused verification failures when the same package name came from different channels with different URLs or checksums (e.g. different mirrors or channel URLs).

Example: `python_abi-3.14-8_cp314.conda` from `prefix.dev` and from another channel could end up in the same cache entry, leading to wrong checksums and cache errors.

## Solution

Packages are now cached via a folder hierarchy.

### Before

```bash
pkgs/
├── urls.txt
├── package_name-version-build.tar.bz2      # tarball
├── package_name-version-build/             # extracted (same base name)
│   └── info/
│       ├── index.json
│       └── repodata_record.json
├── another_package-1.0-0.tar.bz2
├── another_package-1.0-0/
│   └── ...
```

All packages shared the same directory. Collisions occurred when the same package name came from different channels. This is the folder hierarchy used by conda's package cache.

### After

```bash
pkgs/
├── urls.txt
├── <channel>/                                   # e.g. conda-forge, https/repo.example.com/channel
│   └── <platform>/                              # e.g. linux-64, noarch, osx-64
│       ├── package_name-version-build.tar.bz2   # tarball
│       └── package_name-version-build/          # extracted (same base name)
│           └── info/
│               └── repodata_record.json
```

### UX change

The cache layout has changed. Existing caches under the old layout are no longer used. Users will have re-download packages after upgrading.

This is not a breaking change.

## Type of Change

- [ ] Bugfix
- [x] Feature / enhancement
- [ ] CI / Documentation
- [ ] Maintenance

## Checklist

### DIFF
--- a/libmamba/src/core/util.cpp
+++ b/libmamba/src/core/util.cpp
@@ -1674,18 +1674,26 @@ namespace mamba
         std::string copy(str);
 
         // Remove token: /t/token -> remove /t/token entirely
+        // Also handle masked tokens: /t/*****
         if (util::contains(str, "/t/"))
         {
             copy = std::regex_replace(copy, token_regex(), "");
+            // Remove masked token pattern if present (escape * for regex)
+            static const std::regex masked_token_regex{ "/t/\\*{5}" };
+            copy = std::regex_replace(copy, masked_token_regex, "");
         }
 
         // Remove password from basic auth: user:password@domain -> preserve scheme and domain
         // Handle both URLs with scheme (e.g., https://user:pass@domain) and without scheme
         // (e.g., user:pass@domain at start or after whitespace)
+        // Also handle URLs that already have masked passwords (user:*****@domain)
+        // The regex [^@\s]+ will match both real passwords and masked passwords (*****)
+
         // Pattern 1: Match URLs with scheme: (scheme://)(user):(password)@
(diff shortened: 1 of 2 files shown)


libmamba/src/core/package_cache.cpp
```diff
@@ -20,6 +20,24 @@
...
+    auto package_cache_channel_id(const specs::PackageInfo& s) -> std::string
+    {
+        std::string channel = s.channel.empty() ? "no_channel" : s.channel;
+        for (char& c : channel)
+        {
+            if (c == '/' || c == ':' || c == '\\')
+            {
+                c = '_';
+            }
+        }
```
JohanMabille:
```suggestion
        std::ranges::replace_if(channel, [](char c) { return c == '/' || c == ':' || c == '\\'; }, '_');

```

SylvainCorlay:
How does this work for users making use of mamba / conda interchangeably?
Is mamba basically having its own separate cache now?

jjerphan:
Yes, and this is interoperable, conda can continue to use a flat cache at the root (which is totally improper IMO); more details are given in the docstring.

PR #4167, opened 2026-02-19
feat: Support `MSGPACK_OBJECT_ARRAY` for hashes

jjerphan:
# Description

Follow-up of #4141.

It seems that hashes' binary representations use `MSGPACK_OBJECT_ARRAY` with `MSGPACK_OBJECT_POSITIVE_INTEGER` instead of `MSGPACK_OBJECT_BIN` or `MSGPACK_OBJECT_EXT`.

This adds support for other representations, covering the observed one.

## Type of Change

- [ ] Bugfix
- [x] Feature / enhancement
- [ ] CI / Documentation
- [ ] Maintenance

## Checklist

### DIFF
--- a/libmamba/src/core/shards.cpp
+++ b/libmamba/src/core/shards.cpp
@@ -138,12 +138,6 @@ namespace mamba
                     {
                         bytes.push_back(static_cast<std::byte>(elem.via.u64 & 0xFF));
                     }
-                    else if (elem.type == MSGPACK_OBJECT_NEGATIVE_INTEGER)
-                    {
-                        // Handle negative integers (shouldn't happen for bytes, but handle
-                        // gracefully)
-                        bytes.push_back(static_cast<std::byte>(std::abs(elem.via.i64) & 0xFF));
-                    }
                     else
                     {
                         LOG_WARNING
--- a/libmamba/tests/src/core/test_shards.cpp
+++ b/libmamba/tests/src/core/test_shards.cpp
@@ -1228,9 +1228,10 @@ TEST_CASE("Shard parsing - Hash format edge cases")
         msgpack_zone_destroy(unpacked.zone);
     }
 
-    SECTION("Parse sha256 as array with negative integers")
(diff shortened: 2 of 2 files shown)


libmamba/src/core/shards.cpp
```diff
@@ -114,16 +120,62 @@ namespace mamba
...
+                        bytes.push_back(static_cast<std::byte>(std::abs(elem.via.i64) & 0xFF));
+                    }
+                    else
+                    {
+                        LOG_WARNING
+                            << "Array element " << i << " in " << field_name
+                            << " is not a positive integer (type: " << static_cast<int>(elem.type)
+                            << "), cannot convert to bytes";
+                        return std::string();
+                    }
```
jjerphan, line 148: We can simplify this part to if we want to assume that `MSGPACK_OBJECT_POSITIVE_INTEGER` is always used (and remove all the cases above, i.e. `MSGPACK_OBJECT_{STR,BIN,EXT}`).
```suggestion
                for (std::uint32_t i = 0; i < obj.via.array.size; ++i)
                {
                    const msgpack_object& elem = obj.via.array.ptr[i];
                    bytes.push_back(static_cast<std::byte>(elem.via.u64 & 0xFF));
                }
```

JohanMabille: I think we should handle the error case where the `elem.type` is not a positive integer; meaning `MSGPACK_OBJECT_NEGATIVE_INTEGER` should be considered as any other invalid type and we log and return an empty string.

jjerphan: Updated accordingly with 925737d.

jjerphan: Should we still support the other types, such as `MSGPACK_OBJECT_{STR,BIN,EXT}`?

I am +0.5 regarding removing them (this would remove actually unused but potentially future-proof code).

JohanMabille: Let's be conservative and keep them for now.

PR #4168, opened 2026-02-19
feat: Consider `emscripten` a UNIX platform

jjerphan:
# Description

## Type of Change

- [ ] Bugfix
- [x] Feature / enhancement
- [ ] CI / Documentation
- [ ] Maintenance

## Checklist

### DIFF
--- a/libmamba/src/core/virtual_packages.cpp
+++ b/libmamba/src/core/virtual_packages.cpp
@@ -354,16 +354,6 @@ namespace mamba
             return util::windows_version();
         }
 
-        [[nodiscard]] auto overridable_emscripten_version() -> std::string
-        {
-            if (auto override_version = util::get_env("CONDA_OVERRIDE_EMSCRIPTEN"))
-            {
-                return std::move(override_version).value();
-            }
-            // 4.x is assumed to be the latest version used by emscripten-forge.
-            return { "4.x" };
-        }
-
         std::vector<specs::PackageInfo> dist_packages(const std::string& platform)
         {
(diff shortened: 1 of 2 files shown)


libmamba/src/core/virtual_packages.cpp
```diff
@@ -354,6 +354,16 @@ namespace mamba
...
         }
 
+        [[nodiscard]] auto overridable_emscripten_version() -> std::string
+        {
+            if (auto override_version = util::get_env("CONDA_OVERRIDE_EMSCRIPTEN"))
+            {
+                return std::move(override_version).value();
+            }
+            // 4.x is assumed to be the latest version used by emscripten-forge.
+            return { "4.x" };
```
martinRenou: Could this be the same as the `emscripten-abi` version from the solve result? Otherwise not sure it makes sense here.

jjerphan: Yes, as discussed privately, this version is just placed but not actually used (AFAIK).


I'll add a comment to explain and motivates those changes in details.

JohanMabille: If the version is not used, it should be set to something that cannot be considered as a correct version. The comment explaining this is part of the caller instead of this function (which makes sense since it's part of a  more global explanation), so maybe it would be better to pass the default version as an argument of this function?

jjerphan: As discussed privately, we do not need `__emscripten`, so let's keep something simple.

See fce466221a16358c988e1bfc41613bd19b3e502a.

PR #4177, opened 2026-02-27
maint: Adapt `Shards` ownership model

jjerphan:
# Description

Follow-up of https://github.com/mamba-org/mamba/pull/4162#discussion_r2863540467.

## Type of Change

- [ ] Bugfix
- [ ] Feature / enhancement
- [ ] CI / Documentation
- [x] Maintenance

## Checklist

### DIFF
--- a/libmamba/src/core/shard_traversal.cpp
+++ b/libmamba/src/core/shard_traversal.cpp
@@ -92,10 +92,13 @@ namespace mamba
     RepodataSubset::RepodataSubset(std::vector<Shards> shards)
         : m_shards(std::move(shards))
     {
-        for (std::size_t i = 0; i < m_shards.size(); ++i)
-        {
-            m_shards_by_url[m_shards[i].url()] = i;
-        }
+        // Keep shards ordered by URL to ensure deterministic traversal
+        // Sorting here is inexpensive because we only have a few Subdir to explore (typically 2)
+        std::sort(
+            m_shards.begin(),
+            m_shards.end(),
+            [](const Shards& lhs, const Shards& rhs) { return lhs.url() < rhs.url(); }
+        );
     }
 
     void RepodataSubset::reachable(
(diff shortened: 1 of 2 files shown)


libmamba/src/core/shard_traversal.cpp
```diff
@@ -89,12 +89,12 @@ namespace mamba
...
+    RepodataSubset::RepodataSubset(std::vector<Shards> shards)
     {
-        for (const auto& s : m_shards)
+        for (auto& s : shards)
         {
-            m_shards_by_url[s->url()] = s;
+            auto url = s.url();
+            m_url_keyed_shards.emplace(std::move(url), std::move(s));
         }
     }
```
jjerphan, line 102: Actually, this can be quite costly. I think. We would benefit from keeping the `std::vector<Shards>` and not moving structures.

jjerphan: See 3027e45e3491665baef69f8821d249daf7923b7e.

JohanMabille: I don't think we need this map. If the goal is to traverse the shard by URLs order, then we can simply sort the vector upon construction.

JohanMabille (changes requested):

libmamba/src/core/shard_traversal.cpp
jjerphan: Accesses would still be costlier ( $\mathcal{O}(\log n)$ instead of $\mathcal{O}(1)$ ).

jjerphan: How about adding some comment motivating datastructure involved explaining trade-offs?

jjerphan: Actually, the numbers of `Shards` elements stored is equal to the number of subdirs considered; so using a vector should be faster. Let me try.

jjerphan: Fixed with 27d4c5a.

JohanMabille: > Accesses would still be costlier ( O ( log ⁡ n ) instead of O ( 1 ) ).

std::map access is O(log(n)) too.

JohanMabille:
The mamba-content-trust and windows tests failure are unrelated to this PR.

libmamba/src/core/shard_traversal.cpp
jjerphan: I yes, I reported for a `std::unordered_map`.

PR #4191, opened 2026-03-12
feat: Multithreaded Shard Parsing

jjerphan:
# Description

Experimental PR for multi-threaded shard parsing. Not required, a nice to have to explore potentially in the future.

## Type of Change

- [ ] Bugfix
- [x] Feature / enhancement
- [ ] CI / Documentation
- [ ] Maintenance

## Checklist

### DIFF
--- a/libmamba/include/mamba/util/concurrent_map.hpp
+++ b/libmamba/include/mamba/util/concurrent_map.hpp
@@ -1,0 +1,157 @@
+// Copyright (c) 2024, QuantStack and Mamba Contributors
+//
+// Distributed under the terms of the BSD 3-Clause License.
+//
+// The full license is in the file LICENSE, distributed with this software.
+
+#ifndef MAMBA_UTIL_CONCURRENT_MAP_HPP
+#define MAMBA_UTIL_CONCURRENT_MAP_HPP
+
+#include <array>
+#include <functional>
+#include <optional>
+#include <shared_mutex>
+#include <unordered_map>
+
+namespace mamba::util
+{
+
+    /**
+     * A thread-safe map with stripe locking for concurrent access.
+     *
+     * Uses multiple mutexes (stripes) to allow concurrent operations on different keys.
+     * Keys are distributed across stripes via hashing, so operations on different keys
+     * typically lock different mutexes and can proceed in parallel.
+     *
+     * @tparam Key Key type (must be hashable)
+     * @tparam Value Value type
+     * @tparam Hash Hash function for key distribution across stripes
+     * @tparam NumStripes Number of lock stripes (default 64)
+     */
+    template <typename Key, typename Value, typename Hash = std::hash<Key>, std::size_t NumStripes = 64>
+    class concurrent_map
+    {
+    public:
+
+        using key_type = Key;
+        using mapped_type = Value;
+        using value_type = std::pair<const Key, Value>;
+        using map_type = std::unordered_map<Key, Value, Hash>;
+
+        concurrent_map() = default;
+
+        /** Insert or assign a key-value pair. Thread-safe. */
+        void insert_or_assign(const Key& key, Value value)
+        {
+            auto& s = stripe(key);
+            std::lock_guard lock(s.mutex);
+            s.map.insert_or_assign(key, std::move(value));
+        }
+
+        /** Find a key. Returns value if found, std::nullopt otherwise. Thread-safe. */
+        auto find(const Key& key) const -> std::optional<Value>
+        {
+            auto& s = stripe(key);
+            std::shared_lock lock(s.mutex);
+            auto it = s.map.find(key);
+            if (it == s.map.end())
+            {
+                return std::nullopt;
+            }
+            return it->second;
+        }
+
+        /** Check if key exists. Thread-safe. */
+        auto contains(const Key& key) const -> bool
+        {
+            auto& s = stripe(key);
+            std::shared_lock lock(s.mutex);
+            return s.map.find(key) != s.map.end();
+        }
+
+        /**
+         * Apply a function to each key-value pair.
+         * Locks all stripes for the duration. Use for iteration/snapshot.
+         * Thread-safe but blocks all other operations during the call.
+         */
+        template <typename Func>
+        void for_each(Func&& func) const
+        {
+            for (auto& s : m_stripes)
+            {
+                std::shared_lock lock(s.mutex);
+                for (const auto& [k, v] : s.map)
+                {
+                    func(k, v);
+                }
+            }
+        }
+
+        /**
+         * Copy all entries into a standard map.
+         * Useful when a snapshot is needed for further processing.
+         */
+        auto copy() const -> map_type
+        {
+            map_type result;
+            for_each([&result](const Key& k, const Value& v) { result[k] = v; });
+            return result;
+        }
+
+        /** Copy constructor. */
+        concurrent_map(const concurrent_map& other)
+        {
+            other.for_each([this](const Key& k, const Value& v) { insert_or_assign(k, v); });
+        }
+
+        /** Copy assignment. */
+        auto operator=(const concurrent_map& other) -> concurrent_map&
+        {
+            if (this != &other)
+            {
+                clear();
+                other.for_each([this](const Key& k, const Value& v) { insert_or_assign(k, v); });
+            }
+            return *this;
+        }
+
+        /** Move constructor. */
+        concurrent_map(concurrent_map&&) noexcept = default;
+
+        /** Move assignment. */
+        auto operator=(concurrent_map&&) noexcept -> concurrent_map& = default;
+
+        /** Remove all entries. */
+        void clear()
+        {
+            for (auto& s : m_stripes)
+            {
+                std::lock_guard lock(s.mutex);
+                s.map.clear();
+            }
+        }
+
+    private:
+
+        struct stripe_type
+        {
+            mutable std::shared_mutex mutex;
+            map_type map;
+        };
+
+        mutable std::array<stripe_type, NumStripes> m_stripes;
+
+        auto stripe(const Key& key) -> stripe_type&
+        {
+            return m_stripes[Hash{}(key) % NumStripes];
+        }
+
+        auto stripe(const Key& key) const -> stripe_type&
+        {
+            return m_stripes[Hash{}(key) % NumStripes];
+        }
+    };
+
+}  // namespace mamba::util
+
+#endif
--- a/libmamba/include/mamba/core/shards.hpp
+++ b/libmamba/include/mamba/core/shards.hpp
@@ -23,6 +23,7 @@
 #include "mamba/fs/filesystem.hpp"
 #include "mamba/specs/authentication_info.hpp"
 #include "mamba/specs/channel.hpp"
+#include "mamba/util/concurrent_map.hpp"
 
 namespace mamba
 {
@@ -134,8 +135,8 @@ namespace mamba
         /** Optional base mirrors for channel-based downloads. */
         std::optional<std::reference_wrapper<const download::mirror_map>> m_mirrors;
 
-        /** Visited shards, keyed by package name. */
-        std::map<std::string, ShardDict> m_visited;
+        /** Visited shards, keyed by package name. Uses concurrent map for parallel writes. */
+        util::concurrent_map<std::string, ShardDict> m_visited;
 
         /** Cached shards_base_url. */
         mutable std::optional<std::string> m_shards_base_url;
@@ -149,7 +150,7 @@ namespace mamba
         /** Directory for cached shard files: {pkgs_cache_root}/cache/shards */
         fs::u8path m_shard_cache_dir;
 
-        /** Mutex guarding mutable state such as m_visited and cached URLs.
+        /** Mutex guarding cached URLs (m_shards_base_url, m_base_url_cache).
          * Uses unique_ptr to make Shards movable (required for std::vector<Shards>). */
         mutable std::unique_ptr<std::mutex> m_mutex;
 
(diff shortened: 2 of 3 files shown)


libmamba/include/mamba/util/concurrent_map.hpp
```diff
@@ -0,0 +1,157 @@
...
+        };
+
+        mutable std::array<stripe_type, NumStripes> m_stripes;
+
+        auto stripe(const Key& key) -> stripe_type&
+        {
+            return m_stripes[Hash{}(key) % NumStripes];
+        }
+
+        auto stripe(const Key& key) const -> stripe_type&
```
Klaim, line 149:
```suggestion
        auto stripe(const Key& key) const -> const stripe_type&
```

libmamba/include/mamba/util/concurrent_map.hpp
```diff
@@ -0,0 +1,157 @@
...
+            for (auto& s : m_stripes)
+            {
+                std::lock_guard lock(s.mutex);
+                s.map.clear();
+            }
+        }
+
+    private:
+
+        struct stripe_type
```
Klaim, line 136: I would prefer `synchronized_value<map_type, std::shared_mutex>` as it would be easier to prevent unsafe access.

libmamba/include/mamba/util/concurrent_map.hpp
```diff
@@ -0,0 +1,157 @@
...
+        }
+
+        /**
+         * Copy all entries into a standard map.
+         * Useful when a snapshot is needed for further processing.
+         */
+        auto copy() const -> map_type
+        {
+            map_type result;
+            for_each([&result](const Key& k, const Value& v) { result[k] = v; });
```
Klaim, line 97: I think using [`.merge`](https://en.cppreference.com/w/cpp/container/unordered_map/merge.html) instead of going element per element would be more performant and make more sense here.

libmamba/include/mamba/util/concurrent_map.hpp
```diff
@@ -0,0 +1,157 @@
...
+            return result;
+        }
+
+        /** Copy constructor. */
+        concurrent_map(const concurrent_map& other)
+        {
+            other.for_each([this](const Key& k, const Value& v) { insert_or_assign(k, v); });
```
Klaim, line 104: Consider using an algorithm based on [`.merge`](https://en.cppreference.com/w/cpp/container/unordered_map/merge.html) instead of per-element inserts?

libmamba/include/mamba/util/concurrent_map.hpp
```diff
@@ -0,0 +1,157 @@
...
+         */
+        auto copy() const -> map_type
+        {
+            map_type result;
+            for_each([&result](const Key& k, const Value& v) { result[k] = v; });
+            return result;
+        }
+
+        /** Copy constructor. */
+        concurrent_map(const concurrent_map& other)
```
Klaim, line 102: Not locking all the mutexes of `other` while going through it means the copy isn't guaranteed to be a copy at the end of the expression, as values could have been added concurrently. I think you really need to lock everything for the execution of the whole copy.

libmamba/include/mamba/util/concurrent_map.hpp
```diff
@@ -0,0 +1,157 @@
...
+     * Keys are distributed across stripes via hashing, so operations on different keys
+     * typically lock different mutexes and can proceed in parallel.
+     *
+     * @tparam Key Key type (must be hashable)
+     * @tparam Value Value type
+     * @tparam Hash Hash function for key distribution across stripes
+     * @tparam NumStripes Number of lock stripes (default 64)
+     */
+    template <typename Key, typename Value, typename Hash = std::hash<Key>, std::size_t NumStripes = 64>
+    class concurrent_map
```
Klaim, line 32: Did you consider a more precise name to reflect the properties of this kind of concurrent map?

libmamba/include/mamba/util/concurrent_map.hpp
```diff
@@ -0,0 +1,157 @@
...
+     * Uses multiple mutexes (stripes) to allow concurrent operations on different keys.
+     * Keys are distributed across stripes via hashing, so operations on different keys
+     * typically lock different mutexes and can proceed in parallel.
+     *
+     * @tparam Key Key type (must be hashable)
+     * @tparam Value Value type
+     * @tparam Hash Hash function for key distribution across stripes
+     * @tparam NumStripes Number of lock stripes (default 64)
+     */
+    template <typename Key, typename Value, typename Hash = std::hash<Key>, std::size_t NumStripes = 64>
```
Klaim, line 31: Any new templates should use named concepts or at least requirements to the template type parameters.
Here for example you could express the requirements for `Key` to be hashable as 
```
requires(const Key& key) { Hash{}(key) -> std::convertible_to<std::size_t>; }
```
`Value` also needs to be semi-regular at least. Etc.

libmamba/include/mamba/util/concurrent_map.hpp
```diff
@@ -0,0 +1,157 @@
...
+    public:
+
+        using key_type = Key;
+        using mapped_type = Value;
+        using value_type = std::pair<const Key, Value>;
+        using map_type = std::unordered_map<Key, Value, Hash>;
+
+        concurrent_map() = default;
+
+        /** Insert or assign a key-value pair. Thread-safe. */
```
Klaim, line 43: Does not having "Thread-safe" in the comment means if you dont have it in the comments of the functions below, they are not "thread-safe"? If so could you add a note on the class comment, something like "Only the member functions described as Thread-Safe can by concurrent threads." 
Several of the functions below are not marked "thead-safe" and it is not clear to me if it was the intent or not.

libmamba/include/mamba/util/concurrent_map.hpp
```diff
@@ -0,0 +1,157 @@
...
+                    func(k, v);
+                }
+            }
+        }
+
+        /**
+         * Copy all entries into a standard map.
+         * Useful when a snapshot is needed for further processing.
+         */
+        auto copy() const -> map_type
```
Klaim, line 94: I would prefer this function to be named differently as it is not really a copy. Maybe `snapshot` or `capture` or something in that direction?
Also note that this is not a true snapshot as values can be inserted while it is gathering the values.

libmamba/include/mamba/util/concurrent_map.hpp
```diff
@@ -0,0 +1,157 @@
...
+        }
+
+        /** Copy constructor. */
+        concurrent_map(const concurrent_map& other)
+        {
+            other.for_each([this](const Key& k, const Value& v) { insert_or_assign(k, v); });
+        }
+
+        /** Copy assignment. */
+        auto operator=(const concurrent_map& other) -> concurrent_map&
```
Klaim, line 108: Same comments as for copy-constructions.

libmamba/include/mamba/util/concurrent_map.hpp
```diff
@@ -0,0 +1,157 @@
...
+            if (this != &other)
+            {
+                clear();
+                other.for_each([this](const Key& k, const Value& v) { insert_or_assign(k, v); });
+            }
+            return *this;
+        }
+
+        /** Move constructor. */
+        concurrent_map(concurrent_map&&) noexcept = default;
```
Klaim, line 119: Note that while these move operations are not marked as thread-safe, they could be implemented as thread-safe. Here they are don't seem to be, but it's not clear to me if that was the intent.

libmamba/include/mamba/util/concurrent_map.hpp
```diff
@@ -0,0 +1,157 @@
...
+            }
+            return it->second;
+        }
+
+        /** Check if key exists. Thread-safe. */
+        auto contains(const Key& key) const -> bool
+        {
+            auto& s = stripe(key);
+            std::shared_lock lock(s.mutex);
+            return s.map.find(key) != s.map.end();
```
Klaim, line 69: Not sure if it's available in all our supported build toolchains, but in C++20 unordered_map has [`.contains`](https://en.cppreference.com/w/cpp/container/unordered_map/contains.html).

libmamba/include/mamba/util/concurrent_map.hpp
```diff
@@ -0,0 +1,157 @@
...
+        }
+
+        /** Move constructor. */
+        concurrent_map(concurrent_map&&) noexcept = default;
+
+        /** Move assignment. */
+        auto operator=(concurrent_map&&) noexcept -> concurrent_map& = default;
+
+        /** Remove all entries. */
+        void clear()
```
Klaim, line 125: The container is not guaranteed to be empty after calling this function. Is that intentional?

libmamba/include/mamba/core/context.hpp
```diff
@@ -165,8 +166,17 @@ namespace mamba
 
         download::Options download_options() const
         {
+            auto dt = this->threads_params.download_threads;
```
Klaim, line 169: Potential UX improvement that I saw recently in another tool, can be ignored:
Allow negative values, in which case we would use `std::max(hardware_concurrency() - dt, 1)`.
That way users can express that they want all the available computation units except a few to be used.

libmamba/src/api/configuration.cpp
```diff
@@ -1052,15 +1053,37 @@ namespace mamba
 
         void download_threads_hook(std::size_t& value)
         {
-            if (!value)
+            if (value == 0)
             {
-                throw std::runtime_error(
-                    fmt::format("Number of download threads as to be positive (currently set to {})", value)
-                );
+                value = std::thread::hardware_concurrency();
```
Klaim, line 1058:
```suggestion
                value = std::max(std::thread::hardware_concurrency(), 1);
```

libmamba/include/mamba/core/context.hpp
```diff
@@ -165,8 +166,17 @@ namespace mamba
...
+            if (dt == 0)
+            {
+                dt = std::thread::hardware_concurrency();
```
Klaim, line 172:
```suggestion
                dt = std::max(std::thread::hardware_concurrency(), 1);
```

libmamba/src/api/configuration.cpp
Klaim: Also applies to the functions below.

libmamba/src/core/shard_index_loader.cpp
```diff
@@ -503,6 +504,43 @@ namespace mamba
...
+        // without HEAD check or download. This skips network when metadata state
+        // was lost (e.g. fresh shell) but cache file is still valid.
+        fs::u8path cache_path = shard_index_cache_path(subdir);
+        if (fs::exists(cache_path) && shards_ttl > 0)
+        {
+            try
+            {
+                auto last_write = fs::last_write_time(cache_path);
+                auto now = fs::file_time_type::clock::now();
+                auto age = std::chrono::duration_cast<std::chrono::seconds>(now - last_write).count();
```
Klaim, line 517:
```suggestion
                const auto age_secs = std::chrono::duration_cast<std::chrono::seconds>(now - last_write).count();
```

libmamba/src/core/shard_index_loader.cpp
```diff
@@ -503,6 +504,43 @@ namespace mamba
...
     {
+        // Check cache first: if we have a cached index within TTL, use it directly
+        // without HEAD check or download. This skips network when metadata state
+        // was lost (e.g. fresh shell) but cache file is still valid.
+        fs::u8path cache_path = shard_index_cache_path(subdir);
+        if (fs::exists(cache_path) && shards_ttl > 0)
+        {
+            try
+            {
+                auto last_write = fs::last_write_time(cache_path);
```
Klaim, line 515: Add some `const`s for the following values. :)

libmamba/src/core/shard_traversal.cpp
```diff
@@ -43,6 +46,72 @@ namespace mamba
...
+                    const std::size_t start = begin;
+                    const std::size_t end = start + chunk;
+                    begin = end;
+
+                    if (start >= end)
+                    {
+                        continue;
+                    }
+
+                    auto task = std::make_shared<std::packaged_task<void()>>([start, end, func]() mutable
```
Klaim, line 98: If I understand correctly your intent, you don't need to make a shared-pointer, you can move the packaged task into the lambda using:
```
[task=std::move(task)]{ ... }
```

libmamba/src/core/shard_traversal.cpp
```diff
@@ -43,6 +46,72 @@ namespace mamba
...
+                                                                             { func(start, end); });
+                    futures.push_back(task->get_future());
+                    MainExecutor::instance().schedule([task]() { (*task)(); });
```
Klaim, line 101: I think it would help if we have the usual generic future-returning `MainExecutor::schedule` variation. I didn't add it before because we didnt have much use of such function so far but now seems to be a good time. I'll make a short pr and your code will be simpler.

libmamba/src/core/shard_traversal.cpp
```diff
@@ -43,6 +46,72 @@ namespace mamba
...
+                }
+                if (max_workers == 0)
+                {
+                    max_workers = 1;
+                }
+                m_max_workers = max_workers;
+            }
+
+            template <class Func>
+            void parallel_for(std::size_t count, Func&& func) const
```
Klaim, line 67: Did you consider implementing this using `std::for_each(std::execution::par_unseq, ...` or similar algorithm (whatever is adequate) + parallel policy? 
That would avoid having to implement it yourself and the underlying implementation will automatically avoid thread contention.

jjerphan:
Shard parsing is not where the cost is (`libsolv` takes most of the time now).

Closing but might potentially revisit.

PR #4193, opened 2026-03-17
fix: Respect thread affinity concurrency

jjerphan:
# Description

## Type of Change

- [x] Bugfix
- [ ] Feature / enhancement
- [ ] CI / Documentation
- [ ] Maintenance

## Checklist

### DIFF
--- a/libmamba/include/mamba/core/context_params.hpp
+++ b/libmamba/include/mamba/core/context_params.hpp
@@ -42,9 +42,7 @@ namespace mamba
 
     struct ThreadsParams
     {
-        // 0 means: auto, use process-affinity-based concurrency (see
-        // normalize_to_affinity_concurrency()).
-        std::size_t download_threads{ 0 };
+        std::size_t download_threads{ 5 };
         int extract_threads{ 0 };
     };
 


libmamba/include/mamba/core/thread_utils.hpp
```diff
@@ -67,6 +68,20 @@ namespace mamba
...
+     * On platforms where we can query a process affinity mask (currently Linux and Windows),
+     * this takes the current affinity mask into account. For example, running
+     * `taskset -c 0 micromamba` on Linux or constraining the process with
+     * `SetProcessAffinityMask` on Windows will report a single available thread even if
+     * the machine has more cores.
+     *
+     * On other platforms (such as macOS), this falls back to
+     * `std::thread::hardware_concurrency()`. In all cases, the return value is at least 1.
+     */
+    int get_affinity_concurrency();
```
JohanMabille: This should return a `std::size_t` (and that would reduce the number of casts required).

jjerphan: I am/was thinking of managing the semantic of `extract_threads` which uses `int` to be able to have negative values (I am looking about how to best handle it ATM).

libmamba/include/mamba/core/thread_utils.hpp
```diff
@@ -67,6 +68,23 @@ namespace mamba
...
 
+    /**
+     * Cap a requested thread count to the process's effective concurrency.
+     *
+     * The `threads` parameter follows the same semantics as `extract_threads`:
+     * - `threads > 0`: request exactly this many threads;
+     * - `threads == 0`: request the maximum concurrency available to this process;
+     * - `threads < 0`: request (available_concurrency + threads).
+     *
+     * The result is always clamped to the range \[1, available_concurrency\], where
```
Klaim: I'm not sure I'm at ease with *always* clamping here.
My usual expectation is that if when the value is positive the user has total control on that number and we should not clamp it. Whatever their reason, they should be able to set what they want and we should comply. If user decides to assign 4 times the number of logical cores, they should get it whatever the consequences.
That's also what most tools do AFAIK?
After all the affinity and even using `hardware_concurrency` are best guess at optimal maximum, it's not necessarily ideal for all contexts and we should give the user a way to bypass our assumptions?

I'm not sure.

libmamba/include/mamba/core/thread_utils.hpp
```diff
@@ -67,6 +68,23 @@ namespace mamba
...
+     * `available_concurrency` is the number of hardware threads effectively available
+     * to this process. On platforms where we can query a process affinity mask
+     * (currently Linux and Windows), this takes the current affinity mask into account
+     * (e.g. `taskset -c 0 micromamba` or `SetProcessAffinityMask`). On other platforms
+     * (such as macOS), it falls back to `std::thread::hardware_concurrency()`.
+     */
+    std::size_t cap_to_affinity_concurrency(int requested_n_threads = 0);
```
JohanMabille: Why not `std::ptrdiff_t` ? That would be consistent with the `set_max`  signature and would avoid a cast.

libmamba/include/mamba/core/thread_utils.hpp
```diff
@@ -210,19 +228,8 @@ namespace mamba
...
-        else if (value < 0)
-        {
-            new_max = std::thread::hardware_concurrency() + value;
-        }
-        else
-        {
-            new_max = value;
-        }
+        const auto capped = cap_to_affinity_concurrency(static_cast<int>(value));
+        const std::ptrdiff_t new_max = static_cast<std::ptrdiff_t>(capped);
```
JohanMabille: With the previous change taht could be simplified into
```cpp
const auto new_max = static_cast<std::ptrdiff_t>(cap_to_affinityèconcurrency(value));
```

libmamba/src/core/thread_utils.cpp
```diff
@@ -158,6 +167,75 @@ namespace mamba
...
+                }
+                if (count > 0)
+                {
+                    return static_cast<std::size_t>(count);
+                }
+            }
+#endif
+
+            const auto hc = std::thread::hardware_concurrency();
+            return hc > 0 ? static_cast<std::size_t>(hc) : std::size_t{ 1 };
```
JohanMabille: `hardware_concurrency`  returns an `unsigned int`, so this statement is not necessary, you can simply the casted value (too bad that it returns `unsigned int` instead of `size_t` ...)

libmamba/include/mamba/core/thread_utils.hpp
jjerphan: I agree that we should preserve the values set by the users and to only constrain the values otherwise.

libmamba/src/core/thread_utils.cpp
```diff
@@ -158,6 +168,86 @@ namespace mamba
...
+            requested = available;
+        }
+        else if (requested_n_threads < 0)
+        {
+            requested = available + requested_n_threads;
+        }
+        else
+        {
+            requested = requested_n_threads;
+        }
```
JohanMabille: This can be removed since it is handled a early return. I think a regular if / else branch would make this code easier to read (because the 3 cases would be handled in the same structure.

jjerphan:
The test suite now takes a lot more time to run (I need to understand why).

jjerphan:
e0304b3 reverted the change of behavior: now the test suite takes as much time as before.

PR #4195, opened 2026-03-17
feat: Parse optional fields for `ShardPackageRecord`

jjerphan:
# Description

## Type of Change

- [ ] Bugfix
- [x] Feature / enhancement
- [ ] CI / Documentation
- [ ] Maintenance

## Checklist


libmamba/include/mamba/core/shard_types.hpp
```diff
@@ -76,11 +76,23 @@ namespace mamba
         std::size_t build_number = 0;
         std::optional<std::string> sha256;
         std::optional<std::string> md5;
+        /** Optional path to the site-packages directory for Python packages. */
+        std::optional<std::string> python_site_packages_path;
+        /** Deprecated md5 hash for legacy .tar.bz2 archives. */
+        std::optional<std::string> legacy_bz2_md5;
+        /** Deprecated size for legacy .tar.bz2 archives. */
+        std::optional<std::size_t> legacy_bz2_size;
+        std::size_t size = 0;
```
Klaim: Minor nit: Would it be problematic to gather the non-optional values first and the optional values in a second group of members?
Or at least move that line closer to the first members.

libmamba/src/core/shards.cpp
```diff
@@ -253,6 +253,18 @@ namespace mamba
                             record.md5 = hash;
                         }
                     }
+                    else if (key == "python_site_packages_path")
```
Klaim, line 256: Minor remark: looks like this whole sequence of `if` could have been a hash map? That would force usage of `function` though.

jjerphan: How would you use it? What would be the benefits?

Klaim: The benefit would be avoiding potentially 14 string comparisons (if I counted correctly), in exchange of invoking a `std::function`. No idea if it's actually worth it, I just reacted to the potential number of string comparisons in this block 🤡, hence the "minor remark" haha, you can ignore of course.

I was imagining essentially, something like that:
```c++
try{
  std::unordered_map<std::string_view, std::function<void()>> dispatcher { // or a any hash map
     {  "name", [&] { record.name = msgpack_object_to_string(val_obj); },
     { "version", [&]{  record.version = msgpack_object_to_string(val_obj); },
     ...
  };
  
  if(auto handler_it = dispatcher.find(key); handler_it != dispatcher.end())
  {
       (*handler_it)(); // execute the lambda, might be clearer using std::invoke
  }
  
}
```
Because the whole block in a loop it's not clear either if rebuilding the hashmap every cycle would cost a lot (probably), or if it can be written to avoid rebuilding (probably). Anyway I was just wondering.

jjerphan: Yes, we could also use an approach _à la_ [`semi::static_map`](https://github.com/hogliux/semimap) since we statically know the keys.

I would use simplicity. For now, we do not know if this causes an overhead. I think we can optimize it after profiling it or after a first release of the support of sharded repodata.

PR #4198, opened 2026-03-18
feat: Cache mentioned packages by each `Node` during traversal

jjerphan:
# Description

This avoids calling `extract_dependencies_impl` repeatedly.

`shard_mentioned_packages` was removed as not used anymore, and its tests have been adapted for `extract_dependencies_impl`.

## Type of Change

- [ ] Bugfix
- [x] Feature / enhancement
- [ ] CI / Documentation
- [ ] Maintenance

## Checklist

### DIFF
--- a/libmamba/include/mamba/core/shard_traversal.hpp
+++ b/libmamba/include/mamba/core/shard_traversal.hpp
@@ -52,14 +52,6 @@ namespace mamba
     using ShardsByUrl = std::map<std::string, std::size_t>;
 
     /**
-     * Extracts package names from dependency specs in a shard.
-     *
-     * Parses depends and constrains via specs::MatchSpec, returning the
-     * package names for dependency traversal.
-     */
-    [[nodiscard]] auto shard_mentioned_packages(const ShardDict& shard) -> std::vector<std::string>;
-
-    /**
      * Traverses sharded repodata to find reachable packages from root packages.
      *
      * Uses BFS or pipelined strategy to explore dependencies across shards.
@@ -94,6 +86,13 @@ namespace mamba
         std::vector<Shards> m_shards;
 
         NodeMap m_nodes;
+
+        // Cache of dependency names mentioned in a Node's ShardDict, computed on first visit.
+        // Keyed by NodeId to re-iterating on the package records.
+        std::map<NodeId, std::vector<std::string>> m_mentioned_packages;
+
+        const std::vector<std::string>&
+        mentioned_packages(const NodeId& node_id, const ShardDict& shard);
 
         void reachable_bfs(
             const std::vector<std::string>& root_packages,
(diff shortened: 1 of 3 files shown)


libmamba/include/mamba/core/shard_traversal.hpp
```diff
@@ -95,6 +87,13 @@ namespace mamba
 
         NodeMap m_nodes;
 
+        // Cache of dependency names mentioned in a Node's ShardDict, computed on first visit.
+        // Keyed by NodeId to re-iterating on the package records.
+        std::map<NodeId, std::vector<std::string>> m_mentioned_packages;
```
Klaim, line 92: I suspect you chose `map` instead of `unordered_map` because there is no hash function setup for `NodeID` yet?

libmamba/src/core/shard_traversal.cpp
```diff
@@ -53,21 +55,27 @@ namespace mamba
...
+        void add_names_from_record(const ShardPackageRecord& record, std::vector<std::string>& names)
         {
             add_names_from_specs(record.depends, names);
             add_names_from_specs(record.constrains, names);
         }
 
         std::vector<std::string> extract_dependencies_impl(const ShardDict& shard)
         {
-            std::set<std::string> names;
+            // We collect dependency names in a flat vector, then sort + unique.
```
Klaim, line 72: Isn't that basically a `flat_set`? We have an implementation [there](https://github.com/mamba-org/mamba/blob/main/libmamba/include/mamba/util/flat_set.hpp#L29-L37) if that helps. No idea if sorting+unique last is better than checks at insertion in the right place.

libmamba/src/core/shard_traversal.cpp
```diff
@@ -227,18 +232,49 @@ namespace mamba
...
-                        neighbor_id.shard_url,
-                        false,
-                    };
-                    pending.push_back(neighbor_id);
+                    const std::string url = dep_shards.url();
+                    if (!dep_shards.contains(dep))
+                    {
+                        continue;
+                    }
+                    NodeId neighbor_id{ dep, url, dep_shards.shard_url(dep) };
```
Klaim, line 266:
```suggestion
                    const NodeId neighbor_id{ dep, url, dep_shards.shard_url(dep) };
```

libmamba/include/mamba/core/shard_traversal.hpp
jjerphan: Yes, also I prefer to used sorted data-structure to make sure we get something repeatable in case we want to parallelize it then.

I need to perform some more advanced benchmarks to understand the trade offs  between `main` and this PR and possible datastructures for it; I found that this implementation can be faster for some packages with many dependencies e.g. `pyarrow`, but not all of them.

Klaim: I feel like it has been several times in the past year that we felt the need for essentially a `flat_map`. I'll note to potentially add one soon impl next to our current `flat_set`. Worst case, C++26 impls will provide it but we probably wont upgrade for one or two years.

jjerphan:
Not actually worth it.

PR #4201, opened 2026-03-19
fix: JSON should be set to `"success": false` when any transactions failed

Klaim:
# Description

I found some cases while working on #4126 where the JSON output would set `"success": true` while `MTransaction::execution` actually failed, leading to potentially succeeding integration tests which check for that value.

This PR attempts to make the JSON output correct and reveal tests that are actually not passing.

## Type of Change

- [x] Bugfix
- [ ] Feature / enhancement
- [ ] CI / Documentation
- [ ] Maintenance

## Checklist

### DIFF
--- a/libmamba/src/core/transaction.cpp
+++ b/libmamba/src/core/transaction.cpp
@@ -367,7 +367,8 @@ namespace mamba
     bool
     MTransaction::execute(const Context& ctx, ChannelContext& channel_context, PrefixData& prefix)
     {
-        // If an exception is thrown in this function, we must consider the whole operation as a failure.
+        // If an exception is thrown in this function, we must consider the whole operation as a
+        // failure.
         Console::JSonFailureOnException fail_json_on_exception;
 
         // JSON output


jjerphan:
The failure of `TestMultiplePkgCaches::test_no_writable_extracted_dir_corrupted` comes from [those changes](https://github.com/mamba-org/mamba/commit/d5f014c86bd98345364661bb64c1d14adb6ec19f#diff-52f2d47c832785d7ee6ee64e68141acdf6893ab92bc7f59f58284fdc6da76afeR350-R360) from https://github.com/mamba-org/mamba/pull/4123 (mamba does not actually support read-only caches, does it?) and which should be reverted IMO.

See: https://github.com/mamba-org/mamba/pull/4213

Klaim:
@jjerphan Weird: that ci failure wasnt visible in your pr but it's the same sequence of commits here. I'm suspecting a timing issue, although the test is about detecting a "hang". I'll re-run to be sure but I think maybe that timeout is optimistically short?

jjerphan:
We can increase the value for the timeout (we have done that several time in the past).

jjerphan (commented):
Thank you.

Are there way we can test it?

Klaim:
> Are there way we can test it?

I'm not sure if we actually can test it in a way that's always correct. A command that should always fail in the transaction execution?
Any suggestion?

libmamba/src/core/transaction.cpp
```diff
@@ -367,6 +367,9 @@ namespace mamba
     bool
     MTransaction::execute(const Context& ctx, ChannelContext& channel_context, PrefixData& prefix)
     {
+        // If an exception exists this function, we must consider the whole operation a failure.
+        Console::JSonFailureOnException fail_json_on_exception;
```
jjerphan:
```suggestion
        // If an exception is thrown in this function, we must consider the whole operation as a failure.
        Console::JSonFailureOnException fail_json_on_exception;
```

jjerphan:
What if we `SIGINT` an installation of an environment?

Klaim:
> What if we `SIGINT` an installation of an environment?

If I understand correctly, that would rely on timing to trigger the exception in the `MTransaction::execution` function, it wouldnt be very reliable?

How about attempting to get a package from an impossible url? I suspect that might fail before that function though. Maybe with an incorrect build id?

jjerphan:
We should be able to use an explicit environment specification with a faulty package.

Klaim:
> We should be able to use an explicit environment specification with a faulty package.

I suspect we might have one somewhere in the tests, I'll look into it or create it👍🏼

Issue #3763, opened 2025-01-22
Micromamba is not silent on errors even when you pass `--quiet`


simon-ess:
### Troubleshooting docs

- [x] My problem is not solved in the Troubleshooting docs

### Anaconda default channels

- [x] I do NOT use the Anaconda default channels (pkgs/* etc.)

### How did you install Mamba?

Micromamba

### Search tried in issue tracker

quiet

### Latest version of Mamba

- [x] My problem is not solved with the latest version

### Tried in Conda?

I do not have this problem with Conda, just with Mamba

### Describe your issue

Micromamba really likes to tell you when something has gone wrong (see for example my previous issue https://github.com/mamba-org/mamba/issues/3636), even if you tell it not to. Running a command such as
```
$ micromamba list --quiet --json -n not_an_environment
```
yields the extremely verbose, non-json output
```
info     libmamba ****************** Backtrace Start ******************
debug    libmamba Loading configuration
trace    libmamba Compute configurable 'create_base'
trace    libmamba Compute configurable 'no_env'
trace    libmamba Compute configurable 'no_rc'
trace    libmamba Compute configurable 'rc_files'
trace    libmamba Compute configurable 'root_prefix'
trace    libmamba Get RC files configuration from locations up to HomeDir
trace    libmamba Configuration not found at '/root/.mambarc'
trace    libmamba Configuration not found at '/root/.mamba/mambarc.d'
trace    libmamba Configuration not found at '/root/.mamba/mambarc'
trace    libmamba Configuration not found at '/root/.mamba/.mambarc'
trace    libmamba Configuration not found at '/root/.config/mamba/mambarc.d'
trace    libmamba Configuration not found at '/root/.config/mamba/mambarc'
trace    libmamba Configuration not found at '/root/.config/mamba/.mambarc'
trace    libmamba Configuration not found at '/root/.condarc'
trace    libmamba Configuration not found at '/root/.conda/condarc.d'
trace    libmamba Configuration not found at '/root/.conda/condarc'
trace    libmamba Configuration not found at '/root/.conda/.condarc'
trace    libmamba Configuration not found at '/root/.config/conda/condarc.d'
trace    libmamba Configuration not found at '/root/.config/conda/condarc'
trace    libmamba Configuration not found at '/root/.config/conda/.condarc'
trace    libmamba Configuration not found at '/root/micromamba/.mambarc'
trace    libmamba Configuration not found at '/root/micromamba/condarc.d'
[...]

PR #4202, opened 2026-03-19
fix: Don't log if `--json` or `--quiet`

Klaim:
# Description

While working on #4126  when checking issues on the CI I regularly hit failing tests where the json output failed to parse because of an error (usually because of #4201) plus the error was logged while `--json` was used, which should never happen.
This is a known and old issue, mainly visible while PRs are not stabilized yet.

This PR relies on the recently introduced logging system by completely deactivating it when `--json` or `--quiet` is parsed or if the related params passed to the `Context` constructor are `true`. 
- If `--quiet` is used, the logging system is then stopped, which means whatever existing logging handler will be removed, leading to logging records being completely ignored. 
- If `--json` is used, a special log-handler is setup to keep track of the log-records which pass the log-level filter (with no maximum count limit) and these records will be added to an array `"log_history"` as member of the JSON output object.
- `--json` also guarantees that a JSON output will appear even if it's an empty JSON **object** (when returning without a log, which should be rare). 
  - This **breaks compatibility with previous JSON output** as previously some commands resulted in a JSON **array** being output instead of an **object**.

Some consequences of the solution implemented in this PR:
- using current `(micro)mamba` will never output any log-record as long as `--json` or `--quiet` (this doesnt apply to outputs which are not logs, like the json output);
- It is possible for a `libmamba` user code to inject a new log-handler after processing the output params, which will potentially renabled log-record being output if `libmamba`'s code doesnt properly prevent it because of `--json` or `--quiet`; although in the present codebase we never do that and don't expect users to try switching log-handlers after initialization (except for testing).
- It is possible to emit logs before reaching the point where the logging system is stopped, that is at `Context` construction or when `--json` or `--quiet` are parsed (we are not doing that in (micro)mamba).

Some alternative solutions we could have used here but rejected for now:
1. Make the logging system simply ignore all log-records if one of these flags is set, even if a log-handler is installed. This is more drastic, but might be overkill. It also implies some communication between the output and logging systems.
2. Implement the choice of handling log-records or not depending on these flags in the default log-handler. This solution moves the responsibility into the log-handler but it also is quite fragile if `libmamba` users want to use another log-handler (python handling for example).

Note: I also fixed an incorrect log-level filter in `LogHandler_History` which is now revealed by a more complete set of tests.

## Type of Change

- [x] Bugfix
- [ ] Feature / enhancement
- [ ] CI / Documentation
- [ ] Maintenance

## Checklist


Klaim:
Might fix #3763

Klaim:
The failing test is actually not outputting any json, which suggests that the error is not properly handled with `--json`.

Klaim:
@JohanMabille @jjerphan 

I had to go a bit farther for this pr:
1. Some commands like `(micro)mamba config list --json` expects a JSON or YAML output which is generated by some other code than the `Console`, so when they succeed I had to *not output the normal JSON* to avoid output 2 JSON objects. 
  -> I suspect that it would be better to fusion the normal `Console` json output with the generated one? Otherwise no log history in these cases. If yes, how should I handle the YAML case?
2. The errors at `(micro)mamba` arguments parsing step were emitted before switching to the json output mode, as expected because we only know that we will output json after doing that parsing. However that leads to harder to predict output when `--json` is enabled.
  -> I decided to attempt to pre-parse `--json` and `--quiet` so that we setup the JSON output (with log history) as soon as possible. This leads to this output at the moment (with a randomly named flag):
  ```
  {
    "log_history": [
        {
            "level": "critical",
            "location": "E:\\QuantStack\\mamba-origin\\micromamba\\src\\main.cpp:188:9 int __cdecl main(int,char **)",
            "message": "The following argument was not expected: --fgsonsdgniodfsgndfoigndfoigndfo",
            "source": "libmamba"
        }
    ]
}
``` 
The missing part is the message to use `--help`, I can add it. It was injected automatically by CLI11's macro which was completely bypassing our error handling system so I changed that too. 
Is this still consistent with the expectation to get JSON output whatever is happening when requestion `--json`? Or do we expect normal error output on arguments parsing errors? If we do, I can remove the specific commit(s) forcing json.

**Comment by jjerphan** [comment]
>  -> I suspect that it would be better to fusion the normal Console json output with the generated one? Otherwise no log history in these cases. If yes, how should I handle the YAML case?

Yes, the output should be handle via the `Console` IMO, and I would convert YAML output to JSON (#4158 once finished should eventually helped).

> Is this still consistent with the expectation to get JSON output whatever is happening when requestion `--json`? Or do we expect normal error output on arguments parsing errors? If we do, I can remove the specific commit(s) forcing json.

To me the behavior that you describe is consistent with what we are expecting.

**Comment by Klaim** [comment]
> Yes, the output should be handle via the Console IMO, and I would convert YAML output to JSON (https://github.com/mamba-org/mamba/pull/4158 once finished should eventually helped).

It looks like the YAML output is only when `--json` is not used, so I think it should be fine.

**Comment by jjerphan** [comment]
FYI, I rebased this branch onto `main`.

I kept the previous state of the branch on [`backup/no-logging-json-quiet`](https://github.com/mamba-org/mamba/tree/backup/no-logging-json-quiet).

File: libmamba/include/mamba/core/logging_tools.hpp [hunk 1]
```diff
@@ -485,12 +520,25 @@ namespace mamba::logging
...
     {
         if (pimpl)
         {
             auto synched_data = pimpl->data.synchronize();
-            return std::vector<LogRecord>(synched_data->history.begin(), synched_data->history.end());
+            if (and_clear)
+            {
+                std::vector<LogRecord> result(synched_data->history.size());
+                std::ranges::move(synched_data->history, result.begin());
+                synched_data->history.clear();
```
jjerphan, line 533: Could this instruction after `move` lead to an UB?

libmamba/src/api/configuration.cpp
```diff
@@ -2775,9 +2789,17 @@ namespace mamba
...
-        std::cout << dump_text << std::endl;
+        if (config.context().output_params.json)
+        {
+            // merge the output with existing json output
+            auto dump_json = nlohmann::json::parse(dump_text);
+            Console::instance().set_json_output(JSONEdit::from_json_object_members(dump_json));
+        }
+        else
+        {
+            std::cout << dump_text << std::endl;  // outputs YAML
```
jjerphan, line 2800: Is there a limitation for now using the `Console` here?

micromamba/tests/helpers.py
```diff
@@ -315,15 +315,30 @@ def run_env(*args, f=None, **kwargs):
     return res.decode()
 
 
-def umamba_list(*args, **kwargs):
+def pkgs_list_from_json_result(json_result):
+    assert json_result
+    packages = json_result["packages"]
+    # TODO: consider if we check here if there was warnings or errors,
+    # maybe driven by a parameter
```
jjerphan, line 322: How about asserting that there are no warnings or errors here (or just bellow at the place of the other `TODO`)? It would make the tests stricter, but I think that it might be beneficial.

micromamba/src/umamba.cpp
```diff
@@ -31,10 +31,17 @@ set_umamba_command(CLI::App* com, mamba::Configuration& config)
...
     {
-        std::cout << umamba::version() << std::endl;
-        exit(0);
+        if (config.context().output_params.json)
+        {
+            Console::instance().set_json_output("/version"_json_pointer, umamba::version());
+        }
+        else
+        {
+            std::cout << umamba::version() << std::endl;
```
jjerphan, line 42: Why don't we use the `Console` here?

libmamba/include/mamba/core/logging_tools.hpp
```diff
@@ -37,6 +49,21 @@ namespace mamba::logging
...
+            switch (current_level)
+            {
+                case log_level::off:
+                    return true;
+
+                case log_level::all:
+                    return false;
+
+                default:
+                    return current_level > record.level;
```
jjerphan: Looking at [this change](https://github.com/mamba-org/mamba/pull/4202/changes#diff-42be0a002f3733e0a1995f8c588b9100c62776d0b339541a666f4120b6ee1b88R464) and [this other change](https://github.com/mamba-org/mamba/pull/4202/changes#diff-42be0a002f3733e0a1995f8c588b9100c62776d0b339541a666f4120b6ee1b88R639), should this be changed for:
```suggestion
                    return current_level < record.level;
```

or should the first change be adapted with a negation?

libmamba/tests/libmamba_logging/include/mamba/testing/test_logging_common.hpp
```diff
@@ -97,9 +98,16 @@ namespace mamba::logging::testing
             stats->current_params = std::move(new_params);
         }
 
-        auto log(LogRecord) -> void
+        auto log(LogRecord record) -> void
         {
             auto stats = pimpl->stats.synchronize();
+            if (stats->current_params.logging_level > record.level)
```
jjerphan: A candidate for `details::should_be_ignored`?

jjerphan (commented):
LGTM from a functional point of view modulo a few comments and questions.

As discussed privately, there is still some fuzz between directs usage of `std::{cout,cerr}` and of the `Console`, but this goes way beyond the scope of this PR.

libmamba/include/mamba/core/logging_tools.hpp
Klaim: No because the `.clear()` removes the possibility of accessing the moved-from objects. In general in the standard, containers moved-from are allowed to do write operations that dont have a precondition on their content, `clear()` is one of these (it just destroys objects, moved-from or not). A UB would happen if the contained objects would not handle being moved-from correctly, but then the issue would not be in this code but in the implementation of the contained type.
Also as long as nobody else has a pointer into `data->history`, it should be fine. The synchronization locks the mutual access.

If you find some part of the code that breaks the logic I just explained feel free to point it so that I fix it 👍🏽

libmamba/src/api/configuration.cpp
Klaim: I'm not sure. I tried to not change code that wasn't handled by the json case for now.
But I agree it would be good to investigate.

micromamba/tests/helpers.py
Klaim: I agree but I'm not sure if that's what we want in the current set of tests, some might be ok even with warnings.
Do you want me to try and see if we still pass the tests? I didnt try so far, just noted the idea.

libmamba/src/api/configuration.cpp
jjerphan: OK. This needs not be addressed as part of this PR to me.

micromamba/src/umamba.cpp
Klaim: As mentionned previously for this pr I preserved the behavior when it's not the `--json` case but I wondered the same question.

libmamba/include/mamba/core/logging_tools.hpp
Klaim: I 'm surprised the tests didnt catch that...

I'll reverse the negation (function name and results) to make this all easier to read and intuit. 👍🏽

micromamba/tests/helpers.py
jjerphan: Up to you. To me, it make sense to have those checks eventually, but they better not be introduced in this PR as it is a bit out of scope.

libmamba/tests/libmamba_logging/include/mamba/testing/test_logging_common.hpp
Klaim: Good catch, I introduced `should_be_ignored` after tweaking that code and forgot about it.

libmamba/include/mamba/core/logging_tools.hpp
jjerphan: Thank you.

This explanation is clear to me.

micromamba/tests/helpers.py
Klaim: I'll note to do an experimental pr later see if it breaks when we add the check.

libmamba/src/api/configuration.cpp
Klaim: Noted to investigate later if I find the time.

libmamba/include/mamba/core/logging_tools.hpp
Klaim: Actually that check `current_level > record.level` is correct as we want it to be `true` (as in "should be ignored") when `record.level` is inferior to the current log level. It could have been `return record.level < current_level;`. Hence the tests pass.

But it's all confusing to read so I'll change the function name, reverse the checks and results for clarity.

micromamba/src/env.cpp
jjerphan, line 1: We can simplify / fix the explicit environment export in another PR.

micromamba/tests/test_create.py
```diff
@@ -2669,7 +2689,7 @@ def test_create_package_with_non_url_char(tmp_home, tmp_root_prefix):
     assert any(pkg["name"] == "x264" for pkg in res["actions"]["LINK"])
 
 
-@pytest.mark.timeout(30)
```
jjerphan: With the recent changes regarding `should_be_emitted`, might have logs been reduced? On my machine, it now takes 14.47s to run.

Can we try reverting this?

micromamba/src/main.cpp
```diff
@@ -31,19 +33,59 @@
...
+        else if (arg == "--quiet"sv)
+        {
+            output_params.quiet = true;
+        }
+    }
+
+    if (output_params.json or output_params.quiet)
+    {
+        options.output_params = output_params;
+    }
```
jjerphan: Nit: Why use an intermediary `mamba::OutputParams`?

micromamba/src/main.cpp
```diff
@@ -90,15 +131,21 @@ main(int argc, char** argv)
...
+        (error_to_report->append(additional_messages), ...);
         set_sig_interrupted();
     };
 
+    int return_value = EXIT_SUCCESS;
+
     try
     {
-        CLI11_PARSE(app, argc, utf8argv);
+        // Note: do not use CLI11_PARSE macro as it's error handling
```
jjerphan: Nit.
```suggestion
        // Note: do not use CLI11_PARSE macro as its error handling
```

libmamba/src/core/util.cpp
```diff
@@ -76,9 +76,11 @@ namespace mamba
 
     const std::regex& token_regex()
     {
-        // usernames on anaconda.org can have a underscore, which influences the
-        // first two characters
-        static const std::regex token_regex{ "/t/([a-zA-Z0-9-_]{0,2}[a-zA-Z0-9-]*)" };
+        // Usernames on anaconda.org can have a underscore, which influences the
+        // first two characters.
+        // The `+` is there to make sure we dont match `/t/*` with `*` being literal or anything we
+        // dont capture here.
```
jjerphan: Nit.
```suggestion
        // don't capture here.
```

libmamba/src/api/configuration.cpp
```diff
@@ -2287,6 +2300,7 @@ namespace mamba
 
         if (this->at("print_config_only").value<bool>())
         {
+            // TODO: fix this for the case where `--json` is used
```
jjerphan: Should we open an issue which tracks all the `TODO` introduced in this PR such as this one?

libmamba/include/mamba/core/output.hpp
```diff
@@ -183,11 +188,31 @@ namespace mamba
...
+            });
+            // clang-format on
+        }
+
         /** If json output was requested, calling this before destroying a `Console` instance will
             not lead to a json output.
         */
         void cancel_json_print();
 
+        // FIXME: documentation
```
jjerphan, line 213: Did you mean to introduce a small docstring?

micromamba/tests/test_create.py
Klaim: It's the same issue we see in the rest of the CI so I dont think it has anything to do with any of the changes here?
Reverting this will probably work as long as the issue is random.

libmamba/include/mamba/core/output.hpp
Klaim: I forgot to add the docstring yes, will add.

libmamba/src/api/configuration.cpp
Klaim: I'll double check but I thinkt his TODO is not up to date, `print_dump` now outputs to the normal json output, if I'm not mistaken, when `--json` is used, was not the case before.

micromamba/tests/test_create.py
jjerphan: I guess I meant that we can ignore this failure and treat it independently so that the timeout does not get increased in this PR.

micromamba/src/env.cpp
Klaim: I'm not 100% sure I understand what you mean, but to clarify why this file is modified:
this PR is fixing the whole json output so that it all go into the same system and we can properly make sure that no non-json (usually errors) is emitted in that case. The change in this file is necessary to achieve that (along with any other json output code that was not using Console).
Note that the resulting output was *not* changed, it just goes into the proper system but the output json object is the same (with just the `"log_history"` field added, but otherwise same).
Or am I missing something/misunderstanding your comment?

jjerphan: I meant that there are still many direct usages of `stdout` here and some TODO which we can treat later in another PR.

libmamba/include/mamba/core/logging_tools.hpp
```diff
@@ -67,14 +96,15 @@ namespace mamba::logging
             */
             auto push_if_enabled(LogRecord& record) -> bool
```
JohanMabille: I am still uncomfortabe with this API and the move below. Given that this function is called [here](https://github.com/mamba-org/mamba/blob/f78ee8e07dc560083723bf69e3cc79048d02bedf/libmamba/include/mamba/core/logging_tools.hpp#L440) and [here](https://github.com/mamba-org/mamba/blob/f78ee8e07dc560083723bf69e3cc79048d02bedf/libmamba/include/mamba/core/logging_tools.hpp#L598) only, that the LogRecord is passed by value and is not used after the call to `push_if_enabled`, I think it is safe to accept it by rvalue reference here instead of lvalue reference.

micromamba/tests/test_create.py
Klaim: Yes will revert

libmamba/include/mamba/core/logging_tools.hpp
Klaim: When this function returns `false`, in both usage we move the record in another container, so I can't just take by r-value-reference without at least returning the object again so that it can be used by the caller, otherwise at the call point it would look like we move the object twice, which is even weirder.
```c++
if (not backtrace.push_if_enabled(std::move(record)))
{
    something_else(std::move(record)); // looks like UB
}
```

My goal here is to have this algorithm written only once, as having the same logic in both place lead to some annoying errors while developing. But I also agree it's quite weird api.

I'll try to find another way to achieve that goal with a nicer api. I was thinking maybe something in that direction:
```c++
// free function:
auto push_to_backtrace_or(Record&& record, BasicBacktrace& backtrace, std::invocable<Record&&> auto or_func);

// in call site:
push_to_backtrace_or(std::move(record), backtrace, [&](LogRecord&& record){ 
   // do the work if not pushed to backtrace
   something_else(std::move(record));
});
```
The passing of the same object to the lambda is a bit weird but maybe less weird than this current api?

JohanMabille: I was thinking of having `push_to_backtrace` returning the log record if not pushed (can be an optional, or a pair<bool, LogRecord>), but that would make the syntax heavier. Passing a lambda as you suggest seems the best option in the end, I don't find it weird to pass the same object to the lambda; and it is not that obvious at the call site.

micromamba/src/main.cpp
Klaim: `mamba::ContextOptions::output_params` is an `optional` which is empty by default, so here we only set it if one of the parameters is set. The behavior of the program is different if that optional is empty vs not-empty with members set to `false`. That's why we need to set the optional only if the right condition is met.

For reading clarity I will rewrite this with `.emplace`.

JohanMabille (approved):
LGTM, the error on Windows is unrelated to this PR. The failure on OSX is flaky (timeout), restarting the job should fix it.

Klaim:
> LGTM, the error on Windows is unrelated to this PR. The failure on OSX is flaky (timeout), restarting the job should fix it.

Yes, the OSX timeout was observed in other PRs (also on Linux runs, here and there) and I had doubled the timeout in this pr but we agreed with @jjerphan that it should be treated in another pr.

However, restarting the job is not guaranteed to fix it, it seems to happen often.

jjerphan (approved):
LGTM. Can we use conventional commit for this PR title?

Klaim:
> LGTM. Can we use conventional commit for this PR title?

Sure.
Although, with reluctance: I don't like how that "conventional" commit naming is reducing things like this pr which is not "just a bugfix" by misleadingly putting it into a narrow category. I'm also not convince that it helps tooling.
But I dont really mind otherwise, will rename👍🏽

jjerphan:
I agree that this the types of commit of conventional commits are a bit too coarse (I find it useful when considering doing a release), but is is this PR doing something else than fixing a wrong behavior of mamba?

Klaim:
> I agree that this the types of commit of conventional commits are a bit too coarse (I find it useful when considering doing a release), but is is this PR doing something else than fixing a wrong behavior of mamba?

It does fix the wrong general behavior around the json output but the issue is what that implies, the consequences of the fix:
- it changes the output of many `--json` cases (notably the ones which previously resulting in arrays `[ ... ]` instead of an object `{ ... }`)
- it adds loggs where they were not previously, in the json object, which is an important consequence of making sure errors dont appear in json output (we still need them to appear somewhere). Taken alone you could consider this a "feature" in some ways, or an improvement.

I wouldnt call that a feature either so fix is the closest I guess.

Personally for helping with release changelogs (and general communication around releases) in other projects I worke(d) on I tend to prefer distinguishing "changes" from "fixes" only. This means "fixes" are only changing the implementation, not the interfaces, and correcting behaviors. "Changes" are everything else. Some are big, some small, some might or not break the interface or widely change the behavior.
That's just personal preference though.

But really I'm nitpicking and acting like an old man waving his cane to the clouds, don't mind me 🙈 It's not that important :)

jjerphan:
I see and I agree. My assumption is/was that one can split changes into smaller ones fitting the types of conventional commits.

Here "enhancement" (`enh`) could work?

Anyway, it is not really a big deal.

Klaim:
> I see and I agree. My assumption is/was that one can split changes into smaller ones fitting the types of conventional commits.

Yes, IFF it does make sense to split. Not the case here though IMO. The things I pointed are direct consequences of the core changes of the pr, not really separable except if we accept weird situations where `main` commits are allowed to not actually working correctly/pass all tests, which AFAIK we don't.

(to clarify, the only things that could be separated here is the improvement + fixes on `LogHandler_History` etc., which had some issues, but that doesnt change my point about the consequences which stem from the core changes)

> Here "enhancement" (enh) could work?

If we could I would say it's both fix and enhancement. My understanding is that we cannot set both?

jjerphan:
To me, fixes are one kind of enhancements which themselves are more generic (enhancements as improvements of existing aspect of a project).

PR #4205, opened 2026-03-23
fix: Exclude Installed Packages from Root Packages

jjerphan:
# Description

We do not need to include installed packages as part of the root packages when installing new packages.

## Type of Change

- [x] Bugfix
- [ ] Feature / enhancement
- [ ] CI / Documentation
- [ ] Maintenance

## Checklist

### DIFF
--- a/libmamba/src/api/install.cpp
+++ b/libmamba/src/api/install.cpp
@@ -549,14 +549,6 @@ namespace mamba
 
     namespace
     {
-        std::vector<std::string>
-        build_sharded_root_packages(const std::vector<std::string>& raw_specs)
-        {
-            std::vector<std::string> root_packages = extract_package_names_from_specs(raw_specs);
-            add_pip_if_python(root_packages);
-            return root_packages;
-        }
-
         void print_activation_message(const Context& ctx)
         {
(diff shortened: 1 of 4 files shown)


libmamba/src/api/install.cpp
```diff
@@ -627,33 +635,9 @@ namespace mamba
...
-            // When installing python with sharded repodata, also include pip in root packages.
-            // This also avoids pulling old versions of python (1.x) which do not depend on
-            // other packages, which is a choice the solver can make.
-            if (ctx.repodata_use_shards)
-            {
-                add_pip_if_python(root_packages);
-            }
+            std::vector<std::string> root_packages = ctx.repodata_use_shards
+                                                         ? build_sharded_root_packages(raw_specs)
+                                                         : std::vector<std::string>{};
```
JohanMabille, line 632: This also removes the specs from root packages when we do not use the sharded repodata. A simpler change would be to simply remove the code within the "  if (fs::exists(ctx.prefix_params.target_prefix))" branch.

jjerphan: `root_packages` are ignore when the flat repodata is used, probably we can document it in details.

PR #4207, opened 2026-03-24
maint: Reduce logging in `Shards::parse_shard_msgpack`

jjerphan:
# Description

Logging represents more than 60% of the execution of `Shards::parse_shard_msgpack` (logging calls `Console::hide_secrets` which is costly as it uses regex to identify token and secrets) and also produces very verbose logs.

## Type of Change

- [ ] Bugfix
- [ ] Feature / enhancement
- [ ] CI / Documentation
- [x] Maintenance

## Checklist

### DIFF
--- a/libmamba/src/core/shards.cpp
+++ b/libmamba/src/core/shards.cpp
@@ -80,8 +80,9 @@ namespace mamba
                     {
                         result.push_back(msgpack_object_to_string(obj.via.array.ptr[i]));
                     }
-                    catch (const std::exception&)
+                    catch (const std::exception& e)
                     {
+                        LOG_WARNING << "Failed to parse array element " << i << ": " << e.what();
                         // Skip invalid elements
                     }
                 }
@@ -94,9 +95,12 @@ namespace mamba
          * Handles string, binary, extension, and array types, converting bytes to hex strings.
          *
          * @param obj The msgpack object containing the hash
+         * @param field_name The name of the field (for warning messages)
          * @return The hash as a hex string, or empty string if parsing fails
          */
-        auto msgpack_object_to_hash_string(const msgpack_object& obj) -> std::string
+        auto
+        msgpack_object_to_hash_string(const msgpack_object& obj, const std::string& field_name = "")
+            -> std::string
         {
             if (obj.type == MSGPACK_OBJECT_STR)
             {
@@ -138,6 +142,10 @@ namespace mamba
                     }
                     else
                     {
+                        LOG_WARNING
+                            << "Array element " << i << " in " << field_name
+                            << " is not a positive integer (type: " << static_cast<int>(elem.type)
+                            << "), cannot convert to bytes";
                         return std::string();
                     }
                 }
@@ -155,8 +163,13 @@ namespace mamba
                 {
                     return msgpack_object_to_string(obj);
                 }
-                catch (const std::exception&)
+                catch (const std::exception& e)
                 {
+                    std::string error_msg = "Failed to parse "
+                                            + (field_name.empty() ? "hash" : field_name)
+                                            + " from msgpack: unexpected type "
+                                            + std::to_string(static_cast<int>(obj.type));
+                    LOG_WARNING << error_msg << ": " << e.what();
                     // Return empty string - validation will check that at least one checksum is
                     // present
                     return std::string();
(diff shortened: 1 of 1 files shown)


jjerphan:
All `LOG_*` call `Console::hide_secrets` which is costly.

`LOG_DEBUG`'s core code path is as follow:
- `LOG_DEBUG`
- `LOG(mamba::log_level::debug)`
- `mamba::logging::MessageLogger(log_level).stream()`
- `std::ostream::operator<<` chain for message parts (`"..." << x << ...`)
- `mamba::logging::MessageLogger::~MessageLogger()`
- `make_safe_log_record(...)`
- `Console::hide_secrets(...)`
- `prepend(...)`
- branch:
  - if `!message_logger_use_buffer && Console::is_available()`:
    - `MessageLogger::emit(...)`
    - `mamba::logging::log(...)`
    - `call_log_handler_if_existing(&AnyLogHandler::log, ...)`
    - `get_log_handler()`
    - `AnyLogHandler::log(...)`
    - `AnyLogHandler::Wrapper<T>::log(...)`
    - `as_ref(object).log(...)` (concrete registered handler)
  - else (buffered path):
    - `message_logger_buffer->ready_records()`
    - `std::vector<LogRecord>::push_back(...)`
    - later via `MessageLogger::print_buffer(...)`:
      - `MessageLogger::emit(...)` for each buffered record
      - then the same dispatch chain as above

libmamba/src/core/shards.cpp
```diff
@@ -905,16 +862,10 @@ namespace mamba
...
                             map_obj.via.map.ptr[k].val
                         );
-                        LOG_DEBUG << "Parsed " << log_prefix
-                                  << " package record from shard for package '" << package
-                                  << "': " << record.name << "=" << record.version << "-"
-                                  << record.build;
                         target_map[pkg_filename] = record;
                     }
-                    catch (const std::exception& e)
+                    catch (const std::exception&)
```
Klaim: I'm against silently absorbing exceptions. How about returning an error or removing the try/catch altogether?

Klaim: Or maybe record the info in some way and continue.

jjerphan: I guess we can keep the `LOG_WARNING` in those cases indeed.

JohanMabille:
Notice that there are some ideas to improve the performance (see [here](https://github.com/mamba-org/mamba/blob/main/libmamba/include/mamba/core/logging.hpp#L806-L808) for instance), but we can remove the logs now and add them back later if they can be useful.

Or maybe just comment them out and add a TODO at the beginning of the file about this? That would avoid the confusion of seeing empty catch when debugging.

Klaim:
Indeed, I could make the change described in the link for log macros to be far less expensive, that does require changing the syntax of log macros (from `LOG_X() << expr;` to `LOG_X( expr );`) but I believe we can do that safely (and improve their name on the way). Ping me if it's a priority as I already thought how to do it. That change wouldnt however remove the cost of logs which pass the log level filter, these are paying the cost of the way `MessageLogger` works (which could be better) and indeed there is that regex issue.
For the regex issue an improvement would be to replace the regex implementation by something more performant. For some reason I wont get into here, c++ standard library regex impls tends to be far more expensive than different implementations with mostly the same interface. Using something like CTRE would definitely remove some of the cost when the code reaches that point.

jjerphan:
I do not think this is an issue for sharded repodata (I introduced plothera of `LOG_DEBUG` for development which aren't needed and useful for the end-users IMO).

Although I do not think that `Console::hide_secrets` has a massive overhead anymore on the execution, we can create an issue mentioning when logging is always using regular expressions and to discuss possible solutions or mitigations.

What do you think?

libmamba/src/core/shards.cpp
jjerphan: Done with 6e2f872.

Klaim:
> Although I do not think that Console::hide_secrets has a massive overhead anymore on the execution, we can create an issue mentioning when logging is always using regular expressions and to discuss possible solutions or mitigations.

Seems fair to me. 👍🏽

jjerphan:
https://github.com/mamba-org/mamba/issues/4211#issue-4133870658 has just been opened.

PR #4214, opened 2026-03-26
feat: Prefilter Shard Records on Python Minor Version

jjerphan:
# Description

Comes after https://github.com/mamba-org/mamba/pull/4187 and https://github.com/mamba-org/mamba/pull/4220.

Since the minor versions `python` is pinned in users' environments, we can prefilter records in shards using it to only selected relevant builds.

This definitely eases the installation of packages depending on `python`, by reducing the number of clauses within the solver and thus the time spent and RAM used resolving the environments. For instance for `pyarrow`, it reduces the time spent by 5 and the memory footprint by 20.

If `python` is not installed, we first try to prefilter builds depending on a fallback minor version (`python 3.14` proposed), before potentially retrying without (current behavior) if this fallback minor version cannot be used.

## Type of Change

- [ ] Bugfix
- [x] Feature / enhancement
- [ ] CI / Documentation
- [ ] Maintenance

## Checklist

### DIFF
--- a/libmamba/src/api/utils.cpp
+++ b/libmamba/src/api/utils.cpp
@@ -738,7 +738,18 @@ namespace mamba
             {
                 continue;
             }
-            const specs::VersionSpec relaxed = relax_version_spec_to_minor(maybe_ms.value().version());
+            const auto& raw_version_spec = maybe_ms.value().version();
+            // Pins like ``python=2`` or ``python=3`` specify only the major version. Relaxing those
+            // to ``2.0`` / ``3.0`` for shard prefiltering would drop packages whose ``depends``
+            // require a real minor (e.g. ``python >=2.7``). Skip the prefilter for such specs.
+            if (auto maybe_single_v = version_from_single_equality_spec(raw_version_spec))
+            {
+                if (maybe_single_v->version().size() <= std::size_t{ 1 })
+                {
+                    return std::nullopt;
+                }
+            }
+            const specs::VersionSpec relaxed = relax_version_spec_to_minor(raw_version_spec);
             if (auto maybe_v = version_from_single_equality_spec(relaxed))
             {
                 return maybe_v;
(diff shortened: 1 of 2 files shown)


libmamba/src/api/utils.cpp
```diff
@@ -114,6 +118,64 @@ namespace mamba
                 );
             }
         }
+
+        std::optional<specs::Version>
+        installed_python_minor_for_prefix(const fs::u8path& target_prefix)
```
jjerphan, line 126: Are there utilities we could reuse instead?

libmamba/include/mamba/core/shard_python_minor_prefilter.hpp
```diff
@@ -0,0 +1,45 @@
...
+        -> specs::VersionSpec;
+
+    /**
+     * Whether a ``depends`` line for ``python`` is compatible with
+     * ``python_minor_version_for_prefilter``. Uses ``VersionSpec::contains`` on the parsed version
+     * first; if that fails, relaxes exact on ``major.minor`` (see ``relax_version_spec_to_minor``)
+     * and tests again. Non-python dependencies always return true; parse failures return true (no
+     * prefilter).
+     */
+    [[nodiscard]] auto dependency_matches_python_minor_version_for_prefilter(
```
JohanMabille: Maybe a shorter name like `matches_python_minor` ? The argument names are explicit enough to not repeat them in the function name.

libmamba/src/core/shard_python_minor_prefilter.hpp
```diff
@@ -0,0 +1,45 @@
...
+     * and tests again. Non-python dependencies always return true; parse failures return true (no
+     * prefilter).
+     */
+    [[nodiscard]] auto dependency_matches_python_minor_version_for_prefilter(
+        const std::string& dependency_spec,
+        const specs::Version& python_minor_version_for_prefilter
+    ) -> bool;
+}
+
+#endif
```
JohanMabille, line 45: I don't think this header should be in the public API, it looks this is a utility used in the implementation only. It should be safe to move it to src.

libmamba/src/api/utils.cpp
jjerphan: `PrefixData` could be reused, but this would be much more costly (because it parses all the installed packages JSON and convert them to `PackageInfo`). We only need one `specs::Version` here.

PR #4215, opened 2026-03-27
maint: Factorize common parts of the `install`, `update` and `remove` paths

jjerphan:
# Description

I would like to introduce a proper retry mechanism for #4214 like what exists for `install_specs_impl`.

I have seen that there's a lot of duplications and inconsistency between this paths, so this first corrects it.

## Type of Change

- [ ] Bugfix
- [ ] Feature / enhancement
- [ ] CI / Documentation
- [x] Maintenance

## Checklist

### DIFF
--- a/libmamba/include/mamba/api/install.hpp
+++ b/libmamba/include/mamba/api/install.hpp
@@ -52,6 +52,18 @@ namespace mamba
         bool freeze_installed
     ) -> solver::Request;
 
+    /**
+     * Append file/global/python pins to a solver request and apply solver flags.
+     */
+    void add_pins_to_request(
+        solver::Request& request,
+        const Context& ctx,
+        PrefixData& prefix_data,
+        std::vector<std::string> specs,
+        bool no_pin,
+        bool no_py_pin
+    );
+
     void print_request_pins_to(const solver::Request& request, std::ostream& out);
 
     void install_explicit_specs(
--- a/libmamba/src/api/install.cpp
+++ b/libmamba/src/api/install.cpp
@@ -20,6 +20,7 @@
 #include "mamba/core/output.hpp"
 #include "mamba/core/package_cache.hpp"
 #include "mamba/core/package_database_loader.hpp"
+#include "mamba/core/pinning.hpp"
 #include "mamba/core/transaction.hpp"
 #include "mamba/download/downloader.hpp"
 #include "mamba/fs/filesystem.hpp"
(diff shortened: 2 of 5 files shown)


libmamba/src/api/install.cpp
```diff
@@ -598,131 +495,71 @@ namespace mamba
...
+            auto prefix_data = load_prefix_data_and_installed(ctx, channel_context, db);
 
             auto request = create_install_request(
                 prefix_data,
                 raw_specs,
                 InstallRequestOptions{ .freeze_installed = freeze_installed,
                                        .prefix_data_interoperability = ctx.prefix_data_interoperability }
             );
-            add_pins_to_request(request, ctx, prefix_data, raw_specs, no_pin, no_py_pin);
-            request.flags = ctx.solver_flags;
```
jjerphan, line 657:
```
request.flags = ctx.solver_flags
```

was moved at the end of `add_pins_to_request`.

libmamba/src/api/update.cpp
```diff
@@ -138,157 +132,101 @@ namespace mamba
...
+            ctx.solver_flags,
             prefix_data,
             raw_update_specs,
-            /* no_pin= */ config.at("no_pin").value<bool>(),
-            /* no_py_pin = */ config.at("no_py_pin").value<bool>()
+            no_pin,
+            no_py_pin
         );
 
-        request.flags = ctx.solver_flags;
```
jjerphan, line 209:
```
request.flags = ctx.solver_flags
```

was moved at the end of `add_pins_to_request`.

libmamba/src/api/utils.cpp
```diff
@@ -285,4 +297,334 @@ namespace mamba
...
+                request.jobs.emplace_back(
+                    Request::Pin{
+                        specs::MatchSpec::parse(py_pin)
+                            .or_else([](specs::ParseError&& err) { throw std::move(err); })
+                            .value(),
+                    }
+                );
+            }
+        }
+        request.flags = solver_flags;
```
jjerphan: This was consistently used after `add_pins_to_request`, so I just made it part of this function.

libmamba/include/mamba/api/install.hpp
```diff
@@ -52,15 +52,6 @@ namespace mamba
...
     ) -> solver::Request;
 
-    void add_pins_to_request(
-        solver::Request& request,
-        const Context& ctx,
-        PrefixData& prefix_data,
-        std::vector<std::string> specs,
-        bool no_pin,
-        bool no_py_pin
-    );
```
jjerphan, line 62: Moved to `libmamba/include/mamba/api/utils.hpp` as it is also use in `libmamba/src/api/update.cpp`

libmamba/src/api/install.cpp
```diff
@@ -549,36 +476,6 @@ namespace mamba
...
-                                     "Or to execute a single command in this environment, use:\n\n"
-                                     "    "
-                                  << executable
-                                  << " run "
-                                  // Use -n or -p depending on if the env_name is a full prefix or
-                                  // just a name.
-                                  << (environment == ctx.prefix_params.target_prefix ? "-p " : "-n ")
-                                  << environment << " mycommand\n";
-            }
-        }
```
jjerphan, line 580: Also moved to `utils`.

libmamba/include/mamba/api/install.hpp
JohanMabille: This is theoretically a backward incompatible change since `utils.hpp` is in the src folder (and we don't want to change that).

One way to avoid code duplication is to keep this declaration and ave the implementation in install.cpp forward the call to the new function in `utils.hpp`.

jjerphan: Yes, I know; but I wondered whether any other projects use those headers (it does not seem to be the case); we can keep the declaration here; and the definition in the `install.cpp` translation unit.

JohanMabille: That's something we cannot really know until someone comes and complains about the breakage, so let's be conservative here.

PR #4228, opened 2026-04-10
feat: Support excluding recent builds using timestamp (`--exclude-newer`)

### LINKED ISSUE #4254: Add an option to constrain dependency resolution with a relative cooldown period
The purpose is to mitigate OSS supply chain attacks by preventing the installation that have been recently published on conda-forge or other channels. This gives time for security research to discover and report threats and yank package versions with critical vulnerabilities or corrupted with malware.

Options that exist in other package managers:

- `pip`: `--uploaded-prior-to` 
- `uv`: `--exclude-newer`
- `pixi`: `exclude-newer`

Ideally the option should accept relative dates specifications (such as `P7D`) rather than just absolute dates.

jezdez:
### Description

Adds an optional `exclude_newer_timestamp` field to `Database::Settings`, allowing callers to exclude packages with a timestamp newer than the given Unix epoch cutoff during repodata loading.

This enables conda's `--exclude-newer` feature to work with the libmamba solver backend, matching what rattler already supports natively. See [conda/conda#15759](https://github.com/conda/conda/issues/15759) for the full tracking and [conda/ceps#154](https://github.com/conda/ceps/pull/154) for the CEP proposing `indexed_timestamp` in repodata.

This PR adds only the low-level **global** libmamba API: a single cutoff timestamp applied when loading a repo. It intentionally does not implement conda's richer channel- or package-scoped `exclude_newer` policy; `conda-libmamba-solver` applies scoped policy in Python-side repodata filtering when needed and uses this native cutoff for the simple global-policy case.

Fix #4254.

### How it works

The cutoff is applied at load time in both ingestion paths:

- **JSON path** (`mamba_read_json` -> `set_repo_solvables_impl`): after a solvable is parsed, its policy timestamp is compared to the cutoff. If newer, the solvable is removed via `repo.remove_solvable(id, true)` -- the same pattern used for parse failures. For repodata JSON, `indexed_timestamp` is preferred over `timestamp` when present (matching conda semantics).
- **PackageInfo path** (`add_repo_from_packages_impl_loop`): the package timestamp (normalized from ms to seconds) is checked before adding a solvable. If newer, the package is skipped entirely.

The `.solv` cache path is **not** filtered. Callers (e.g. `conda-libmamba-solver`) are expected to invalidate the cache when `exclude_newer_timestamp` changes, the same way they handle `repodata_filters`.

### Other changes

- Moved `MAX_CONDA_TIMESTAMP` from the anonymous namespace in `helpers.cpp` to `helpers.hpp` and added `normalize_conda_timestamp` for shared timestamp normalization.
- Python bindings: `Database.__init__()` accepts an optional `exclude_newer_timestamp` keyword argument.

### Context

- `--exclude-newer` is a supply chain security feature that excludes recently published packages, giving security vendors time to flag malicious uploads. It's already supported by uv, pip, pixi, and rattler.
- CEP draft proposing `indexed_timestamp`: [conda/ceps#154](https://github.com/conda/ceps/pull/154)
- Andrew Nesbitt's overview: https://nesbitt.io/2026/03/04/package-managers-need-to-cool-down.html
- conda implementation: [conda/conda#15761](https://github.com/conda/conda/pull/15761)
- rattler support: [conda/rattler#654](https://github.com/conda/rattler/pull/654)
- micromamba CLI tracking: [mamba-org/mamba#4254](https://github.com/mamba-org/mamba/issues/4254)

### Test plan

- [x] C++ tests for `add_repo_from_packages`, millisecond normalization, repodata JSON filtering, unfiltered vs filtered `package_count`, and `indexed_timestamp` precedence
- [x] Python binding tests for packages API, repodata JSON, and `None` default

### DIFF
--- a/libmamba/src/solver/libsolv/helpers.cpp
+++ b/libmamba/src/solver/libsolv/helpers.cpp
@@ -40,10 +40,6 @@
 
 namespace mamba::solver::libsolv
 {
-    // Beyond this value, the timestamp would be in milliseconds and therefore should be converted
-    // to seconds.
-    inline constexpr auto MAX_CONDA_TIMESTAMP = 253402300799ULL;
-
     void set_solvable(
         solv::ObjPool& pool,
         solv::ObjSolvableView solv,
@@ -419,7 +415,8 @@ namespace mamba::solver::libsolv
             const std::optional<nlohmann::json>& signatures,
             Filter&& filter,
             OnParsed&& on_parsed,
-            MatchSpecParser parser
+            MatchSpecParser parser,
+            std::optional<std::uint64_t> exclude_newer_timestamp = std::nullopt
         )
         {
             auto packages_as_object = packages.get_object();
@@ -442,7 +439,15 @@ namespace mamba::solver::libsolv
                     );
                     if (parsed)
                     {
-                        on_parsed(filename);
+                        if (exclude_newer_timestamp
+                            && solv.timestamp() > static_cast<std::int64_t>(*exclude_newer_timestamp))
+                        {
+                            repo.remove_solvable(id, /* reuse_id= */ true);
+                        }
+                        else
+                        {
+                            on_parsed(filename);
+                        }
                     }
                     else
                     {
@@ -462,7 +467,8 @@ namespace mamba::solver::libsolv
             const std::string& default_subdir,
             JSONObject& packages,
             const std::optional<nlohmann::json>& signatures,
-            MatchSpecParser parser
+            MatchSpecParser parser,
+            std::optional<std::uint64_t> exclude_newer_timestamp = std::nullopt
         )
         {
             return set_repo_solvables_impl(
@@ -475,7 +481,8 @@ namespace mamba::solver::libsolv
                 signatures,
                 /* filter= */ [](const auto&) { return true; },
                 /* on_parsed= */ [](const auto&) {},
-                parser
+                parser,
+                exclude_newer_timestamp
             );
         }
 
@@ -488,7 +495,8 @@ namespace mamba::solver::libsolv
             const std::string& default_subdir,
             JSONObject& packages,
             const std::optional<nlohmann::json>& signatures,
-            MatchSpecParser parser
+            MatchSpecParser parser,
+            std::optional<std::uint64_t> exclude_newer_timestamp = std::nullopt
         ) -> util::flat_set<std::string>
         {
             auto filenames = util::flat_set<std::string>();
@@ -504,7 +512,8 @@ namespace mamba::solver::libsolv
                 /* on_parsed= */
                 [&](const auto& fn)
                 { filenames.insert(std::string(specs::strip_archive_extension(fn))); },
-                parser
+                parser,
+                exclude_newer_timestamp
             );
             // Sort only once
             return filenames;
@@ -520,7 +529,8 @@ namespace mamba::solver::libsolv
             JSONObject& packages,
             const std::optional<nlohmann::json>& signatures,
             const SortedStringRange& added,
-            MatchSpecParser parser
+            MatchSpecParser parser,
+            std::optional<std::uint64_t> exclude_newer_timestamp = std::nullopt
         )
         {
             return set_repo_solvables_impl(
@@ -534,7 +544,8 @@ namespace mamba::solver::libsolv
                 /* filter= */
                 [&](const auto& fn) { return !added.contains(specs::strip_archive_extension(fn)); },
                 /* on_parsed= */ [&](const auto&) {},
-                parser
+                parser,
+                exclude_newer_timestamp
             );
         }
     }
@@ -599,7 +610,8 @@ namespace mamba::solver::libsolv
         const std::string& channel_id,
         PackageTypes package_types,
         MatchSpecParser ms_parser,
-        bool verify_artifacts
+        bool verify_artifacts,
+        std::optional<std::uint64_t> exclude_newer_timestamp
     ) -> expected_t<solv::ObjRepoView>
     {
         LOG_INFO << "Reading repodata.json file " << filename << " for repo " << repo.name()
@@ -715,7 +727,8 @@ namespace mamba::solver::libsolv
                     default_subdir,
                     pkgs,
                     json_signatures,
-                    ms_parser
+                    ms_parser,
+                    exclude_newer_timestamp
                 );
             }
             if (auto pkgs = repodata_doc["packages"]; !pkgs.error())
@@ -729,7 +742,8 @@ namespace mamba::solver::libsolv
                     pkgs,
                     json_signatures,
                     added,
-                    ms_parser
+                    ms_parser,
+                    exclude_newer_timestamp
                 );
             }
         }
@@ -746,7 +760,8 @@ namespace mamba::solver::libsolv
                     default_subdir,
                     pkgs,
                     json_signatures,
-                    ms_parser
+                    ms_parser,
+                    exclude_newer_timestamp
                 );
             }
 
@@ -761,7 +776,8 @@ namespace mamba::solver::libsolv
                     default_subdir,
                     pkgs,
                     json_signatures,
-                    ms_parser
+                    ms_parser,
+                    exclude_newer_timestamp
                 );
             }
         }
--- a/libmamba/src/solver/libsolv/helpers.cpp
+++ b/libmamba/src/solver/libsolv/helpers.cpp
@@ -1500,8 +1500,7 @@ namespace mamba::solver::libsolv
                             return true;
                         }
                     }
-                    if constexpr (std::is_same_v<Action, Solution::Reinstall>
-                                  || std::is_same_v<Action, Solution::Omit>)
+                    if constexpr (std::is_same_v<Action, Solution::Reinstall> || std::is_same_v<Action, Solution::Omit>)
                     {
                         if (action.what.name == pkg_name)
                         {
--- a/libmamba/src/solver/libsolv/helpers.cpp
+++ b/libmamba/src/solver/libsolv/helpers.cpp
@@ -195,16 +195,15 @@ namespace mamba::solver::libsolv
         template <class JSONObject>
         [[nodiscard]] auto set_solvable(
             solv::ObjPool& pool,
-            // const std::string& repo_url_str,
             const specs::CondaURL& repo_url,
             const std::string& channel_id,
             solv::ObjSolvableView solv,
-
             const std::string& filename,
             JSONObject&& pkg,
             const std::optional<nlohmann::json>& signatures,
             const std::string& default_subdir,
-            MatchSpecParser parser
+            MatchSpecParser parser,
+            std::uint64_t* out_timestamp = nullptr
         ) -> bool
         {
             // Not available from RepoDataPackage
@@ -319,7 +318,12 @@ namespace mamba::solver::libsolv
             if (auto timestamp = pkg["timestamp"]; !timestamp.error())
             {
                 const auto time = timestamp.get_uint64().value_unsafe();
-                solv.set_timestamp((time > MAX_CONDA_TIMESTAMP) ? (time / 1000) : time);
+                const auto normalized = (time > MAX_CONDA_TIMESTAMP) ? (time / 1000) : time;
+                solv.set_timestamp(normalized);
+                if (out_timestamp)
+                {
+                    *out_timestamp = normalized;
+                }
             }
 
             if (auto depends = pkg["depends"].get_array(); !depends.error())
@@ -426,6 +430,7 @@ namespace mamba::solver::libsolv
                 if (filter(filename))
                 {
                     auto [id, solv] = repo.add_solvable();
+                    std::uint64_t pkg_timestamp = 0;
                     const bool parsed = set_solvable(
                         pool,
                         repo_url,
@@ -435,12 +440,12 @@ namespace mamba::solver::libsolv
                         pkg_field.value(),
                         signatures,
                         default_subdir,
-                        parser
+                        parser,
+                        &pkg_timestamp
                     );
                     if (parsed)
                     {
-                        if (exclude_newer_timestamp
-                            && solv.timestamp() > static_cast<std::int64_t>(*exclude_newer_timestamp))
+                        if (exclude_newer_timestamp && pkg_timestamp > *exclude_newer_timestamp)
                         {
                             repo.remove_solvable(id, /* reuse_id= */ true);
                         }
--- a/libmamba/src/solver/libsolv/helpers.cpp
+++ b/libmamba/src/solver/libsolv/helpers.cpp
@@ -40,10 +40,6 @@
 
 namespace mamba::solver::libsolv
 {
-    // Beyond this value, the timestamp would be in milliseconds and therefore should be converted
-    // to seconds.
-    inline constexpr auto MAX_CONDA_TIMESTAMP = 253402300799ULL;
-
     void set_solvable(
         solv::ObjPool& pool,
         solv::ObjSolvableView solv,
@@ -443,7 +439,8 @@ namespace mamba::solver::libsolv
             const std::optional<nlohmann::json>& signatures,
             Filter&& filter,
             OnParsed&& on_parsed,
-            MatchSpecParser parser
+            MatchSpecParser parser,
+            std::optional<std::uint64_t> exclude_newer_timestamp = std::nullopt
         )
         {
             auto packages_as_object = packages.get_object();
@@ -466,7 +463,15 @@ namespace mamba::solver::libsolv
                     );
                     if (parsed)
                     {
-                        on_parsed(filename);
+                        if (exclude_newer_timestamp
+                            && solv.timestamp() > static_cast<std::int64_t>(*exclude_newer_timestamp))
+                        {
+                            repo.remove_solvable(id, /* reuse_id= */ true);
+                        }
+                        else
+                        {
+                            on_parsed(filename);
+                        }
                     }
                     else
                     {
@@ -486,7 +491,8 @@ namespace mamba::solver::libsolv
             const std::string& default_subdir,
             JSONObject& packages,
             const std::optional<nlohmann::json>& signatures,
-            MatchSpecParser parser
+            MatchSpecParser parser,
+            std::optional<std::uint64_t> exclude_newer_timestamp = std::nullopt
         )
         {
             return set_repo_solvables_impl(
@@ -499,7 +505,8 @@ namespace mamba::solver::libsolv
                 signatures,
                 /* filter= */ [](const auto&) { return true; },
                 /* on_parsed= */ [](const auto&) {},
-                parser
+                parser,
+                exclude_newer_timestamp
             );
         }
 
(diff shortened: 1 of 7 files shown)


jjerphan:
Thank you for this contribution, @jezdez. :slightly_smiling_face: 

Please let us know when this can be reviewed.

libmamba/src/solver/libsolv/helpers.cpp
```diff
@@ -1479,8 +1500,7 @@ namespace mamba::solver::libsolv
                             return true;
                         }
                     }
-                    if constexpr (std::is_same_v<Action, Solution::Reinstall>
-                                  || std::is_same_v<Action, Solution::Omit>)
+                    if constexpr (std::is_same_v<Action, Solution::Reinstall> || std::is_same_v<Action, Solution::Omit>)
```
jjerphan: It seems like a spurious change which needs to be reverted.

libmambapy/bindings/solver_libsolv.cpp
jjerphan, line 1: Given those changes, I think this would qualify for the next minor version (i.e. 2.6.0 or 2.7.0 if this PR is merged soon).

libmamba/src/solver/libsolv/helpers.cpp
```diff
@@ -438,11 +440,19 @@ namespace mamba::solver::libsolv
...
-                        parser
+                        parser,
+                        &pkg_timestamp
                     );
                     if (parsed)
                     {
-                        on_parsed(filename);
+                        if (exclude_newer_timestamp && pkg_timestamp > *exclude_newer_timestamp)
+                        {
+                            repo.remove_solvable(id, /* reuse_id= */ true);
```
jjerphan, line 483: Side-note: Removing solvable here is the most appropriate to me. I think that we might even be able (in the case of sharded repodata usage) to filter builds on timestamp at the earliest (see #4214); yet I am not sure that there will be massive benefits (more maintenance and complexity for probably not the most reduced overhead).

libmamba/tests/src/solver/libsolv/test_database.cpp
```diff
@@ -218,6 +264,25 @@ namespace
             REQUIRE(found_python);
         }
 
+        SECTION("exclude_newer_timestamp filters packages from repodata JSON")
```
jjerphan, line 268: Can you extend this test to check that the `package_count` is greater when `exclude_newer_timestamp` is not specified?

jjerphan (commented):
Thank you, @jezdez.

I just have a few comments. I also will let other maintainers review this PR.

SylvainCorlay:
It could be interesting to have a variant of this offering some kind of "as-of" for reproducibility without a lock.

jjerphan:
Hello @jezdez,

Would you authorize us to finish this PR? :slightly_smiling_face:

libmamba/src/solver/libsolv/helpers.cpp
jezdez: Fixed in fbf2a8b3 by restoring the original line wrapping, so this formatting-only change no longer appears in the PR diff.

libmambapy/bindings/solver_libsolv.cpp
jezdez: Agreed. Since this adds Database settings/API surface and a Python binding argument, this should target the next minor release rather than a patch release.

libmamba/src/solver/libsolv/helpers.cpp
jezdez: Agreed. I kept the native libmamba implementation at the solvable insertion/removal point for this PR. The conda-side scoped policy uses Python repodata filtering before libmamba; earlier shard-level filtering can be revisited separately once the shard APIs settle.

libmamba/tests/src/solver/libsolv/test_database.cpp
jezdez: Added in fbf2a8b3: the test now builds an unfiltered database from the same repodata and asserts its package_count is greater than the filtered result. I also added indexed_timestamp precedence coverage to match conda’s updated policy.

jezdez:
Thanks @jjerphan, I’d appreciate the handover if you’re still willing to help finish this PR.

One thing worth keeping explicit while doing that: this PR currently adds only the low-level/global libmamba API, `exclude_newer_timestamp`, so it can filter a repo against one cutoff timestamp. That intentionally differs from the current conda implementation, which now has a richer policy with global, channel-specific, and package-specific `exclude_newer` settings.

For conda, the downstream solver integrations use libmamba’s native global cutoff only for the simple global-policy case. When conda has channel or package overrides, conda-libmamba-solver applies the conda policy in Python-side repodata filtering before handing records to libmamba. So I would not try to squeeze scoped policy into this PR unless standalone mamba/micromamba wants its own matching UX and public API; that likely deserves a separate design/API change.

I also pushed fbf2a8b3 to address the review comments and align the native filter with conda’s timestamp semantics by preferring `indexed_timestamp` over `timestamp` for the cutoff decision.

jjerphan:
Thank you for explaining this.

To me it makes sense to respect the policies which has been adopted by conda and to implement what is necessary to support them (or at least first the global `exclude-newer`) in mamba and micromamba; in this regards, I would have the support of the configuration elements be supported by this PR.

jjerphan:
Everything should have been implemented in the last commits (it makes sense to me to have everything in a `feat` PR, especially given that the diff is OK, but I am not against proceeding in several steps). I think we could add many more tests though. What do people think?

libmamba/CMakeLists.txt
```diff
@@ -226,6 +226,7 @@ set(
     ${LIBMAMBA_SOURCE_DIR}/core/env_lockfile_conda.cpp
     ${LIBMAMBA_SOURCE_DIR}/core/env_lockfile_mambajs.cpp
     ${LIBMAMBA_SOURCE_DIR}/core/environments_manager.cpp
+    ${LIBMAMBA_SOURCE_DIR}/core/exclude_newer.cpp
```
Hind-M: We are missing the header `exclude_newer.hpp` in the `CMakeLists.txt`.

micromamba/src/common_options.cpp
```diff
@@ -390,6 +390,13 @@ init_install_options(CLI::App* subcom, Configuration& config)
         allow_downgrade.description()
     );
 
+    auto& exclude_newer = config.at("exclude_newer");
+    subcom->add_option(
+        "--exclude-newer",
+        exclude_newer.get_cli_config<std::string>(),
+        exclude_newer.description()
+    );
+
```
Hind-M: Shouldn't we add an option for `exclude_newer_package` as well?

libmambapy/tests/test_solver_libsolv.py
```diff
@@ -182,6 +182,90 @@ def test_Database_RepoInfo_from_packages(add_pip_as_python_dependency, matchspec
...
+        repodata_file,
+        "https://example.com/linux-64",
+        "test-channel",
+    )
+    assert repo is not None
+    assert repo.package_count() == 1
+    pkgs = db.packages_in_repo(repo)
+    assert pkgs[0].name == "included-pkg"
+
+
```
Hind-M, line 268: Should we add tests using environment variables `CONDA/MAMBA_EXCLUDE_NEWER/_PACKAGE`?

libmambapy/tests/test_solver_libsolv.py
```diff
@@ -182,6 +182,90 @@ def test_Database_RepoInfo_from_packages(add_pip_as_python_dependency, matchspec
...
+                    "included-pkg-1.0-bld.tar.bz2": {
+                        "name": "included-pkg",
+                        "version": "1.0",
+                        "build": "bld",
+                        "build_number": 0,
+                        "subdir": "linux-64",
+                        "depends": [],
+                        "timestamp": 3000,
+                        "indexed_timestamp": 1000,
+                    },
```
Hind-M, line 247: Should we add tests with only `timestamp` available (no `indexed_timestamp`) and with none of them?

libmamba/src/core/exclude_newer.cpp
```diff
@@ -0,0 +1,395 @@
...
+            while (!value.empty() && std::isspace(static_cast<unsigned char>(value.front())))
+            {
+                value.remove_prefix(1);
+            }
+            while (!value.empty() && std::isspace(static_cast<unsigned char>(value.back())))
+            {
+                value.remove_suffix(1);
+            }
+            return value;
+        }
```
Hind-M: This can be removed and replaced with the `strip` function from `util/string.hpp`. If we only want to strip spaces and not any non-graphic chars, we can do something like this:
`util::strip_if(value, [](unsigned char c) { return std::isspace(c); });`

libmamba/src/core/exclude_newer.cpp
```diff
@@ -0,0 +1,395 @@
...
+            {
+                out.emplace(name, std::nullopt);
+            }
+            else
+            {
+                out.emplace(name, resolve_exclude_newer_cutoff(trimmed, now_seconds));
+            }
+        }
+        return out;
+    }
```
Hind-M: I haven't looked at all these functions in detail, but it might be worth checking whether they can be refactored or whether some existing utility functions can be reused.

libmamba/src/api/utils.cpp
```diff
@@ -575,7 +597,20 @@ namespace mamba
     )
     {
         populate_context_channels_from_specs(raw_specs, ctx);
-        auto db = make_solver_database(ctx.experimental_matchspec_parsing, channel_context);
+
+        if ((!ctx.exclude_newer.empty() || !ctx.exclude_newer_package.empty())
+            && !ctx.experimental_repodata_parsing)
```
Hind-M: Not related to this PR but is `experimental_repodata_parsing` still experimental? If not, maybe it's time to qualify it as permanent and remove that prefix?
Maybe use `mamba_repodata_parsing` instead?
This can be done in another PR and maybe open an issue to track that.
Same for `experimental_matchspec_parsing`.

libmamba/include/mamba/core/context.hpp
```diff
@@ -128,6 +128,8 @@ namespace mamba
 
         // solver options
         solver::Request::Flags solver_flags = {};
+        std::string exclude_newer;
+        std::map<std::string, std::string> exclude_newer_package;
```
Hind-M: We should use a struct here for these (and pass it as an argument in functions instead of passing both). Could be `SecurityParams` for example.

micromamba/src/common_options.cpp
jjerphan: `exclude_newer_package` is, as far as I understand [this issue](https://github.com/conda/conda/issues/15759) and from [the current implementation proposal for `conda`](https://github.com/conda/conda/pull/15761/changes#diff-0fd0c17ecce803151675fdeeb56dc04525e9324fa3d6ff4f6c080570c681abfeR408-R409), only part of the configuration.

libmamba/src/core/exclude_newer.cpp
jjerphan, line 1: I wonder whether we should use elements of `std::chrono` now since we support C++20.

micromamba/src/common_options.cpp
jjerphan: We could introduce it on the CLI, but I do not really know how could one specify per-package policies.

libmambapy/tests/test_solver_libsolv.py
jjerphan: `timestamp` must be specified, but we can test the behavior when it is not present.

libmamba/tests/CMakeLists.txt
```diff
@@ -95,6 +95,7 @@ set(
     src/core/test_cpp.cpp
     src/core/test_env_file_reading.cpp
     src/core/test_env_lockfile.cpp
+    src/core/test_exclude_newer.cpp
```
Hind-M: I just realized this is breaking the alphabetical order of the listed files, should be moved a bit further.

libmamba/CMakeLists.txt
Hind-M: Same.

libmamba/CMakeLists.txt
```diff
@@ -381,6 +382,7 @@ set(
     ${LIBMAMBA_INCLUDE_DIR}/mamba/core/download_progress_bar.hpp
     ${LIBMAMBA_INCLUDE_DIR}/mamba/core/env_lockfile.hpp
     ${LIBMAMBA_INCLUDE_DIR}/mamba/core/environments_manager.hpp
+    ${LIBMAMBA_INCLUDE_DIR}/mamba/core/exclude_newer.hpp
```
Hind-M: Same.

libmamba/include/mamba/core/context.hpp
```diff
@@ -128,8 +129,7 @@ namespace mamba
 
         // solver options
         solver::Request::Flags solver_flags = {};
-        std::string exclude_newer;
-        std::map<std::string, std::string> exclude_newer_package;
+        ExcludeNewerPolicy exclude_newer_policy;
```
Hind-M: Not sure if we really want to include the whole `ExcludeNewerPolicy` in the context. Seems ok for now (not too heavy), but maybe others in the team would mind.

Hind-M:
I'm wondering if we should add a section in the documentation explaining the usage.
It could be done in a follow-up PR and that would be part of a broader security section (what `mamba` provides in that regard).

libmamba/include/mamba/core/context.hpp
jjerphan: I guess this is the best tradeoff to keep up with the ecosystem.

Hind-M (approved):

jjerphan:
I would wait for it to have at least one other review and preferably two.

libmamba/include/mamba/core/exclude_newer.hpp
```diff
@@ -0,0 +1,121 @@
...
+#include <string_view>
+#include <unordered_map>
+
+namespace mamba
+{
+    namespace detail
+    {
+        struct ExcludeNewerPackageHash
+        {
+            using is_transparent = void;
```
JohanMabille: This is unused and should be removed.

libmamba/include/mamba/core/exclude_newer.hpp
```diff
@@ -0,0 +1,121 @@
...
+
+            [[nodiscard]] auto operator()(std::string_view value) const noexcept -> std::size_t
+            {
+                return std::hash<std::string_view>{}(value);
+            }
+        };
+
+        struct ExcludeNewerPackageEqual
+        {
+            using is_transparent = void;
```
JohanMabille: Same remark here.

libmamba/include/mamba/core/exclude_newer.hpp
```diff
@@ -0,0 +1,121 @@
...
+     * - ``std::nullopt`` exempts the package from the global policy (``false`` in config)
+     * - a timestamp value applies a package-specific cutoff
+     *
+     * Packages not listed fall back to the global cutoff.
+     */
+    using ExcludeNewerPackageCutoffs = std::unordered_map<
+        std::string,
+        std::optional<std::uint64_t>,
+        detail::ExcludeNewerPackageHash,
+        detail::ExcludeNewerPackageEqual>;
```
JohanMabille: It seems that the custom hasher and equality comparison implementations are similar to the standard ones for std::string, so why not simply use `std::unordered_map<std::stirng, std::optional<uint64_t>>` ?

libmamba/src/core/exclude_newer.cpp
```diff
@@ -0,0 +1,392 @@
...
+            const auto [ptr, ec] = std::from_chars(value.data(), value.data() + value.size(), number);
+            if (ec != std::errc())
+            {
+                return std::nullopt;
+            }
+            value.remove_prefix(static_cast<std::size_t>(ptr - value.data()));
+            return number;
+        }
+
+        [[nodiscard]] auto parse_fixed_uint(std::string_view value) -> std::optional<int>
```
JohanMabille: Why returning `int` instead of `std::uint64_t` ? This seems inconsistent with the other functions.

libmamba/src/core/exclude_newer.cpp
```diff
@@ -0,0 +1,392 @@
...
+            const auto day = parse_fixed_uint(value.substr(8, 2));
+            if (!year || !month || !day)
+            {
+                return std::nullopt;
+            }
+
+            std::tm tm = {};
+            tm.tm_year = *year - 1900;
+            tm.tm_mon = *month - 1;
+            tm.tm_mday = *day + 1;
```
JohanMabille: This does not work if `*day` is the last day of the month.
A solution would be to add a 24h offset if the tm is a simple date,  that would allow to factorize this function and the next one in a single function.

libmamba/src/core/exclude_newer.cpp
```diff
@@ -0,0 +1,392 @@
...
+            auto remaining = value;
+            const auto amount = parse_uint_prefix(remaining);
+            if (!amount)
+            {
+                return std::nullopt;
+            }
+            while (!remaining.empty() && util::is_space(remaining.front()))
+            {
+                remaining.remove_prefix(1);
+            }
```
JohanMabille: Maybe use `util::lstrip_if`  instead?

libmamba/src/core/exclude_newer.cpp
```diff
@@ -0,0 +1,392 @@
...
+                return 0;
+            }
+            return now_seconds - duration_seconds;
+        }
+
+        [[nodiscard]] auto
+        resolve_exclude_newer_cutoff_impl(std::string_view value, std::uint64_t now_seconds)
+            -> std::optional<std::uint64_t>
+        {
+            value = util::strip_if(value, [](char c) { return util::is_space(c); });
```
JohanMabille: Nitpicking: `util::strip_if(value, util::is_space)` should work and would make the code easier to read.

libmamba/src/core/exclude_newer.cpp
```diff
@@ -0,0 +1,392 @@
...
+
+            throw invalid_exclude_newer(value);
+        }
+
+    }  // namespace
+
+    auto resolve_exclude_newer_cutoff(std::string_view value, std::uint64_t now_seconds)
+        -> std::optional<std::uint64_t>
+    {
+        return resolve_exclude_newer_cutoff_impl(value, now_seconds);
```
JohanMabille: The implementation of `resolve_exclude_newer_cutoff_impl`  should be here, no need for an impl function that is called only here.

libmamba/src/core/exclude_newer.cpp
jjerphan: What do people think?

JohanMabille (changes requested):

JohanMabille:
More generally, using `std::chrono::parse` and the standard date and duration primitive from the C++ standard library would make the code more high level and easier to maintain.

libmamba/src/core/exclude_newer.cpp
JohanMabille: We need to check if we can parse all the formats, but if so, that would make the code much simpler.

jjerphan:
As discussed privately, I opened #4344 to have this present PR a simpler and smaller one to review, focusing on what @jezdez initially intended.

libmamba/src/core/exclude_newer.cpp
jjerphan: Done in #4344.

libmamba/src/solver/libsolv/helpers.hpp
```diff
@@ -37,6 +38,15 @@ namespace mamba::fs
 
 namespace mamba::solver::libsolv
 {
+    // Beyond this value, the timestamp would be in milliseconds and therefore should be
+    // converted to seconds.
+    inline constexpr std::uint64_t MAX_CONDA_TIMESTAMP = 253402300799ULL;
```
JohanMabille: Nitpicking: do we need this constant accessible elsewhere or should it be scoped to `normalize_conda_timestamp` ?

JohanMabille (approved):
LGTM

PR #4282, opened 2026-05-20
fix: Harden `noarch:python` entry point linking

jjerphan:
# Description

The motivation is to enforce strict trust boundaries between package metadata and filesystem writes: entry-point definitions are now validated, target paths are constrained to the environment prefix, and failures clearly identify the faulty distribution so maintainers can fix their package metadata. We also ensure transaction behavior is safer and more predictable when invalid entry points are encountered, so users do not end up with silently unsafe outcomes.

## Type of Change

- [x] Bugfix
- [ ] Feature / enhancement
- [ ] CI / Documentation
- [ ] Maintenance

## Checklist

### DIFF
--- a/libmamba/include/mamba/util/string.hpp
+++ b/libmamba/include/mamba/util/string.hpp
@@ -83,10 +83,10 @@ namespace mamba::util
     [[nodiscard]] auto contains(char c1, char c2) -> bool;
 
     /**
-     * Check if the string contains any of the given characters.
+     * Check if the string contains any of the given substrings.
      */
-    template <typename CharRange>
-    [[nodiscard]] auto contains_any(std::string_view str, const CharRange& chars) -> bool;
+    template <typename StrRange>
+    [[nodiscard]] auto contains_any(std::string_view str, const StrRange& substrings) -> bool;
 
     /**
      * Check if any of the strings starts with the prefix.
@@ -438,15 +438,18 @@ namespace mamba::util
     extern template bool starts_with_any(std::string_view, const std::vector<std::string>&);
     extern template bool starts_with_any(std::string_view, const std::vector<std::string_view>&);
 
-    template <typename CharRange>
-    auto contains_any(std::string_view str, const CharRange& chars) -> bool
+    template <typename StrRange>
+    auto contains_any(std::string_view str, const StrRange& substrings) -> bool
     {
         return std::any_of(
-            chars.begin(),
-            chars.end(),
-            [&str](const char c) { return contains(str, c); }
+            substrings.cbegin(),
+            substrings.cend(),
+            [&str](const auto& sub) { return contains(str, sub); }
         );
     }
+
+    extern template bool contains_any(std::string_view, const std::vector<std::string>&);
+    extern template bool contains_any(std::string_view, const std::vector<std::string_view>&);
 
     /***************************************
      *  Implementation of strip functions  *
--- a/libmamba/include/mamba/util/string.hpp
+++ b/libmamba/include/mamba/util/string.hpp
@@ -83,6 +83,12 @@ namespace mamba::util
     [[nodiscard]] auto contains(char c1, char c2) -> bool;
 
     /**
+     * Check if the string contains any of the given characters.
+     */
+    template <typename CharRange>
+    [[nodiscard]] auto contains_any(std::string_view str, const CharRange& chars) -> bool;
+
+    /**
      * Check if any of the strings starts with the prefix.
      */
     template <typename StrRange>
@@ -431,6 +437,16 @@ namespace mamba::util
 
     extern template bool starts_with_any(std::string_view, const std::vector<std::string>&);
     extern template bool starts_with_any(std::string_view, const std::vector<std::string_view>&);
+
+    template <typename CharRange>
+    auto contains_any(std::string_view str, const CharRange& chars) -> bool
+    {
+        return std::any_of(
+            chars.begin(),
+            chars.end(),
+            [&str](const char c) { return contains(str, c); }
+        );
+    }
 
     /***************************************
      *  Implementation of strip functions  *
--- a/libmamba/src/core/link.cpp
+++ b/libmamba/src/core/link.cpp
@@ -4,6 +4,7 @@
 //
 // The full license is in the file LICENSE, distributed with this software.
 
+#include <algorithm>
 #include <array>
 #include <cctype>
 #include <iostream>
@@ -49,9 +50,8 @@ namespace mamba
             python_identifier_chain_regex(R"(^[A-Za-z_][A-Za-z0-9_]*(\.[A-Za-z_][A-Za-z0-9_]*)*$)");
 
         constexpr std::array forbidden_entry_point_command_substrings = {
-            std::string_view("/"),  std::string_view("\\"), std::string_view("\0"),
-            std::string_view("\n"), std::string_view("\r"), std::string_view("\t"),
-            std::string_view(" "),
+            std::string_view("/"),  std::string_view("\\"), std::string_view("\n"),
+            std::string_view("\r"), std::string_view("\t"), std::string_view(" "),
         };
 
         auto check_python_identifier_chain(std::string_view value, std::string_view field_name)
@@ -84,7 +84,12 @@ namespace mamba
                 );
             }
 
-            if (util::contains_any(command, forbidden_entry_point_command_substrings))
+            if (util::contains_any(command, forbidden_entry_point_command_substrings)
+                || std::any_of(
+                    command.begin(),
+                    command.end(),
+                    [](char c) { return util::is_control(c); }
+                ))
             {
                 return make_unexpected(
                     fmt::format(
@@ -269,7 +274,7 @@ namespace mamba
 
         if (errors.size() == 1)
         {
-            return tl::unexpected(std::move(errors[0]));
+            return tl::unexpected(std::move(errors.front()));
         }
         if (errors.size() > 1)
         {
--- a/libmamba/src/core/link.cpp
+++ b/libmamba/src/core/link.cpp
@@ -5,11 +5,13 @@
 // The full license is in the file LICENSE, distributed with this software.
 
 #include <iostream>
+#include <iterator>
 #include <regex>
 #include <string>
 #include <tuple>
 #include <vector>
 
+#include <fmt/format.h>
 #include <reproc++/reproc.hpp>
 #include <reproc++/run.hpp>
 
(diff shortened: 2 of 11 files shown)


libmamba/src/core/link.cpp
```diff
@@ -34,7 +36,119 @@
...
+        const std::regex MENU_PATH_REGEX("^menu[/\\\\].*\\.json$", std::regex_constants::icase);
+
+        const std::regex
+            python_identifier_chain_regex(R"(^[A-Za-z_][A-Za-z0-9_]*(\.[A-Za-z_][A-Za-z0-9_]*)*$)");
+
+        void validate_python_identifier_chain(const std::string& value, const char* field_name)
+        {
+            if (!std::regex_match(value, python_identifier_chain_regex))
+            {
+                throw std::runtime_error(
```
Klaim: We should not use runtime_error directly, always use mamba_error. Same remark for all this code.

libmamba/src/core/link.cpp
```diff
@@ -34,7 +36,119 @@
...
 {
-    static const std::regex MENU_PATH_REGEX("^menu[/\\\\].*\\.json$", std::regex_constants::icase);
+    namespace
+    {
+        const std::regex MENU_PATH_REGEX("^menu[/\\\\].*\\.json$", std::regex_constants::icase);
+
+        const std::regex
+            python_identifier_chain_regex(R"(^[A-Za-z_][A-Za-z0-9_]*(\.[A-Za-z_][A-Za-z0-9_]*)*$)");
+
+        void validate_python_identifier_chain(const std::string& value, const char* field_name)
```
Klaim: I prefer validation and parsing functions to return an `expected` so that the caller decides if it's an error and how it should be thrown or reported etc. This also allows to accumulate errors before returning, if that makes sense in context.

Klaim: `field_name` should be a `string_view`

libmamba/src/core/link.cpp
```diff
@@ -34,7 +36,119 @@
...
+        void validate_entry_point_command(const std::string& command)
+        {
+            if (command.empty() || command == "." || command == "..")
+            {
+                throw std::runtime_error(
+                    fmt::format("Invalid entry point command name '{}': must be a simple file name", command)
+                );
+            }
+
+            if (command.find('/') != std::string::npos || command.find('\\') != std::string::npos
```
Klaim: I would prefer if this kind of code uses [`contains`](https://github.com/mamba-org/mamba/blob/16de9374209052d0d5e33639f923c9b5b2d565ec/libmamba/src/util/string.cpp#L234) or an overload of it that goes through a sequence of things to check if contained in the given string.

libmamba/src/core/link.cpp
```diff
@@ -34,7 +36,119 @@
...
+                    fmt::format(
+                        "Invalid entry point {} '{}': expected a dotted chain of Python identifiers",
+                        field_name,
+                        value
+                    )
+                );
+            }
+        }
+
+        void validate_entry_point_command(const std::string& command)
```
Klaim: Same remakr about error handling.

libmamba/src/core/link.cpp
```diff
@@ -34,7 +36,119 @@
...
+            const fs::u8path canon_prefix = fs::weakly_canonical(prefix, ec);
+            if (ec)
+            {
+                throw std::runtime_error(
+                    fmt::format("Cannot resolve environment prefix '{}': {}", prefix.string(), ec.message())
+                );
+            }
+
+            const auto [prefix_it, path_it] = std::mismatch(
+                canon_prefix.std_path().begin(),
```
Klaim: This code looks like this one: https://github.com/mamba-org/mamba/blob/16de9374209052d0d5e33639f923c9b5b2d565ec/libmamba/src/core/package_handling.cpp#L206

libmamba/src/core/link.cpp
```diff
@@ -156,9 +289,18 @@ namespace mamba
         std::string win_script = path.string() + "-script.py";
         std::string win_script_gen_str = path.generic_string() + "-script.py";
         fs::u8path script_path = target_prefix / win_script;
+        fs::u8path script_exe = path;
+        script_exe.replace_extension("exe");
+        fs::u8path script_exe_path = target_prefix / script_exe;
 #else
         fs::u8path script_path = target_prefix / path;
```
Klaim: `const`, same for windows versoin

libmamba/src/core/link.cpp
Klaim: Did you check what's mamba's behavior when an exception is thrown from this function? Does it actually cancels the whole operation?

libmamba/tests/src/core/test_link_entry_points.cpp
```diff
@@ -0,0 +1,51 @@
...
+
+#include "core/link.hpp"
+
+namespace mamba
+{
+    namespace
+    {
+        TEST_CASE("parse_entry_point accepts valid definitions")
+        {
+            const auto parsed = parse_entry_point("pip = pip._internal.cli.main:main");
```
Klaim, line 20: Arent spaces not accepted?

libmamba/src/core/link.cpp
```diff
@@ -34,7 +36,119 @@
...
+            {
+                throw std::runtime_error(
+                    fmt::format(
+                        "Invalid entry point command name '{}': absolute paths are not allowed",
+                        command
+                    )
+                );
+            }
+
+#if _WIN32
```
Klaim: Does this means that if a windows path is passed while running on a linux, this is acceptable?

jjerphan: It is not, I guess we can also include this check on other platforms.

libmamba/tests/src/core/test_link_entry_points.cpp
jjerphan: They are not for the command name. I'll add a comment.

libmamba/src/core/link.cpp
jjerphan: I added a commit so that it rolls back with a clear message indicating which package has a faulty entrypoint.

libmamba/include/mamba/util/string.hpp
```diff
@@ -82,6 +82,12 @@ namespace mamba::util
     [[nodiscard]] auto contains(std::string_view str, char c) -> bool;
     [[nodiscard]] auto contains(char c1, char c2) -> bool;
 
+    /**
+     * Check if the string contains any of the given characters.
+     */
+    template <typename CharRange>
+    [[nodiscard]] auto contains_any(std::string_view str, const CharRange& chars) -> bool;
```
Klaim: I don't think we want to check "characters" but really ranges of characters, to support unicode (UTF-8 to be precise) here. The second parameter should be a range of strings (or string-like types) , not a range of chars.

Maybe [`includes`](https://en.cppreference.com/cpp/algorithm/ranges/includes) (C++20 but allows holes in sequences) or [`contains_subrange`](https://en.cppreference.com/cpp/algorithm/ranges/contains) (C++23) would help here, not sure.

libmamba/src/core/link.cpp
```diff
@@ -92,15 +226,54 @@ namespace mamba
...
+            {
+                errors.push_back(check.error());
+            }
+        };
+
+        record_error(check_entry_point_command(result.command));
+        record_error(check_python_identifier_chain(result.module, "module"));
+        record_error(check_python_identifier_chain(result.func, "callable"));
+
+        if (errors.size() == 1)
```
Klaim, line 276: Why is this special case (1 error) treated specially?

libmamba/src/core/transaction.cpp
```diff
@@ -365,6 +366,37 @@ namespace mamba
...
+            TransactionRollback& rollback,
+            const Context& ctx,
+            const specs::PackageInfo& pkg,
+            std::string_view phase,
+            const std::exception& cause
+        )
+        {
+            rollback.rollback(ctx);
+
+            const auto error_code = dynamic_cast<const mamba_error*>(&cause) != nullptr
```
Klaim: Note: it's cheaper on runtime to check the typeid and compare than to do an dynamic cast. Consider doing the comparison through typeid, and then static cast?
I'm not 100% sure that would be always correct but it would avoid unnecessary runtime costs.
Although we can also ignore that if this code is not supposed to happen often and is not in a hot path.

libmamba/src/core/transaction.cpp
```diff
@@ -614,7 +646,14 @@ namespace mamba
...
                 UnlinkPackage up(pkg, cache_path, &transaction_context);
-                up.execute();
+                try
+                {
+                    up.execute();
+                }
+                catch (const std::exception& e)
+                {
+                    rethrow_transaction_cancelled_after_rollback(rollback, ctx, pkg, "unlinking", e);
+                }
```
Klaim, line 731: In both calls to execute, could you add a catch "anything" that error-logs that something unknown/unexpected happened?

libmamba/src/core/link.cpp
jjerphan: Indeed, it is done natively.

jjerphan: Actually, there are still important differences, so the branch is not redundant:

 - `mamba_aggregated_error` always sets error code to aggregated (constructor sets base code to `mamba_error_code::aggregated)`, even for one item.
 - Returning the original single `mamba_error` preserves the precise code (`invalid_spec`) and exact type/metadata.
 - Nit: Aggregated errors also append the bug-report footer by default, which is noisier for a single validation failure.

I guess we could adapt `mamba_aggregated_error`, but it might be out of the scope of this PR.

Klaim: Agreed, lets continue with that.

libmamba/include/mamba/util/string.hpp
```diff
@@ -432,6 +438,19 @@ namespace mamba::util
...
+    auto contains_any(std::string_view str, const StrRange& substrings) -> bool
+    {
+        return std::any_of(
+            substrings.cbegin(),
+            substrings.cend(),
+            [&str](const auto& sub) { return contains(str, sub); }
+        );
+    }
+
+    extern template bool contains_any(std::string_view, const std::vector<std::string>&);
```
Klaim, line 451: Is that purely for optimization? or did you have to add these specializations?

libmamba/src/core/link.cpp
```diff
@@ -34,7 +42,140 @@
...
-    static const std::regex MENU_PATH_REGEX("^menu[/\\\\].*\\.json$", std::regex_constants::icase);
+    namespace
+    {
+        const std::regex MENU_PATH_REGEX("^menu[/\\\\].*\\.json$", std::regex_constants::icase);
+
+        const std::regex
+            python_identifier_chain_regex(R"(^[A-Za-z_][A-Za-z0-9_]*(\.[A-Za-z_][A-Za-z0-9_]*)*$)");
+
+        constexpr std::array forbidden_entry_point_command_substrings = {
+            std::string_view("/"),  std::string_view("\\"), std::string_view("\n"),
```
Klaim: BTW you could use [`"/"sv`](https://en.cppreference.com/cpp/string/basic_string_view/operator%22%22sv) too if that helps with verbosity. I think that just requires adding `using namespace std::literals;` in the scope.

libmamba/src/core/transaction.cpp
```diff
@@ -365,6 +366,104 @@ namespace mamba
...
+                    phase,
+                    pkg.name,
+                    pkg.build_string,
+                    cause.what()
+                ),
+                mamba_error_code::internal_failure
+            );
+        }
+
+        [[noreturn]] void rethrow_transaction_cancelled_after_rollback_unknown(
```
Klaim: I think it can also be an overload? (same name)

Klaim (approved):
LGTM, my comments are minor suggestions or wondering why a choice was made but you can ignore them for merging 👍🏽

libmamba/include/mamba/util/string.hpp
jjerphan: I just kept the conventions of the file which I guess are there for optimization?

I am not against removing them.

libmamba/src/core/transaction.cpp
jjerphan: Indeed, 86541120322a741b2cbd1aefc5aa250abce8497b performs it and factorizes the logic.

libmamba/src/core/transaction.cpp
```diff
@@ -614,7 +721,18 @@ namespace mamba
...
+                    up.execute();
+                }
+                catch (const mamba_error& e)
+                {
+                    rethrow_transaction_cancelled_after_rollback(rollback, ctx, pkg, "unlinking", e);
+                }
+                catch (...)
+                {
+                    handle_unexpected_package_execute_exception(rollback, ctx, pkg, "unlinking");
+                }
```
jjerphan, line 735: Should we handle them differently (with another message than "An unknown error occurred")?

jjerphan:
@Klaim: I let you merge when the state is good for you.

libmamba/include/mamba/util/string.hpp
Klaim: it's fine, I was just wondering why.

libmamba/src/core/transaction.cpp
Klaim: I dont have an idea, the only thing we can do is log that something happened. Ideally we would print a callstack trace too to help figure out what's happening, but currently we can only do that by calling directly spdlog and we really dont want to do it that way. C++23 provides a way to get and print call stacks. Boost also (the stacktrace library was the prototype for the standard).

If you want to add more information, you could try to print the type name of the exception object, but you dont have direct access to it, you would have to use `std::current_exception()` which provides an opaque pointer to that object, but cannot be used like a pointer, the only way to use it is to rethrow it to re-catch it with the actual type (see [the example there](https://en.cppreference.com/cpp/error/current_exception)).
Otherwise if we had access to the object we could just use `typeid(e).name()` and print that.

Klaim: Let's maybe try to improve this kind of case (which is common in this codebase) once we have access to call stack?

jjerphan: I guess we could improve it, but is it a priority?

I am not against it, but I think that I would have our time focussed on defining what would best improve users' life and aiming at it. I think we need to discuss priorities within the team.

Klaim: > I guess we could improve it, but is it a priority?

I think it's a wrong way to frame this kind of issues and tasks, at least in the case of such tools project which have a long-term maintenance: it's not about helping users but about helping ourselves figure out problems quickly (by providing more information when unexpected cases occurs). 
It will never become a "priority" if we only focus on what the users reports, for example, and not on what would help us fix users problem as fast as possible.

But also: no it's not urgent, I'm not saying to do it soon. It would mostly be a way to help our maintenance work in the future. I mentionned that because I know that we have a lot of `catch(...)` in the codebase which might hide issues that happen in user's usage that we are not aware of. Taking note of it to track it later is enough for now, I suppose.

I agree with that we need to discuss priorities to clarify them, it would be nice to also have a regular priority issue sorting session for example, however that tends to burn even more dev time (in my experience). I disagree with ignoring internal/maintenance/implementation-details issues not visible to users: if we can make our work faster and easier, and make this software more stable, such work should not be ignored, otherwise the maintenance pain will only rise up. 🙈

jjerphan: Everything is about trade-offs I guess.

I guess the best would be that we continue the discussion privately within QuantStack.

PR #4283, opened 2026-05-21
fix: Guard Root Packages' Expansion

### LINKED ISSUE #4277: Micromamba 2.6.x: major performance degradation on non-sharded channels
### Troubleshooting docs

### Anaconda default channels

Micromamba

 #4274 (related but different code path — same release, sharded conda-forge, narrowed by python pin)   #4227 (the PR that introduced the regression I'm reporting)   #4244 (rename of repodata_use_shards → use_sharded_repodata)

### Latest version of Mamba

### Tried in Conda?

Not applicable

## Summary
  Solving a one-spec environment (`python=3.11`) against a non-sharded conda
  channel takes **~600 s** on micromamba 2.6.1 and **~30 s** on micromamba
  2.5.0, both with cold caches. The same spec against conda-forge (sharded)
  resolves in ~1.4 s on 2.6.1 and ~15 s on 2.5.0.

  | Channel | python= | 2.5.0 cold | 2.6.1 cold | Ratio |
  |---|---|---:|---:|---:|
  | non-sharded (Artifactory monolithic) | 3.11 | 29.85 s | 626.69 s | **21.0×** |
  | non-sharded | 3.12 | 17.57 s | 644.96 s | 36.7× |
  | non-sharded | 3.14 | 17.58 s | 651.95 s | 37.1× |
  | conda-forge | 3.11 | 15.00 s | 1.48 s | 0.10× |
  | conda-forge | 3.12 | 14.83 s | 1.41 s | 0.10× |
  | conda-forge | 3.14 | 14.32 s | 1.41 s | 0.10× |

  The conda-forge side is exactly the sharded fast-path win you'd expect. 
  The non-sharded side is a regression of the same magnitude in the opposite direction.

  ## Root cause
_Note an LLM was used for the code analysis so take it with a grain of salt_

  A verbose run shows: HEAD probes for `repodata_shards.msgpack.zst` 404 as expected (non-sharded server), then `repodata.json.zst` is fetched, decompressed, and parsed into libsolv in ~16 s ("Done (11.6 sec)" /
  "(4.8 sec)" for linux-64 / noarch). Then logging falls silent while the process pegs one core at 100% CPU for ~10 minutes before the solver runs.

  That silent phase is `expand_shard_root_packages_from_full_repodata_repos` in `libmamba/src/api/channel_loader.cpp`, added in PR #4227 ("fix: Support mixture of flat and sharded repodata"). It runs whenever `ctx.use_sharded_repodata && !root_packages.empty()`:

  ```cpp
  const bool shard_then_expand =
      ctx.use_sharded_repodata && !root_packages.empty();
  if (shard_then_expand) {
      for (std::size_t i = 0; i < subdirs.size(); ++i)
          try_load(i, /*full_repodata_only_pass=*/true);  // loads flat repos
      expand_shard_root_packages_from_full_repodata_repos(
          database, full_repos_for_shard_roots, root_packages);
      // … then a second pass loads shard subdirs
  }
  ```

  The expansion BFS does a sequential scan of every record in every flat
  repo, per dependency name discovered [code](https://github.com/mamba-org/mamba/blob/2.6.1/libmamba/src/api/channel_loader.cpp#L544-L571):

  ```cpp
  while (!frontier.empty()) {
      const std::string pkg_name = std::move(frontier.back());
      frontier.pop_back();
      for (const auto& repo : full_repos)
          database.for_each_package_in_repo(repo,
              [&](const specs::PackageInfo& pkg) {
                  if (pkg.name != pkg_name) return;  // O(N) scan per name
                  for (const auto& dep : pkg.dependencies) add_from_spec(dep);
                  for (const auto& c   : pkg.constrains)   add_from_spec(c);
              });
  }
  ```

  For a flat repo with ~100k records and a transitive-dep closure of
  hundreds of names (typical for `python`), that's tens of millions of name
  comparisons — minutes of single-threaded CPU work, every invocation, for
  a query that should resolve in milliseconds.

  **The pass is intended to seed `root_packages` for *sharded* loads on
  *other* channels.** The guard checks only `use_sharded_repodata &&
  !root_packages.empty()` — it doesn't check whether any sharded subdir is
  actually present. When the channel set is pure-flat, the expansion's
  output is consumed by nobody and the entire cost is wasted.

  ## Toggle test (confirms diagnosis)

  Setting `use_sharded_repodata: false` in `~/.mambarc` (or any condarc):

  | `use_sharded_repodata` | source | elapsed |
  |---|---|---:|
  | `true` (default) | — | 626.69 s |
  | `false` | `--rc-file /tmp/mmrc.yaml` | 44.19 s |
  | `false` | `~/.mambarc` | **29.66 s** |

  Verbose log under the disabled flag contains no `repodata_shards.*`
  requests, no "Shard index check failed" lines, and no `expand_shard_…`
  activity — only "Using Flat Repodata for …" markers. So the flag
  short-circuits the shard code path at runtime, not merely at config
  read.

  ## Suggested fix

  Gate the expansion on actual sharded presence in the load list:

  ```cpp
  const bool any_sharded = std::any_of(
      subdirs.begin(), subdirs.end(),
      [&](const auto& s) {
          return s.metadata().has_up_to_date_shards(ctx.repodata_shards_ttl);
      });
  const bool shard_then_expand =
      ctx.use_sharded_repodata && !root_packages.empty() && any_sharded;
  ```

  Independently, the BFS itself would benefit from a `name → records` hash
  index built once before the loop, reducing per-name lookup from O(N) to
  O(1) and making the expansion cheap even when it does run on mixed
  channel sets.

  ## Side finding

  `use_sharded_repodata` is registered with `.set_rc_configurable()` only
  in `libmamba/src/api/configuration.cpp` — no env binding. Setting
  `MAMBA_USE_SHARDED_REPODATA=false` is silently ignored:
  `micromamba config list --all -s` still reports
  `use_sharded_repodata: true # 'default'`. RC files work; env vars do
  not. Probably worth a small follow-up either way.

```shell
libmamba version : 2.6.1
       micromamba version : 2.6.1
             curl version : libcurl/8.20.0 OpenSSL/3.5.6 zlib/1.3.2 zstd/1.5.7 libssh2/1.11.1 nghttp2/1.68.1 mit-krb5/1.22.2
       libarchive version : libarchive 3.8.7 zlib/1.3.2 bz2lib/1.0.8 libzstd/1.5.7 libb2/bundled
         envs directories : /home/<user>/micromamba/envs
            package cache : /home/<user>/micromamba/pkgs
                            /home/<user>/.mamba/pkgs
         virtual packages : __unix=0=0
                            __linux=5.14.0=0
                            __glibc=2.34=0
                            __archspec=1=x86_64_v4
         base environment : /home/<user>/micromamba
                 platform : linux-64
```

### Logs

```shell
Reproducer (verbose output, redacted channel URL, with shards enabled — the failing case):

  $ rm -rf ~/micromamba/pkgs envs/_t
  $ time micromamba -vvv create -y --dry-run -p envs/_t \
      -c <non-sharded-channel> --override-channels python=3.11
  …
  debug    libmamba Pre-filtering shards using explicitly requested python minor version: 3.11
  info     libmamba Searching index cache file for repo '<non-sharded-channel>/linux-64'
  info     libmamba Valid cache found  for '<non-sharded-channel>/linux-64': 0
  info     libcurl  > HEAD /<…>/linux-64/repodata.json.zst HTTP/1.1
  info     libcurl  > HEAD /<…>/linux-64/repodata_shards.msgpack.zst HTTP/1.1
  info     libcurl  > HEAD /<…>/noarch/repodata.json.zst HTTP/1.1
  info     libcurl  > HEAD /<…>/noarch/repodata_shards.msgpack.zst HTTP/1.1
  debug    libmamba Shard index check failed: <…>/noarch/repodata_shards.msgpack.zst [404]
  debug    libmamba Shard index check failed: <…>/linux-64/repodata_shards.msgpack.zst [404]
  info     libcurl  > GET /<…>/linux-64/repodata.json.zst HTTP/1.1
  info     libcurl  > GET /<…>/noarch/repodata.json.zst HTTP/1.1
  ⚠ Shard Index for <non-sharded-channel>/linux-64 not available, falling back to flat repodata
  Using Flat Repodata for <non-sharded-channel>/linux-64   ✔ Starting
  Using Flat Repodata for <non-sharded-channel>/linux-64   ✔ Done (11.6 sec)
  ⚠ Shard Index for <non-sharded-channel>/noarch not available, falling back to flat repodata
  Using Flat Repodata for <non-sharded-channel>/noarch     ✔ Starting
  Using Flat Repodata for <non-sharded-channel>/noarch     ✔ Done (4.8 sec)
  # <<< process then sits at 100% CPU on a single core for ~10 minutes
  #     with no further log output, before the solver runs >>>
  real    10m26.69s

  Same command with use_sharded_repodata: false in ~/.mambarc:

  real    0m29.66s
  # log shows only "Using Flat Repodata …" markers,
  # no HEAD repodata_shards.* requests, no "Shard index check failed",
  # no expand_shard_… activity
```

### environment.yml

```yaml
name: repro
  channels:
    - <a non-sharded conda channel serving monolithic repodata.json>  # e.g. an Artifactory conda proxy
  dependencies:
    - python=3.11

  (The bug is triggered by the channel topology, not the env contents — any single non-trivial spec against a non-sharded channel
  reproduces it.)
```

### ~/.condarc

```yaml
changeps1: false
channel_priority: strict
```

jjerphan:
# Description

Fix https://github.com/mamba-org/mamba/issues/4277.

## Type of Change

- [x] Bugfix
- [ ] Feature / enhancement
- [ ] CI / Documentation
- [ ] Maintenance

## Checklist

### DIFF
--- a/libmamba/tests/src/core/test_channel_loader.cpp
+++ b/libmamba/tests/src/core/test_channel_loader.cpp
@@ -22,23 +22,13 @@
 #include "mamba/solver/libsolv/repo_info.hpp"
 
 #include "mambatests.hpp"
+#include "mambatests_utils.hpp"
 
 using namespace mamba;
+using mambatests::make_simple_channel;
 
 namespace
 {
-    [[nodiscard]] auto make_simple_channel(std::string_view chan) -> specs::Channel
-    {
-        const auto resolve_params = ChannelContext::ChannelResolveParams{
-            { "linux-64", "noarch" },
-            specs::CondaURL::parse("https://conda.anaconda.org").value()
-        };
-
-        return specs::Channel::resolve(specs::UnresolvedChannel::parse(chan).value(), resolve_params)
-            .value()
-            .front();
-    }
-
     [[nodiscard]] auto make_subdir_loader(
         const specs::Channel& channel,
         const fs::u8path& cache_root,
--- a/libmamba/tests/src/core/test_channel_loader.cpp
+++ b/libmamba/tests/src/core/test_channel_loader.cpp
@@ -4,20 +4,89 @@
 //
 // The full license is in the file LICENSE, distributed with this software.
 
+#include <chrono>
+#include <fstream>
 #include <map>
+#include <sstream>
 
 #include <catch2/catch_all.hpp>
+#include <nlohmann/json.hpp>
 
 #include "mamba/api/channel_loader.hpp"
 #include "mamba/core/channel_context.hpp"
 #include "mamba/core/context.hpp"
 #include "mamba/core/package_cache.hpp"
+#include "mamba/core/subdir_index.hpp"
 #include "mamba/core/util.hpp"
 #include "mamba/solver/libsolv/database.hpp"
+#include "mamba/solver/libsolv/repo_info.hpp"
 
 #include "mambatests.hpp"
 
 using namespace mamba;
+
+namespace
+{
+    [[nodiscard]] auto make_simple_channel(std::string_view chan) -> specs::Channel
+    {
+        const auto resolve_params = ChannelContext::ChannelResolveParams{
+            { "linux-64", "noarch" },
+            specs::CondaURL::parse("https://conda.anaconda.org").value()
+        };
+
+        return specs::Channel::resolve(specs::UnresolvedChannel::parse(chan).value(), resolve_params)
+            .value()
+            .front();
+    }
+
+    [[nodiscard]] auto make_subdir_loader(
+        const specs::Channel& channel,
+        const fs::u8path& cache_root,
+        const std::string& platform = "linux-64"
+    ) -> SubdirIndexLoader
+    {
+        auto caches = MultiPackageCache({ cache_root }, ValidationParams{});
+        return SubdirIndexLoader::create({}, channel, platform, caches).value();
+    }
+
+    void write_chain_repodata(const fs::u8path& repodata_path, std::size_t package_count)
+    {
+        nlohmann::json repodata;
+        repodata["info"]["subdir"] = "linux-64";
+        repodata["repodata_version"] = 1;
+        repodata["packages"] = nlohmann::json::object();
+
+        for (std::size_t i = 0; i < package_count; ++i)
+        {
+            nlohmann::json record;
+            record["name"] = "pkg-" + std::to_string(i);
+            record["version"] = "1.0";
+            record["build"] = "0";
+            record["build_number"] = 0;
+            record["subdir"] = "linux-64";
+            record["depends"] = nlohmann::json::array();
+            if (i > 0)
+            {
+                record["depends"].push_back("pkg-" + std::to_string(i - 1) + " >=1.0");
+            }
+            repodata["packages"]["pkg-" + std::to_string(i) + "-1.0-0.tar.bz2"] = std::move(record);
+        }
+
+        nlohmann::json python;
+        python["name"] = "python";
+        python["version"] = "3.11";
+        python["build"] = "0";
+        python["build_number"] = 0;
+        python["subdir"] = "linux-64";
+        python["depends"] = nlohmann::json::array(
+            { "pkg-" + std::to_string(package_count - 1) + " >=1.0" }
+        );
+        repodata["packages"]["python-3.11-0.tar.bz2"] = std::move(python);
+
+        std::ofstream out(repodata_path.string());
+        out << repodata.dump();
+    }
+}
 
 TEST_CASE("init_channels", "[mamba::api][channel_loader]")
 {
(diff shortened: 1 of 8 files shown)


libmamba/tests/src/core/test_channel_loader.cpp
```diff
@@ -4,21 +4,90 @@
...
+    {
+        const auto resolve_params = ChannelContext::ChannelResolveParams{
+            { "linux-64", "noarch" },
+            specs::CondaURL::parse("https://conda.anaconda.org").value()
+        };
+
+        return specs::Channel::resolve(specs::UnresolvedChannel::parse(chan).value(), resolve_params)
+            .value()
+            .front();
+    }
```
Hind-M: This seems to be common to a bunch of test files. We should declare it in a common header file in `tests/include` and leave implementation here (or in `test_subdir_index.cpp` and remove the others).

jjerphan: What do you think of 382be05?

Hind-M (approved):

PR #4284, opened 2026-05-21
JSON output api rewritten to avoid JSON flattening

Klaim:
# Description

This changes the API in `mamba::Console` to specify the JSON output in a way to enables us to implement it without JSON "flattening". The flattening was destroying information about empty arrays and dictionary, which lead to issues with completing #4202 due to ending up with null values instead of empty arrays for example.

For the using code the main change is that there is not anymore a stack-based or object-merge way to inject information on the json object. Now values have to be explicitly set at specific addresses of in the output json object.

This PR does not attempt to fix the issue of comands implementing the json output through direct call to the standard ouput followed by exit (like with `list` command) which does not call the `Console` json output system and leads to inconsistencies in what the json output looks like. This is issue is completely covered in #4202 

There is an opportunity to make that new api thread-safe but because we dont need it at the moment I didnt implement it. Instead I took in count in the API design possibility of adding a locking mechanism in the future without having to change the API. I also added an error when we detect concurrent calls to that API, which currently never triggers but might in the future, so at least we will be warned.

I attempted to add some tests to that API but because of `Console` being a singleton, I ended up relying on the current python tests instead. A way to do it properly would be to extract the json output mechaism from `Console` so that at least it can be tested in isolation, but I decided to not spend more time on that for now. Tell me if you think it would be worth the time.

## Type of Change

- [x] Bugfix
- [x] Feature / enhancement
- [ ] CI / Documentation
- [ ] Maintenance

## Checklist


libmamba/src/core/output.cpp
```diff
@@ -387,7 +393,7 @@ namespace mamba
 
     void Console::print_in_place(std::string_view str, bool finalize, bool force_print)
     {
-        if (force_print || !(context().output_params.quiet || context().output_params.json))
+        if (force_print or !(context().output_params.quiet or context().output_params.json))
```
JohanMabille: Nitpicking: not instead of ! to be consistent with previous change.

libmamba/src/core/transaction.cpp
```diff
@@ -223,8 +223,10 @@ namespace mamba
         // if no action required, don't even start logging them
         if (!empty())
         {
-            Console::instance().json_down("actions");
-            Console::instance().json_write({ { "PREFIX", ctx.prefix_params.target_prefix.string() } });
+            Console::instance().set_json_output(
+                { .to_assign{
+                    { "/actions/PREFIX"_json_pointer, ctx.prefix_params.target_prefix.string() } } }
```
JohanMabille: NItpicking: formatting (last two curly braces on new line)

libmamba/src/core/transaction.cpp
```diff
@@ -479,19 +480,14 @@ namespace mamba
...
-        // back to the top level if any action was required
-        if (!empty())
-        {
-            Console::instance().json_up();
-        }
-        Console::instance().json_write(
-            { { "dry_run", ctx.dry_run }, { "prefix", ctx.prefix_params.target_prefix.string() } }
+        Console::instance().set_json_output(
+            { .to_assign{ { "/dry_run"_json_pointer, ctx.dry_run },
+                          { "/prefix"_json_pointer, ctx.prefix_params.target_prefix.string() } } }
```
JohanMabille: Same here.

JohanMabille (commented):
LGTM, only some nitpicking comments ;)

libmamba/src/core/transaction.cpp
Klaim: I'll try but clang-format might not let me, maybe I'll use the no-formatting comment.

JohanMabille: Ha, I don't want this to be a blocker, it was just nitpicking, feel free to ignore this comment if the formatter is resisting too much ;)

JohanMabille (approved):

JohanMabille:
Let's wait for 2.7.0 before merging this one.

PR #3625, opened 2024-11-21
ci: add brew toolchain test

henryiii:
Followup to #3613, adds CI tests for the brew toolchain. These take 4 and 5 minutes, and add the only
CI jobs with AppleClang and GCC <13 (11.4, specifically).

PR #4295, opened 2026-05-26
ci: Remove `linuxbrew` workflow

jjerphan:
# Description

Reason: [Homebrew on Linux is deprecated](https://github.com/linuxbrew).

## Type of Change

- [ ] Bugfix
- [ ] Feature / enhancement
- [x] CI / Documentation
- [ ] Maintenance

## Checklist

### DIFF
--- a/.github/workflows/brew.yml
+++ b/.github/workflows/brew.yml
@@ -1,47 +1,9 @@
-name: Homebrew and Linuxbrew toolchains
+name: Homebrew toolchain
 
 on:
   workflow_call:
 
 jobs:
-  build_linuxbrew:
-    name: Build on linuxbrew
-    runs-on: ubuntu-latest
-    container:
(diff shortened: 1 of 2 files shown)


Klaim (approved):

Hind-M:
It seems this [workflow](https://github.com/mamba-org/mamba/pull/3625) was added after deprecating https://github.com/linuxbrew which was actually just moved to https://github.com/Homebrew.
So I don't think the issue is deprecation, but something else...

jjerphan:
Independently from failures, do we want to maintain this workflow? To me, this just adds to the maintenance burden without significant benefits; who is using Homebrew on Linux? Why would we want to maintain this workflow?

Hind-M:
Well yes that's the question we need to answer. If it was added for some reason then we should keep it.

Issue #4297, opened 2026-05-27
ci: Ubuntu integration test are now slower


Hind-M:
The CI jobs are noticed to be slow from a couple of versions back of mamba.
Maybe specifically on linux but not necessarily.
As @Klaim mentioned internally, this [call](https://github.com/mamba-org/mamba/blob/main/libmamba/src/api/utils.cpp#L499) takes most of the time, and this needs more investigation.

Hind-M:
cc @jjerphan

jjerphan:
It only seems to affect Linux, this is also in line with other recently reported issues like https://github.com/mamba-org/mamba/issues/4272.

jjerphan:
With shards:
```yaml
channels:
  - conda-forge
dependencies:
  - python=3.9
  # - xeus-cling=0.6.0
  - xtensor=0.20.8
  - xtensor-blas=0.16.1
  - notebook
```

resolves in 2.4 seconds whilst

```yaml
channels:
  - conda-forge
dependencies:
  - python=3.9
  - xeus-cling=0.6.0
  - xtensor=0.20.8
  - xtensor-blas=0.16.1
  - notebook
```

resolves in 50 seconds.

Klaim:
Are we sure it's specific to linux?

I cannot confirm on Windows because `xeus-cling` cannot be solved (I get an error because the package doesnt exist), so the related ci test which runs only on linux anyway might be misleading.
I tried (still on Windows) the first env link provided in #4272 but stopped after more than 2 minutes passed (that env is massive)

It would make sense if it's a general speed regression not specific to linux but we only see it there because more people use linux + mamba?

jjerphan:
I have identified https://github.com/mamba-org/mamba/commit/395941f9aad3b6c64602455c9f03c21945efc204 as one culprit for the slowness (if we remove it, the environment resolves in less than 4 seconds), but this logic is required for other environments to be resolvable.

jjerphan:
Fixed by https://github.com/mamba-org/mamba/pull/4298.

Issue #4274, opened 2026-05-13
Huge regression in solving time between 2.5.0 and 2.6.0


JGobeil:
### Troubleshooting docs

- [x] My problem is not solved in the Troubleshooting docs

### Anaconda default channels

- [x] I do NOT use the Anaconda default channels (pkgs/* etc.)

### How did you install Mamba?

Mambaforge or latest Miniforge

### Search tried in issue tracker

 #4272, #4263 

### Latest version of Mamba

- [x] My problem is not solved with the latest version

### Tried in Conda?

I do not have this problem with Conda, just with Mamba

### Describe your issue

Version 2.6.0 is ~10x solver to solve environments than 2.5.0.

This is on a fresh docker image (docker.io/library/ubuntu:latest) with a fresh installation of miniforge.

Time to solve the environment (environment file below):
- conda 26.3.2: 2.7s
- mamba 2.5.0:  53s
- mamba 2.6.0:  5s

The problem seems to be worse on an unclean cache. On my laptop, after mamba update, `mamba` was taking forever to solve the environment. It was a better after cleaning all caches, but was still really slower than before.

Maybe link to #4272 and #4263 

Tests run: 

``` bash
(base) root@8f5d80fd02ed:/# conda --version 
conda 26.3.2

(base) root@8f5d80fd02ed:/# time conda env create --dry-run --quiet -n test -f environment.yml
Channels:
 - conda-forge
Platform: linux-64
Collecting package metadata (repodata.json): ...working... done
Solving environment: ...working... done
name:
channels:
  - conda-forge
dependencies:
  - conda-forge/noarch::_x86_64-microarch-level==4=3_zen4
  - conda-forge/noarch::ca-certificates==2026.4.22=hbd8a1cb_0
  - conda-forge/noarch::font-ttf-dejavu-sans-mono==2.37=hab24e00_0
  - conda-forge/noarch::font-ttf-inconsolata==3.000=h77eed37_0
  - conda-forge/noarch::font-ttf-source-code-pro==2.038=h77eed37_0
  - conda-forge/noarch::font-ttf-ubuntu==0.83=h77eed37_3
  - conda-forge/linux-64::hicolor-icon-theme==0.17=ha770c72_3
  - conda-forge/noarch::kernel-headers_linux-64==6.12.0=he073ed8_6
  - conda-forge/noarch::libgcc-devel_linux-64==14.3.0=hf649bbc_119
[...]

PR #4298, opened 2026-05-27
fix: Adapt root packages expansion

### LINKED ISSUE #4272: micromamba 2.6.0 can't resolve
I give the latest a list of packages to create a new env with and it takes hours and MANY GBs of memory to resolve (or more likely I run out of RAM/patience and aborts).
Downgraded to 2.5.0 same task --- resolves in minutes and uses a fraction of the RAM.

jjerphan:
# Description

Fix #4272, #4274, #4297, and [this issue](https://gist.github.com/minrk/5fdabeb7ab8cd2c69cbc27588fed1903).

- Reduces sharded-repodata solve regressions by removing expensive shard-pass root expansion and using indexed full-repodata root expansion.
- Uses a name-to-record mapping during shard root expansion from full repodata, improving runtime by avoiding repeated full scans.
- Adds sharded non-regression integration tests for:
  - Apache Arrow docs env (`apache_arrow_conda_env_sphinx.txt`)
  - Issue #4274 env (`issue_4274.env.txt`)
  - Omni env (`omni.env.txt`, issue #4277 reference)
- Reverts time out of `test_env_logging_overhead_regression`.

## Type of Change

- [x] Bugfix
- [ ] Feature / enhancement
- [ ] CI / Documentation
- [ ] Maintenance

## Checklist


Hind-M:
I'm not sure I understand what this PR is doing. AFAICT, #4245 was reverted, then we use a mapping instead of doing things in place (is that really necessary?).
So we don't need changes on #4245 anymore? What was its purpose in the first place?

micromamba/tests/test_create.py
```diff
@@ -918,7 +918,7 @@ def test_clone_environment_with_many_packages(tmp_home, tmp_root_prefix, tmp_pat
 
 # Only run this test on Linux, as it is the only platform where xeus-cling
 # (which is part of the environment) is available.
-@pytest.mark.timeout(120)
+@pytest.mark.timeout(20)
```
Hind-M, line 921: How much time does it take now? If it's slightly under 20 sec, I think it would be wise to add some margin and set the timeout to 30 sec for example.

jjerphan: This test takes less than 10 seconds on my machine to run:

```
 → export TEST_MAMBA_EXE=/home/jjerphan/dev/mamba/build/micromamba/micromamba
unset CONDARC  # Interferes with test
python -m pytest micromamba/tests/test_create.py -k test_env_logging_overhead_regression
===================================================================================== test session starts =====================================================================================
platform linux -- Python 3.12.12, pytest-9.0.2, pluggy-1.6.0
rootdir: /home/jjerphan/dev/mamba
configfile: pyproject.toml
plugins: asyncio-1.3.0, cov-7.0.0, rerunfailures-16.1, timeout-2.4.0, xprocess-1.0.2
asyncio: mode=Mode.STRICT, debug=False, asyncio_default_fixture_loop_scope=None, asyncio_default_test_loop_scope=function
collected 1954 items / 1953 deselected / 1 selected                                                                                                                                           

micromamba/tests/test_create.py .                                                                                                                                                       [100%]

============================================================================= 1 passed, 1953 deselected in 9.82s ==============================================================================
```

jjerphan:
> I'm not sure I understand what this PR is doing. AFAICT, https://github.com/mamba-org/mamba/pull/4245 was reverted, then we use a mapping instead of doing things in place (is that really necessary?). So we don't need changes on https://github.com/mamba-org/mamba/pull/4245 anymore? What was its purpose in the first place?

Yes, https://github.com/mamba-org/mamba/pull/4245 was a correction for a last minute issue I observed just after we decided releasing 2.6.0 (see [the non-regression test](https://github.com/mamba-org/mamba/pull/4245/changes#diff-3f66b5f68eaf75cecccf1852b8783880c334bb5a5eda67260f465b363ed22cb8R624-R681)). But it is in fact too broad and comes with massive runtime for larger environment.

It is way too gross, and we can in fact only use `expand_shard_root_packages_from_full_repodata_repos` (whose runtime was roughly quadratic in the number of packages in the worst case (supposing the number of dependencies for each package is a small constant) and which is now linear thanks to the mapping constructed at its start).

Hind-M:
Ok, so this is only related to the sharded case right? IIUC the mentioned issues are related to the non sharded case as well? How is this fixing that?

jjerphan:
Root packages' expansion is necessary when a mixture of channel with shards and without shards are used. So this also was a problem when channels without shards are used (such as `bioconda`) with channels with shards (such as `conda-forge`).

This also was a problem in some configurations when channels with shards are used.

(#4277 is different: currently root packages' expansion is run, but it must not, and a guard like the one proposed in #4285 or #4283 will prevent it.)

libmamba/tests/src/core/test_sharded_repodata_integration.cpp
```diff
@@ -449,6 +450,43 @@ namespace
...
+        {
+            if (line.starts_with("http://") || line.starts_with("https://"))
+            {
+                urls.push_back(line);
+            }
+        }
+        return urls;
+    }
+
+    auto read_yaml_dependencies_specs(const fs::u8path& path) -> std::vector<std::string>
```
Hind-M: Why not use `dependencies` from `read_yaml_file` directly? This function is not robust as it's supposing a certain indentation for example.

libmamba/tests/src/core/test_sharded_repodata_integration.cpp
```diff
@@ -449,6 +450,43 @@ namespace
             REQUIRE(is_newer);
         }
     }
+
+    auto read_explicit_urls(const fs::u8path& path) -> std::vector<std::string>
```
Hind-M: Maybe move this to some utility file (`util.{hpp,cpp}`)?

libmamba/tests/src/core/test_sharded_repodata_integration.cpp
```diff
@@ -781,6 +819,257 @@ TEST_CASE("Sharded repodata - solve pyjs-obspy env specs on emscripten", "[mamba
...
+            ctx,
+            channel_context,
+            update_specs,
+            /*is_retry=*/false,
+            /*no_py_pin=*/false
+        );
+        (void) package_caches;
+
+        auto prefix_data = load_prefix_data_and_installed(ctx, channel_context, db);
+        (void) prefix_data;
```
Hind-M:
```suggestion
        load_prefix_data_and_installed(ctx, channel_context, db);
```
As `load_prefix_data_and_installed` is not marked `[[nodiscard]]`.

libmamba/tests/src/core/utils.cpp
```diff
@@ -0,0 +1,25 @@
...
+//
+// The full license is in the file LICENSE, distributed with this software.
+
+#include "mamba/core/util.hpp"
+
+#include "utils.hpp"
+
+namespace mambatests
+{
+    auto read_explicit_urls(const mamba::fs::u8path& path) -> std::vector<std::string>
```
Hind-M: I meant put it directly within the existing files `libmamba/src/api/utils.{hpp, cpp}` even if this is used only in tests for now.

But if you prefer to put it in the tests folder, then either put it in `mambatests_utils.hpp` or rename the new file you added to `test_utils.hpp` and move it to `tests/include/` (no need for a `.cpp` file and it would be confusing with the other tests, if you want to separate the implementation then we can keep it in `test_sharded_repodata_integration.cpp` directly).

But I'm more in favor of the first option.

Hind-M (approved):

JGobeil:
Thanks a lot for fixing that!

jjerphan:
Thank you. Do you confirm that 2.7.0 resolved the issue?

PR #4304, opened 2026-06-04
fix: Restore cross-channel expansion

jjerphan:
# Description: Reintroduce cross-channel expansion

When sharded repodata is enabled and at least one channel has shards, channel loading runs in two phases:

1. **Flat-first pass** — subdirs **without** a shard index (e.g. **bioconda**) which loads the full `repodata.json`.
2. **Shard pass** — subdirs **with** shards (e.g. **conda-forge**, **emscripten-forge-dev**) load only shards reachable from `root_packages`.

### Flat channels: `expand_shard_root_packages_from_full_repodata_repos`

After the flat-first pass, dependency names reachable from the current roots are collected from those **full-repodata** repos (indexed BFS by package name). That extends `root_packages` before the shard pass so sharded channels (e.g. conda-forge) can fetch the right shards when a flat channel already contributed the closure (typical **conda-forge + bioconda** case, e.g. omni env).

### All-sharded channels: `expand_shard_root_packages_from_shard_loaded_packages`

When **no** flat channel was loaded in the first pass (`full_repos_for_shard_roots` is empty), shard-loaded package metadata is used to extend `root_packages` and, if needed, **one** extra shard pass runs. That fixes **shard-only** cross-channel / cross-subdir cases (e.g. **xeus-python** on emscripten-forge-dev with **noarch** deps on conda-forge) where flat repodata never ran.

### Why shard-metadata closure is skipped when flat channels are present?

If bioconda (or any non-sharded channel) was already loaded flat, roots are seeded via `expand_shard_root_packages_from_full_repodata_repos`. Running shard-metadata expansion and a follow-up shard reload on top of that would duplicate work and can explode cost (large env files + conda-forge shards). So `expand_shard_root_packages_from_shard_loaded_packages` and the second shard pass are **disabled** whenever the flat-first pass loaded at least one repo.

Pure flat channel sets still skip the whole path via `should_shard_then_expand_roots()` (no shards at all).

---

Also adapt the visibility of some functions.

## Type of Change

- [x] Bugfix
- [ ] Feature / enhancement
- [ ] CI / Documentation
- [ ] Maintenance

## Checklist

### DIFF
--- a/libmamba/include/mamba/api/channel_loader.hpp
+++ b/libmamba/include/mamba/api/channel_loader.hpp
@@ -7,12 +7,14 @@
 #ifndef MAMBA_API_CHANNEL_LOADER_HPP
 #define MAMBA_API_CHANNEL_LOADER_HPP
 
+#include <map>
 #include <optional>
 #include <set>
 #include <string>
 #include <vector>
 
 #include "mamba/core/error_handling.hpp"
+#include "mamba/specs/package_info.hpp"
 #include "mamba/specs/version.hpp"
 
 namespace mamba
@@ -38,7 +40,8 @@ namespace mamba
      *
      * @param ctx Context (use_sharded_repodata, shard TTL, download params, etc.).
      * @param database Libsolv database to add repos into.
-     * @param root_packages Root package names for reachability (e.g. install specs).
+     * @param root_packages Root package names for reachability (e.g. install specs); may be
+     *                      extended with dependency names discovered in loaded shard packages.
      * @param subdirs All subdir loaders; \p subdir_idx is the one to load.
      * @param subdir_idx Index of the subdir to load in \p subdirs.
      * @param loaded_subdirs_with_shards Set of subdir names already loaded via shards (updated).
@@ -50,7 +53,7 @@ namespace mamba
     auto load_subdir_with_shards(
         Context& ctx,
         solver::libsolv::Database& database,
-        const std::vector<std::string>& root_packages,
+        std::vector<std::string>& root_packages,
         std::vector<SubdirIndexLoader>& subdirs,
         std::size_t subdir_idx,
         std::set<std::string>& loaded_subdirs_with_shards,
@@ -84,6 +87,18 @@ namespace mamba
     void expand_shard_root_packages_from_full_repodata_repos(
         const solver::libsolv::Database& database,
         const std::vector<solver::libsolv::RepoInfo>& full_repos,
+        std::vector<std::string>& root_packages
+    );
+
+    /**
+     * Extend \p root_packages with dependency names from shard-loaded package records.
+     *
+     * Only walks packages already fetched for the current shard subset (typically tens to
+     * hundreds of records), not full repodata. Used for cross-channel / cross-subdir shard
+     * closure (#4245).
+     */
+    void expand_shard_root_packages_from_shard_loaded_packages(
+        const std::map<std::string, std::vector<specs::PackageInfo>>& packages_by_url,
         std::vector<std::string>& root_packages
     );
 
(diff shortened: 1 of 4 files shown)


libmamba/include/mamba/api/channel_loader.hpp
```diff
@@ -38,24 +40,29 @@ namespace mamba
...
      */
     auto load_subdir_with_shards(
         Context& ctx,
         solver::libsolv::Database& database,
-        const std::vector<std::string>& root_packages,
+        std::vector<std::string>& root_packages,
         std::vector<SubdirIndexLoader>& subdirs,
         std::size_t subdir_idx,
-        std::set<std::string>& loaded_subdirs_with_shards,
+        std::map<std::string, solver::libsolv::RepoInfo>& loaded_subdirs_with_shards,
```
JohanMabille: This is another API backward incompatible change.

libmamba/src/api/channel_loader.cpp
```diff
@@ -654,13 +675,51 @@ namespace mamba
...
+            // loaded in the first pass (e.g. bioconda), and when nothing new appeared.
+            if (expand_shard_roots_from_loaded_shards
+                && root_packages.size() > roots_after_full_repodata_pass)
+            {
+                LOG_DEBUG << "Shard root packages expanded by "
+                          << (root_packages.size() - roots_after_full_repodata_pass)
+                          << " additional name(s) during shard pass; re-running shard pass once.";
+                for (auto& [subdir_name, repo] : loaded_subdirs_with_shards)
+                {
+                    static_cast<void>(subdir_name);
```
JohanMabille: Useless, prefer anonymous '_' in the structured bindings.

libmamba/include/mamba/api/channel_loader.hpp
```diff
@@ -38,24 +40,29 @@ namespace mamba
...
      * (from \c prepare_solver_context).
+     * @param expand_shard_roots_from_loaded_shards When true, extend \p root_packages from
+     *        dependency names in loaded shard records (all-sharded channel sets only).
      * @return The repo for the requested subdir, or unexpected mamba_error on failure.
      */
     auto load_subdir_with_shards(
         Context& ctx,
         solver::libsolv::Database& database,
-        const std::vector<std::string>& root_packages,
+        std::vector<std::string>& root_packages,
```
JohanMabille: This is an API backward incompatible change, so theoretically we should release a major version. However, we already released minor with API breakages in the past because libmamba is not directly used, downstream projects usually depend on libmambapy. However, in order to stabilize public APIs for future usages, I would go for removing this function from the public API since it is designed for internal usage only.

JohanMabille (changes requested):

libmamba/src/api/channel_loader.cpp
```diff
@@ -35,8 +35,61 @@
...
+        solver::libsolv::Database& database,
+        std::vector<std::string>& root_packages,
+        std::vector<SubdirIndexLoader>& subdirs,
+        std::size_t subdir_idx,
+        std::map<std::string, solver::libsolv::RepoInfo>& loaded_subdirs_with_shards,
+        const std::vector<solver::libsolv::Priorities>& priorities,
+        std::optional<specs::Version> python_minor_version_for_prefilter,
+        bool expand_shard_roots_from_loaded_shards
+    ) -> expected_t<solver::libsolv::RepoInfo>;
+
```
jjerphan: I did not want to have massive movement of code, but just make this declaration private; yet we can get ride of this declaration and reorder implementations if needed.

JohanMabille (approved):

PR #4308, opened 2026-06-08
fix: Handle `nodefaults` sentinel consistently

jjerphan:
# Description

See https://github.com/geojupyter/jupytergis/pull/1483#issuecomment-4645916705.

## Type of Change

- [x] Bugfix
- [ ] Feature / enhancement
- [ ] CI / Documentation
- [ ] Maintenance

## Checklist

### DIFF
--- a/libmamba/src/api/install.cpp
+++ b/libmamba/src/api/install.cpp
@@ -6,6 +6,7 @@
 
 #include <algorithm>
 #include <stdexcept>
+#include <string_view>
 
 #include <fmt/format.h>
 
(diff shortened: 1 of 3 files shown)


libmamba/src/api/install.cpp
```diff
@@ -1170,21 +1171,49 @@ namespace mamba
             }
         }
 
+        namespace
+        {
+            constexpr std::string_view NODEEFAULTS = "nodefaults";
```
JohanMabille: Typo, should probably be named `NODEFAULTS`

libmamba/src/api/install.cpp
```diff
@@ -1170,21 +1171,49 @@ namespace mamba
...
             auto& config_channels = config.at("channels");
-            std::vector<std::string> cli_channels;
 
             if (config_channels.cli_configured())
             {
-                cli_channels = config_channels.cli_value<std::vector<std::string>>();
-                auto it = find(cli_channels.begin(), cli_channels.end(), "nodefaults");
-                if (it != cli_channels.end())
+                auto cli_channels = config_channels.cli_value<std::vector<std::string>>();
+                if (contains_channel(cli_channels, NODEEFAULTS))
```
JohanMabille: I'm still confused why we assign the `cli_channels` to `channels` only if it contains nodefaults.

libmamba/src/api/install.cpp
```diff
@@ -1170,21 +1171,49 @@ namespace mamba
             }
         }
 
+        namespace
+        {
+            constexpr std::string_view NODEEFAULTS = "nodefaults";
+
+            void erase_channel(std::vector<std::string>& channels, std::string_view name)
+            {
+                channels.erase(
```
Klaim: Not sure it's available in all our standard library implementation but if it is using [`std::erase_if`](https://en.cppreference.com/cpp/container/vector/erase2) would be better.

libmamba/src/api/install.cpp
```diff
@@ -1170,21 +1171,49 @@ namespace mamba
...
+            }
+
+            bool contains_channel(const std::vector<std::string>& channels, std::string_view name)
+            {
+                return find_if(
+                           channels.begin(),
+                           channels.end(),
+                           [&](const std::string& channel) { return channel == name; }
+                       )
+                       != channels.end();
```
JohanMabille: `string` and `string_view` can be comapred, so this could be simplified into
```cpp
return std::find(channels.begin(), channels.end(), name) != channels.end();
```

libmamba/src/api/install.cpp
jjerphan: I think that this is to respect conda's behavior which ignores channels from the configuration and only take the channels of the CLI if `nodefaults` is specified as part of them via CLI arguments. I do not find this intuitive at all.

jjerphan: Actually, conda does not behave this way.

To me, the change has been wrongly introduced by https://github.com/mamba-org/mamba/pull/1773 breaking the semantic of both `nodefaults` and `--override-channels`.

JohanMabille (approved):

PR #4312, opened 2026-06-10
fix: Lock virtual packages in libsolv's repo

### LINKED ISSUE #4311: mamba ignoring run constraints on virtual packages
### Troubleshooting docs

### Anaconda default channels

Micromamba

virtual package constraints

### Latest version of Mamba

### Tried in Conda?

### Describe your issue

```
$ CONDA_OVERRIDE_CUDA=12.0 micromamba create -n test cuda-version=13.3
Using Cached Shard Index for conda-forge/linux-64                                                   ✔ Done
Using Cached Shard Index for conda-forge/noarch                                                     ✔ Done
Fetching and Parsing Packages' Shards                                                     ✔ Done (0.1 sec)

Resolving Environment                                                                     ✔ Done (0.0 sec)

Transaction

  Prefix: /home/dching/.local/share/mamba/envs/test

  Updating specs:

   - cuda-version=13.3


  Package         Version  Build       Channel         Size
─────────────────────────────────────────────────────────────
  Install:
─────────────────────────────────────────────────────────────

  + cuda-version     13.3  hcbadf70_3  conda-forge     22kB

  Summary:

  Install: 1 packages

  Total download: 22kB

─────────────────────────────────────────────────────────────
```

mamba is ignoring run constraints on virtual packages. For example, the cuda-version=13 package [has a runtime constraint on __cuda >=13](https://conda-metadata-app.streamlit.app/?q=conda-forge%2Fnoarch%2Fcuda-version-13.3-hcbadf70_3.conda). This means that if the system reports `__cuda=12.0`, `cuda-version>=13` should not be installable. However, this is not the case. As shown above, mamba installs `cuda-version>=13` even though `__cuda=12.0`. An error related to unsolvability to `__cuda` should be raised and the package installation should fail.

```shell
$ CONDA_OVERRIDE_CUDA=12.0 micromamba info

       libmamba version : 2.8.1
     micromamba version : 2.8.1
           curl version : libcurl/8.20.0 OpenSSL/3.5.6 zlib/1.3.2 zstd/1.5.7 libssh2/1.11.1 nghttp2/1.68.1 mit-krb5/1.22.2
     libarchive version : libarchive 3.8.7 zlib/1.3.2 bz2lib/1.0.8 libzstd/1.5.7 libb2/bundled
       envs directories : /home/dching/.local/share/mamba/envs
          package cache : /home/dching/.local/share/mamba/pkgs
                          /home/dching/.mamba/pkgs
            environment : base
           env location : /home/dching/.local/share/mamba
      user config files : /home/dching/.mambarc
 populated config files : .condarc
                          /home/dching/.condarc
       virtual packages : __unix=0=0
                          __linux=6.6.114=0
                          __glibc=2.39=0
                          __archspec=1=x86_64_v3
                          __cuda=12.0=0
               channels : https://conda.anaconda.org/conda-forge/linux-64
       base environment : /home/dching/.local/share/mamba
               platform : linux-64
```

### Logs

```shell

```

### environment.yml

```yaml

```

### ~/.condarc

```yaml
$ cat ~/.condarc
channels:
  - conda-forge
  - nodefaults
solver: classic
channel_priority: strict
```

jjerphan:
# Description

Fix #4311.

## Type of Change

- [x] Bugfix
- [ ] Feature / enhancement
- [ ] CI / Documentation
- [ ] Maintenance

## Checklist

### DIFF
--- a/libmamba/include/mamba/solver/libsolv/database.hpp
+++ b/libmamba/include/mamba/solver/libsolv/database.hpp
@@ -11,6 +11,7 @@
 #include <memory>
 #include <optional>
 #include <string_view>
+#include <vector>
 
 #include "mamba/core/error_handling.hpp"
 #include "mamba/solver/libsolv/parameters.hpp"
@@ -22,6 +23,7 @@
 namespace solv
 {
     class ObjPool;
+    class ObjQueue;
 }
 
 namespace mamba
@@ -118,6 +120,25 @@ namespace mamba::solver::libsolv
         [[nodiscard]] auto installed_repo() const -> std::optional<RepoInfo>;
 
         void set_installed_repo(RepoInfo repo);
+
+        /**
+         * Add a virtual package (name starting with ``__``) to a repository.
+         *
+         * Solver lock jobs are recorded immediately so virtual packages stay in the solution and
+         * run constraints on them remain enforceable.
+         */
+        auto add_virtual_package(const RepoInfo& repo, const specs::PackageInfo& pkg) -> SolvableId;
+
+        /** Finalize repository changes after adding solvables outside @ref add_repo_from_packages.
+         */
+        void internalize_repo(const RepoInfo& repo);
+
+        /** Forget solver jobs from @ref add_virtual_package, e.g. before reloading installed
+         * packages. */
+        void clear_virtual_package_lock_jobs();
+
+        /** Solver jobs recorded by @ref add_virtual_package. */
+        [[nodiscard]] auto virtual_package_lock_jobs() const -> const solv::ObjQueue&;
 
         void set_repo_priority(RepoInfo repo, Priorities priorities);
 
--- a/libmamba/include/mamba/solver/libsolv/database.hpp
+++ b/libmamba/include/mamba/solver/libsolv/database.hpp
@@ -121,23 +121,21 @@ namespace mamba::solver::libsolv
 
         void set_installed_repo(RepoInfo repo);
 
-        /**
-         * Add a virtual package (name starting with ``__``) to a repository.
-         *
-         * Solver lock jobs are recorded immediately so virtual packages stay in the solution and
-         * run constraints on them remain enforceable.
-         */
-        auto add_virtual_package(const RepoInfo& repo, const specs::PackageInfo& pkg) -> SolvableId;
+        template <typename Iter>
+        void add_virtual_packages(const RepoInfo& repo, Iter first_package, Iter last_package);
+
+        template <typename Range>
+        void add_virtual_packages(const RepoInfo& repo, const Range& packages);
 
         /** Finalize repository changes after adding solvables outside @ref add_repo_from_packages.
          */
         void internalize_repo(const RepoInfo& repo);
 
-        /** Forget solver jobs from @ref add_virtual_package, e.g. before reloading installed
+        /** Forget solver jobs from @ref add_virtual_packages, e.g. before reloading installed
          * packages. */
         void clear_virtual_package_lock_jobs();
 
-        /** Solver jobs recorded by @ref add_virtual_package. */
+        /** Solver jobs recorded by @ref add_virtual_packages. */
         [[nodiscard]] auto virtual_package_lock_jobs() const -> const solv::ObjQueue&;
 
         void set_repo_priority(RepoInfo repo, Priorities priorities);
@@ -185,6 +183,7 @@ namespace mamba::solver::libsolv
         auto add_repo_from_packages_impl_pre(std::string_view name) -> RepoInfo;
         void add_repo_from_packages_impl_loop(const RepoInfo& repo, const specs::PackageInfo& pkg);
         void add_repo_from_packages_impl_post(const RepoInfo& repo, PipAsPythonDependency add);
+        void add_virtual_package_impl(const RepoInfo& repo, const specs::PackageInfo& pkg);
 
         enum class PackageId : int;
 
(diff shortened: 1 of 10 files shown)


AntoinePrv (commented):
Can we not lock them when they are added via add_virtual_package (even if it requires tracking them separately up until later)?

This kind of *after-the-fact* re-analyze and fix is IMHO what contributes to code rot (to which our libsolv code is very vulnerable).

libmamba/src/solver/libsolv/database.cpp
```diff
@@ -315,6 +317,39 @@ namespace mamba::solver::libsolv
         pool().set_installed_repo(repo.id());
     }
 
+    void Database::add_virtual_package_impl(const RepoInfo& repo, const specs::PackageInfo& pkg)
+    {
+        if (!util::starts_with(pkg.name, "__"))
+        {
+            throw std::invalid_argument(
```
Klaim: Can we use `mamba_error` in this part of the code?

libmamba/src/solver/libsolv/helpers.cpp
```diff
@@ -1761,11 +1767,20 @@ namespace mamba::solver::libsolv
...
     {
         auto solv_jobs = solv::ObjQueue();
+        if (!virtual_package_lock_jobs.empty())
+        {
+            solv_jobs.insert(
+                solv_jobs.begin(),
+                virtual_package_lock_jobs.begin(),
+                virtual_package_lock_jobs.end()
+            );
+        }
```
jjerphan: Not sure if there's a nicer way.

Klaim: yeah the nicer way is [`.append_range`](https://en.cppreference.com/cpp/container/vector/append_range) or [`.insert_range`](https://en.cppreference.com/cpp/container/vector/insert_range) but requires C++23 😭
I wonder if we didnt already add an equivalent function in utilities though

Klaim: Didnt find an equivalent in utils.
BTW, I have [this "algorithm" header I started in that pr](https://github.com/mamba-org/mamba/pull/4126/changes#diff-3aa5266a7174293d9c09d2f195c42bb43f7f10d79dfeb60046fcfe4b827e7248) , to add algorithm that exist in higher version of C++ but not yet in ours. I needed the contains algorithm.
We could put it in a separate pr and add free-function versions of append_range and insert_range to then easilly replace them later?

jjerphan: I meant, I do not really like to pass those objects by left-value reference, adding another parameter to `request_to_decision_queue`, before inserting them in another `solv::ObjQueue`.

Klaim: Actually that should be just a simple copy? Why isnt `solv_job` just a value (as in copy) parameter of the function?

jjerphan: Pass-by-value would only add a redundant copy on the way in without simplifying the logic, isn't it?

jjerphan: Actually no since we can move it then.

libmamba/src/solver/libsolv/helpers.cpp
JohanMabille, line 51: Maybe mark this parameter as deprecated since it is not used anymore in the function?

Or move the implementation in an aoverload with only 3 parameters, replace the calls in the codebase, have this one forward the call and mark it as deprecated.

This also raises the question about the configurability of the MatchSpecParser if we force it to Libsolv.

jjerphan: Is the configurability of `MatchSpecParser` useful now that mamba's parser is not experimental anymore?

JohanMabille: Well you may want to be able to choose your parser, maybe the name of the configurable can be updated.
My remark was because I did not find where else the parser was used, but it is in the solver, so you can ignore my initial comment.

JohanMabille (approved):

libmamba/src/solver/libsolv/helpers.cpp
jjerphan: I think it was a way for user to fallback on the previous parser when the new one was introduced, but I doubt user would want to know or to care about a particular parser.

> Or move the implementation in an aoverload with only 3 parameters, replace the calls in the codebase, have this one forward the call and mark it as deprecated.

I prefer this approach, but I guess that it better be tackled in the PR resolving #4320, for now I'll mark the parameter as deprecated.

jjerphan: `parser` is actually still used in this function, and there are other parts of the codebase which use `MatchSpecParser::Libsolv`, _e.g._:

https://github.com/mamba-org/mamba/blob/be477514d8361a7ffcd15c8d38dfa465ff881647/libmamba/src/solver/libsolv/helpers.cpp#L983-L988

libmamba/include/mamba/solver/libsolv/database.hpp
```diff
@@ -119,6 +121,23 @@ namespace mamba::solver::libsolv
 
         void set_installed_repo(RepoInfo repo);
 
+        template <typename Iter>
```
Klaim: I would prefer if any new template would be constrained with concepts/requires . Here I suspect `forward_iterator` would work? Also the value type needs to be `PackageInfo` so that's another constraint.

I can see the `add_repo_from_packages` also dont have these but now would be a good time to see if we can improve that.

libmamba/include/mamba/solver/libsolv/database.hpp
```diff
@@ -119,6 +121,23 @@ namespace mamba::solver::libsolv
...
+        void add_virtual_packages(const RepoInfo& repo, Iter first_package, Iter last_package);
+
+        template <typename Range>
```
Klaim: Same remark about templates constrained, here the range could be `forward_range` + value type constraint.

Klaim (approved):
LGTM although I dont have a clear understanding of virtual packages, I focused on the correctness of the structure.
As improvement I would prefer if every time we introduce new templates we make sure all parameters are constrained (and the improve the previously written templates where we can while we're at it).

PR #4313, opened 2026-06-11
Handle pre/post-link/unlink scripts

### LINKED ISSUE #3761: run_script not called for pre-unlink scripts
run_script is called for post-link scripts but despite the comment in the code and the docs describing similar for pre-unlink there is no call present and thus pre-unlink scripts are not executed.

Is this a bug/oversight or by design?

Thanks!

### LINKED ISSUE #4310: Handle linking scripts correctly in `mamba`
Mamba doesn't handle all pre/post-link/unlink scripts thoroughly.

cf. #3761 and https://github.com/conda/ceps/issues/169

Hind-M:
# Description

Fix #4310
Fix #3761 

## Type of Change

- [x] Bugfix
- [ ] Feature / enhancement
- [ ] CI / Documentation
- [ ] Maintenance

## Checklist

### DIFF
--- a/libmamba/src/core/link.cpp
+++ b/libmamba/src/core/link.cpp
@@ -541,6 +541,7 @@ namespace mamba
     bool run_script(
         const TransactionParams& transaction_params,
         const PrefixParams& prefix_params,
+        const fs::u8path& script_prefix,
         const specs::PackageInfo& pkg_info,
         const std::string& action = "post-link",
         const std::string& env_prefix = "",
@@ -550,12 +551,12 @@ namespace mamba
         fs::u8path path;
         if (util::on_win)
         {
-            path = prefix_params.target_prefix / get_bin_directory_short_path()
+            path = script_prefix / get_bin_directory_short_path()
                    / util::concat(".", pkg_info.name, "-", action, ".bat");
         }
         else
         {
-            path = prefix_params.target_prefix / get_bin_directory_short_path()
+            path = script_prefix / get_bin_directory_short_path()
                    / util::concat(".", pkg_info.name, "-", action, ".sh");
         }
 
@@ -566,13 +567,22 @@ namespace mamba
             return true;
         }
 
-        // TODO impl.
-        std::map<std::string, std::string> envmap;  // = env::copy();
-
         if (action == "pre-link")
         {
-            throw std::runtime_error("mamba does not support pre-link scripts");
+            LOG_WARNING << "Special Note: Pre-link scripts are particularly high-risk as they can "
+                        << "modify the package cache, potentially affecting all environments on this system.";
         }
+
+        if (action == "post-unlink")
+        {
+            LOG_WARNING << "post-unlink scripts are deprecated and therefore won't be executed!";
+            return true;
+        }
+
+        LOG_WARNING << "Executing " << action << " script for package '" << pkg_info.name << "'.";
+
+        // TODO impl.
+        std::map<std::string, std::string> envmap;  // = env::copy();
 
         // script_caller = None
         std::vector<std::string> command_args;
@@ -799,6 +809,16 @@ namespace mamba
             return true;
         }
 
+        run_script(
+            m_context->transaction_params(),
+            m_context->prefix_params(),
+            m_context->prefix_params().target_prefix,
+            m_pkg_info,
+            "pre-unlink",
+            "",
+            false
+        );
+
         std::ifstream json_file = open_ifstream(json);
         if (!json_file.good())
         {
@@ -847,6 +867,16 @@ namespace mamba
         }
 
         json_file.close();
+
+        run_script(
+            m_context->transaction_params(),
+            m_context->prefix_params(),
+            m_context->prefix_params().target_prefix,
+            m_pkg_info,
+            "post-unlink",
+            "",
+            true
+        );
 
         fs::remove(json);
 
@@ -1167,6 +1197,16 @@ namespace mamba
         nlohmann::json index_json, out_json;
         LOG_TRACE << "Preparing linking from '" << m_source.string() << "'";
 
+        run_script(
+            m_context->transaction_params(),
+            m_context->prefix_params(),
+            m_source,
+            m_pkg_info,
+            "pre-link",
+            "",
+            false
+        );
+
         LOG_TRACE << "Opening: " << m_source / "info" / "paths.json";
         auto paths_data = read_paths(m_source);
 
@@ -1431,6 +1471,7 @@ namespace mamba
         run_script(
             m_context->transaction_params(),
             m_context->prefix_params(),
+            m_context->prefix_params().target_prefix,
             m_pkg_info,
             "post-link",
             "",
--- a/libmamba/src/core/link.cpp
+++ b/libmamba/src/core/link.cpp
@@ -569,8 +569,9 @@ namespace mamba
 
         if (action == "pre-link")
         {
-            LOG_WARNING << "Special Note: Pre-link scripts are particularly high-risk as they can "
-                        << "modify the package cache, potentially affecting all environments on this system.";
+            LOG_WARNING
+                << "Special Note: Pre-link scripts are particularly high-risk as they can "
+                << "modify the package cache, potentially affecting all environments on this system.";
         }
 
         if (action == "post-unlink")
--- a/libmamba/src/core/link.cpp
+++ b/libmamba/src/core/link.cpp
@@ -567,11 +567,13 @@ namespace mamba
             return true;
         }
 
+        std::map<std::string, std::string> envmap;
         if (action == "pre-link")
         {
             LOG_WARNING
                 << "Special Note: Pre-link scripts are particularly high-risk as they can "
                 << "modify the package cache, potentially affecting all environments on this system.";
+            envmap["SOURCE_DIR"] = script_prefix.string();
         }
 
         if (action == "post-unlink")
@@ -581,9 +583,6 @@ namespace mamba
         }
 
         LOG_WARNING << "Executing " << action << " script for package '" << pkg_info.name << "'.";
-
-        // TODO impl.
-        std::map<std::string, std::string> envmap;  // = env::copy();
 
         // script_caller = None
         std::vector<std::string> command_args;
--- a/libmamba/src/core/transaction.cpp
+++ b/libmamba/src/core/transaction.cpp
@@ -1172,8 +1172,9 @@ namespace mamba
             return true;
         }
 
-        LOG_WARNING << "Security Warning: This transaction includes executing package scripts (pre/post-link/unlink) if present. "
-                    << "These scripts can contain arbitrary code. Please ensure you trust the package sources.";
+        LOG_WARNING
+            << "Security Warning: This transaction includes executing package scripts (pre/post-link/unlink) if present. "
+            << "These scripts can contain arbitrary code. Please ensure you trust the package sources.";
 
         return Console::prompt("Confirm changes", 'y');
     }
--- a/libmamba/src/core/transaction.cpp
+++ b/libmamba/src/core/transaction.cpp
@@ -1172,12 +1172,9 @@ namespace mamba
             return true;
         }
 
-        if (!ctx.output_params.json)
-        {
-            LOG_WARNING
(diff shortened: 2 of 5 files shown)


libmamba/src/core/link.cpp
```diff
@@ -566,14 +567,24 @@ namespace mamba
...
+        if (action == "post-unlink")
+        {
+            LOG_WARNING << "post-unlink scripts are deprecated and therefore won't be executed!";
+            return true;
+        }
+
+        LOG_WARNING << "Executing " << action << " script for package '" << pkg_info.name << "'.";
+
+        // TODO impl.
+        std::map<std::string, std::string> envmap;  // = env::copy();
```
JohanMabille: Should we address this after this PR is merged?
Also, is it possible to declare it just before we fill it?

libmamba/src/core/transaction.cpp
```diff
@@ -1172,6 +1172,10 @@ namespace mamba
             return true;
         }
 
+        LOG_WARNING
+            << "Security Warning: This transaction includes executing package scripts (pre/post-link/unlink) if present. "
+            << "These scripts can contain arbitrary code. Please ensure you trust the package sources.";
+
```
jjerphan, line 1178: I think that:
 - this might be scary for user if we do not provide more context (for instance we could link the final version of [the subsection of CEP34](https://github.com/conda/ceps/pull/173/changes#diff-154a3bbc344a85365a316dfb77251bd08a10d5ffa58f4a0f620efe43dbf4388eR299-R306) once approved.
 - that there must be a way for users to disable this warning

Must or should users still must be able to opt-out from using those links? I think it would be valuable, but I am not sure the CEP specifies it properly (let me reread it).

libmamba/src/core/link.cpp
```diff
@@ -566,14 +567,24 @@ namespace mamba
...
+            LOG_WARNING
+                << "Special Note: Pre-link scripts are particularly high-risk as they can "
+                << "modify the package cache, potentially affecting all environments on this system.";
         }
 
+        if (action == "post-unlink")
+        {
+            LOG_WARNING << "post-unlink scripts are deprecated and therefore won't be executed!";
+            return true;
+        }
```
jjerphan, line 583: I wonder whether warnings must be reported at the end of the transaction also (might people miss them in the long linking and unlinking logs?).

jjerphan (commented):
Thank you.

I think that we might need to think of ways for users to opt-out of warnings (which will be annoying for them) and from this new behavior.

libmamba/src/core/link.cpp
jjerphan, line 649: Based on [this mention in CEP 34](https://github.com/conda/ceps/pull/173/changes#diff-154a3bbc344a85365a316dfb77251bd08a10d5ffa58f4a0f620efe43dbf4388eR276)

> The following environment variables MUST be exposed to the script runtime:
> - ...
> - `PREFIX`, `SOURCE_DIR`: Path to the target environment.
> - ...

should `SOURCE_DIR` be set also?

libmamba/src/core/link.cpp
jjerphan: Also, to meet [this requirement from CEP 34](https://github.com/conda/ceps/pull/173/changes#diff-154a3bbc344a85365a316dfb77251bd08a10d5ffa58f4a0f620efe43dbf4388eR281)

> The `PATH` environment variable MUST include the directory where the script is located as its first element.


should `PATH` be adapted?

libmamba/src/core/link.cpp
jjerphan, line 583: Should we have a deprecation period (i.e. mention that their support this will be removed in a future version while still executing them)?

Hind-M: There is a kind of global warning added in the PR at the transaction level, when the user is asked to confirm the transaction execution. I think that's enough and we shouldn't add more than that, otherwise it would be too noisy.

Hind-M: I was wondering about that too but I don't think the `post-unlink` should be executed since it was deprecated a long time ago. And we haven't had any issues reported regarding that so far.

libmamba/src/core/link.cpp
Hind-M: Yes, I added it but only in the case of `pre-link` (like `conda` does).

libmamba/src/core/link.cpp
Hind-M: > should PATH be adapted?

Yes it's already handled here: 
`envmap["PATH"] = util::concat(path.parent_path().string(), util::pathsep(), PATH);`

Hind-M: > Should we address this after this PR is merged?

I did a bit of digging in the code history and I think we can remove the `TODO` comment as I don't think this is referring to anything obvious missing.
I believe the `env::copy` comment was a leftover as well since `reproc` was not used from day one and it seems to have the same behavior as:
```
options.env.behavior = reproc::env::extend;
options.env.extra = envmap;
```
Therefore, I removed both comments.

> Also, is it possible to declare it just before we fill it?

Done, although setting the additional `SOURCE_DIR` in the `envmap` forces the declaration to be earlier.

Hind-M:
> I think that we might need to think of ways for users to opt-out of warnings (which will be annoying for them) and from this new behavior.

Yes I was thinking about that too as they do in `Pixi`, I think we can add a way to opt-out the scripts execution and in that case not print the warnings. I can do that in a follow-up PR.

libmamba/src/core/transaction.cpp
Hind-M: See the [comment](https://github.com/mamba-org/mamba/pull/4313#issuecomment-4720376655).
I think anything else is too much.
We might consider adding a section in the documentation for more details.

libmamba/src/core/link.cpp
jjerphan: It might break a few packages' distributions (we probably should search in the main channels for their presence). I think that the risk is low _a priori_.

Hind-M: Yeah last time I checked there weren't any in the main channels (`conda-forge`, `bioconda` and `defaults`) hence I guess the deprecation.

libmamba/src/core/transaction.cpp
```diff
@@ -1172,6 +1172,13 @@ namespace mamba
             return true;
         }
 
+        if (!ctx.output_params.json)
```
Hind-M: @Klaim Can you please confirm this is the right way to go after #4202?

Klaim: If you mean is it the right way to emit a warning with `--json` : yes, it will appear like this in the output
```json
{
   // ...
   "log_history" : [ 
       // ... 
       {  // LogRecord
           "message": "Security Warning: This transaction includes executing package scripts (pre/post-link/unlink) if present. These scripts can contain arbitrary code. Please ensure you trust the package sources.",
           "level": // whatever warning is
            "location": // path to this source file and line
       },
       // ...
     ],
   // ...
}
```

However it is not clear to me why that warning would appear only in `--json` case?

While the confirmation request will appear if `--json` is used, that warning will not appear in the terminal. If you intend to make it appear in the terminal (which is contradictory but the confirmation is too) then it needs to be done either through the `Console` or standard output.
Either way that means that the output is not json.

Klaim: Maybe a better way would be to not do the confirmation at all if `--json` is used.
I think  that would make more sense for all operations (`create` for example) where confirmation can happen. It's contradictory to request json output and still have some non-json in the output.

Klaim: Woops didnt see the negation

Hind-M: Ok, I'm using `find_message_in_json_logs` now in the failing test as discussed outside this PR.

JohanMabille (approved):

libmamba/tests/src/core/test_link_scripts.cpp
```diff
@@ -0,0 +1,172 @@
...
+                                      / (get_self_exe_path().stem().string() + ".bat");
+                auto out = open_ofstream(conda_bat_path);
+                out << "@exit /b 0\n";
+            }
+            else
+            {
+                fs::create_directories(prefix / "bin");
+                auto conda_path = prefix / "bin" / "conda";
+                auto out = open_ofstream(conda_path);
+                out << "#!/bin/sh\n";
```
Klaim, line 60: In case you want to make the script more readable: you could use Raw String Litterals `R("script here")` and write the whole script in there in one block.

libmamba/tests/src/core/test_link_scripts.cpp
```diff
@@ -0,0 +1,172 @@
...
+
+                return TransactionContext(
+                    tx_params,
+                    { "3.14.4", "3.14.4" },
+                    "lib/python3.14/site-packages",
+                    {}
+                );
+            };
+
+            std::string script_ext = util::on_win ? ".bat" : ".sh";
```
Klaim, line 95:
```suggestion
            const std::string script_ext = util::on_win ? ".bat" : ".sh";
```

Klaim: And same for the following lines

micromamba/tests/test_create.py
```diff
@@ -2426,10 +2426,13 @@ def test_create_from_mirror_with_prefix(tmp_home, tmp_root_prefix, tmp_path):
...
 
-    # Verify no warnings in output
-    combined_output = result.stdout + result.stderr
-    assert "warning" not in combined_output.lower(), (
-        f"Unexpected warning in output:\nstdout:\n{result.stdout}\nstderr:\n{result.stderr}"
+    # Verify warning in output
+    assert (
+        helpers.find_message_in_json_logs(
+            res,
+            "Security Warning: This transaction includes executing package scripts (pre/post-link/unlink) if present.",
```
Klaim, line 2433: Note that `find_message_in_json_logs` takes a string that is searched for inside the messages, so if you prefer to avoid having a specific message but at least a specific section (to allow tweaking the message later) you could pass just the important part of the message to find. As you prefer.

libmamba/src/core/link.cpp
```diff
@@ -566,14 +567,23 @@ namespace mamba
...
-            throw std::runtime_error("mamba does not support pre-link scripts");
+            LOG_WARNING
+                << "Special Note: Pre-link scripts are particularly high-risk as they can "
+                << "modify the package cache, potentially affecting all environments on this system.";
+            envmap["SOURCE_DIR"] = script_prefix.string();
         }
 
+        if (action == "post-unlink")
+        {
+            LOG_WARNING << "post-unlink scripts are deprecated and therefore won't be executed!";
```
Klaim, line 581: Shouldnt this be an error, through thrown exception? (I might not have enough context)

libmamba/src/core/link.cpp
```diff
@@ -566,14 +567,23 @@ namespace mamba
             return true;
         }
 
-        // TODO impl.
-        std::map<std::string, std::string> envmap;  // = env::copy();
-
+        std::map<std::string, std::string> envmap;
```
Klaim, line 570: I think `unordered_map` here works? Except if you need the ordering, it's generally a better default. (I can see it was `map` before but while we're at it...)

Klaim (approved):
LGTM, my comments are mostly minor improvements, notes and questions which answers probably wont change the validity.

micromamba/tests/test_create.py
Hind-M: Yes I only included the first sentence of the message :)

libmamba/src/core/link.cpp
Hind-M: If there is a `post-unlink `script, the package should still be removed anyway and just warn the user that the script wasn't executed because of deprecation.
This shouldn't be happening as this script type doesn't exist anymore on main channels (`conda-forge`, `bioconda`, `defaults`), but we never know.

libmamba/src/core/link.cpp
Hind-M: Yes, I will change this and address the other remaining comments in the [follow-up PR](https://github.com/mamba-org/mamba/pull/4325).

Issue #4322, opened 2026-06-17
"Invalid package cache" ruins installed env


mathause:
### Troubleshooting docs

- [x] My problem is not solved in the Troubleshooting docs

### Anaconda default channels

- [x] I do NOT use the Anaconda default channels (pkgs/* etc.)

### How did you install Mamba?

Mambaforge or latest Miniforge

### Search tried in issue tracker

yes (although "cache" 

### Latest version of Mamba

- [x] My problem is not solved with the latest version

### Tried in Conda?

I do not have this problem with Conda, just with Mamba

### Describe your issue

A `critical libmamba Package cache error` aborted a `mamba update --all` command. Mamba successfully unlinked all necessary packages but then failed while linking. Thus, many packages were not re-installed, leading to a broken environment.

0. This happened for two different system where I updated base at +- the same time.
1. Unfortunately cannot provide a reproducer, as I cannot force a cache error.
2. mamba _knew_ before installing that there is a cache problem (see log) but continued anyways. Should it abort early?
3. mamba should probably not abort but just skip the failing package?
4. Luckily I still had the output - so I could diff which packages were missing.

### mamba info / micromamba info

```shell
libmamba version : 2.8.1
          mamba version : 2.8.1
           curl version : libcurl/8.20.0 OpenSSL/3.6.3 zlib/1.3.2 zstd/1.5.7 libssh2/1.11.1 nghttp2/1.68.1 mit-krb5/1.22.2
     libarchive version : libarchive 3.8.7 zlib/1.3.2 liblzma/5.8.3 bz2lib/1.0.8 liblz4/1.10.0 libzstd/1.5.7 liblzo2/2.10 openssl/3.5.6 libb2/bundled
       envs directories : /home/mathause/.conda/envs
                          /home/mathause/.mambabase/envs
          package cache : /home/mathause/.conda/pkgs
            environment : base (active)
           env location : /home/mathause/.mambabase
      user config files : /home/mathause/.mambarc
 populated config files : /home/mathause/.condarc
                          /home/mathause/.mambabase/.condarc
       virtual packages : __unix=0=0
                          __linux=6.8.0=0
                          __glibc=2.35=0
                          __archspec=1=x86_64_v3
               channels : https://conda.anaconda.org/conda-forge/linux-64
                          https://conda.anaconda.org/conda-forge/noarch
       base environment : /home/mathause/.mambabase
               platform : linux-64
```

### Logs

```shell
mamba update --all
[...]

PR #4328, opened 2026-06-22
fix: Clear stale negative cache entries

jjerphan:
# Description

Tentative fix for #4322.

Fix failures observed in Code Coverage workflows.

## Type of Change

- [x] Bugfix
- [ ] Feature / enhancement
- [ ] CI / Documentation
- [ ] Maintenance

## Checklist


libmamba/include/mamba/core/package_fetcher.hpp
```diff
@@ -116,6 +116,7 @@ namespace mamba
         void update_monitor(progress_callback_t* cb, PackageExtractEvent event) const;
 
         specs::PackageInfo m_package_info;
+        MultiPackageCache* m_caches = nullptr;
```
Hind-M, line 119: Hmm is this a leftover? It doesn't seem to be used and I'm not sure I understand why it would be needed.

libmamba/tests/src/core/test_transaction.cpp
```diff
@@ -0,0 +1,401 @@
...
+                m_ctx->prefix_data_interoperability = m_saved_prefix_data_interoperability;
+                m_ctx->threads_params.extract_threads = m_saved_extract_threads;
+            }
+
+            Context* m_ctx = nullptr;
+            PrefixParams m_saved_prefix_params;
+            std::vector<fs::u8path> m_saved_pkgs_dirs;
+            bool m_saved_prefix_data_interoperability = false;
+            int m_saved_extract_threads = 0;
+        };
```
Hind-M: This seems overrated and doesn't seem to be used either (unused declared `context_guard`).

jjerphan: The constructor and the destructor of `ScopedContextOverride` are handling the logic of the interaction with the context guaranteeing that the state will be the original one when the scope change.

So `context_guard` is declared (also defined) unused by any other instructions, yet not useless.

We have this pattern at several places in tests.

libmamba/include/mamba/core/package_fetcher.hpp
jjerphan: The `MultiPackageCache` is actually used via the `PackageFetchers` (see their instantiations bellow).

libmamba/include/mamba/core/package_fetcher.hpp
```diff
@@ -116,6 +117,7 @@ namespace mamba
         void update_monitor(progress_callback_t* cb, PackageExtractEvent event) const;
 
         specs::PackageInfo m_package_info;
+        std::shared_ptr<MultiPackageCache> m_caches;
```
JohanMabille: No need for a shared pointer here, this should be a raw pointer (given the way it is initialized in PackageFetcher constructor).

JohanMabille (changes requested):

JohanMabille (approved):

libmamba/tests/src/core/test_transaction.cpp
Hind-M: > We have this pattern at several places in tests.

In that case, we may want to consider having a global common struct for all tests instead of rewriting it in each test (even if not all parameters are involved each time).
To be evaluated...

jjerphan: I propose https://github.com/mamba-org/mamba/pull/4334.

libmamba/tests/src/core/test_transaction.cpp
```diff
@@ -0,0 +1,401 @@
...
+#include "core/link.hpp"
+#include "core/transaction_context.hpp"
+
+#include "mambatests.hpp"
+
+namespace mamba
+{
+    namespace
+    {
+        struct ScopedContextOverride
```
Hind-M: Can we use the available `ScopedContextChange` instead?

Hind-M (approved):

PR #4342, opened 2026-06-29
fix: Long paths support activation hint on Windows

### LINKED ISSUE #4275: Change of pkgs cache directory made it go over the windows limit
### Troubleshooting docs

### Anaconda default channels

windows limit

### Latest version of Mamba

### Tried in Conda?

I didn't try

### Describe your issue

[change of pkgs cache directory made it go over the windows limit](https://github.com/conda-forge/miniforge/issues/873#issuecomment-4407622491)

```shell

```

### Logs

```shell

```

### environment.yml

```yaml

```

### ~/.condarc

```yaml

```

jjerphan:
# Description

Fix https://github.com/mamba-org/mamba/issues/4275.

## Type of Change

- [x] Bugfix
- [ ] Feature / enhancement
- [ ] CI / Documentation
- [ ] Maintenance

## Checklist

### DIFF
--- a/libmamba/include/mamba/core/util_os.hpp
+++ b/libmamba/include/mamba/core/util_os.hpp
@@ -38,7 +38,30 @@ namespace mamba
 #endif
 
     void run_as_admin(const std::string& args);
+
+#ifdef _WIN32
     bool enable_long_paths_support(bool force, Palette palette = Palette::no_color());
+
+    /**
+     * Return whether the current process can use paths longer than MAX_PATH.
+     *
+     * Requires Windows 10 version 1607 or later, ``LongPathsEnabled`` in the registry, and a
+     * ``longPathAware`` application manifest. See Microsoft's documentation:
+     * https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=registry#enable-long-paths-in-windows-10-version-1607-and-later
+     */
+    [[nodiscard]] auto are_long_paths_enabled() -> bool;
+
+    /**
+     * Human-readable hint when long paths are disabled; empty string otherwise.
+     */
+    [[nodiscard]] auto format_long_paths_support_diagnostic() -> std::string;
+
+    /**
+     * Log a long-path diagnostic when @p error is a path-length related filesystem error and long
+     * paths are disabled.
+     */
+    void log_long_paths_support_hint_if_relevant(const std::exception& error);
+#endif
 
     void init_console();
     void reset_console();
--- a/libmamba/include/mamba/core/util_os.hpp
+++ b/libmamba/include/mamba/core/util_os.hpp
@@ -46,15 +46,16 @@ namespace mamba
      * Return whether the current process can use paths longer than MAX_PATH.
      *
      * Requires Windows 10 version 1607 or later, ``LongPathsEnabled`` in the registry, and a
-     * ``longPathAware`` application manifest. See Microsoft's documentation:
+     * ``longPathAware`` application manifest (micromamba ships ``longpath.manifest`` for this).
+     * See Microsoft's documentation:
      * https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=registry#enable-long-paths-in-windows-10-version-1607-and-later
      */
     [[nodiscard]] auto are_long_paths_enabled() -> bool;
 
     /**
-     * Human-readable hint when long paths are disabled; empty string otherwise.
+     * Human-readable diagnostics for each unmet long-path requirement; empty string when enabled.
      */
-    [[nodiscard]] auto format_long_paths_support_diagnostic() -> std::string;
+    [[nodiscard]] auto long_paths_support_diagnostic() -> std::string;
 
     /**
      * Log a long-path diagnostic when @p error is a path-length related filesystem error and long
--- a/libmamba/src/core/transaction.cpp
+++ b/libmamba/src/core/transaction.cpp
@@ -468,23 +468,13 @@ namespace mamba
                 LOG_ERROR << "Unexpected error while " << phase << " package '" << pkg.name
                           << "': " << e.what();
 #ifdef _WIN32
-                if (phase == "linking" || phase == "unlinking")
+                if ((phase == "linking" || phase == "unlinking") && !are_long_paths_enabled())
                 {
-                    if (!are_long_paths_enabled())
-                    {
-                        LOG_WARNING << "Windows long path support is disabled, which can cause "
-                                    << phase << " failures with deep cache or prefix paths. "
-                                    << format_long_paths_support_diagnostic();
-                    }
-                    else
-                    {
-                        log_long_paths_support_hint_if_relevant(e);
-                    }
+                    LOG_WARNING << "Windows long path support is disabled, which can cause "
+                                << phase << " failures with deep cache or prefix paths. "
+                                << long_paths_support_diagnostic();
                 }
-                else
-                {
-                    log_long_paths_support_hint_if_relevant(e);
-                }
+                log_long_paths_support_hint_if_relevant(e);
 #endif
                 rethrow_transaction_cancelled_after_rollback(rollback, ctx, pkg, phase, e);
             }
--- a/libmamba/src/core/transaction.cpp
+++ b/libmamba/src/core/transaction.cpp
@@ -30,6 +30,7 @@
 #include "mamba/core/thread_utils.hpp"
 #include "mamba/core/transaction.hpp"
 #include "mamba/core/util.hpp"
+#include "mamba/core/util_os.hpp"
 #include "mamba/core/util_scope.hpp"
 #include "mamba/solver/libsolv/database.hpp"
 #include "mamba/specs/match_spec.hpp"
@@ -466,6 +467,25 @@ namespace mamba
             {
                 LOG_ERROR << "Unexpected error while " << phase << " package '" << pkg.name
                           << "': " << e.what();
+#ifdef _WIN32
+                if (phase == "linking" || phase == "unlinking")
+                {
+                    if (!are_long_paths_enabled())
+                    {
+                        LOG_WARNING << "Windows long path support is disabled, which can cause "
+                                    << phase << " failures with deep cache or prefix paths. "
+                                    << format_long_paths_support_diagnostic();
+                    }
+                    else
+                    {
+                        log_long_paths_support_hint_if_relevant(e);
+                    }
+                }
+                else
+                {
+                    log_long_paths_support_hint_if_relevant(e);
+                }
+#endif
                 rethrow_transaction_cancelled_after_rollback(rollback, ctx, pkg, phase, e);
             }
             catch (...)
--- a/libmamba/src/core/util_os.cpp
+++ b/libmamba/src/core/util_os.cpp
@@ -21,6 +21,7 @@
 #include <intrin.h>
 #include <io.h>
 #include <shlobj_core.h>
+#include <versionhelpers.h>
 #include <windows.h>
 // Incomplete header included last
 #include <tlhelp32.h>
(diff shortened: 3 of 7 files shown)


libmamba/include/mamba/core/util_os.hpp
```diff
@@ -38,8 +38,31 @@ namespace mamba
...
     void run_as_admin(const std::string& args);
+
+#ifdef _WIN32
     bool enable_long_paths_support(bool force, Palette palette = Palette::no_color());
 
+    /**
+     * Return whether the current process can use paths longer than MAX_PATH.
+     *
+     * Requires Windows 10 version 1607 or later, ``LongPathsEnabled`` in the registry, and a
+     * ``longPathAware`` application manifest. See Microsoft's documentation:
```
Klaim: note that the manifest is 1 way to handle that, the one we chose. That sentence should reflect that.

libmamba/include/mamba/core/util_os.hpp
```diff
@@ -38,8 +38,31 @@ namespace mamba
...
+     * https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=registry#enable-long-paths-in-windows-10-version-1607-and-later
+     */
+    [[nodiscard]] auto are_long_paths_enabled() -> bool;
+
+    /**
+     * Human-readable hint when long paths are disabled; empty string otherwise.
+     */
+    [[nodiscard]] auto format_long_paths_support_diagnostic() -> std::string;
```
Klaim: why "format"?

libmamba/src/core/transaction.cpp
```diff
@@ -467,6 +468,25 @@ namespace mamba
...
+#ifdef _WIN32
+                if (phase == "linking" || phase == "unlinking")
+                {
+                    if (!are_long_paths_enabled())
+                    {
+                        LOG_WARNING << "Windows long path support is disabled, which can cause "
+                                    << phase << " failures with deep cache or prefix paths. "
+                                    << format_long_paths_support_diagnostic();
+                    }
+                    else
```
Klaim: I dont think both `else`  are necessary? We can have that warning and the log specific to the error.
Something like
```c++
if ((phase == "linking" or phase == "unlinking") and not are_long_paths_enabled())
      LOG_WARNING << "Windows long path support is disabled, which can cause "
                                << phase << " failures with deep cache or prefix paths. "
                                << format_long_paths_support_diagnostic();
}
log_long_paths_support_hint_if_relevant(e);
```

libmamba/src/core/util_os.cpp
```diff
@@ -149,6 +150,178 @@ namespace mamba
...
 
 #ifdef _WIN32
+    namespace
+    {
+        // https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation
+        static constexpr std::string_view k_windows_long_paths_doc_url = "https://learn.microsoft.com/en-us/windows/win32/fileio/"
+                                                                         "maximum-file-path-limitation?tabs=registry"
+                                                                         "#enable-long-paths-in-windows-10-version-1607-and-later";
+
+        static constexpr wchar_t k_file_system_registry_path[] = L"SYSTEM\\CurrentControlSet\\Control\\FileSystem";
```
Klaim: Why is this an array?

libmamba/src/core/util_os.cpp
```diff
@@ -149,6 +150,178 @@ namespace mamba
     }
 
 #ifdef _WIN32
+    namespace
+    {
+        // https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation
+        static constexpr std::string_view k_windows_long_paths_doc_url = "https://learn.microsoft.com/en-us/windows/win32/fileio/"
```
Klaim: I dont believe we have a "k_" prefix policy for constants?

libmamba/src/core/util_os.cpp
```diff
@@ -149,6 +150,178 @@ namespace mamba
...
+        {
+            bool enabled = false;
+            bool os_supports = false;
+            bool registry_enabled = false;
+            bool process_long_path_aware = false;
+        };
+
+        auto windows_supports_long_paths_registry() -> bool
+        {
+            const auto win_ver = util::windows_version();
```
Klaim: `util::windows_version()` seems designed to return a printable name. I would prefer if we use something that doesnt force us to split strings to get a version, like windows's [`GetVersion()`](https://learn.microsoft.com/en-us/windows/win32/api/sysinfoapi/nf-sysinfoapi-getversion)

libmamba/src/core/util_os.cpp
```diff
@@ -149,6 +150,178 @@ namespace mamba
...
+        {
+            const auto kernel32 = ::GetModuleHandleW(L"kernel32.dll");
+            if (kernel32 == nullptr)
+            {
+                return false;
+            }
+
+            using AreLongPathsEnabledFn = BOOL(WINAPI*)();
+            const auto are_long_paths_enabled = reinterpret_cast<AreLongPathsEnabledFn>(
+                ::GetProcAddress(kernel32, "AreLongPathsEnabled")
```
Klaim: This way to get a function seems very fishy. Where does that comes from?

libmamba/src/core/util_os.cpp
```diff
@@ -149,6 +150,178 @@ namespace mamba
...
+            return are_long_paths_enabled() != FALSE;
+        }
+
+        auto query_long_paths_support() -> LongPathsSupportInfo
+        {
+            LongPathsSupportInfo info;
+            info.os_supports = windows_supports_long_paths_registry();
+            if (!info.os_supports)
+            {
+                return info;
```
Klaim: That returns incorrect info? We didnt check the other properties yet.

libmamba/src/core/util_os.cpp
```diff
@@ -149,6 +150,178 @@ namespace mamba
...
+            }
+
+            try
+            {
+                info.registry_enabled = read_long_paths_registry_enabled();
+            }
+            catch (const winreg::RegException&)
+            {
+                return info;
```
Klaim: If there is a registration error at this point, I don't think it's a good idea to absorb it and return incorrect information?
I think I would prefer if there was no try/catch.

libmamba/src/core/util_os.cpp
```diff
@@ -149,6 +150,178 @@ namespace mamba
...
+        static constexpr std::string_view k_windows_long_paths_doc_url = "https://learn.microsoft.com/en-us/windows/win32/fileio/"
+                                                                         "maximum-file-path-limitation?tabs=registry"
+                                                                         "#enable-long-paths-in-windows-10-version-1607-and-later";
+
+        static constexpr wchar_t k_file_system_registry_path[] = L"SYSTEM\\CurrentControlSet\\Control\\FileSystem";
+        static constexpr std::string_view k_file_system_registry_subkey = "SYSTEM\\CurrentControlSet\\Control\\FileSystem";
+
+        struct LongPathsSupportInfo
+        {
+            bool enabled = false;
```
Klaim: `enabled` (or more like `is_enabled`?) should be a function:
```c++
bool enabled() { return os_supports and registry_enabled and process_long_path_aware; }
```
That way it is never de-synchronized when one of the members changes.

libmamba/src/core/util_os.cpp
```diff
@@ -149,6 +150,178 @@ namespace mamba
...
+        auto make_long_paths_support_diagnostic(const LongPathsSupportInfo& info) -> std::string
+        {
+            if (info.enabled)
+            {
+                return {};
+            }
+
+            if (!info.os_supports)
+            {
+                return fmt::format(
```
Klaim: I would prefer if the result was the cummulated warnings instead of only one at a time. We need to provide feedback for every requirements.

libmamba/src/core/util_os.cpp
```diff
@@ -149,6 +150,178 @@ namespace mamba
...
+                k_windows_long_paths_doc_url
+            );
+        }
+
+        auto is_path_length_related_error(const std::error_code& ec) -> bool
+        {
+            // ERROR_FILENAME_EXCEED_RANGE; see
+            // https://learn.microsoft.com/en-us/windows/win32/debug/system-error-codes--0-499-
+#ifndef ERROR_FILENAME_EXCEED_RANGE
+#define ERROR_FILENAME_EXCEED_RANGE 206
```
Klaim, line 357:
```suggestion
#    define ERROR_FILENAME_EXCEED_RANGE 206
```
or are we constrained by the current formatting policy?

libmamba/src/core/util_os.cpp
```diff
@@ -149,6 +150,178 @@ namespace mamba
...
+#endif
+            return ec.category() == std::system_category()
+                   && ec.value() == static_cast<int>(ERROR_FILENAME_EXCEED_RANGE);
```
Klaim: the `static_cast` seems unnecessary, arent's these both `int` by design?

libmamba/src/core/util_os.cpp
```diff
@@ -188,6 +361,25 @@ namespace mamba
     bool enable_long_paths_support(bool force, Palette palette)
     {
         // Needs to be set system-wide & can only be run as admin ...
+        const auto status = query_long_paths_support();
+        if (status.enabled)
+        {
+            const auto win_ver = util::windows_version();
+            auto out = Console::stream();
```
Klaim: Why console output and not a log? The decision is not clear to me here (maybe add a comment to clarify?)

libmamba/tests/src/core/test_package_cache.cpp
```diff
@@ -153,5 +153,27 @@ namespace mamba
...
+            constexpr auto segment = "very_long_directory_name_segment";
+
+            fs::u8path long_pkgs_dir = temp_dir.path();
+            while ((long_pkgs_dir / segment / "pkgs").generic_string().size()
+                   <= min_pkgs_dir_path_length)
+            {
+                long_pkgs_dir /= segment;
+            }
+            long_pkgs_dir /= "pkgs";
+            fs::create_directories(long_pkgs_dir);
```
Klaim, line 171: Note that if long-path is not enabled at this point (both OS-level and for this process), this call will fail already.
Not saying we need to change anything but it's a bit weird to think about XD

libmamba/include/mamba/core/util_os.hpp
jjerphan: According to the document, the manifest is one of the two conditions, the registry must also be set by the end-users.

libmamba/include/mamba/core/util_os.hpp
jjerphan: It is mostly formatting strings to get a diagnostic. We can definitely rename it.

libmamba/src/core/transaction.cpp
jjerphan: It mostly was to avoid repetitions, but we can remove them.

Klaim: I think about these warnings as complementary, personally.

libmamba/include/mamba/core/util_os.hpp
Klaim: Yes I was talking about the alternative choice to use the string stuffs (using `"\\?\"`), but it's irrelevant.

libmamba/src/core/util_os.cpp
jjerphan: I think it was `auto` before which probably meant that `long` could be used.

libmamba/src/core/util_os.cpp
```diff
@@ -149,6 +152,249 @@ namespace mamba
...
+            if (written > 0 && large_buffer[written - 1] == L'\0')
+            {
+                large_buffer.resize(written - 1);
+            }
+            return large_buffer == L"true";
+        }
+
+        auto query_long_paths_support() -> LongPathsSupportInfo
+        {
+            LongPathsSupportInfo info;
```
Klaim: Very minor improvement: initialize the struct directly with the data.

libmamba/src/core/util_os.cpp
```diff
@@ -149,6 +152,249 @@ namespace mamba
...
+
+            if (messages.empty())
+            {
+                return fmt::format(
+                    "Windows long path support appears to be disabled. See {}",
+                    windows_long_paths_doc_url
+                );
+            }
+
+            return util::join(" ", messages);
```
Klaim: Do we really want to put them in 1 text line?
Would using line-jump as a separator be a problem? Or maybe returning the vector so that the user code decides?

libmamba/src/core/util_os.cpp
Klaim: Looks like it got back to the previous way, I'll assume that current formatting policy is doing this (not to my taste but it's not a big deal).

Klaim (approved):
LGTM, my comments are just minor nitpicks and can be ignored.
Note that I didn't try this PR locally (on Windows) yet, but I don't think I can trigger these errors without changing my system. We'll have to make sure to ask feedback from users.

libmamba/src/core/util_os.cpp
```diff
@@ -149,6 +152,249 @@ namespace mamba
...
+    namespace
+    {
+        // https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation
+        static constexpr std::string_view windows_long_paths_doc_url = "https://learn.microsoft.com/en-us/windows/win32/fileio/"
+                                                                       "maximum-file-path-limitation?tabs=registry"
+                                                                       "#enable-long-paths-in-windows-10-version-1607-and-later";
+
+        static constexpr auto file_system_registry_path = L"SYSTEM\\CurrentControlSet\\Control\\FileSystem";
+        static constexpr std::string_view file_system_registry_subkey = "SYSTEM\\CurrentControlSet\\Control\\FileSystem";
+        static constexpr auto long_path_windows_settings_namespace = L"http://schemas.microsoft.com/SMI/2016/WindowsSettings";
```
JohanMabille: Nitpickin: since these variables are all in an anonymous namespace, I don't think we need them to be `static`

JohanMabille (approved):

PR #4359, opened 2026-07-24
Refactor and clean up

Hind-M:
# Description

## Type of Change

- [ ] Bugfix
- [x] Feature / enhancement
- [ ] CI / Documentation
- [ ] Maintenance

## Checklist

### DIFF
--- a/libmamba/include/mamba/core/db_utils.hpp
+++ b/libmamba/include/mamba/core/db_utils.hpp
@@ -1,0 +1,55 @@
+// Copyright (c) 2026, QuantStack and Mamba Contributors
+//
+// Distributed under the terms of the BSD 3-Clause License.
+//
+// The full license is in the file LICENSE, distributed with this software.
+
+#ifndef MAMBA_CORE_DB_UTILS_HPP
+#define MAMBA_CORE_DB_UTILS_HPP
+
+#include <optional>
+#include <span>
+#include <string_view>
+#include <tuple>
+
+#include "mamba/solver/libsolv/database.hpp"
+#include "mamba/specs/match_spec.hpp"
+#include "mamba/specs/package_info.hpp"
+#include "mamba/specs/version.hpp"
+
+namespace mamba
+{
+    struct PkgInfoCmp
+    {
+        auto operator()(const specs::PackageInfo* lhs, const specs::PackageInfo* rhs) const -> bool
+        {
+            auto attrs = [](const auto& pkg)
+            {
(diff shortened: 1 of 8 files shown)


libmamba/include/mamba/core/db_utils.hpp
```diff
@@ -0,0 +1,55 @@
...
+#include "mamba/solver/libsolv/database.hpp"
+#include "mamba/specs/match_spec.hpp"
+#include "mamba/specs/package_info.hpp"
+#include "mamba/specs/version.hpp"
+
+namespace mamba
+{
+    struct PkgInfoCmp
+    {
+        auto operator()(const specs::PackageInfo* lhs, const specs::PackageInfo* rhs) const -> bool
```
JohanMabille, line 24: Arguments should be passed by reference instead of pointer.

libmamba/include/mamba/core/db_utils.hpp
```diff
@@ -0,0 +1,55 @@
...
+#include <tuple>
+
+#include "mamba/solver/libsolv/database.hpp"
+#include "mamba/specs/match_spec.hpp"
+#include "mamba/specs/package_info.hpp"
+#include "mamba/specs/version.hpp"
+
+namespace mamba
+{
+    struct PkgInfoCmp
```
JohanMabille, line 22: This structure seems to be used in `db_utils.cpp` only, would it make sense to define it there instead of here?

JohanMabille (changes requested):

libmamba/include/mamba/core/db_utils.hpp
Hind-M: It's also used in [query.cpp](https://github.com/mamba-org/mamba/blob/main/libmamba/src/core/query.cpp#L154) (not visible in the PR files diff).

libmamba/include/mamba/core/db_utils.hpp
Hind-M: Because `PkgInfoCmp`is also used in [query.cpp](https://github.com/mamba-org/mamba/blob/main/libmamba/src/core/query.cpp#L154), as `using VisitedMap = std::map<specs::PackageInfo*, node_id, PkgInfoCmp>;`.
Pointers are needed as references are not allowed as map keys because they are not objects.
We can use `reference_wrapper` but that would be more verbose/complex and wouldn't add any benefits.

JohanMabille (approved):

PR #4357, opened 2026-07-24
ci: Add `win-arm64` static builds workflows

jjerphan:
# Description

## Type of Change

- [ ] Bugfix
- [ ] Feature / enhancement
- [x] CI / Documentation
- [ ] Maintenance

## Checklist

- [x] My code follows the general style and conventions of the codebase, ensuring consistency
- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [x] I have run `pre-commit run --all` locally in the source folder and confirmed that there are no linter errors.
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing tests pass locally with my changes

PR #4361, opened 2026-07-28
fix: Handle many missing top-level packages

jjerphan:
# Description

Failures first observed in https://github.com/mamba-org/mamba/pull/4357.

## Type of Change

- [x] Bugfix
- [ ] Feature / enhancement
- [ ] CI / Documentation
- [ ] Maintenance

## Checklist

### DIFF
--- a/libmamba/src/solver/problems_graph.cpp
+++ b/libmamba/src/solver/problems_graph.cpp
@@ -9,7 +9,6 @@
 #include <limits>
 #include <map>
 #include <sstream>
-#include <stdexcept>
 #include <string_view>
 #include <tuple>
 #include <type_traits>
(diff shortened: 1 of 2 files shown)


JohanMabille (approved):

libmamba/src/solver/problems_graph.cpp
```diff
@@ -378,6 +377,28 @@ namespace mamba::solver
             return std::visit([](const auto& n) -> std::string_view { return name_or(n, ""); }, node);
         }
 
+        /**
+         * The name of a CompressedProblemsGraph::node_t.
+         */
+        auto compressed_node_name(const CompressedProblemsGraph::node_t& node) -> std::string_view
+        {
+            return std::visit(
+                [](const auto& n) -> std::string_view
```
Klaim: I think overloads would be clearer here, itś a lot of code just to say "if it's that type"
```c++
struct { 
   auto operator()(const CompressedProblemGrpah::RootNode& root_node) const -> std::string_view
   {
        return "";
   }
   auto operator()(const auto& node const -> std::string_view
   {
        return node.name();
   }
}
```
Or if we had an `overloaded()` callable combiner (not sure if we already have that in the codebase, but it's common when using variants/visit):
```c++
std::visit( std::overloaded([](const RootNode&){ return ""; }, [](const auto& node){ return node.name(); }), node);

File: libmamba/src/solver/problems_graph.cpp [hunk 2]
```diff
@@ -649,22 +679,17 @@ namespace mamba::solver
     {
         if (first < last)
         {
+            const auto first_name = std::string(invoke_name(*first));
```
Klaim: Why is the `std::string` ncessary here?

Hind-M (approved):

libmamba/src/solver/problems_graph.cpp
Hind-M: It doesn't seem necessary indeed.

JohanMabille:
Ok let me address the comments before we merge

Klaim (approved):
LGTM

PR #4364, opened 2026-07-28
fix: Honor `override_virtual_packages` configuration

### LINKED ISSUE #4125: Micromamba is ignoring override_virtual_packages specified in .mambarc file
### Troubleshooting docs

### Anaconda default channels

archspec

### Latest version of Mamba

### Tried in Conda?

### Describe your issue

override_virtual_packages defined in .mambarc is parsed by micromamba but not applied to virtual packages such as archspec, glibc, or cuda.

Even with valid syntax and confirmed config loading, micromamba continues to report autodetected virtual package values. Environment-variable overrides (e.g. CONDA_OVERRIDE_ARCHSPEC) do work, which suggests the issue is specific to config-based overrides.

Create a .mambarc with virtual package overrides:

channels:
  - conda-forge
override_virtual_packages:
  archspec: "1 x86_64_v4"
  cuda: "13.1"
  glibc: "2.15"

```shell
micromamba info | sed -n '/virtual packages/,+10p'

virtual packages :
  __glibc=2.17=0
  __archspec=1=x86_64_v3
```

### Logs

```shell

```

### environment.yml

```yaml

```

### ~/.condarc

```yaml

```

jjerphan:
# Description

Fix #4125.

## Type of Change

- [x] Bugfix
- [ ] Feature / enhancement
- [ ] CI / Documentation
- [ ] Maintenance

## Checklist


libmamba/include/mamba/core/context.hpp
```diff
@@ -259,6 +259,11 @@ namespace mamba
 
         std::vector<std::string> pinned_packages = {};
 
+        // Keys are virtual package names without the `__` prefix (e.g. "cuda", "glibc").
+        // Names with a `__` prefix are also accepted and normalized on lookup.
+        // Environment variables `CONDA_OVERRIDE_<NAME>` take precedence over this map.
+        std::map<std::string, std::string> override_virtual_packages = {};
```
Klaim: Prefer `unordered_map` (and in the future `flat_map`) by default for associative containers. Or is there a good reason to use map?
Also, `find` is never used with this object and it is processed linearly, so for me this should be (`flat_map` but we can't) `vector<SomeStruct>`.

libmamba/src/core/virtual_packages.cpp
```diff
@@ -29,12 +29,35 @@ namespace mamba
...
-                return override_version.value();
+                return env_override;
+            }
+
+            for (const auto& [key, value] : overrides)
+            {
+                const auto normalized = util::starts_with(key, "__")
+                                            ? std::string_view(key).substr(2)
+                                            : std::string_view(key);
+                if (normalized == name)
```
Klaim: IFF overrides was an `unordered_map`, just doing `find(name) != end or find("__"+name) != end` would make more sense than this loop.

libmamba/src/core/virtual_packages.cpp
```diff
@@ -29,12 +29,33 @@ namespace mamba
...
-            auto override_version = util::get_env("CONDA_OVERRIDE_GLIBC");
-            if (override_version)
+            // Environment variables take precedence over config (conda-compatible).
+            if (auto env_override = util::get_env("CONDA_OVERRIDE_" + util::to_upper(name)))
             {
-                return override_version.value();
+                return env_override;
+            }
+
+            if (const auto it = overrides.find(std::string{ name }); it != overrides.end())
```
Klaim: I believe the `std::string` conversion is not necessary because of transparent comparison? (as in the object is passed directly in `it->first == name` so as long as the types are comparable there is no conversion needed (see [case 3/4 in there](https://en.cppreference.com/cpp/container/unordered_map/find)))

libmamba/include/mamba/core/context.hpp
```diff
@@ -259,6 +260,11 @@ namespace mamba
 
         std::vector<std::string> pinned_packages = {};
 
+        // Keys are virtual package names without the `__` prefix (e.g. "cuda", "glibc").
+        // Names with a `__` prefix are also accepted and normalized on lookup.
+        // Environment variables `CONDA_OVERRIDE_<NAME>` take precedence over this map.
+        std::unordered_map<std::string, std::string> override_virtual_packages = {};
```
Klaim: Very minor improvment suggestion: using an alias so that any future change in the exact type doesnt impact all the using code.

Klaim (approved):
LGTM overall, I'm not familiar withthe concept of virtual packages though, maybe another review would be best

libmamba/src/api/configuration.cpp
```diff
@@ -1597,6 +1597,26 @@ namespace mamba
...
+                   .set_rc_configurable()
+                   .description("Override values for virtual packages")
+                   .long_description(unindent(R"(
+                        A dictionary of virtual package overrides. Keys are virtual package names
+                        without the leading `__` (e.g. `cuda`, `glibc`, `archspec`), though names
+                        with `__` are also accepted. Values override the detected version (or the
+                        build string for `archspec`).
+
+                        Environment variables of the form `CONDA_OVERRIDE_<NAME>` take precedence
+                        over this setting.
```
Hind-M, line 1611: How is `conda` behaving regarding this? Names with and without `__` are both accepted? I suppose `CONDA_OVERRIDE_` takes precedence?

libmamba/include/mamba/core/virtual_packages.hpp
```diff
@@ -7,20 +7,31 @@
...
 #include "mamba/specs/package_info.hpp"
 
 namespace mamba
 {
-    class Context;
-
-    std::vector<specs::PackageInfo> get_virtual_packages(const std::string& platform);
+    std::vector<specs::PackageInfo> get_virtual_packages(
+        const std::string& platform,
+        const std::unordered_map<std::string, std::string>& override_virtual_packages = {}
```
Hind-M: I think we should not set a default for `override_virtual_packages` in this kind of functions. Making it explicit avoids omissions and therefore potential bugs.
Same for `cuda_version` and `dist_packages`.
In tests, if applicable, we should just set it as empty explicitly.

libmamba/include/mamba/core/virtual_packages.hpp
```diff
@@ -7,20 +7,31 @@
...
     {
-        std::string cuda_version();
+        /** Resolve a virtual-package override: `CONDA_OVERRIDE_<NAME>` then config map. */
+        [[nodiscard]] auto get_virtual_package_override(
+            std::string_view name,
+            const std::unordered_map<std::string, std::string>& overrides
+        ) -> std::optional<std::string>;
+
+        std::string
+        cuda_version(const std::unordered_map<std::string, std::string>& override_virtual_packages = {});
```
Hind-M: Same remark for optional `override_virtual_packages`.

libmamba/include/mamba/core/virtual_packages.hpp
```diff
@@ -29,7 +40,10 @@ namespace mamba
             std::string build_string = ""
         ) -> specs::PackageInfo;
 
-        std::vector<specs::PackageInfo> dist_packages(const std::string& platform);
+        std::vector<specs::PackageInfo> dist_packages(
+            const std::string& platform,
+            const std::unordered_map<std::string, std::string>& override_virtual_packages = {}
```
Hind-M: Same remark for optional `override_virtual_packages`.

libmamba/src/api/configuration.cpp
```diff
@@ -1597,6 +1597,27 @@ namespace mamba
...
+
+                        Environment variables of the form `CONDA_OVERRIDE_<NAME>` take
+                        precedence over this setting, as in conda.
+
+                        Example:
+                          override_virtual_packages:
+                            cuda: "12.8"
+                            glibc: "2.17"
+                            archspec: "x86_64_v3"
+                   )")));
```
Hind-M, line 1618: I'm not sure we need to specify explicitly that the behavior matches `conda`'s in the description. My previous question was more to validate the behavior regarding this PR with what `conda` does, not necessarily to write it down (feels a bit weird IMHO, we won't mention that in every command/option/flag...).

libmamba/tests/src/core/test_configuration.cpp
```diff
@@ -861,6 +861,43 @@ namespace mamba
...
+                    { "glibc", "2.15" },
+                    { "archspec", "x86_64_v4" },
+                };
+                REQUIRE(actual == expected);
+                REQUIRE(ctx.override_virtual_packages == expected);
+            }
+
+            TEST_CASE_METHOD(Configuration, "override_virtual_packages_dunder_keys")
+            {
+                // conda accepts both `cuda` and `__cuda` as keys.
```
Hind-M: This is weird, we are not in `conda` codebase, maybe just:

```suggestion
                // Both `cuda` and `__cuda` keys are accepted.
```

jjerphan: This was to mention that conda at the origin of this behavior.

Hind-M: Well maybe something more explicit like:
```
// Accept both `cuda` and `__cuda` as keys (like `conda`).
```

dev/environment-dev.yml
```diff
@@ -65,3 +65,5 @@ dependencies:
   - sphinx-book-theme
   - myst-parser
   - sphinx-copybutton
+  # For conda parity tests
+  - conda
```
Hind-M: I don't think we want this even just for tests...

libmamba/tests/src/core/test_virtual_packages.cpp
```diff
@@ -116,41 +141,186 @@ namespace mamba
...
+                REQUIRE(pkgs[4].name == "__cuda");
+                REQUIRE(pkgs[4].version == "13.1");
+
+                pkgs = get_virtual_packages("osx-64", overrides);
+                REQUIRE(pkgs.size() == 4);
+                REQUIRE(pkgs[0].name == "__unix");
+                REQUIRE(pkgs[1].name == "__osx");
+                REQUIRE(pkgs[1].version == "12.1");
+                REQUIRE(pkgs[2].name == "__archspec");
+                REQUIRE(pkgs[2].build_string == "x86_64_v4");
```
Hind-M, line 195: Maybe add this to be consistent with the previous test on `linux-64`: `REQUIRE(pkgs[2].version == "1");`?

micromamba/tests/test_virtual_pkgs.py
```diff
@@ -29,3 +29,77 @@ def test_virtual_linux(self):
...
+  cuda: "13.1"
+"""
+        )
+        infos = info(
+            "--rc-file",
+            str(rc_file),
+            env={**os.environ, "CONDA_OVERRIDE_CUDA": ""},
+        )
+
+        assert "__cuda=" not in infos
```
Hind-M: I don't think we need these tests as they seem to be redundant with the already added tests in `test_configuration.cpp`. (If no added value, we should avoid adding more python tests).

libmamba/tests/src/core/test_virtual_packages.cpp
```diff
@@ -21,6 +25,25 @@ namespace mamba
...
+            require_virtual_package(const std::vector<specs::PackageInfo>& pkgs, std::string_view name)
+                -> const specs::PackageInfo&
+            {
+                const auto it = std::ranges::find_if(
+                    pkgs,
+                    [&](const auto& pkg) { return pkg.name == name; }
+                );
+                REQUIRE(it != pkgs.end());
+                return *it;
+            }
```
Hind-M: This seems to be the same as the one defined in `test_configuration.cpp`. We should put it in a common header, maybe `mambatests_utils.hpp`.

libmamba/tests/src/core/test_virtual_packages.cpp
```diff
@@ -116,41 +132,187 @@ namespace mamba
...
+
+            TEST_CASE("conda_parity_empty_env_overrides_config")
+            {
+                mambatests::EnvironmentCleaner env_clean(mambatests::CleanMambaEnv{});
+                util::set_env("CONDA_OVERRIDE_CUDA", "");
+                const override_virtual_packages_map overrides = { { "cuda", "13.1" } };
+
+                const auto pkgs = get_virtual_packages("linux-64", overrides);
+                REQUIRE_FALSE(has_virtual_package(pkgs, "__cuda"));
```
Hind-M, line 314: This is failing.

jjerphan: c9a01a0 is a tentative fix.

libmamba/src/util/environment.cpp
```diff
@@ -24,63 +24,54 @@ namespace mamba::util
     namespace
     {
 
-        // See:
-        // https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/getenv-s-wgetenv-s?view=msvc-170
+        // CRT and Win32 environment APIs are not thread-safe.
```
Hind-M: Maybe keep the reference for more details?

libmamba/src/util/environment.cpp
```diff
@@ -24,63 +24,54 @@ namespace mamba::util
...
-        {
-            throw std::runtime_error(
-                fmt::format(R"(Failed to acquire environment variable "{}" : errcode = {})", key, error_code)
-
-            );
-        };
-
-        std::scoped_lock ready_to_execute{ env_mutex };  // Calls to getenv_s kinds of
-                                                         // functions are not thread-safe, this
-                                                         // is to prevent related issues.
```
Hind-M: If we have no context, the previous comment is more explicit/clearer.

libmamba/src/util/environment.cpp
```diff
@@ -24,63 +24,54 @@ namespace mamba::util
...
+        ::SetLastError(ERROR_SUCCESS);
+        const DWORD needed = ::GetEnvironmentVariableW(unicode_key.c_str(), nullptr, 0);
+        if (needed == 0)
         {
-            return {};
+            if (::GetLastError() == ERROR_ENVVAR_NOT_FOUND)
+            {
+                return {};
+            }
+            return std::string{};
```
Hind-M, line 48: Shouldn't we throw if there are other errors like before (previous behavior)?

libmamba/src/util/environment.cpp
```diff
@@ -24,63 +24,54 @@ namespace mamba::util
...
+        ::SetLastError(ERROR_SUCCESS);
+        const DWORD nchars = ::GetEnvironmentVariableW(unicode_key.c_str(), value.data(), needed);
+        if (nchars == 0)
         {
-            on_failed(error_code);
+            if (::GetLastError() == ERROR_ENVVAR_NOT_FOUND)
+            {
+                return {};
+            }
+            return std::string{};
```
Hind-M, line 60: Same?
Or maybe just throw if `::GetLastError() != ERROR_SUCCESS` after checking it's not `ERROR_ENVVAR_NOT_FOUND`, to avoid going through all the errors (if the behavior is supposed to be the same for all of them of course).

Hind-M (approved):

PR #4369, opened 2026-08-03
fix: Explain missing packages when string ids collide with solvables

jjerphan:
# Description

Fix regressions introduced by https://github.com/mamba-org/mamba/pull/4312 observed in the workflows of https://github.com/conda/conda-libmamba-solver/pull/957/changes/f73d38db300bec4ef894df83a697fa6cc4ad7137.

`pool_dependency_to_string` treated any id that happened to be a valid solvable index as a lock/verify job id. String ids and solvable indices share `libsolv`'s numeric Id space, so a missing-package install job whose name was interned early produced an empty problems graph and only `"unsupported request"`.

Detect `SOLVER_SOLVABLE` jobs via the job select mask instead, and keep plain string dependencies for `JOB_UNSUPPORTED`.

## Type of Change

- [x] Bugfix
- [ ] Feature / enhancement
- [ ] CI / Documentation
- [ ] Maintenance

## Checklist


libmamba/src/solver/libsolv/unsolvable.cpp
```diff
@@ -395,6 +400,32 @@ namespace mamba::solver::libsolv
                 return pkg;
             };
 
+            // Lock/verify jobs use SOLVER_SOLVABLE; dep_id is then a solvable id.
```
Klaim: Could you add backticks around code so that it's a bit clearer (notably is `;` code or prose?)

libmamba/src/solver/libsolv/unsolvable.cpp
```diff
@@ -395,6 +400,32 @@ namespace mamba::solver::libsolv
...
+            // Detect that via the job select mask (stored in target_id), not by
+            // probing get_solvable(dep_id), which collides with string ids.
+            // Returns true when the problem was handled as such a job.
+            auto try_add_root_locked_solvable = [&](const SolverProblem& problem) -> bool
```
Klaim:
```suggestion
            const auto try_add_root_locked_solvable = [&](const SolverProblem& problem) -> bool
```

libmamba/tests/src/solver/libsolv/test_solver.cpp
```diff
@@ -1469,4 +1470,78 @@ namespace
...
+        auto early_repo = db.add_repo_from_packages(
+            std::array{ specs::PackageInfo("missing-pkg", "1.0", "h0", 0) },
+            "early",
+            libsolv::PipAsPythonDependency::No
+        );
+        db.remove_repo(early_repo);
+
+        auto pkgs = std::vector<specs::PackageInfo>{};
+        pkgs.reserve(512);
+        for (int i = 0; i < 300; ++i)
```
Klaim: Why did you reserve 512 then instantiate only 300?
Could you make these constants named so that it's clearer what is there role here?

Klaim (approved):
LGTM but note that I'm not familiar with that specific logic so I might not be able to spot a logic error compared to the intention. I suspect @AntoinePrv might be best knowledgeable about this part.

libmamba/tests/src/solver/libsolv/test_solver.cpp
jjerphan: Probably some leftovers. :man_facepalming:

Hind-M (approved):

PR #4370, opened 2026-08-03
Fix shell init with `cmd.exe`

### LINKED ISSUE #4077: Silent failure on Window after "shell hook"
### Troubleshooting docs

### Anaconda default channels

Micromamba

Silent failure

### Latest version of Mamba

### Tried in Conda?

Not applicable

### Describe your issue

OS: Windows 10 LTSC

1. **Control**: Run `micromamba.exe` on a system having no Mamba/Micromamba-related files/settigns under %USERPROFILE% - displays help message.
2. Run `micromamba.exe shell hook -s cmd.exe` - directory `%APPDATA%\mamba` is created with scripts.
3. Run `micromamba.exe` again - fails silently. No output to console is produced whatsoever, though %ERRORLEVEL% is set to non-zero value.
4. Delete `%APPDATA%\mamba` and Micromamba works again as before.

Also, even when solely use `micromamba.exe create`, `%APPDATA%\mamba` is somehow created, causing subsequrnt failure, if not deleted. Perhaps, it is created by Mamba during subsequent `mamba.bat install` operations?

Hind-M:
# Description

Fix #4077 

## Type of Change

- [x] Bugfix
- [ ] Feature / enhancement
- [ ] CI / Documentation
- [ ] Maintenance

## Checklist


micromamba/src/main.cpp
```diff
@@ -83,158 +84,198 @@ decide_log_handler(const ContextOptions& options) -> mamba::logging::AnyLogHandl
...
+        return EXIT_FAILURE;
     }
 
-    if (error_to_report)
+    catch (...)
     {
-        LOG_CRITICAL << error_to_report.value();
-        return_value = EXIT_FAILURE;
+        report_error(argc, argv, "Unhandled non-standard exception");
+        return EXIT_FAILURE;
```
Hind-M: @Klaim Can you please confirm this is the way to go to print errors in case of `json` and `quiet` flags?
See `report_error` function (we can't use `Console` since everything is instantiated in the `try` block and this is the outer `catch` part).
Note that I added this outer `try`/`catch` for the code in the `main` function because construction of `Context` and other objects may throw.

libmamba/include/mamba/core/output.hpp
```diff
@@ -25,6 +25,11 @@ namespace mamba
 
     std::string cut_repo_name(std::string_view reponame);
 
+    namespace logging
+    {
+        auto to_json(const logging::LogRecord& record) -> nlohmann::json;
```
Hind-M, line 30: `to_json` has many overloads, so I put it in a specific namespace (we can also decide to pick another name) . We can also move this to `logging.{hpp, cpp}` instead but since `logging` is not pulling `nlohmann_json`, I left it here for now.

Hind-M:
Note that the output part definitely needs more work and testing in the future to simplify things (+ check `set_terminate` possibility). Added some general `TODO` comments.

libmamba/include/mamba/core/output.hpp
Klaim: I think it's ok in that namespace for now, even if not ideal in that header.

> We can also move this to logging.{hpp, cpp} instead but since logging is not pulling nlohmann_json, I left it here for now.

Yes I would prefer to avoid making `logging.hpp` dragging an additional dependency, in particular a header-only heavy library.
We coudl also add a header `logging-json.hpp` or something but that seems overkill right now.

libmamba/src/core/output.cpp
```diff
@@ -275,11 +275,10 @@ namespace mamba
      * Console *
      ***********/
 
-
-    namespace
+    // TODO Think about moving the json machinery somewhere independent from `Console`?
+    // maybe `logging.{hpp, cpp}`? or some other new files which would include `logging.hpp`
```
Klaim: New header file that's a bridge between the logging system and some json library would make sense, but it's ok to do it another time. logging.hpp is a big no-go for me (maybe reflect that in this comment?).

Klaim:
> Note that the output part definitely needs more work and testing in the future to simplify things (+ check `set_terminate` possibility). Added some general `TODO` comments.

Why not do it now? I'm not at ease with this big try/catch and would prefer either a RAII solution or terminate to handle all the unexpected failure cases (details discussed privately).
I think it's best to not go with a temporary solution for this kind of things. I can help with the pr if you want.

micromamba/src/main.cpp
```diff
@@ -83,13 +86,82 @@ decide_log_handler(const ContextOptions& options) -> mamba::logging::AnyLogHandl
...
+    }
+    else if (not options or not options->output_params or not options->output_params->quiet)
+    {
+        std::cerr << message << std::endl;
+    }
+}
+
+namespace
+{
+    bool constructed_console = false;
```
Klaim: I think this needs to be atomic as the terminate handler has no guarantee as to which thread it will be called from.
No need to protect the options because they are set only once before setting the handler.

Hind-M:
@Klaim This can be reviewed again. Thanks!

micromamba/src/main.cpp
```diff
@@ -83,13 +88,91 @@ decide_log_handler(const ContextOptions& options) -> mamba::logging::AnyLogHandl
...
+
+        nlohmann::json output{
+            { "success", false },
+            { "log_history", nlohmann::json::array({ record_json }) },
+        };
+        std::cout << output.dump(4) << std::endl;
+    }
+    else if (not options or not options->output_params or not options->output_params->quiet)
+    {
+        std::cerr << message << std::endl;
```
Klaim:
```suggestion
        std::cerr << "critical error: " << message << std::endl;
```
Or maybe just `"crictical: "`?
Just so that it's not confused as anything else than an error report. It's already more obvious in the other cases.

micromamba/src/main.cpp
```diff
@@ -83,13 +88,91 @@ decide_log_handler(const ContextOptions& options) -> mamba::logging::AnyLogHandl
...
 {
+    pre_config_options = decide_preconfig_context_options(argc, argv);
+    std::set_terminate(mamba_terminate_handler);
+
     mamba::MainExecutor scoped_threads;
-    const auto pre_config_options = decide_preconfig_context_options(argc, argv);
-    mamba::Context ctx{ pre_config_options, decide_log_handler(pre_config_options) };
+    mamba::Context ctx{ pre_config_options.value(), decide_log_handler(pre_config_options.value()) };
     mamba::Console console{ ctx };
+    constructed_console = true;  // TODO think of a more elegant way for this
```
Klaim:
```suggestion
    constructed_console = true;  // TODO: think of a more elegant way for this
```
I wish `std::once_flag` had a specified interface, that would have been perfect here. Anyway yeah.

Klaim (approved):
minor tweaks requested but LGTM

micromamba/src/main.cpp
```diff
@@ -83,13 +88,91 @@ decide_log_handler(const ContextOptions& options) -> mamba::logging::AnyLogHandl
...
+void
+report_error(
+    const std::optional<ContextOptions>& options,
+    const std::string& message,
+    const bool constructed_console
+)
+{
+    if (Console::is_available())
+    {
+        LOG_CRITICAL << message;
```
jjerphan, line 101: How about using early-returns here and in other branches?

micromamba/src/main.cpp
```diff
@@ -83,13 +88,91 @@ decide_log_handler(const ContextOptions& options) -> mamba::logging::AnyLogHandl
...
+    }
+    // `constructed_console` is used to avoid printing 2 json objects
+    else if (options and options->output_params and options->output_params->json
+             and not constructed_console)
```
jjerphan: How about defining:

```cpp
    const bool must_output_json = options and options->output_params and options->output_params->json;
```

And using it then, here:

```suggestion
    else if (must_output_json and not constructed_console)
```

and in the next branch (see the other suggestion)?

micromamba/src/main.cpp
```diff
@@ -83,13 +88,91 @@ decide_log_handler(const ContextOptions& options) -> mamba::logging::AnyLogHandl
...
+        };
+        const nlohmann::json record_json = logging::to_json(log_record);
+
+        nlohmann::json output{
+            { "success", false },
+            { "log_history", nlohmann::json::array({ record_json }) },
+        };
+        std::cout << output.dump(4) << std::endl;
+    }
+    else if (not options or not options->output_params or not options->output_params->quiet)
```
jjerphan:
```suggestion
    else if (not must_output_json)
```

micromamba/src/main.cpp
```diff
@@ -83,13 +88,91 @@ decide_log_handler(const ContextOptions& options) -> mamba::logging::AnyLogHandl
...
+    const std::optional<ContextOptions>& options,
+    const std::string& message,
+    const bool constructed_console
+)
+{
+    if (Console::is_available())
+    {
+        LOG_CRITICAL << message;
+    }
+    // `constructed_console` is used to avoid printing 2 json objects
```
jjerphan: I understand `report_error` allow reporting errors when the console and logging system might not have been (properly) initialized, but the this logic is not the most motivated entirely to me.

For instance, I do not understand the origin of 2 JSON objects being printed and why `constructed_console` is required here over class methods on `Console` such as `Console::is_available` and `Console::can_report_status`. I would also prefer using something else that a global boolean accessible without encapsulation as you also noted.

libmamba/src/core/shell_init.cpp
```diff
@@ -800,10 +800,20 @@ namespace mamba
     {
         const ShellInitPathsWindowsCmd paths{ root_prefix };
 
+        // The default root prefix is validated on the next invocation, so it must
+        // be a valid conda prefix (with `pkgs`, `conda-meta` or `envs` folder) from the start.
+        // It used to contain only the hook scripts -> every following run aborted silently.
+        std::error_code ec;
+        fs::create_directories(root_prefix / "conda-meta", ec);
```
jjerphan, line 807: Looking at:

https://github.com/mamba-org/mamba/blob/f1539d0d7592c7cfc914bafb189e90cbcdaf7128/libmamba/src/core/environments_manager.cpp#L19-L22

https://github.com/mamba-org/mamba/blob/f1539d0d7592c7cfc914bafb189e90cbcdaf7128/libmamba/include/mamba/core/environments_manager.hpp#L20

https://github.com/mamba-org/mamba/blob/f1539d0d7592c7cfc914bafb189e90cbcdaf7128/libmamba/src/api/install.cpp#L954-L961

Shouldn't it here use:

```suggestion
        path::touch(root_prefix / "conda-meta" / "history", true);
```

?

micromamba/src/main.cpp
Klaim: We discussed this with @Hind-M and I suggested to handle this case, sorry for not writing the reasoning down first. Here:

Once terminate handler `mamba_terminate_handler` is set, it could be called at any time, which includes:
- before the `Console` (which is used by the logging macros) is initalized but after the logging system is ready -> covered here ;
- while  `Console` is usable -> covered in the previous branch;
- after `Console` is destroyed -> covered in the following branch;
- any time after `main()`, while static objects are destroyed -> covered in the following branch.

We must cover all the cases.

When `Console` is destroyed with json output enabled, it will dump the json, that is the standard ouput will be written in already, we can't modify any more the logs from that output after that point. If the terminate handler is called after that dump but before main is finished ,then outputing a json object again will result in the output with 2 json objects. That's what we're trying to avoid here with these different checks.

Does that makes more sense?

> I would also prefer using something else that a global boolean accessible without encapsulation as you also noted.

Do you have a suggestion? The only similar thing I can think about is `std::once_flag` but it's not exactly what we need here.

micromamba/src/main.cpp
Hind-M: Here we use `options->output_params->json` but then further below we use `options->output_params->quiet` so not the same condition.

micromamba/src/main.cpp
Hind-M: See [previous comment](https://github.com/mamba-org/mamba/pull/4370#discussion_r3776785799).

libmamba/src/core/shell_init.cpp
Hind-M: `history` file is created and used only with actual environments (i.e target env). In the case of this PR, `shell hook` only uses and validates the `root prefix` as it writes hook scripts in there, so [this](https://github.com/mamba-org/mamba/blob/main/libmamba/src/api/configuration.cpp#L616) was failing.

micromamba/src/main.cpp
Hind-M: To add some pointers that would help understand:

When `Console` is destroyed [here](https://github.com/mamba-org/mamba/blob/main/libmamba/src/core/output.cpp#L384-L394), the json is dumped as mentioned above and everything is lost (can't access `Console` obviously so can't use its methods), so the `constructed_console` flag is basically saying `Console` was there at some point, so we can't create another json object (there should be only one `json` output at the end of execution) => use `cerr` (the last `if` branch) otherwise, we can/should create a `json` object with the corresponding error (in the case of `--json` of course).

jjerphan: OK thank you for the precision.

What do you think of adding an inline comment to make the intend clear?

micromamba/src/main.cpp
```diff
@@ -83,13 +88,102 @@ decide_log_handler(const ContextOptions& options) -> mamba::logging::AnyLogHandl
...
 {
+    pre_config_options = decide_preconfig_context_options(argc, argv);
+    std::set_terminate(mamba_terminate_handler);
+
     mamba::MainExecutor scoped_threads;
-    const auto pre_config_options = decide_preconfig_context_options(argc, argv);
-    mamba::Context ctx{ pre_config_options, decide_log_handler(pre_config_options) };
+    mamba::Context ctx{ pre_config_options.value(), decide_log_handler(pre_config_options.value()) };
     mamba::Console console{ ctx };
+    constructed_console = true;  // TODO: think of a more elegant way for this
```
jjerphan: As an alternative, we could use in `Console`:
```cpp
enum class ConsoleLifetime { never_created, available, destroyed };

static ConsoleLifetime lifetime();  // never_created | available | destroyed
static bool is_available();         // lifetime() == available
```

and in place of `constructed_console`, we could check that:

```cpp
Console::lifetime() == destroyed
```

jjerphan (approved):
LGTM modulo a suggestion for improving the case where the `Console` was destroyed replacing the global atomic.

micromamba/src/main.cpp
Hind-M: I don't have a strong opinion on this.
@Klaim What do you think?

Klaim: I'm not sure:
1. By moving the static object out of main's translation unit we increase the chances that it be destroyed before a terminate handler is called (because order of destruction and construction of static objects are only guaranteed in the same translation unit, also known as Static Initialization Fiasco). I'm not 100% sure if having the static object in main's translation unit gives some guarantees, I believe not, but I think it's more likely that that value and the optional are alive at the same time. Cant completely confirm either 🤔
2. I don't see why it would help to expose globally, publicly, in the library's interface, this information which is kind of specific to micromamba's `main()`, but at the same time it doesnt really hurt.

The enum can help with clarity.

In any case the stored value still needs to be atomic. This is because termination handlers can be called from any thread, even threads not handled by the program.

I think we could try, it's at least easy to revert. 
I'll let Hind decide given these points. 🙈

Hind-M: Hum, I see. Since the whole issue was that errors were not printed and everything failed silently, and all the changes in this PR regarding error reporting are meant to capture as much information as possible in all cases, I would say to stick with the global bool in `main()` for now (functionality over clarity).
I think that we will need to consider all these possibilities holistically when trying to move things around/refactor/simplify in the future... (will reference this discussion in the `TODO` comment as well).

libmamba/src/core/output.cpp
```diff
@@ -275,11 +275,11 @@ namespace mamba
      * Console *
      ***********/
 
-
-    namespace
+    // TODO Think about moving the json machinery somewhere independent from `Console`?
```
Klaim:
```suggestion
    // TODO: Think about moving the json machinery somewhere independent from `Console`?
```
(todo-highlighting tools usually requires the colon to make these highlighted, you can ignore this suggestion if you don't care 🙈)

micromamba/src/main.cpp
```diff
@@ -83,13 +88,104 @@ decide_log_handler(const ContextOptions& options) -> mamba::logging::AnyLogHandl
...
+        LOG_CRITICAL << message;
+        return;
+    }
+
+    // If `Console` was constructed and then destroyed
+    // (i.e `not Console::is_available()` and `constructed_console`),
+    // we should use `cerr` (last `if` branch) and not create another json
+    // object (already one dumped in `Console` destructor, if `--json` is given).
+    // `cerr` is also used in the following cases: no `--json` and no `--quiet`.
+    // Otherwise, we must create a json object (unique in this case)
```
Klaim:
```suggestion
    // In the opposite case, when Console has not be created and logging macros are not yet usable, we must still create a json object (unique in this case)
```
(need formatting 🙈)

micromamba/src/main.cpp
```diff
@@ -83,13 +88,104 @@ decide_log_handler(const ContextOptions& options) -> mamba::logging::AnyLogHandl
...
 {
+    pre_config_options = decide_preconfig_context_options(argc, argv);
+    std::set_terminate(mamba_terminate_handler);
+
     mamba::MainExecutor scoped_threads;
-    const auto pre_config_options = decide_preconfig_context_options(argc, argv);
-    mamba::Context ctx{ pre_config_options, decide_log_handler(pre_config_options) };
+    mamba::Context ctx{ pre_config_options.value(), decide_log_handler(pre_config_options.value()) };
     mamba::Console console{ ctx };
+    // TODO: think of a more elegant way for this
```
Klaim:
```suggestion
    // TODO: think of a more elegant way to track console's lifetime stages.
```

Just so if the comment is moved around we know what it was refering to.

Klaim (approved):
LGTM with ignorable nitpicks

libmamba/src/core/output.cpp
Hind-M: Sorry about this, I will try to remember that in the future (`TODO` is highlighted in my case without the colon).

PR #4350, opened 2026-07-20
fix: Add `deactivate` command to the help message

faze-geek:
# Description

Adds visibility to one of the missing commands mentioned in #3535 

## Type of Change

- [x] Bugfix
- [ ] Feature / enhancement
- [ ] CI / Documentation
- [ ] Maintenance

## Checklist

- [x] My code follows the general style and conventions of the codebase, ensuring consistency
- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [x] I have run `pre-commit run --all` locally in the source folder and confirmed that there are no linter errors.
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing tests pass locally with my changes

PR #4371, opened 2026-08-04
fix: Remove duplicate deactivate from shell completion

henriquesimoes:
# Description

When implementing the shell completion, the "deactivate" command is introduced manually in add_activate_completion(). When first introduced by 821de6b5 (also complete for deactivate command (#1557), 2022-03-03) this was needed (although probably in the wrong function), because micromamba didn't have a proper "deactivate" command.

However, commit 5206c2a9 (fix: Add `deactivate` command to the help message (#4350), 2026-07-29) added this missing command to set_umamba_command() (that is also used for handling the "completer" command), but it didn't remove the extra "deactivate". Because of that, running the autocompletion triggers the following error:

    terminate called after throwing an instance of 'CLI::OptionAlreadyAdded'
    what():  subcommand name or alias matches existing subcommand: deactivate is already added

Drop the manual and redundant addition of "deactivate" command in the completer implementation, so the autocompletion gets back to working as expected.

Fixes: 5206c2a9 (fix: Add `deactivate` command to the help message (#4350), 2026-07-29)

## Type of Change

- [x] Bugfix
- [ ] Feature / enhancement
- [ ] CI / Documentation
- [ ] Maintenance

## Checklist


jjerphan (approved):
Good catch! I can reproduce it.

henriquesimoes:
@jjerphan, thanks for the review and for adding the missing test to avoid future regressions. ;)

jjerphan:
Thank you for this fix, @henriquesimoes.

We have been lucky to include it as part of [2.9.0](https://github.com/mamba-org/mamba/releases/tag/2.9.0). :)

PR #4379, opened 2026-08-11
fix: Use `stop_reason::program_exit` when destructing  `Context`

### LINKED ISSUE #1319: Test crash when importing from conda_interface.py with mamba 2.8.1
### Checklist

- [x] I added a descriptive title
- [x] I searched open reports and couldn't find a duplicate

### What happened?

While working in https://github.com/conda/constructor/pull/1314 I ran into segmentation fault when running `pytests` when the Python interpreter is shutting down.

```
(test) robin@:~/repos$ PYTHONFAULTHANDLER=1 pytest -q constructor/tests/ -m "not examples"
................................................................sssssssssssssssss.sssssssssss..................ssss....s..ssssssssssss.........................sssssssssssssssssssss [ 77%]
ssssssssssssssssssssssssssssssss...........s.......                                                                                                                                  [100%]

[pytest-split] Stored test durations in /.../.test_durations

===================================================================================== warnings summary =====================================================================================
tests/test_main.py::test_dry_run
tests/test_main.py::test_installer_type_flag_valid
  /home/robin/miniconda3/envs/test2/lib/python3.11/contextlib.py:105: PendingDeprecationWarning: conda.common.io.env_vars is pending deprecation and will be removed in 27.3.
    self.gen = func(*args, **kwds)

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
132 passed, 99 skipped, 54 deselected, 2 warnings in 6.34s
Segmentation fault (core dumped)
```

After a lot of debugging, it seems to be a cocktail of components required to get this to happen. First I noticed this in our CI runners **only** on macOS and Ubuntu for Python 3.10 and Python 3.11 (every other job passed). It also seems to be triggered when you import a new function from `conda_interface.py`, even though the function in this situation does nothing.

## To reproduce
```
robin@:~/repos/constructor$ cat /etc/os-release
PRETTY_NAME="Ubuntu 24.04.4 LTS"
NAME="Ubuntu"
VERSION_ID="24.04"
VERSION="24.04.4 LTS (Noble Numbat)"
...
```
1. Add these changes https://github.com/conda/constructor/commit/2febbe7b5011696c224405edebaff8b005c234b5. Note how I have added a dummy function, and simply imported it to trigger the error.
2. `conda create -n test python=3.11 constructor pytest pytest-mock libmamba=2.8.1 pytest-split -c conda-forge`
3. `pip install -e ./constructor`
4. `PYTHONFAULTHANDLER=1 pytest -q constructor/tests/ -m "not examples"`

I can mention that if I install `libmamba=2.3.2` there is **no** segmentation fault.

### Additional Context

I also ran this a bit via `gdb` earlier:
```
  Thread 1 "python" received signal SIGSEGV
  #0  ?? () in libmambapy/bindings...so
  #1  ?? () in libmambapy/bindings...so
  #2  mamba::logging::stop_logging(mamba::logging::stop_reason) in libmamba.so.4
  #3  mamba::Context::~Context() in libmamba.so.4
  #4  ?? () in libmambapy/bindings...so
  #5  __run_exit_handlers (...) at exit.c:108
  #6  __GI_exit
  #7-9 libc startup / _start
```
I believe that this tells us the crash happens during process exit (`__run_exit_handlers` → `exit()`), not during any test itself. I spent a lot of time trying to create a minimal reproducer based on this information but I had no luck - it's also strange how this is triggered when I add the changes in the commit (link 1) above).

Finally, if it helps anyone reading this issue, the disassembly at the crash instruction (from Claude Code):
```
  mov (%rbx),%rdi      ; load pointer from vector element
  => mov (%rdi),%rax    ; CRASH: dereference that pointer to get vtable
     call *0x18(%rax)   ; virtual function call through vtable
     add $0x10,%rbx     ; advance to next vector element (16-byte stride)
     cmp %rbx,%rbp
     jne ...            ; loop
```

### LINKED ISSUE #4378: Crash when Python interpreter shuts down
### Troubleshooting docs

### Anaconda default channels

segfault, segmentation fault

### Latest version of Mamba

### Tried in Conda?

Not applicable

### Describe your issue

I ran into an issue last week which shows up as a Segmentation fault when the Python interpreter is shutting down. I spent a lot of time trying to create a minimal reproducer to simplify debugging on your end but did not manage to. The issue is filed with more context here https://github.com/conda/constructor/issues/1319. Primarily the interesting part is:
```
  Thread 1 "python" received signal SIGSEGV
  #0  ?? () in libmambapy/bindings...so
  #1  ?? () in libmambapy/bindings...so
  #2  mamba::logging::stop_logging(mamba::logging::stop_reason) in libmamba.so.4
  #3  mamba::Context::~Context() in libmamba.so.4
  #4  ?? () in libmambapy/bindings...so
  #5  __run_exit_handlers (...) at exit.c:108
  #6  __GI_exit
  #7-9 libc startup / _start
```

I first observed it during the CI testing for Python `3.10` and `3.11` (does not happen with `3.12`/`3.13`) and I was also able to reproduce it locally.
I verified that it also happens with `libmamba=2.9.0`. Note that when I see this error, all my tests pass, but after `pytest` has written the test summary and the interpreter and everything is shutting down it crashes as seen in the issue.

This is not blocking me but thought it might be good to share it with you.

```shell

```

### Logs

```shell

```

### environment.yml

```yaml

```

### ~/.condarc

```yaml

```

jjerphan:
# Description

Fix https://github.com/mamba-org/mamba/issues/4378.
Fix https://github.com/conda/constructor/issues/1319.

`Context::~Context` called `stop_logging` with manual_stop, which flushes and drops `spdlog` loggers even during interpreter/`atexit` teardown, when `spdlog`'s static state may already be gone. Use `program_exit` instead, and drop leftover loggers when restarting.

## Type of Change

- [x] Bugfix
- [ ] Feature / enhancement
- [ ] CI / Documentation
- [ ] Maintenance

## Checklist

### DIFF
--- a/libmamba-spdlog/include/mamba/spdlog/logging_spdlog_impl.hpp
+++ b/libmamba-spdlog/include/mamba/spdlog/logging_spdlog_impl.hpp
@@ -92,6 +92,11 @@ namespace mamba::logging::spdlogimpl
 
         const auto main_source = sources.front();
 
+        // A previous `Context` may have been destroyed with `stop_reason::program_exit`,
+        // which intentionally leaves spdlog loggers registered. Clear them before
+        // re-registering so starting logging again after such a shutdown is safe.
+        spdlog::drop_all();
+
         spdlog::set_default_logger(
             std::make_shared<Logger>(name_of(main_source), params.log_pattern, "\n")
         );
--- a/libmamba-spdlog/tests/test_logging_spdlog.cpp
+++ b/libmamba-spdlog/tests/test_logging_spdlog.cpp
@@ -93,6 +93,38 @@ namespace mamba::logging
         }
     }
 
+    TEST_CASE("LogHandler_spdlog restart after program_exit stop")
+    {
+        // Regression: Context::~Context stops logging with program_exit, which leaves
+        // spdlog loggers registered. Starting logging again must still succeed
+        // (mamba-org/mamba#4378).
+        spdlogimpl::LogHandler_spdlog handler{ testing_options };
+        on_scope_exit _{ [&] { handler.stop_log_handling(stop_reason::manual_stop); } };
+
+        handler.start_log_handling({}, testing::testing_log_sources());
+        REQUIRE(handler.is_started());
+        handler.log(
+            LogRecord{
+                .message = "before program_exit stop",
+                .level = log_level::warn,
+                .source = log_source::tests,
+            }
+        );
+
+        handler.stop_log_handling(stop_reason::program_exit);
+        REQUIRE(not handler.is_started());
+
+        handler.start_log_handling({}, testing::testing_log_sources());
+        REQUIRE(handler.is_started());
+        handler.log(
+            LogRecord{
+                .message = "after restart",
+                .level = log_level::warn,
+                .source = log_source::tests,
+            }
+        );
+    }
+
     TEST_CASE("LogHandler_spdlog concurrency")
     {
         spdlogimpl::LogHandler_spdlog handler{ testing_options };
--- a/libmamba/src/core/context.cpp
+++ b/libmamba/src/core/context.cpp
@@ -130,7 +130,19 @@ namespace mamba
 
     Context::~Context()
     {
-        logging::stop_logging();
+        // Use `program_exit` rather than the default `manual_stop`.
+        //
+        // `Context` often outlives `main()` / is destroyed from Python during
+        // interpreter or process shutdown (`atexit`, module teardown). In that
+        // situation, touching spdlog (flush / `drop_all`) is unsafe: its static
+        // registry or sinks may already have been destroyed, which leads to a
+        // segfault in `stop_logging` (see mamba-org/mamba#4378,
+        // conda/constructor#1319).
+        //
+        // `program_exit` lets the log-handler skip that cleanup and leave it to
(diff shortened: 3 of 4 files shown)


Klaim:
I think that's not the right solution, will discuss it in private as it's quite delicate/complex to get right.

Klaim:
Reporting here a summary of the discussions and investigations we did in private, for later reference: 
This "termination reason" system was initially introduced to handle the case where `exit()` was called in `(micro)mamba` in which case we don't know the order of calls to spdlog's termination and must let it handle it's resource itself instead of stopping it explicitly.
We agreed that a far better way to handle these cases would be to banish `exit()` calls from the library and program and remove that termination reason system in favor of always explicitly (via RAII) stop spdlog or whatever logging implementation is used. This is already done properly in `(micro)mamba`'s `main()` so we don't have to touch that. The main issue is to remove the `exit()` calls without changing the behavior of the program in a problematic way.

jjerphan:
Closing in preference of https://github.com/mamba-org/mamba/pull/4399.

Klaim:
For completeness: in preference of both #4397 and #4399 which together should remove the possibility of the crash, we hope.

PR #4381, opened 2026-08-17
maint: Use cxx-compiler <2

jjerphan:
# Description

The newly released 2.0 version of `cxx-compiler` removes environment variables used to resolve compilers.

See https://github.com/conda-forge/compilers-feedstock/pull/78 and https://github.com/conda-forge/conda-forge.github.io/issues/2595.

## Type of Change

- [ ] Bugfix
- [ ] Feature / enhancement
- [ ] CI / Documentation
- [x] Maintenance

## Checklist

- [x] My code follows the general style and conventions of the codebase, ensuring consistency
- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [x] I have run `pre-commit run --all` locally in the source folder and confirmed that there are no linter errors.
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing tests pass locally with my changes

PR #4382, opened 2026-08-17
maint: Adapt build setup for `cxx-compiler >=2`

jjerphan:
# Description

Follow-up of https://github.com/mamba-org/mamba/pull/4381.

Comes after https://github.com/mamba-org/mamba/pull/4386 and https://github.com/mamba-org/mamba/pull/4387.

## Type of Change

- [ ] Bugfix
- [ ] Feature / enhancement
- [ ] CI / Documentation
- [x] Maintenance

## Checklist


libmamba/src/core/util_os.cpp
jjerphan, line 1: We can contribute this change separately.

.github/workflows/coverage.yml
```diff
@@ -69,12 +69,6 @@ jobs:
             ccache-libmamba-${{ github.runner.os }}
       - name: Build mamba with coverage
         run: |
-          # Ensure ar/ranlib are found: conda compiler activation may not set AR when
-          # mixing envs (dev + micromamba-static); CMAKE_AR from preset uses $env{AR}
-          # which can resolve to a bare name, causing "x86_64-conda-linux-gnu-ar: not found"
-          # when building the static libmamba (micromamba requires BUILD_STATIC).
-          export AR="${CONDA_PREFIX}/bin/x86_64-conda-linux-gnu-ar"
-          export RANLIB="${CONDA_PREFIX}/bin/x86_64-conda-linux-gnu-ranlib"
```
Hind-M, line 77: Why not leave them?

dev/CMakePresetsUnix.json
```diff
@@ -27,12 +14,9 @@
...
       "cacheVariables": {
-        "CMAKE_AR": "$env{GCC_AR}",
-        "CMAKE_CXX_COMPILER": "$env{GXX}",
-        "CMAKE_C_COMPILER": "$env{GCC}",
-        "CMAKE_Fortran_COMPILER": "$env{GFORTRAN}",
-        "CMAKE_LINKER": "$env{LD_GOLD}",
-        "CMAKE_RANLIB": "$env{GCC_RANLIB}"
+        "CMAKE_CXX_COMPILER": "g++",
+        "CMAKE_C_COMPILER": "gcc",
+        "CMAKE_Fortran_COMPILER": "gfortran"
```
Hind-M: Not sure these work on platforms other that `linux`?

dev/environment-dev.yml
```diff
@@ -4,6 +4,8 @@ dependencies:
   # libmamba build dependencies
   # Workaround `mamba-org/mamba#4043`
   - sel(linux): sysroot_linux-64==2.17
+  # compilers >=2.0 install unprefixed compilers on PATH and no longer set CC/CXX.
+  # See conda-forge/compilers-feedstock#78 and conda-forge/conda-forge.github.io#2595.
```
Hind-M: Can we have full urls here instead (I know it can be more verbose but it's more convenient to just use the link instead of looking for the right one)?

dev/environment-micromamba-static.yml
```diff
@@ -4,11 +4,10 @@ dependencies:
   - python >=3.12
   # libmamba build dependencies
   # Workaround `mamba-org/mamba#4043`
-  # TODO: Understand the linkage issue with GCC 14
-  - sel(linux): cxx-compiler <1.11 #       GCC 13
   - sel(linux): sysroot_linux-64==2.17
-  - sel(osx): cxx-compiler
-  - sel(win): cxx-compiler
+  # compilers >=2.0 install unprefixed compilers on PATH and no longer set CC/CXX.
+  # See conda-forge/compilers-feedstock#78 and conda-forge/conda-forge.github.io#2595.
```
Hind-M: Same.

Hind-M (commented):
I don't get it, are the changes in the codebase related to `cxx-compiler >= 2`?

jjerphan:
> are the changes in the codebase related to cxx-compiler >= 2?

Those are required, but some can be contributed independently for separing concerns.

.github/workflows/coverage.yml
jjerphan: They aren't used anymore; why should we keep them?

dev/CMakePresetsUnix.json
jjerphan: I also agree, this currently preserve the previous values.

CMakeLists.txt
```diff
@@ -8,6 +8,16 @@ cmake_minimum_required(VERSION 3.16)
...
 project(mamba)
 
+# cxx-compiler >=2 no longer injects LDFLAGS. GNU ld does not search -L for DT_NEEDED of linked
+# shared libraries; it needs -rpath-link at link time.
+if(
+    UNIX
+    AND NOT APPLE
+    AND DEFINED ENV{CONDA_PREFIX}
+)
+    add_link_options("-Wl,-rpath-link,$ENV{CONDA_PREFIX}/lib")
```
isuruf: We do set this in the gcc specs file. Can you open an issue with a minimal reproducer?

.github/workflows/coverage.yml
Hind-M: Well that's the question, they were apparently used to avoid issues with static build, are they set now explicitly with `cxx-compiler >= 2`?

dev/CMakePresetsUnix.json
Hind-M: > this currently preserve the previous values.

Well not on every `unix` platform (for which this preset is intended). These should be removed if not covering all possibilities.

Hind-M:
> > are the changes in the codebase related to cxx-compiler >= 2?
> 
> Those are required, but some can be contributed independently for separing concerns.

I would prefer to strictly separate concerns yes (clearer and easier to track if debugging later).

jjerphan:
Following https://github.com/mamba-org/mamba/pull/4382#issuecomment-5339850672, the unrelated bugfixes are now in dedicated PRs:

- https://github.com/mamba-org/mamba/pull/4386 (HTTP header value parsing)
- https://github.com/mamba-org/mamba/pull/4387 (macOS ad-hoc codesign spawn)

jjerphan:
The `coverage.yml` `AR`/`RANLIB` change is now in a dedicated PR as well to test whether the environment variables are still needed for `cxx-compiler<2`: https://github.com/mamba-org/mamba/pull/4388

CMakeLists.txt
jjerphan: Done with https://github.com/conda-forge/compilers-feedstock/issues/80.

.github/workflows/coverage.yml
jjerphan: `cxx-compiler >=2` installs unprefixed `ar`/`ranlib` on `PATH` and no longer sets `AR`/`RANLIB`, so those hardcoded paths are unused (and would point at binaries that are no longer installed).

dev/CMakePresetsUnix.json
jjerphan: `conda-unix` now picks unprefixed compilers from the host OS, matching `cxx-compiler >=2`:

- Linux: `gcc` / `g++` / `gfortran`
- macOS: `clang` / `clang++`

The GNU-named preset no longer hardcodes Linux-only compiler names; it inherits that host selection. LLVM remains an explicit `clang++` override.

jjerphan: I've updated the setup with beafa81.

dev/conda_unix_compilers.cmake
```diff
@@ -0,0 +1,29 @@
...
+        CMAKE_C_COMPILER
+        gcc
+        CACHE STRING "C compiler"
+    )
+    set(
+        CMAKE_CXX_COMPILER
+        g++
+        CACHE STRING "C++ compiler"
+    )
+endif()
```
Hind-M, line 29: I think this doesn't guarantee that the `conda` prefix installed compiler is used (could be an installed compiler elsewhere). Maybe use `$CONDA_PREFIX/bin/clang` instead? (same for other compilers)?
Not sure about the different paths on platforms - to be checked.

jjerphan: Since `$CONDA_PREFIX/bin` is in `PATH`, this should not be required. Also IIUC, the recent changes with `compiler 2.0` eases setups so that one does not need to specify explicit paths.

jjerphan: Hello @isuruf,

We would like to have your guidance here after the `cxx-compiler` 2.0 release.

Is it safe to just use `g++` or `clang++` if `CONDA_PREFIX` is set? What would you recommend setting?

h-vetinari: `cxx-compiler` is now just a [wrapper](https://conda-metadata-app.streamlit.app/?q=conda-forge%2Flinux-64%2Fcxx-compiler-2.0.0-hdab9b82_0.conda&with_broken=true) around `gxx` resp. `clangxx` on unix. The latter packages [come with](https://conda-metadata-app.streamlit.app/?q=conda-forge%2Flinux-64%2Fgxx-16.2.0-hc6a0c74_4.conda&with_broken=true) a [bare symlink](https://conda-metadata-app.streamlit.app/?q=conda-forge%2Fosx-arm64%2Fclangxx-22.1.8-default_nocfg_h70e6b6c_9.conda&with_broken=true), so indeed, it's safe to use `g++` and `clang++`.

Hind-M (approved):

cmake/CompilerWarnings.cmake
```diff
@@ -141,6 +141,10 @@ function(mamba_target_add_compile_warnings target)
         -Wlogical-op
         # Warn if you perform a cast to the same type
         -Wuseless-cast
+        # GCC 13+ false positives on std::variant of types containing std::string (libstdc++
+        # basic_string inlined through variant). Keep the warning, do not fail the build. See GCC PR
+        # 105329 / PR 109561.
+        -Wno-error=maybe-uninitialized
```
Klaim, line 147: I am uncomfortable removing this warning in general (it's also an error starting c++26 (erroneous behavior)). Could we add that flag only if that specific set of compiler versions are used?

dev/conda_unix_compilers.cmake
JohanMabille: Mmh not sure why we need this file, is it just fo avoid passing CMAKE_CXX_COMPILER=g++ to the cmake command line? In that case I would prefer to drop it and explicitly pass the compiler to cmake.

cmake/CompilerWarnings.cmake
jjerphan: Done with 02b468a.

dev/conda_unix_compilers.cmake
jjerphan: Done with 1db86a5.

JohanMabille (approved):

jjerphan:
Waiting for @Klaim's approval regarding https://github.com/mamba-org/mamba/pull/4382#discussion_r3901978861 before merging.

Klaim (approved):
LGTM 👍🏽

PR #4384, opened 2026-08-18
ci: Disable coverage env cache to avoid pygments package-cache errors

jjerphan:
## Summary

- The Code Coverage workflow sporadically failed while linking `pygments` after a `cache-environment` hit, with a stale negative extracted-dir cache in micromamba 2.9.0 (#4328 is on \`main\` but not in that release).
- Combining \`environment-dev.yml\`, extra, and static into one \`micromamba create\` is unsatisfiable (shared vs static libraries, \`python>=3.12\` vs \`bcrypt==4.0.1\`).
- Workaround: keep the sequential extra/static installs and set \`cache-environment: false\` so \`pkgs/\` is present when Python is upgraded.

## Type of Change

- [ ] Bugfix
- [ ] Feature / enhancement
- [x] CI / Documentation
- [ ] Maintenance

## Test plan

- [ ] Code Coverage creates \`build_env\` from \`environment-dev.yml\`, then extra, then static, without a solver conflict
- [ ] Static install (\`python>=3.12\`) links pygments successfully
- [ ] Coverage job proceeds past env setup

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -69,12 +69,6 @@ jobs:
             ccache-libmamba-${{ github.runner.os }}
       - name: Build mamba with coverage
         run: |
-          # Ensure ar/ranlib are found: conda compiler activation may not set AR when
-          # mixing envs (dev + micromamba-static); CMAKE_AR from preset uses $env{AR}
-          # which can resolve to a bare name, causing "x86_64-conda-linux-gnu-ar: not found"
-          # when building the static libmamba (micromamba requires BUILD_STATIC).
-          export AR="${CONDA_PREFIX}/bin/x86_64-conda-linux-gnu-ar"
-          export RANLIB="${CONDA_PREFIX}/bin/x86_64-conda-linux-gnu-ranlib"
           cmake -B build/ -G Ninja \
             --preset mamba-unix-shared-debug \
             -D CMAKE_CXX_COMPILER_LAUNCHER=sccache \
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -48,6 +48,10 @@ jobs:
         with:
           environment-file: ./dev/environment-dev.yml
           environment-name: build_env
+          # Most verbose micromamba logs to debug sporadic
+          # "Cannot find a valid extracted directory cache" failures
+          # after environment cache restore.
+          log-level: trace
           cache-environment: true
           cache-environment-key: >-
             coverage-${{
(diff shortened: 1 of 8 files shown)


jjerphan:
First Code Coverage run was a **cache miss** (new cache key includes this PR number), so it did not exercise the failure path.

What we learned:

- `setup-micromamba` created `build_env` from `environment-dev.yml`, then saved that env immediately (`Saved cache with ID 6739797613`) **before** the extra/static installs.
- After create, `pkgs/` used the hierarchical layout and **did** contain both the tarball and extracted dir for pygments (`pygments-2.21.0-pyhcf101f3_0`, not 2.20.0 from the original failure). Extra and static installs therefore succeeded.
- On a later cache **hit**, only the env prefix is restored; `pkgs/` is not. That is the situation from the original failure, where `python>=3.12` then reinstalls noarch packages such as pygments at link time.

Re-running Code Coverage now to force a cache hit and capture TRACE logs on that path.

jjerphan:
## Cache-hit reproduction and root cause

Rerun of [Code Coverage](https://github.com/mamba-org/mamba/actions/runs/32119113322/job/95657454553) was a **cache hit** and reproduced the failure.

### What happened

1. `setup-micromamba` restored `build_env` only. `pkgs/` did **not** exist. pygments **was** already in the prefix (`conda-meta/pygments-2.21.0-pyhcf101f3_0.json`, Python 3.11).
2. Extra install succeeded (new packages were fetched and linked).
3. Static install (`python>=3.12`) planned a **reinstall** of noarch pygments because Python was changing. At 09:07:27 both the extracted dir and tarball were reported invalid (they were not on disk).
4. Download and extract **succeeded**:
   - tarball: `/home/runner/micromamba/pkgs/https/conda.anaconda.org/conda-forge/noarch/pygments-2.21.0-pyhcf101f3_0.conda`
   - extracted: `.../noarch/pygments-2.21.0-pyhcf101f3_0` at 09:07:28.918
5. pygments was unlinked from the prefix (09:07:31.868).
6. Linking failed at 09:07:35.474 **without** another `Verify cache` line:

   ```
   error    libmamba Cannot find a valid extracted directory cache for 'pygments-2.21.0-pyhcf101f3_0.conda'
   critical libmamba Package cache error.
   ```

7. The post-failure dump showed **both** the tarball and extracted directory still present.

The missing `Verify cache` immediately before the error is the tell: `PackageCacheData::has_valid_extracted_dir` returned a **stale negative** (`m_valid_extracted_dir[pkg] == false` from the planning lookup) and never re-checked the disk.

This is in **released micromamba 2.9.0** used by `setup-micromamba@v3`, not in the sources this job compiles later. `#4328` (`90aba132`, *Clear stale negative cache entries*) is already on `main` and is **not** in 2.9.0: it clears the query cache after extract and re-resolves the extracted path at link time. Coverage CI will keep hitting 2.9.0 until that ships.

### Why it is sporadic

`cache-environment` saves the env prefix **immediately after create** from `environment-dev.yml` (Python 3.11), before extra/static installs. A cache **miss** leaves `pkgs/` populated, so the Python 3.12 reinstall finds a valid extracted pygments dir. A cache **hit** restores the prefix without `pkgs/`, so that reinstall has to download/extract in the same transaction that still holds the negative cache entry.

### CI workaround (pushed)

Create the full coverage env in one `micromamba create` (`environment-dev.yml` + extra + static via `create-args`). The cached prefix then already has `python>=3.12`, so a later restore no longer upgrades Python and reinstalls noarch packages against an empty `pkgs/`. Trace dumps are removed; they did their job.

Unix ubuntu **Build binaries** failures on this PR are unrelated (`No CMAKE_C_COMPILER could be found`).

jjerphan:
The one-shot `create` failed because the three env files are **not jointly solvable**. The command was:

```
micromamba create -f ./dev/environment-dev.yml -n build_env \
  -f ./dev/environment-dev-extra.yml -f ./dev/environment-micromamba-static.yml
```

That asks the solver to install shared `yaml-cpp` / `reproc-cpp` **and** their `*-static` counterparts, plus unpinned `python` with `bcrypt==4.0.1` **and** `python>=3.12`. Sequential `micromamba install` can replace packages; a single create cannot.

The `ENOENT` on `micromamba-shell` is post-cleanup after that failed create (shell init never finished).

Pushed a different workaround: restore the extra/static install steps and set `cache-environment: false` so `pkgs/` is always populated and we never take the 2.9.0 stale-cache relink path.

jjerphan:
Closing as the issue should be resolved with https://github.com/mamba-org/mamba/pull/4328.

PR #4387, opened 2026-08-19
fix: Make macOS ad-hoc codesign spawn reliably

jjerphan:
# Description

Extracted from https://github.com/mamba-org/mamba/pull/4382 following https://github.com/mamba-org/mamba/pull/4382#issuecomment-5339850672.

`reproc` forks and then closes every FD up to `RLIMIT_NOFILE`. On GitHub Actions that limit is huge, so the child aborts with `EINVAL` and linking rewritten Mach-O binaries such as ncurses fails. Spawn `/usr/bin/codesign` directly, as conda and rattler do.

## Type of Change

- [x] Bugfix
- [ ] Feature / enhancement
- [ ] CI / Documentation
- [ ] Maintenance

## Checklist


libmamba/src/core/util_os.cpp
```diff
@@ -745,8 +754,64 @@ namespace mamba
...
+        }
+
+        int wstatus = 0;
+        if (waitpid(pid, &wstatus, 0) < 0)
+        {
+            throw std::runtime_error(
+                std::string("Could not codesign executable: ") + std::strerror(errno)
+            );
+        }
+#else
```
Hind-M: Can you add comments to explain the steps? (initialization, redirecting stdout/stderr etc).

libmamba/src/core/util_os.cpp
```diff
@@ -745,8 +754,64 @@ namespace mamba
...
+            const_cast<char*>(path_str.c_str()),
+            nullptr,
+        };
+
+        posix_spawn_file_actions_t file_actions;
+        int rc = posix_spawn_file_actions_init(&file_actions);
+        if (rc != 0)
+        {
+            throw std::runtime_error(
+                std::string("Could not codesign executable: ") + std::strerror(rc)
```
Hind-M: The error here can be misleading, as we didn't run `codesign` yet. Maybe some error message on initialization failure? (Having different and specific error messages would definitely help debugging, so we should change error messages below accordingly when possible).

libmamba/src/core/util_os.cpp
```diff
@@ -745,8 +754,64 @@ namespace mamba
...
+        }
+
+        int devnull = -1;
+        if (!verbose)
+        {
+            devnull = ::open("/dev/null", O_RDWR | O_CLOEXEC);
+            if (devnull >= 0)
+            {
+                posix_spawn_file_actions_adddup2(&file_actions, devnull, STDOUT_FILENO);
+                posix_spawn_file_actions_adddup2(&file_actions, devnull, STDERR_FILENO);
```
Hind-M: Maybe check return codes of `posix_spawn_file_actions_adddup2` on both redirections to warn/error out about it? Not sure if it's worth it.

libmamba/src/core/util_os.cpp
```diff
@@ -745,8 +754,64 @@ namespace mamba
...
+        }
+        if (rc != 0)
+        {
+            throw std::runtime_error(
+                std::string("Could not codesign executable: ") + std::strerror(rc)
+            );
+        }
+
+        int wstatus = 0;
+        if (waitpid(pid, &wstatus, 0) < 0)
```
Hind-M, line 827: Maybe check termination status, i.e checking `wstatus` with `WIFEXITED` and `WIFSIGNALED`?

libmamba/src/core/util_os.cpp
Hind-M: It looks like `codesign` function is only used in `#if defined(__APPLE__) `blocks. If confirmed, we should just define it only there and remove the useless `else` part handling other platforms.

libmamba/src/core/util_os.cpp
```diff
@@ -10,8 +12,11 @@
 #include <sys/utsname.h>
 #include <unistd.h>
 #if defined(__APPLE__)
+#include <fcntl.h>
 #include <libproc.h>
 #include <mach-o/dyld.h>
+#include <spawn.h>
+#include <sys/wait.h>
```
Hind-M, line 20: Btw if the headers are added/used for `codesign` only, consider putting them under the `__APPLE__` directive as well.

libmamba/include/mamba/core/util_os.hpp
```diff
@@ -79,7 +79,9 @@ namespace mamba
     int get_console_width();
     int get_console_height();
 
+#if defined(__APPLE__)
     void codesign(const fs::u8path& path, bool verbose = false);
+#endif
 }
```
jjerphan: Ideally we would like not to expose this declaration for other platforms, but I am not sure we are able to now without modifying the ABI.

libmamba/src/core/util_os.cpp
jjerphan: Those headers are already under this directive l.14; but probably you are referring to something else?

Hind-M: Oh yes I meant all of them including `<cerrno>` and `<cstring>`.

libmamba/include/mamba/core/util_os.hpp
Hind-M: Right, we should ask other team members for their opinions, but this change would probably be deferred.

jjerphan: I would keep this interface and have it a co-op on other platform than macOS (as before). What do you think?

Hind-M: Ok.

libmamba/src/core/util_os.cpp
```diff
@@ -743,23 +752,110 @@ namespace mamba
...
 
-    void codesign(const fs::u8path& path, bool verbose)
+    void codesign([[maybe_unused]] const fs::u8path& path, [[maybe_unused]] bool verbose)
     {
-        reproc::options options;
-        options.env.behavior = reproc::env::empty;
+#if defined(__APPLE__)
+        // Do not use reproc here. It forks and then fcntl/closes every FD up to
+        // RLIMIT_NOFILE; on GitHub Actions that limit is huge, the child aborts,
+        // and the parent sees EINVAL. conda and rattler just posix_spawn codesign
```
Hind-M:
```suggestion
        // and the parent sees EINVAL. `conda` and `rattler` just `posix_spawn` `codesign`
```

libmamba/src/core/util_os.cpp
```diff
@@ -743,23 +752,110 @@ namespace mamba
...
+        // with the inherited environment.
+
+        // Ad-hoc sign: /usr/bin/codesign -s - -f <path>
```
Hind-M:
```suggestion
        // Ad-hoc sign: `/usr/bin/codesign -s - -f <path>`
```

libmamba/src/core/util_os.cpp
```diff
@@ -743,23 +752,110 @@ namespace mamba
...
+                        mamba_error_code::internal_failure
+                    );
+                }
+            }
         }
 
-        const std::vector<std::string> cmd = { "/usr/bin/codesign", "-s", "-", "-f", path.string() };
-        auto [status, ec] = reproc::run(cmd, options);
-        if (ec)
+        // Spawn codesign with the inherited environment, then drop parent-side FDs.
```
Hind-M:
```suggestion
        // Spawn `codesign` with the inherited environment, then drop parent-side FDs.
```

Hind-M (approved):

PR #3125, opened 2024-01-12
Added mirrored channels

JohanMabille:
First PR for plugging the mirrors

PR #4395, opened 2026-09-03
fix: Assign channel priority in `ctx.channels` order

black-desk:
# Description

Mirrored channels were processed in a separate first pass before regular
channels, causing them to always receive a higher priority regardless of
their position in the channels list. This broke strict channel priority
when a channel listed later (for example, conda-forge) had mirror
configuration while a channel listed earlier did not.

Process all channels in ctx.channels order and expand mirror
configuration at the configured channel's position. A mirrored channel
that is not listed in channels is also no longer loaded implicitly into
the solver.

The integration regression test resolves an overlapping package from an
earlier regular channel and a later mirrored channel under strict
priority. An additional unit test checks that mirror-only configuration
does not add repositories to the solver.

Bug introduced in 67f36d4a4815 (#3125).

Assisted-by: LLM

## Checklist


JohanMabille (approved):
Thanks for the fix!

PR #4397, opened 2026-09-04
fix: remove any usage of `exit()`

Klaim:
# Description

Removes any call to `exit()` or similar from libmamba and mamba exe.
This makes the handling of static objects clearer (everything depends on main and singletons explicit lifetimes) and helps solve related issues.
It also gives back the control of program exit to the program using `libmamba` when the exit call was done in the library.
Discussed [there](https://github.com/mamba-org/mamba/pull/4379#issuecomment-5538429616), needed to remove the logging termination reason system while also solving [this crash](https://github.com/conda/constructor/issues/1319).

To replace some of the exit calls, this PR:
- adds a mechanism in `(micro)mamba` executable to request a specific exit value, mainly used by `run` subcommand;
- adds a C++ exception type only used as a request to exit the program early, [discussed there](https://github.com/mamba-org/mamba/pull/4397#issuecomment-5583903107).

IMPORTANT: This also changes the output of `--version` to something like `(micro)mamba v2.x.y`. This is due to the exit call that was removed hiding the rest of the execution which had to be adjusted to output correctly.

- [x] Bugfix
- [ ] Feature / enhancement
- [ ] CI / Documentation
- [ ] Maintenance

## Checklist

### DIFF
--- a/libmamba/include/mamba/core/error_handling.hpp
+++ b/libmamba/include/mamba/core/error_handling.hpp
@@ -76,6 +76,16 @@ namespace mamba
         bool m_with_bug_report_message = true;
     };
 
+    /// Only used when we need to exit the program ASAP
+    /// but:
+    //  - we still need to go exit from `main()`;
+    /// - we want the program in control of what to do exactly.
+    /// This should NOT be considered an error and should
+    /// not result in an error report.
+    class mamba_early_exit_request : public std::exception
+    {
+    };
+
     /********************************
      * wrappers around tl::expected *
      ********************************/
--- a/libmamba/src/api/configuration.cpp
+++ b/libmamba/src/api/configuration.cpp
@@ -2297,7 +2297,7 @@ namespace mamba
             int dump_opts = MAMBA_SHOW_CONFIG_VALUES | MAMBA_SHOW_CONFIG_SRCS
                             | MAMBA_SHOW_ALL_CONFIGS;
             print_dump(*this, dump_opts);
-            exit(0);
+            return;
         }
 
         m_context.set_log_level(m_context.output_params.logging_level);
--- a/libmamba/src/api/configuration.cpp
+++ b/libmamba/src/api/configuration.cpp
@@ -2297,7 +2297,7 @@ namespace mamba
             int dump_opts = MAMBA_SHOW_CONFIG_VALUES | MAMBA_SHOW_CONFIG_SRCS
                             | MAMBA_SHOW_ALL_CONFIGS;
             print_dump(*this, dump_opts);
-            return;
+            throw mamba_early_exit_request{};
         }
 
         m_context.set_log_level(m_context.output_params.logging_level);
--- a/libmamba/src/core/run.cpp
+++ b/libmamba/src/core/run.cpp
@@ -243,7 +243,7 @@ namespace mamba
     }
 
 #ifndef _WIN32
-    void daemonize()
+    auto daemonize() -> std::optional<int>
     {
         pid_t pid, sid;
         int fd;
@@ -251,20 +251,20 @@ namespace mamba
         // already a daemon
         if (getppid() == 1)
         {
-            return;
+            return {};
         }
 
         // fork parent process
         pid = fork();
         if (pid < 0)
         {
-            exit(1);
+            return 1;
         }
 
         // exit parent process
         if (pid > 0)
         {
-            exit(0);
+            return 0;
         }
 
         // at this point we are executing as the child process
@@ -272,7 +272,7 @@ namespace mamba
         sid = setsid();
         if (sid < 0)
         {
-            exit(1);
+            return 1;
         }
 
         fd = open("/dev/null", O_RDWR, 0);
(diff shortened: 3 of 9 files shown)


Klaim:
# To Reviewers

I don't like at all to use C++ exceptions as a execution flow mechanism but I feel I'm being forced:
- I considered making `Configuration::load()` return a boolean value to express if we should or not continue execution after it's call but it is called a bit more than 50 places in the project and that dissuades me to add `if`s everywhere.
- At a minimum the distinct exception allows the caller of the library to decide what to do. Before this PR, calling `Configuration::load()` could result in an exit of the program.

I think the execution of the configuration dump is probably not done where it should be done and that leads to this mess. If you have suggestions as to how to not have to use either exception or exit call to do the configuration dump quickly without compromising the rest of the configuration behavior, I'm interested.

libmamba/include/mamba/core/error_handling.hpp
```diff
@@ -76,6 +76,16 @@ namespace mamba
         bool m_with_bug_report_message = true;
     };
 
+    /// Only used when we need to exit the program ASAP
```
Klaim, line 79: Reviewers, see [my comment there](https://github.com/mamba-org/mamba/pull/4397#issuecomment-5583903107).

libmamba/src/api/configuration.cpp
```diff
@@ -2297,7 +2297,7 @@ namespace mamba
             int dump_opts = MAMBA_SHOW_CONFIG_VALUES | MAMBA_SHOW_CONFIG_SRCS
                             | MAMBA_SHOW_ALL_CONFIGS;
             print_dump(*this, dump_opts);
-            exit(0);
+            throw mamba_early_exit_request{};
```
Klaim, line 2300: see: https://github.com/mamba-org/mamba/pull/4397#issuecomment-5583903107

micromamba/src/umamba.cpp
```diff
@@ -15,6 +18,31 @@
...
+    }
+
+    auto app_name() -> std::string
+    {
+        return mamba::get_self_exe_path().stem().string();
+    }
+
+    auto app_name_with_version() -> std::string
+    {
+        return std::format("{} v{}", umamba::app_name(), umamba::version());
```
jjerphan, line 42: This is a breaking change; one would have:

```
mamba v2.9.0
```

instead of:

```
2.9.0
```

isn't it?

libmamba/src/core/run.cpp
```diff
@@ -400,7 +402,11 @@ namespace mamba
                 "Running wrapped script {} in the background\n",
                 fmt::join(command, " ")
             );
-            daemonize();
+            auto result = daemonize();
+            if (result)
+            {
+                return *result;
+            }
```
jjerphan:
```suggestion
            const auto& exit_code = daemonize();
            if (exit_code)
            {
                return *exit_code;
            }
```

jjerphan (commented):
LGTM modulo a comment.

micromamba/src/umamba.cpp
Klaim: You mean this function used in `--version`? Indeed, I guess people could be parsing this. Let's keep this function called as name of the app so that it's visible in `--help` and change `--version` back to just the value. 👍🏽

libmamba/src/core/run.cpp
Klaim: The reference is not useful here and the name can be confused as usually exit codes are just ints and tested like that it looks like we are checking if it's zero. I'll just rename to `maybe_exit_code` (as it's an optional) and keep the const 👍🏽

micromamba/src/umamba.cpp
jjerphan: Yes, I know that at least [`repo2docker` uses it](https://github.com/jupyterhub/repo2docker/blob/main/tests/conda/py35-binder-dir/verify#L20-L24) ([2.0 changed the output for mamba](https://github.com/jupyterhub/repo2docker/commit/f41329fb3076d51ee876d09c5139367a0cfaa01a#diff-5d7c40c95bf9ac2a2eb6deb2cb6b6c4e1bbd15a34d28181c88d6c162d9c636cfR21-R22)).

Klaim: Done: `--version` will return `umamba::version()` now and this function is still used in the name of the app for cli11.

libmamba/src/api/configuration.cpp
JohanMabille: Actually I don't know who or which tool is using this configuration value. And I think it should be a command (like dump_config) or a flag for the config command, rather than a value in the configuration file. But let's keep this refactoring for later, that would be a breaking change.

JohanMabille (approved):

jjerphan (approved):

libmamba/src/api/configuration.cpp
Klaim: My assumption is that it's specified by conda's cli but I might be wrong.

PR #4399, opened 2026-09-09
feat: Remove stop-reason mechanism from logging system

Klaim:
# Description

The stop-reason mechanism in the logging system of `mamba` was initially introduced because of the behavior of `spdlog` when a call to `exit()` was done. In essence, `spdlog` will do the proper work of de-initializing itself at some point after `main()` but we cannot guarantee when if `exit()` is called. The proper way to use `spdlog` or any library that relies on static objects is to explicitly end their usage through RAII and scope that usage within the boundaries of `main()`.
Our usage of `exit()` prevented that scenario so we had to add a way to inform the currently active log handler (in `mamba`'s logging system) why it was being stopped so that in the specific case of `spdlog` we would either explicitly stop `spdlog` or do nothing because we dont know if `spdlog` internals were destroyed or not (`exit()` case).

However, that mechanism doesnt work well in contexts where libmamba is used inside another program than `(micro)mamba`, as reported by  #4378 . [A workaround](https://github.com/mamba-org/mamba/pull/4379) was being developed but we discussed better alternatives and realized that if we didn't have `exit()` calls, we could simply always assume that we can stop `spdlog` explicitly without any issue.

#4397 removed any call to `exit()` from the library or executable, forcing execution to go through `main()` properly (which also helps with reasoning about the flow in the program).
Now we have no reason anymore to keep the stop-reason mechanism which was kind of workwaround. This PR removes that mechanism and simplifies the logic of the logging system.

This PR also adds a test for `libmambapy` for the [reported crash](https://github.com/mamba-org/mamba/issues/4378), that test comes from [the workaround pr](https://github.com/mamba-org/mamba/pull/4379).

## Type of Change

- [ ] Bugfix
- [x] Feature / enhancement
- [ ] CI / Documentation
- [ ] Maintenance

## Checklist

### DIFF
--- a/libmamba-spdlog/include/mamba/spdlog/logging_spdlog.hpp
+++ b/libmamba-spdlog/include/mamba/spdlog/logging_spdlog.hpp
@@ -75,7 +75,7 @@ namespace mamba::logging::spdlogimpl
         */
         ///@{
         auto start_log_handling(LoggingParams params, std::vector<log_source> sources) -> void;
-        auto stop_log_handling(stop_reason reason) -> void;
+        auto stop_log_handling() -> void;
 
         auto set_log_level(log_level new_level) -> void;
         auto set_params(LoggingParams new_params) -> void;
--- a/libmamba-spdlog/include/mamba/spdlog/logging_spdlog_impl.hpp
+++ b/libmamba-spdlog/include/mamba/spdlog/logging_spdlog_impl.hpp
@@ -120,34 +120,24 @@ namespace mamba::logging::spdlogimpl
         pimpl->is_active = true;
     }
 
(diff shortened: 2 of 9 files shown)


jjerphan (approved):

Hind-M (approved):

jjerphan:
Should we close https://github.com/mamba-org/mamba/pull/4379 and mark https://github.com/mamba-org/mamba/issues/4378 as resolved?

Klaim:
> Should we close #4379 and mark #4378 as resolved?

Yes 👍🏽  and probably suggest to reopen the crash issue if next version doenst fix it.

